PR #1, opened 2021-06-04
Initial development setup

pdyraga:
Hardhat-based development environment with Solidity and JS linting based on Keep rules. Configuration for Slither static analyzer from Trail Of Bits included as well. The setup is based on the one from https://github.com/keep-network/coverage-pools/.

PR #2, opened 2021-06-04
GitHub Actions CI jobs

pdyraga:
Depends on #1 

The following jobs are defined:
- contracts build and test,
- contracts linting for `.sol` code and `.js` tests,
- slither static analyzer for contracts.

This solution does not cache `node_modules` and it needs to be addressed separately. 

Apologies for force-pushes, this is still a working draft and I need to test things.

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -1,0 +1,112 @@
+name: Solidity
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "solidity/**"
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  contracts-detect-changes:
+    runs-on: ubuntu-latest
+    outputs:
+      path-filter: ${{ steps.filter.outputs.path-filter }}
+    steps:
+      - uses: actions/checkout@v2
+        if: github.event_name == 'pull_request'
+        
+      - uses: dorny/paths-filter@v2
+        if: github.event_name == 'pull_request'
+        id: filter
+        with:
+          filters: |
+            path-filter:
+              - './solidity/**'
+  contracts-build-and-test:
+    needs: contracts-detect-changes
+    if: |
+      github.event_name != 'pull_request'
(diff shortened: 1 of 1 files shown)


solidity/.github/workflows/contracts.yml
```diff
@@ -0,0 +1,122 @@
+name: Solidity
```
nkuba, line 1: We need to move the configuration file to `.github/workflows/contracts.yml`.

pdyraga: 🤦

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,122 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./solidity
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "12.x"
```
nkuba, line 43: If we're starting with a new repo, could we give version `14` a try?

solidity/.github/workflows/contracts.yml
nkuba: ... and remove this file :)

.github/workflows/contracts.yml
pdyraga: Sure, why not.

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,112 @@
...
+        with:
+          node-version: "14.x"
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8.5
+
+      - name: Install Solidity
+        env:
+          SOLC_VERSION: 0.7.6 # according to solidity.version in hardhat.config.js
```
nkuba, line 97: We should extract the version from `hardhat.config.js` and use it here as a variable. But it's for the future...

nkuba (approved):

PR #3, opened 2021-06-04
Burnable TBTC ERC20 token with permit

pdyraga:
TBTC is now a burnable ERC20 token with EIP2612 permit functionality.
User can authorize a transfer of their token with a signature conforming
EIP712 standard instead of an on-chain transaction from their address.
Anyone can submit this signature on the user's behalf by calling the
permit function, as specified in EP2612, paying gas fees, and possibly
performing other actions in the same transaction.

By the time of implementing this contract, to my knowledge, there is no other
audited EIP2612 token implementation than Uniswap V2 on which this code is
based on. OpenZeppelin has quite recently added permit to their ERC20
but this code is in a draft state, was released in the most recent
v4.0.0, and to my understanding was not audited yet and may have
breaking changes in the future. On the contrary, Uniswap implementation
was audited and is even referenced from the EIP as an implementation
example.

This implementation mostly mirrors `UniswapV2ERC20`, and is [identical to
the implementation we use in coverage pools](https://github.com/keep-network/coverage-pools/blob/main/contracts/ERC20WithPermit.sol). It adds some tweaks from
OpenZeppelin ERC20 implementation such as more meaningful revert messages
and `Approval` event emitted on calls to `transferFrom`. This allows
applications to reconstruct the allowance for all accounts just by
listening to the event.

Unit tests were based on OpenZeppelin ERC20 unit tests with our own
tests implementation for EIP2612 and EIP712.

In the future, we will add more functionalities to the token, such as
`approveAndCall` and recovering tokens mistakenly sent to TBTC token
address but all of that in a separate PR.

### DIFF
--- a/solidity/contracts/token/ERC20WithPermit.sol
+++ b/solidity/contracts/token/ERC20WithPermit.sol
@@ -21,22 +21,24 @@ import "./IERC20WithPermit.sol";
 contract ERC20WithPermit is IERC20WithPermit, Ownable {
     using SafeMath for uint256;
 
-    uint8 public constant decimals = 18;
-
-    string public name;
-    string public symbol;
-
-    uint256 public override totalSupply;
     mapping(address => uint256) public override balanceOf;
     mapping(address => mapping(address => uint256)) public override allowance;
 
-    uint256 public cachedChainId;
-    bytes32 public cachedDomainSeparator;
+    mapping(address => uint256) public override nonces;
+
+    uint256 public immutable cachedChainId;
+    bytes32 public immutable cachedDomainSeparator;
 
     // keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)");
     bytes32 public constant override PERMIT_TYPEHASH =
         0x6e71edae12b1b97f4d1f60370fef10105fa2faae0126114a169c64845d6126c9;
-    mapping(address => uint256) public override nonces;
+
+    uint256 public override totalSupply;
+
+    string public name;
+    string public symbol;
+
+    uint8 public constant decimals = 18;
 
     constructor(string memory _name, string memory _symbol) {
         name = _name;
--- a/solidity/contracts/token/ERC20WithPermit.sol
+++ b/solidity/contracts/token/ERC20WithPermit.sol
@@ -1,0 +1,202 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity <0.9.0;
+
+import "@openzeppelin/contracts/math/SafeMath.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+import "./IERC20WithPermit.sol";
+
+/// @title  ERC20WithPermit
+/// @notice Burnable ERC20 token with EIP2612 permit functionality. User can
+///         authorize a transfer of their token with a signature conforming
+///         EIP712 standard instead of an on-chain transaction from their
+///         address. Anyone can submit this signature on the user's behalf by
+///         calling the permit function, as specified in EIP2612 standard,
+///         paying gas fees, and possibly performing other actions in the same
+///         transaction.
+/// @dev    By the time of implementing this contract there is no other audited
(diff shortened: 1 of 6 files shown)


solidity/contracts/token/ERC20WithPermit.sol
```diff
@@ -0,0 +1,221 @@
...
+    string public symbol;
+
+    uint256 public override totalSupply;
+    mapping(address => uint256) public override balanceOf;
+    mapping(address => mapping(address => uint256)) public override allowance;
+
+    uint256 public cachedChainId;
+    bytes32 public cachedDomainSeparator;
+
+    // keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)");
```
beaushinkle, line 36: I see this comment in the original uniswap implementation and the covpools implementation. What does it mean?

solidity/contracts/token/ERC20WithPermit.sol
```diff
@@ -0,0 +1,221 @@
...
+    bytes32 public constant override PERMIT_TYPEHASH =
+        0x6e71edae12b1b97f4d1f60370fef10105fa2faae0126114a169c64845d6126c9;
```
beaushinkle, line 38: What does this refer to?

pdyraga: [EIP 2612](https://eips.ethereum.org/EIPS/eip-2612) permit message is defined as:
```
keccak256(abi.encodePacked(
   hex"1901",
   DOMAIN_SEPARATOR,
   keccak256(abi.encode(
            keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)"),
            owner,
            spender,
            value,
            nonce,
            deadline))
))
```

`PERMIT_TYPEHASH` is precomuted keccak256 so that we do not need to compute it for every `permit` call. 

In other words, the value of `PERMIT_TYPEHASH` is the result of the expression in the comment above.

beaushinkle: that's smart!

solidity/test/token/ERC20WithPermit.test.js
```diff
@@ -0,0 +1,1008 @@
...
+  })
+
+  describe("permit_typehash", () => {
+    it("should be keccak256 of EIP2612 Permit message", async () => {
+      const expected = ethers.utils.keccak256(
+        ethers.utils.toUtf8Bytes(
+          "Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)"
+        )
+      )
+      expect(await token.PERMIT_TYPEHASH()).to.equal(expected)
```
beaushinkle, line 53: ahhhhhhhh, I get it

beaushinkle:
Okay, after comparing our tests and implementation with the openzeppelin work, I see that this is *mostly* the smart contract and testing version of boilerplate / importing a library. I'm good to go here. I'll leave it to @pdyraga as to whether or not we want to seek more reviewers or if we're good to merge

solidity/contracts/token/TBTCToken.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity <0.9.0;
+
+import "./ERC20WithPermit.sol";
+
+contract TBTCToken is ERC20WithPermit {
+    constructor() ERC20WithPermit("TBTC v2 migration", "TBTC") {}
```
dimpar, line 8: just double checking.. symbol stays the same as in V1 - TBTC?

pdyraga: Correct. But there will be a way to change the name (description). This will be done in a separate PR.

dimpar:
I've compared `ERC20WithPermit` and tests with the one that we have in [coverage pool](https://github.com/keep-network/coverage-pools) - they match. Lgtm

dimpar (approved):

PR #4, opened 2021-06-10
Allow TBTC contract owner to recover misfunded ERC20/ERC721 tokens

pdyraga:
In case someone sends tokens to TBTC token contract address, these
tokens can be recovered by the governance. We observed such cases of
misfunds for tBTC v1 so to become more user-friendly, we are giving
rescue tools to tBTC v2 governance.

In practice, recovery functions will have to be proxied through the
vending machine because the vending machine will be the owner of TBTC
token contract.

### DIFF
--- a/solidity/test/token/MisfundRecovery.test.js
+++ b/solidity/test/token/MisfundRecovery.test.js
@@ -1,0 +1,88 @@
+const { to1e18 } = require("../helpers/contract-test-helpers")
+
+const { expect } = require("chai")
+
+describe("MisfundRecovery", () => {
+  let recoveryOwner
+  let thirdParty
+
+  let randomERC20
+  let randomERC721
+
+  let recovery
+
+  beforeEach(async () => {
+    ;[deployer, recoveryOwner, thirdParty] = await ethers.getSigners()
+
+    const MisfundRecovery = await ethers.getContractFactory("MisfundRecovery")
+    recovery = await MisfundRecovery.deploy()
+    await recovery.deployed()
+
+    const TestERC20 = await ethers.getContractFactory("TestERC20")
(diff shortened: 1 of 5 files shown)


solidity/test/token/MisfundRecovery.test.js
```diff
@@ -0,0 +1,88 @@
...
+    await randomERC721.deployed()
+
+    await recovery.connect(deployer).transferOwnership(recoveryOwner.address)
+  })
+
+  describe("recoverERC20", () => {
+    const amount = to1e18(725)
+
+    beforeEach(async () => {
+      await randomERC20.mint(recovery.address, amount)
```
dimpar, line 36: In the description of this PR it says that we should be able to recover tokens from the TBTC token contract. I see that TBTCToken contract inherits from MisfundRecovery contract, but don't we want to test the actual TBTCToken contract around recover functionality?

pdyraga: I was also thinking about it and I was still on the fence when implementing the code. We can test the inheriting contract but then... why not to test all `ERC20WithPermit` functionalities against `TBTCToken`? 🤷 I have finally came up with the rule that if some function is implemented in `TBTCToken` - just like `approveAndCall` in #5, then I put tests in `TBTCToken.test.js`. If something is inherited from the abstract contract, then the abstract contract is tested. It is not perfect, but it's how most tokens implemented based on Open Zeppelin abstractions work and I think it is better than repeating those tests.

dimpar: Ok, this rule makes sense, especially if we're planning to inherit multiple times from the abstract contracts.

dimpar (approved):

PR #5, opened 2021-06-10
Added `approveAndCall` to TBTC token

pdyraga:
Depends on #4 

`approveAndCall` is not a part of ERC20 - it was specified in [ERC827](https://github.com/ethereum/eips/issues/827) and
slightly diverged for some implementations. This one is identical to
`approveAndCall` behavior of [KEEP token](https://github.com/keep-network/keep-core/blob/v1.0.1/solidity/contracts/KeepToken.sol#L29).

`approveAndCall` executes `receiveApproval` function on the spender as
specified in `IReceiveApproval` interface. Approves spender to withdraw
from the caller multiple times, up to the value amount. If this function
is called again, it overwrites the current allowance with the provided
value. Reverts if the approval operation reverted or if `receiveApproval`
call on the spender reverted.

### DIFF
--- a/solidity/contracts/test/ReceiveApprovalStub.sol
+++ b/solidity/contracts/test/ReceiveApprovalStub.sol
@@ -2,8 +2,6 @@
 
 pragma solidity <0.9.0;
 
-import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
-
 import "../token/TBTCToken.sol";
 
 contract ReceiveApprovalStub is IReceiveApproval {
--- a/solidity/contracts/test/ReceiveApprovalStub.sol
+++ b/solidity/contracts/test/ReceiveApprovalStub.sol
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity <0.9.0;
+
+import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
+
+import "../token/TBTCToken.sol";
+
+contract ReceiveApprovalStub is IReceiveApproval {
+    bool public shouldRevert;
+
+    event ApprovalReceived(
+        address from,
+        uint256 value,
+        address token,
+        bytes extraData
+    );
+
+    function receiveApproval(
+        address from,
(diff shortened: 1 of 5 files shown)


solidity/contracts/test/ReceiveApprovalStub.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity <0.9.0;
+
+import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
```
dimpar, line 5: I think we can remove this import

solidity/contracts/token/TBTCToken.sol
```diff
@@ -5,6 +5,37 @@ pragma solidity <0.9.0;
 import "./ERC20WithPermit.sol";
 import "./MisfundRecovery.sol";
 
+interface IReceiveApproval {
```
dimpar, line 8: In TBTC v1 this interface is named ITokenRecipient. Was there a reason why we don't want to name it the same?

pdyraga: In retrospect, I think `ITokenRecipient` is too broad. It is about a specific way of receiving approval and *maybe* receiving tokens if the implementation decides for that.

solidity/contracts/test/ReceiveApprovalStub.sol
pdyraga: 🔥 in https://github.com/keep-network/tbtc-v2/pull/5/commits/d671d1a73c27b5b2dce30126cab6b12a9d873db3

solidity/contracts/token/TBTCToken.sol
dimpar: Sure, I'm okay with both, but what I liked about `ITokenRecipient` is that it tells me right away in its name that this contract will receive tokens. `IReceiveApproval(spender).receiveApproval` also looks a bit redundant.

dimpar (approved):

Issue #7, opened 2021-07-05
Yearn Finance integration


lukasz-zimnoch:
This issue describes the work which should be done to integrate with [Yearn Finance](https://yearn.finance/). The target is to have a Yearn vault, similar to the existing [Curve tBTC Pool yVault (crvTBTC)](https://yearn.watch/vault/0x23D3D0f1c697247d5e0a9efB37d8b0ED0C464f7f) created for tBTC v1, and a new one which will be able to handle Saddle incentives. Here are the steps that need to be done:

**Pre-deploy actions**

1. Deploy the tBTC v2 token
2. Create a Curve pool for tBTC v2 token like the one we have for [v1](https://github.com/curvefi/curve-contract/tree/master/contracts/pools/tbtc). Worth noting the v1 pool uses the [`LiquidityGaugeReward`](https://github.com/curvefi/curve-dao-contracts/blob/master/contracts/gauges/LiquidityGaugeReward.vy) contract as a gauge. This gauge stakes LP tokens in the Synthetix staking rewards contract and handles distribution of KEEP token as additional rewards. Presumably, we'll want to replay that in v2 pool.
3. Create a Saddle pool for tBTC v2 token.
4. Create an experimental Yearn vault for the LP token of the tBTC v2 Curve pool. This is what we have for tBTC v1. The new vault should be probably named `crvTBTCv2`. An experimental vault can be created using [`newExperimentalVault`](https://github.com/yearn/yearn-vaults/blob/0964695205e62c6eb4820473a54cb39b215cc3a6/contracts/Registry.vy#L238) method of the [Yearn registry contract](https://etherscan.io/address/v2.registry.ychad.eth). Experimental vault is a kind of „unofficial” vault which is not listed by the registry.
5. Create and experimental Yearn vault for the LP token of the tBTC v2 Saddle pool. The new vault should be probably named `saddleTBTCv2`.

**Strategies implementation**

[...]

PR #8, opened 2021-07-06
Yearn: Curve voter proxy strategy

lukasz-zimnoch:
Refs: #7 

**Introduction**

This pull request consists of two parts. The first part introduces a basic `yearn` submodule intended to be a bag for components used for Yearn Finance integration. The second one adds the first strategy named `CurveVoterProxyStrategy` meant to be used with the Yearn vault for tBTC v2.

**The `yearn` submodule**

The `yearn` submodule has been created in order to hold all contracts required by the Yearn integration. This module uses Hardhat and Yarn in the same way the protocol contracts module (`solidity` directory) does. A separate module for Yearn integration contracts was chosen because:

- Yearn integration contracts require third-party dependencies which shouldn't be included into the protocol contracts package. Specifically, strategies contracts need to inherit from the `BaseStrategy.sol` base contract and that dependency could not be broken using interfaces.
- Yearn integration contracts require an older version of Solidity (`0.6.12`) and older versions of several auxilliary packages (e.g. OpenZeppelin). This is because strategies need to use the same version of Solidity and packages as used by the `BaseStartegy` contract they inherit from.
- Separate module improves readability and isolates unrelated contracts

A separated CI workflow named `Yearn` was also created. It uses the `contracts-build-and-test` and `contracts-lint` jobs like the protocol contract module does. However, it doesn't use `contracts-slither` because integration contracts rely heavily on third-party code and must follow imposed conventions. This fact causes a lot of Slither problems that can be resolved only by ignoring them. In that case, there is no value-added provided by Slither scans.

Regarding testing, the plan assumes to use only system tests. Unit tests don't make much sense in that case as Yearn integration contracts use a lot of third-party code which is already unit tested. We are mostly interested to make sure the new contracts work well from the perspective of the entire Yearn ecosystem. System tests leveraging Hardhat mainnet forking feature seem to be the best option here.

Last but not least, the aforementioned `BaseStrategy.sol` contract needed by custom strategies is provided in the official [yearn-vaults](https://github.com/yearn/yearn-vaults) repository. However, there is no official npm package for that repo. There is one existing [`@yearnvaults/contracts`](https://www.npmjs.com/package/@yearnvaults/contracts) package but it is based on an outdated fork that doesn't have a lot of security and bug fixes compared to the latest official version. A git-based import of the original `yearn-vault` repo could be an option but it doesn't contain a the `version` parameter in their `package.json` which leads to an error when trying to add that repo using `yarn add`. Because of that, we [forked](https://github.com/keep-network/yearn-vaults) the `yearn-vault` repo under the `keep-network` organization, provided missing pieces in their `package.json`, and we use it from within `yearn` submodule.

**The `CurveVoterProxyStrategy` strategy contract**

This strategy is used in the existing vault for tBTC v1 ([Curve tBTC pool yVault](https://yearn.watch/vault/0x23D3D0f1c697247d5e0a9efB37d8b0ED0C464f7f)). The current implementation for tBTC v2 is based on the following repositories, especially on the last one:

- [General Yearn strategy template](https://github.com/yearn/brownie-strategy-mix)
- [Curve voter proxy strategy template](https://github.com/orbxball/curve-voter-proxy)
- [Curve voter proxy implementation for tBTC v1 vault](https://github.com/orbxball/btc-curve-voter-proxy)

Basically, the current implementation for tBTC v2 vault uses the same logic as the strategy for [tBTC v1 vault](https://github.com/orbxball/btc-curve-voter-proxy/blob/master/archived/tbtc.sol). However, it tries to simplify the contract structure and attempts to achieve better readability and documentation. It also uses a newer version of the `yearn-vault` base contract. Last but not least, it introduces support for additional rewards provided by Curve pool gauges leveraging Synthetix staking rewards. Additional rewards are supported by the [tBTC v1 Curve pool gauge](https://github.com/curvefi/curve-contract/tree/master/contracts/pools/tbtc) (see the usage of `LiquidityGaugeReward`) and will be probably supported by the tBTC v2 pool gauge as well. To read a detailed description about how this strategy works in details, please see the [documentation of the `CurveVoterProxyStrategy` contract](https://github.com/keep-network/tbtc-v2/blob/04043f4e2e8459fa5306831b918f6025bb94ec19/yearn/contracts/CurveVoterProxyStrategy.sol#L68)


lukasz-zimnoch:
Two TODOs need to be addressed in the future:
- Price oracle for `ethToWant` method. The `ethToWant` is used from within `harvestTrigger` method which is called by keeper bots to make a determination if calling it is "worth it" for the keeper. The price oracle is needed to convert the wei call cost estimation to a cost denominated in the vault's underlying token. Because neither the tBTC v2 nor its Curve pool exists yet, this one can not be addressed at the moment. However, it does not block the entire strategy which can be fully operational. It just affects the integration with the Keep3r network.
- Contract addresses for the system test scenario. The tBTC v2 token and its Curve pool are not deployed yet. Because of that, the system test just uses the tBTC v1 token and its Curve pool to test the strategy contract. Once the v2 components land on mainnet, all relevant addresses should be changed.

dimpar:
> However, there is no official npm package for that repo.

Do you think we can open an issue and ask them to deploy a package? Might be useful to get the latest from `npm` and not "track" their official repo. Maybe there's a small chance they make it happen? :)

lukasz-zimnoch:
> > However, there is no official npm package for that repo.
> 
> Do you think we can open an issue and ask them to deploy a package? Might be useful to get the latest from `npm` and not "track" their official repo. Maybe there's a small chance they make it happen? :)

Yep, this can be a good idea. However, I think that even if we keep tracking their official repo, this won't be a big pain for us. Those contracts are meant to be updated very very rarely.

yearn/test/system/curve-voter-proxy-strategy.test.js
```diff
@@ -0,0 +1,343 @@
...
+
+    // Deploy a new experimental vault accepting tBTC v2 Curve pool LP tokens.
+    const tx = await registry.newExperimentalVault(
+      tbtcCurvePoolLPToken.address,
+      vaultGovernance.address,
+      vaultGovernance.address, // set governance to be the guardian as well
+      vaultGovernance.address, // set governance to be the rewards target as well
+      vaultName,
+      vaultSymbol,
+      releaseDelta
```
dimpar, line 127: Not sure if I'm following this `releaseDelta`. There's another `newExperimentalVault()` which won't take this param and also all the tests pass. From the equation on line 117 the result is `0`, but if you change it changes to other number because of `registry.numReleases()` it will fail.

tomaszslabon:
I didn't find any issues with this PR.
Would it make sense to test more (non-trivial) functions from `CurveVoterProxyStrategy`?
For example, we could write tests for `migrate` (if we had another strategy)?
Or any other function that uses overridden methods we provide.

tomaszslabon (commented):
I didn't find any issues with this PR.
Would it make sense to test more (non-trivial) functions from `CurveVoterProxyStrategy`?
For example, we could write tests for `migrate` (if we had another strategy)?
Or any other function that uses overridden methods we provide.

lukasz-zimnoch:
> I didn't find any issues with this PR.
> Would it make sense to test more (non-trivial) functions from `CurveVoterProxyStrategy`?
> For example, we could write tests for `migrate` (if we had another strategy)?
> Or any other function that uses overridden methods we provide.

Yeah, we will probably need some additional system tests covering cases like that. However, I think we should address this in the end. We'll have more strategies and a better understanding of how everything works.

yearn/test/system/curve-voter-proxy-strategy.test.js
```diff
@@ -0,0 +1,343 @@
...
+  let tbtcCurvePoolGaugeReward
+  let tbtcCurvePoolGaugeRewardDistributor
+  let synthetixCurveRewards
+  let synthetixCurveRewardsOwner
+  let strategyProxyGovernance
+  let vault
+  let strategy
+
+  before(async () => {
+    await resetFork(12786839)
```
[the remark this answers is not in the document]
lukasz-zimnoch: Just the latest block from the moment the test was created. We need to set the same starting block in order to get a consistent state of the world at the beginning of testing.

yearn/test/system/curve-voter-proxy-strategy.test.js
lukasz-zimnoch: Yeah, that needs an additional explanation. The `newExperimentalVault` is defined in the Yearn's [`Registry`](https://github.com/yearn/yearn-vaults/blob/b60d2ef87645ef35a3d78a33d12591d70c7e9880/contracts/Registry.vy#L245) contract. This is a Vyper contract and it allows to set default function parameters. The `releaseDelta` param is the last parameter with a default value of `0`. However, once the Vyper code gets compiled, it produces two different signatures of the `newExperimentalVault` method - with and without the default parameter. Here we use the one which allows passing our own `releaseDelta`. We are doing this because we always need to deploy the same version of the vault contract. If new releases are added by the Yearn team, we are safe because we'll provide a right delta pointing to the same release we're using from the beginning.

yearn/contracts/CurveVoterProxyStrategy.sol
```diff
@@ -0,0 +1,511 @@
...
+    uint256 public keepCRV;
+
+    constructor(
+        address _vault,
+        address _tbtcCurvePoolDepositor,
+        address _tbtcCurvePoolGauge,
+        address _tbtcCurvePoolGaugeReward // optional, zero is valid value
+    ) public BaseStrategy(_vault) {
+        // Strategy settings.
+        minReportDelay = 6 hours;
```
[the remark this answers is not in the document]
lukasz-zimnoch: This is not a constant. See the definition in `BaseStrategy.sol`.

yearn/test/system/curve-voter-proxy-strategy.test.js
```diff
@@ -0,0 +1,343 @@
...
+
+    // Get a handle to the Synthetix Curve rewards contract used by the
+    // tBTC v2 Curve pool gauge.
+    synthetixCurveRewards = await ethers.getContractAt(
+      "ICurveRewards",
+      synthetixCurveRewardsAddress
+    )
+
+    // Allocate 100k.
+    const rewardsAllocation = to1ePrecision(100000, 18)
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/8/commits/401d9ddf0defac757ce7e6d0a2ca9fffdeef0cd1

yearn/contracts/CurveVoterProxyStrategy.sol
```diff
@@ -0,0 +1,511 @@
...
+        // Harvest the CRV rewards from the Curve pool's gauge.
+        IStrategyProxy(strategyProxy).harvest(tbtcCurvePoolGauge);
+
+        // Buy WBTC using harvested CRV tokens.
+        uint256 crvBalance = IERC20(crv).balanceOf(address(this));
+        if (crvBalance > 0) {
+            // Deposit a portion of CRV to the voter to gain CRV boost.
+            crvBalance = adjustCRV(crvBalance);
+
+            IERC20(crv).safeApprove(dex, 0);
```
dimpar, line 390: why do we need to approve `0` and right after `crvBalance` for the same `crv` token?

lukasz-zimnoch: Please see: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/fa64a1ced0b70ab89073d5d0b6e01b0778f7e7d6/contracts/token/ERC20/IERC20.sol#L42

dimpar: Right, but the link points to `approve`, not [safeApprove](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/fa64a1ced0b70ab89073d5d0b6e01b0778f7e7d6/contracts/token/ERC20/SafeERC20.sol#L37)

yearn/contracts/CurveVoterProxyStrategy.sol
```diff
@@ -0,0 +1,511 @@
...
+            IERC20(crv).safeApprove(dex, crvBalance);
+
+            address[] memory path = new address[](3);
+            path[0] = crv;
+            path[1] = weth;
+            path[2] = wbtc;
+
+            IUniswapV2Router(dex).swapExactTokensForTokens(
```
dimpar, line 398: I think we can pass just 0 instead of `uint256(0)`. The last param is `deadline` and the docs says "Unix timestamp after which the transaction will revert.". If the transaction is happening now, won't it be reverted immediately?

yearn/contracts/CurveVoterProxyStrategy.sol
```diff
@@ -0,0 +1,511 @@
...
+            IStrategyProxy(strategyProxy).claimRewards(
+                tbtcCurvePoolGauge,
+                tbtcCurvePoolGaugeReward
+            );
+
+            uint256 rewardBalance = IERC20(tbtcCurvePoolGaugeReward).balanceOf(
+                address(this)
+            );
+            if (rewardBalance > 0) {
+                IERC20(tbtcCurvePoolGaugeReward).safeApprove(dex, 0);
```
dimpar, line 418: Same concerns as above for the `safeApprove()` and `swapExactTokensForTokens() - deadline param`

yearn/contracts/CurveVoterProxyStrategy.sol
```diff
@@ -0,0 +1,511 @@
...
+
+        // Calculate the profit after obtaining new vault's underlying tokens.
+        profit = want.balanceOf(address(this)).sub(initialWantBalance);
+
+        // Check the profit and loss in the context of strategy debt.
+        uint256 totalAssets = estimatedTotalAssets();
+        uint256 totalDebt = vault.strategies(address(this)).totalDebt;
+        if (totalAssets < totalDebt) {
+            loss = totalDebt - totalAssets;
+            profit = 0;
```
dimpar, line 459: in this case we overwrite what was written on line 452. I guess it's no biggy, but maybe it will save us some gas if we do
```
if (totalAssets < totalDebt) {
//
profit = 0;
} else {
profit = want.balanceOf(address(this)).sub(initialWantBalance);
}

dimpar:
Finished reviewing

yearn/contracts/CurveVoterProxyStrategy.sol
lukasz-zimnoch: But `safeApprove` doesn't solve the problem: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/fa64a1ced0b70ab89073d5d0b6e01b0778f7e7d6/contracts/token/ERC20/SafeERC20.sol#L38

mhluongo: Indeed. Does `safeIncreaseApproval()` cover it?

yearn/contracts/CurveVoterProxyStrategy.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/8/commits/897d8bcddabf6791d3ccab496ee4a68024ac0c1b. However I think we are talking about `minAmount` parameter, not the `deadline`. The `deadline` is indeed the last one but we pass `now` here. `now` means that the swap is valid only within the block in which it was created.

yearn/contracts/CurveVoterProxyStrategy.sol
lukasz-zimnoch: Same answer as above :)

yearn/contracts/CurveVoterProxyStrategy.sol
lukasz-zimnoch: Yep, makes sense. Please see https://github.com/keep-network/tbtc-v2/pull/8/commits/0790f4daead4c0b6cfbaee8c405f4815d76851a7

yearn/contracts/CurveVoterProxyStrategy.sol
dimpar: I mentioned two points: one for `uint256(0)` which you resolved and another one was for `deadline`, which you answered. Thx, looks good!

yearn/contracts/CurveVoterProxyStrategy.sol
dimpar: It looks like [safeIncreaseAllowance()](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v3.4.0/contracts/token/ERC20/SafeERC20.sol#L48) is available in `3.4.0`

lukasz-zimnoch: Good catch! I browsed an older version by mistake 🤦 I'll use it.

lukasz-zimnoch: Here it is: https://github.com/keep-network/tbtc-v2/pull/8/commits/9419c237815fcd015d7c9b766e699ca8cb0bf02f

dimpar (approved):

dimpar:
One more comment @lukasz-zimnoch and I'm ready to merge. Do you think we can add `*Stub` suffix for interfaces in `/test` dir do differentiate easily in tests which one is a real contract and which one is a Stub?

lukasz-zimnoch:
> One more comment @lukasz-zimnoch and I'm ready to merge. Do you think we can add `*Stub` suffix for interfaces in `/test` dir to differentiate easily in tests which one is a real contract and which one is a Stub?

They are all real contracts. They are in `/test` because those interfaces are used by tests exlusively.

PR #10, opened 2021-07-13
Yearn: Convex strategy

lukasz-zimnoch:
Refs: #7
Depends on: #8 

**Introduction**

This pull request introduces another Yearn strategy named `ConvexStrategy`, meant to be used with the Yearn vault for tBTC v2. It follows all conventions introduced in the previous PR (#8) and is quite similar to the first `CurveVoterProxyStrategy` but uses [Convex Finance](https://www.convexfinance.com/) to generate yield.

**The `ConvexStrategy` strategy contract**

This strategy is used in the existing vault for tBTC v1 ([Curve tBTC pool yVault](https://yearn.watch/vault/0x23D3D0f1c697247d5e0a9efB37d8b0ED0C464f7f)). The current implementation for tBTC v2 is based on the following repositories, especially on the last one:

- [General Yearn strategy template](https://github.com/yearn/brownie-strategy-mix)
- [Convex implementation for tBTC v1 vault](https://github.com/orbxball/btc-curve-convex)

Basically, the current implementation for tBTC v2 vault uses the same logic as the strategy for [tBTC v1 vault](https://github.com/orbxball/btc-curve-convex/blob/master/archived/tbtc.sol). However, it tries to simplify the contract structure and attempts to achieve better readability and documentation. It also uses a newer version of the `yearn-vault` base contract. Last but not least, it introduces support for extra rewards provided by Convex reward pools sitting on top of Curve gauges which stake their deposits into the Synthetix staker rewards contract. To read a detailed description about how this strategy works in details, please see the [documentation of the `ConvexStrategy` contract](https://github.com/keep-network/tbtc-v2/blob/aacc3130ca68a22fb245016262723c0d2a9188c0/yearn/contracts/ConvexStrategy.sol#L99)

### DIFF
--- a/yearn/contracts/ConvexStrategy.sol
+++ b/yearn/contracts/ConvexStrategy.sol
@@ -549,15 +549,14 @@ contract ConvexStrategy is BaseStrategy {
             );
         }
 
-        // Calculate the profit after obtaining new vault's underlying tokens.
-        profit = want.balanceOf(address(this)).sub(initialWantBalance);
-
         // Check the profit and loss in the context of strategy debt.
         uint256 totalAssets = estimatedTotalAssets();
         uint256 totalDebt = vault.strategies(address(this)).totalDebt;
         if (totalAssets < totalDebt) {
             loss = totalDebt - totalAssets;
             profit = 0;
+        } else {
+            profit = want.balanceOf(address(this)).sub(initialWantBalance);
         }
 
         // Repay some vault debt if needed.
--- a/yearn/contracts/ConvexStrategy.sol
+++ b/yearn/contracts/ConvexStrategy.sol
@@ -1,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.6.12;
+pragma experimental ABIEncoderV2;
+
+import "@keep-network/yearn-vaults/contracts/BaseStrategy.sol";
+
+contract ConvexStrategy is BaseStrategy {
+}
--- a/yearn/contracts/ConvexStrategy.sol
+++ b/yearn/contracts/ConvexStrategy.sol
@@ -481,7 +481,7 @@ contract ConvexStrategy is BaseStrategy {
 
             IUniswapV2Router(uniswap).swapExactTokensForTokens(
                 crvBalance,
-                uint256(0),
+                0,
                 path,
                 address(this),
                 now
@@ -501,7 +501,7 @@ contract ConvexStrategy is BaseStrategy {
 
             IUniswapV2Router(sushiswap).swapExactTokensForTokens(
                 cvxBalance,
-                uint256(0),
+                0,
                 path,
                 address(this),
                 now
@@ -525,7 +525,7 @@ contract ConvexStrategy is BaseStrategy {
 
                 IUniswapV2Router(uniswap).swapExactTokensForTokens(
                     extraRewardBalance,
-                    uint256(0),
+                    0,
                     path,
                     address(this),
                     now
--- a/yearn/contracts/ConvexStrategy.sol
+++ b/yearn/contracts/ConvexStrategy.sol
@@ -17,12 +17,12 @@ interface IConvexBooster {
         external
         view
         returns (
-            address,
-            address,
-            address,
-            address,
-            address,
-            bool
+            address lpToken,
+            address convexDepositToken,
+            address curvePoolGauge,
+            address convexRewardPool,
+            address convexExtraRewardStash,
+            bool shutdown
         );
 
     function deposit(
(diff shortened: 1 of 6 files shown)


yearn/test/system/convex-strategy.test.js
```diff
@@ -0,0 +1,255 @@
...
+    await strategy.deployed()
+
+    // Add ConvexStrategy to the vault.
+    await vault.addStrategy(
+      strategy.address,
+      10000, // 100% debt ratio
+      0, // zero min debt per harvest
+      BigNumber.from(2).pow(256).sub(1), // infinite max debt per harvest
+      1000 // 10% performance fee
+    )
```
tomaszslabon, line 148: Perhaps constants?

yearn/contracts/ConvexStrategy.sol
```diff
@@ -0,0 +1,614 @@
...
+            IERC20(wbtc).safeApprove(tbtcCurvePoolDepositor, 0);
+            IERC20(wbtc).safeApprove(tbtcCurvePoolDepositor, wbtcBalance);
+            ICurvePool(tbtcCurvePoolDepositor).add_liquidity(
+                [0, 0, wbtcBalance, 0],
+                0
+            );
+        }
+
+        // Calculate the profit after obtaining new vault's underlying tokens.
+        profit = want.balanceOf(address(this)).sub(initialWantBalance);
```
dimpar, line 553: I think we can call `balanceOfWant()` instead.

yearn/test/system/convex-strategy.test.js
lukasz-zimnoch: Sure. I'll address this during the refactoring I'm planning to do at the end.

lukasz-zimnoch: Please see: https://github.com/keep-network/tbtc-v2/pull/10/commits/3006e09515af28c1d15e54df2095d6ddb460253e

yearn/contracts/ConvexStrategy.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/10/commits/7d20ec730332e0af308f37dfeeebf06c0db2b998

lukasz-zimnoch:
Ready for the full review! Keeping it draft until #8 lands on main.

yearn/contracts/ConvexStrategy.sol
```diff
@@ -0,0 +1,612 @@
...
+            IERC20(crv).safeIncreaseAllowance(uniswap, crvBalance);
+
+            address[] memory path = new address[](3);
+            path[0] = crv;
+            path[1] = weth;
+            path[2] = wbtc;
+
+            IUniswapV2Router(uniswap).swapExactTokensForTokens(
+                crvBalance,
+                uint256(0),
```
dimpar, line 484: uint256(0) -> 0. We have a couple of occurances of this contract.

yearn/contracts/ConvexStrategy.sol
```diff
@@ -0,0 +1,612 @@
...
+    ///         such as transferring any reserve or LP tokens, CDPs, or other
+    ///         tokens or stores of value. This strategy implements the
+    ///         aforementioned behavior by withdrawing all vault's underlying
+    ///         tokens from the Convex reward pool, claiming all rewards
+    ///         accumulated so far, and transferring those rewards to the new
+    ///         strategy.
+    /// @param newStrategy Address of the new strategy meant to replace the
+    ///        current one.
+    function prepareMigration(address newStrategy) internal override {
+        // Just withdraw the vault's underlying token from the Convex reward pool.
```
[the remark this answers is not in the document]
lukasz-zimnoch: Nope, it is up to date. We have two kinds of tokens here:
- the vault's underlying token, i.e. the Curve pool LP token
- reward tokens distributed by the Convex reward pool, i.e. CRV, CVX, and extra reward token (KEEP) 

The `withdrawAllAndUnwrap(true)` call withdraws the vault's underlying token and claims (`true` parameter does this) all reward tokens. Now, the vault's underlying token should not be transferred manually in this method because this is done by the parent `migrate` method which calls `prepareMigration`. However, all other tokens like our reward tokens must be transferred manually and this is what we're doing here. 

I tried to describe that in this comment but maybe it is still not clear enough. If you have a suggestion on how we can improve it, I'd be happy to do it.

yearn/contracts/ConvexStrategy.sol
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/10/commits/23b2b4753aea07ae8f6e17c0bdebe75974a7d41b

yearn/contracts/ConvexStrategy.sol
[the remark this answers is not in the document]
dimpar: Ok, I see now, maybe I would say instead of "There is no need to transfer those tokens to the new strategy" -> "There is no need to transfer underlying tokens to the new strategy... " I thought by "those" it meant rewards tokens as well.

dimpar (approved):

PR #13, opened 2021-07-20
Yearn: Saddle strategy

tomaszslabon:
Refs: #7 
This PR adds another Yearn strategy named `SaddleStrategy`, meant to be used with the Yearn vault for tBTC v2.
It follows the conventions used in another strategies `CurveVoterProxyStrategy` and `ConvexStrategy`.
It uses [LPRewardsTBTCSaddle](https://etherscan.io/address/0x78aa83BD6c9DE5De0A2231366900AB060A482edd) to generate yield. 

TODO:
- system tests - check what else may be checked (e.g. migration)?
  
  
Note: it is necessary to `setGated` to `false` for the Saddle LP rewards contract. Otherwise only externally-owned accounts can call [stake](https://github.com/keep-network/keep-ecdsa/blob/96cf2541764d019237906c4e0fa7051d15f6b6cd/solidity/contracts/LPRewards.sol#L253) on it.


yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+    // Address of the WETH token contract.
+    address public constant weth = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+    // Address of the WBTC token contract.
+    address public constant wbtc = 0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599;
+    // Address of the Uniswap V2 router contract.
+    address public constant uniswap =
+        0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D;
+    // Address of the Saddle Swap contract, taken from
+    // https://docs.saddle.finance/contracts
+    address public constant swap = 0x4f6A43Ad7cba042606dECaCA730d4CE0A57ac62e;
```
lukasz-zimnoch, line 100: What do you think about naming it `tbtcSaddlePoolSwap` and make it a variable set using the constructor? Regarding the name, `swap` seems too broad and it took me a while to figure out what it refers to. Regarding making it a variable, we don't know yet how tBTC v2 will be added to Saddle. It may be added to the existing BTC pool but it may require a new pool (thus a new swap contract) as well. Making it settable via the constructor is safer.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+    // Address of the tBTC v2 Saddle reward pool.
+    address public lpRewardsTbtcV2Saddle;
```
lukasz-zimnoch, line 102: What do you say about naming it `tbtcSaddleLPRewards`? This way we follow the convention for tBTC-related settings from other strategies. I'm aware the original contract is named this way but I think that change will make it easier to understand.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+interface IERC20Metadata {
+    function symbol() external view returns (string memory);
+}
+
+/// @notice Interface for the Saddle Swap contract.
+/// @dev This is an interface with just a few function signatures of the pool.
+///      For more info and function description please see:
+///      https://github.com/saddle-finance/saddle-contract/blob/master/contracts/Swap.sol
+///      https://github.com/saddle-finance/saddle-contract/blob/master/contracts/guarded/SwapGuarded.sol
+interface ISaddleSwap {
```
lukasz-zimnoch, line 22: `ISaddlePoolSwap`? A little bit longer but clearly explains its purpose.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+    function estimatedTotalAssets() public view override returns (uint256) {
+        return balanceOfWant().add(balanceOfPool());
+    }
+
+    /// @notice This method is defined in the BaseStrategy contract and is meant
+    ///         to provide an accurate conversion from amtInWei (denominated in wei)
+    ///         to want token (using the native decimal characteristics of want token).
+    /// @param amtInWei The amount (in wei/1e-18 ETH) to convert to want tokens.
+    /// @return The amount in want tokens.
+    function ethToWant(uint256 amtInWei)
```
lukasz-zimnoch, line 157: Nitpick: Let's move it at the bottom of the contract as we do in other strategies. This method is not strictly related to strategy cashflow.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+
+    /// @notice Withdraws a portion of the vault's underlying token from
+    ///         the Saddle reward pool.
+    /// @param amount Amount to withdraw.
+    /// @return Amount withdrawn.
+    function withdrawSome(uint256 amount) internal returns (uint256) {
+        amount = Math.min(amount, balanceOfPool());
+        uint256 initialWantBalance = balanceOfWant();
+        // Withdraw some vault's underlying tokens but do not claim the rewards
+        // accumulated so far.
```
lukasz-zimnoch, line 196: Outdated comment.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+    /// @notice This method is defined in the BaseStrategy contract and is meant
+    ///         to liquidate up to amountNeeded of want token of this strategy's
+    ///         positions, irregardless of slippage. Any excess will be
+    ///         re-invested with adjustPosition. This function should return
+    ///         the amount of want tokens made available by the liquidation.
+    ///         If there is a difference between them, loss indicates whether
+    ///         the difference is due to a realized loss, or if there is some
+    ///         other situation at play (e.g. locked funds). This function is
+    ///         used during emergency exit instead of prepareReturn to
+    ///         liquidate all of the strategy's positions back to the vault.
```
lukasz-zimnoch, line 210: I think the part 
```
This function is used during emergency exit instead of prepareReturn to liquidate all of the strategy's positions back to the vault.
```
should refer to `liquidateAllPositions` (see `BaseStrategy.sol` and its `@dev` doc section in this contract) so we can remove this part from here. This mistake was made in existing strategies because we switched Yearn version during the development. The older version didn't contain `liquidateAllPositions` yet and its work was done by `liquidatePosition`. Can we also change that for `CurveVoterProxyStrategy` and `ConvexStrategy` by the way? 🙏

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+    ///         such as transferring any reserve or LP tokens, CDPs, or other
+    ///         tokens or stores of value. This strategy implements the
+    ///         aforementioned behavior by withdrawing all vault's underlying
+    ///         tokens from the tBTC v2 Saddle reward pool.
+    /// @param newStrategy Address of the new strategy meant to replace the
+    ///        current one.
+    function prepareMigration(address newStrategy) internal override {
+        // Withdraw all the LP tokens from the tBTC v2 Saddle reward pool.
+        // These tokens will be transferred to the new strategy in the `migrate`
+        // function.
```
lukasz-zimnoch, line 272: Maybe: 
```
// Just withdraw the vault's underlying token from the tBTC v2 Saddle reward pool.
// There is no need to transfer those tokens to the new strategy
// right here as this is done in the BaseStrategy's migrate() method.
```
???

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+        ILPRewardsTBTCV2Saddle(lpRewardsTbtcV2Saddle).withdraw(balanceOfPool());
+        // Get all the earned KEEP tokens and transfer them to the new strategy.
+        ILPRewardsTBTCV2Saddle(lpRewardsTbtcV2Saddle).getReward();
```
lukasz-zimnoch, line 275: While writing system tests, let's make sure the reward is actually obtained. Maybe we should `getReward` before `withdraw`?

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+        IERC20(keep).safeTransfer(
+            newStrategy,
+            IERC20(keep).balanceOf(address(this))
+        );
+        //TODO: Check if both the LP Token and KEEP token should be migrated to
+        // the new strategy, or just the LP token.
```
lukasz-zimnoch, line 281: If we can migrate both without hassle, we should do it I think.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    /// @notice This method is defined in the BaseStrategy contract and is meant
+    ///         to perform any strategy unwinding or other calls necessary to
+    ///         capture the free return this strategy has generated since the
+    ///         last time its core position(s) were adjusted. Examples include
+    ///         unwrapping extra rewards. This call is only used during normal
+    ///         operation of a strategy, and should be optimized to minimize
+    ///         losses as much as possible.
```
lukasz-zimnoch, line 290: Please remember about describing how this method makes a profit.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+    function prepareReturn(uint256 debtOutstanding)
+        internal
+        override
+        returns (
+            uint256 profit,
+            uint256 loss,
+            uint256 debtPayment
+        )
+    {
+        /* solhint-disable not-rely-on-time */
```
lukasz-zimnoch, line 311: I think we don't need this. `not-rely-on-time` is considered as warning in `.solhint.json`.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+        }
+
+        // Deposit acquired WBTC to the Saddle pool to gain additional
+        // vault's underlying tokens.
+        uint256 wbtcBalance = IERC20(wbtc).balanceOf(address(this));
+        if (wbtcBalance > 0) {
+            IERC20(wbtc).safeIncreaseAllowance(swap, wbtcBalance);
+
+            uint256[] memory amounts = new uint256[](4);
+            amounts[1] = wbtcBalance;
```
lukasz-zimnoch, line 345: Can we replace it with a direct function parameter like:
```
[0, wbtcBalance, 0, 0]
```
passed to the `addLiquidity`? This way it looks more concise.

lukasz-zimnoch (changes requested):

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,395 @@
...
+        uint256 deadline,
+        bytes32[] calldata merkleProof
+    ) external returns (uint256);
+}
+
+/// @notice Interface for the LPRewardsTBTCSaddle reward pool contract.
+/// @dev This is an interface with just a few functions. For more info and
+///      function description please see:
+///      https://github.com/keep-network/keep-ecdsa/blob/main/solidity/contracts/LPRewards.sol
+interface ILPRewardsTBTCV2Saddle {
```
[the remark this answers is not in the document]
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/13/commits/227ed9911ad590b61480bb840375083958d2233b.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/13/commits/227ed9911ad590b61480bb840375083958d2233b.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/13/commits/227ed9911ad590b61480bb840375083958d2233b.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/13/commits/227ed9911ad590b61480bb840375083958d2233b.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/13/commits/227ed9911ad590b61480bb840375083958d2233b.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Removed in https://github.com/keep-network/tbtc-v2/pull/13/commits/5267c11fefbf0a07b3e4b19a6af70b943e735bf6.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: I added a short description in https://github.com/keep-network/tbtc-v2/pull/13/commits/5267c11fefbf0a07b3e4b19a6af70b943e735bf6.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: I removed the comment https://github.com/keep-network/tbtc-v2/pull/13/commits/5267c11fefbf0a07b3e4b19a6af70b943e735bf6. The code seems to be rather self-explanatory.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Unfortunately not:
`[0, wbtcBalance, 0, 0]` has type `uint256[4] memory`, but the function needs `uint256[] memory`.
It can only be done with `new`.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: I don't think it is a problem we call `getReward` after `withdraw`.
The `LPRewardsTBTCSaddle` contract even has an [exit](https://github.com/keep-network/keep-ecdsa/blob/96cf2541764d019237906c4e0fa7051d15f6b6cd/solidity/contracts/LPRewards.sol#L143) function that calls these two functions in the same order.
But I will check this behaviour in a system test.

lukasz-zimnoch: Ok good!

yearn/contracts/SaddleStrategy.sol
lukasz-zimnoch: Ok, got it!

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,393 @@
...
+interface ISaddlePoolSwap {
+    function addLiquidity(
+        uint256[] calldata amounts,
+        uint256 minToMint,
+        uint256 deadline,
+        bytes32[] calldata merkleProof
+    ) external returns (uint256);
+}
+
+/// @notice Interface for the LPRewardsTBTCSaddle reward pool contract.
```
lukasz-zimnoch, line 31: `Interface for the LPRewardsTBTCSaddle reward pool contract.` -> `Interface for the LPRewards contracts.` ?

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,393 @@
...
+import "@openzeppelin/contracts/math/Math.sol";
+
+/// @notice Interface for the optional metadata functions from the ERC20
+///         standard.
+interface IERC20Metadata {
+    function symbol() external view returns (string memory);
+}
+
+/// @notice Interface for the Saddle Swap contract.
+/// @dev This is an interface with just a few function signatures of the pool.
```
lukasz-zimnoch, line 18: `...of the pool.` -> `...of the pool swap contract.`?

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,393 @@
...
+    address public constant keep = 0x85Eee30c52B0b379b046Fb0F85F4f3Dc3009aFEC;
+    // Address of the WETH token contract.
+    address public constant weth = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+    // Address of the WBTC token contract.
+    address public constant wbtc = 0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599;
+    // Address of the Uniswap V2 router contract.
+    address public constant uniswap =
+        0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D;
+
+    // Address of the Saddle Swap contract, see https://docs.saddle.finance/contracts
```
lukasz-zimnoch, line 99: `...of the Saddle Swap contract..` -> `...of the tBTC v2 Saddle pool swap contract...`?

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,393 @@
...
+            amounts[1] = wbtcBalance;
+
+            bytes32[] memory merkleProof = new bytes32[](1);
+
+            //TODO: Check what the values of minToMint, deadline and merkleProof
+            // should be
+            ISaddlePoolSwap(tbtcSaddlePoolSwap).addLiquidity(
+                amounts,
+                0,
+                uint256(-1),
```
lukasz-zimnoch, line 336: Why `uint256(-1)` here?

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/13/commits/5f2ce7710c6f13266ff93f1403dab0131da10e68.

lukasz-zimnoch: Perfect! Thanks.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/13/commits/5267c11fefbf0a07b3e4b19a6af70b943e735bf6.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: The [addLiquidity](https://github.com/saddle-finance/saddle-contract/blob/2b6c3c2a25b1b78d62cb048cbbc6bf0700e2f711/contracts/guarded/SwapGuarded.sol#L435) function has a parameter called `deadline` which defines `latest timestamp to accept this transaction`. 

I decided to use `uint256(-1)` (which will wrap to the max value an `uint256` can take) for now, so that effectively there is no deadline (the authors of the Saddle swap contract (do that in tests)[https://github.com/saddle-finance/saddle-contract/blob/2b6c3c2a25b1b78d62cb048cbbc6bf0700e2f711/contracts/helper/test/TestSwapReturnValues.sol#L62] as well).

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/13/commits/0449be392c7e1f92acb63f46b6526b1901c30467.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/13/commits/0449be392c7e1f92acb63f46b6526b1901c30467.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/13/commits/0449be392c7e1f92acb63f46b6526b1901c30467.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Do you know if there is any risk the new strategy may be of different "type" - e.g. rewarding with a different token then KEEP?

lukasz-zimnoch: Good point. However, I think in that case it will be considered as a separate new strategy rather than a new version of this one.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,396 @@
...
+        // Deposit acquired WBTC to the Saddle pool to gain additional
+        // vault's underlying tokens.
+        uint256 wbtcBalance = IERC20(wbtc).balanceOf(address(this));
+        if (wbtcBalance > 0) {
+            IERC20(wbtc).safeIncreaseAllowance(tbtcSaddlePoolSwap, wbtcBalance);
+
+            uint256[] memory amounts = new uint256[](4);
+            amounts[1] = wbtcBalance;
+
+            bytes32[] memory merkleProof = new bytes32[](1);
```
lukasz-zimnoch, line 332: Don't we need any value here?

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,396 @@
...
+    /// @param amtInWei The amount (in wei/1e-18 ETH) to convert to want tokens.
+    /// @return The amount in want tokens.
+    function ethToWant(uint256 amtInWei)
+        public
+        view
+        virtual
+        override
+        returns (uint256)
+    {
+        // TODO: Create an accurate price oracle.
```
lukasz-zimnoch, line 393: We can implement this just like it was proposed in https://github.com/keep-network/tbtc-v2/pull/15 using Saddle's `calculateTokenAmount` method.

yearn/test/system/constants.js
```diff
@@ -42,4 +42,14 @@ module.exports.tbtc = {
...
   synthetixCurveRewardsOwnerAddress:
     "0xb3726e69da808a689f2607939a2d9e958724fc2a",
+  // Address of the Saddle Pool Swap
+  saddlePoolSwapAddress: "0x4f6A43Ad7cba042606dECaCA730d4CE0A57ac62e",
+  // Address of the tBTC v2 Saddle pool LP token.
+  saddlePoolLPTokenAddress: "0xC28DF698475dEC994BE00C9C9D8658A548e6304F",
+  // Example address which holds an amount of the tBTC v2 Saddle pool LP token.
+  saddlePoolLPTokenHolderAddress: "0x1f032a27b369299c73b331c9fc1e80978db45b15",
+  // Address of the tBTC v2 Saddle pool (LPRewardsTBTCSaddle),
+  saddlePoolAddress: "0x78aa83BD6c9DE5De0A2231366900AB060A482edd",
```
lukasz-zimnoch, line 52: It should be rather:
```suggestion
  // Address of the tBTC v2 Saddle reward pool. (LPRewardsTBTCSaddle),
  saddleLPRewards: "0x78aa83BD6c9DE5De0A2231366900AB060A482edd",
```

yearn/test/system/constants.js
```diff
@@ -42,4 +42,14 @@ module.exports.tbtc = {
...
+  // Address of the owner of the tBTC v2 Saddle pool
+  saddlePoolOwner: "0xb3726e69da808a689f2607939a2d9e958724fc2a",
```
lukasz-zimnoch, line 54:
```suggestion
  // Address of the tBTC v2 Saddle reward pool owner.
  saddleLPRewardsOwner: "0xb3726e69da808a689f2607939a2d9e958724fc2a",
```

yearn/test/system/fixtures.js
```diff
@@ -21,3 +21,24 @@ module.exports.curveStrategyFixture = {
   // The fee the strategist will receive based on this Vault's performance.
   strategyPerformanceFee: 1000, // 10%
 }
+
+module.exports.saddleStrategyFixture = {
+  // Name of the vault for tBTCv2 Curve pool.
+  vaultName: "Curve tBTCv2 Pool yVault",
+  // Symbol of the vault for tBTCv2 Curve pool.
+  vaultSymbol: "yvCurve-tBTCv2",
+  // Total deposit limit of the vault for tBTCv2 Curve pool.
```
lukasz-zimnoch, line 30:
```suggestion
  // Name of the vault for tBTCv2 Saddle pool.
  vaultName: "Saddle tBTCv2 Pool yVault",
  // Symbol of the vault for tBTCv2 Saddle pool.
  vaultSymbol: "yvSaddle-tBTCv2",
  // Total deposit limit of the vault for tBTCv2 Saddle pool.
```

yearn/test/system/fixtures.js
```diff
@@ -21,3 +21,24 @@ module.exports.curveStrategyFixture = {
...
+  vaultDepositLimit: to1ePrecision(300, 15), // 0.3
+  // Amount of the deposit made by the depositor.
+  vaultDepositAmount: to1ePrecision(300, 15), // 0.3
+  // Amount of Synthetix staking rewards which should be allocated.
+  synthetixRewardsAllocation: to1e18(100000), // 100k
```
lukasz-zimnoch, line 35: Not used in Saddle strategy.

yearn/test/system/saddle-strategy.test.js
```diff
@@ -0,0 +1,195 @@
...
+
+    // Setup roles.
+    vaultGovernance = await ethers.getSigner(0)
+
+    vaultDepositor = await impersonateAccount(
+      tbtc.saddlePoolLPTokenHolderAddress,
+      vaultGovernance
+    )
+
+    saddlePoolGovernance = await impersonateAccount(
```
lukasz-zimnoch, line 35: When we change `tbtc.saddlePoolOwner` this should be `saddleLPRewardsGovernance`. Don't confuse Saddle pool with Saddle LP rewards contract.

yearn/test/system/saddle-strategy.test.js
```diff
@@ -0,0 +1,195 @@
...
+      vaultGovernance
+    )
+
+    // Get tBTC v2 Saddle pool handle and set `gated` to false to allow
+    // non-externally-owned accounts to perform staking
+    saddlePool = await ethers.getContractAt(
+      "ILPRewards",
+      tbtc.saddlePoolAddress
+    )
+    await saddlePool.connect(saddlePoolGovernance).setGated(false)
```
lukasz-zimnoch, line 46: This is not Saddle pool. This is Saddle LP rewards.

yearn/test/system/constants.js
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/13/commits/045e79a6b12147241e18295132a5992350505be2.

yearn/test/system/constants.js
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/commit/045e79a6b12147241e18295132a5992350505be2.

yearn/test/system/fixtures.js
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/commit/045e79a6b12147241e18295132a5992350505be2.

yearn/test/system/fixtures.js
tomaszslabon: Removed in https://github.com/keep-network/tbtc-v2/commit/045e79a6b12147241e18295132a5992350505be2.

yearn/test/system/saddle-strategy.test.js
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/commit/045e79a6b12147241e18295132a5992350505be2.

yearn/test/system/saddle-strategy.test.js
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/commit/045e79a6b12147241e18295132a5992350505be2.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,396 @@
...
+        uint256 deadline,
+        bytes32[] calldata merkleProof
+    ) external returns (uint256);
+}
+
+/// @notice Interface for the LPRewards contracts.
+/// @dev This is an interface with just a few functions. For more info and
+///      function description please see:
+///      https://github.com/keep-network/keep-ecdsa/blob/main/solidity/contracts/LPRewards.sol
+interface ILPRewards {
```
dimpar, line 35: What are your thoughts about adding dependency `"@keep-network/keep-ecdsa": "1.6.0"` in `package.json` for `LPRewards` instead? We have an npm package and we'll be able to import `LPRewards` instead of defining this interface here.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,396 @@
...
+    address public constant wbtc = 0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599;
+    // Address of the Uniswap V2 router contract.
+    address public constant uniswap =
+        0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D;
+
+    // Address of the tBTC v2 Saddle pool swap contract, see
+    // https://docs.saddle.finance/contracts
+    address public tbtcSaddlePoolSwap;
+    // Address of the tBTC v2 Saddle reward pool.
+    address public tbtcSaddleLPRewards;
```
dimpar, line 105: Can we add 'Pool' at the end, ie `tbtcSaddleLPRewardsPool`?

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,396 @@
...
+
+    /// @return Balance of the vault's underlying token under management.
+    function balanceOfWant() public view returns (uint256) {
+        return want.balanceOf(address(this));
+    }
+
+    /// @return Balance of the vault's underlying token staked into the tBTC v2
+    ///         Saddle reward pool.
+    function balanceOfPool() public view returns (uint256) {
+        return ILPRewards(tbtcSaddleLPRewards).balanceOf(address(this));
```
dimpar, line 145: shouldn't it be `want.balanceOf(ILPRewards(tbtcSaddleLPRewards));`?

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,396 @@
...
+    function liquidateAllPositions()
+        internal
+        override
+        returns (uint256 amountFreed)
+    {
+        ILPRewards(tbtcSaddleLPRewards).withdraw(balanceOfPool());
+
+        // Yearn docs doesn't specify clear enough what exactly should be
+        // returned here. It may be either the total balance after
+        // withdrawAllAndUnwrap or just the amount withdrawn. Currently opting
```
dimpar, line 234: - [... after `withdraw` or just ...] 
- And since we're returning the entire balance and not only Pool's rewards portion, then we should probably reword @return param to "Total balance of want tokens". "Amount that got freed" denotes what was freed from the `tbtcSaddleLPRewards` pool and not the entire balance.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,396 @@
...
+        if (wbtcBalance > 0) {
+            IERC20(wbtc).safeIncreaseAllowance(tbtcSaddlePoolSwap, wbtcBalance);
+
+            uint256[] memory amounts = new uint256[](4);
+            amounts[1] = wbtcBalance;
+
+            bytes32[] memory merkleProof = new bytes32[](1);
+
+            //TODO: Check what the values of minToMint, deadline and merkleProof
+            // should be
```
dimpar, line 335: According to [their docs](https://github.com/saddle-finance/saddle-contract/blob/master/contracts/guarded/SwapGuarded.sol#L431) merkleProof is ignored when the contract is not guarded. And `isGuarded()` returns `false`, so I think we should not worry so much about this `merkleProof` param.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,396 @@
...
+/// @dev Implementation is based on:
+///      - General Yearn strategy template
+///        https://github.com/yearn/brownie-strategy-mix
+contract SaddleStrategy is BaseStrategy {
+    using SafeERC20 for IERC20;
+    using Address for address;
+    using SafeMath for uint256;
+
+    // Address of the KEEP token contract.
+    address public constant keep = 0x85Eee30c52B0b379b046Fb0F85F4f3Dc3009aFEC;
```
dimpar, line 92: I wonder if we shouldn't make all the token names capital letters across all the contracts. I think this is a standard. cc @lukasz-zimnoch

yearn/contracts/SaddleStrategy.sol
lukasz-zimnoch: `ILPRewards` is an interface for all `LPRewards`-like contracts. In this case, we are talking with `LPRewardsTBTCSaddle` using this interface. The `LPRewardsTBTCSaddle` is for tBTC v1 Saddle pool. For tBTC v2, we will have (probably) a new pool and a new `LPRewards` contract. Because of that, I think it is better to keep a small interface instead of opting for a package dependency.

yearn/contracts/SaddleStrategy.sol
lukasz-zimnoch: Up to you guys. Both versions work for me.

yearn/contracts/SaddleStrategy.sol
lukasz-zimnoch: Maybe leave it as is and just modify the comment to `Address of the tBTC v2 Saddle LP rewards contract`? I'm afraid adding the `pool` suffix may cause confusion with the Saddle pool.

yearn/contracts/SaddleStrategy.sol
dimpar: >  For tBTC v2, we will have (probably) a new pool and a new LPRewards contract

Okay, I agree that in this case it's better to stick with an interface.  Maybe I'd add a TODO to verify functions for a new `LPRewards`, cause I take that now we "assume" that this is what it's going to look like.

lukasz-zimnoch: Most likely, it will be just a new instance of the same contract but yeah, we could add a TODO just in case.

yearn/contracts/SaddleStrategy.sol
dimpar: Ok, let's leave it as-is and just change a comment if you agree with that Tomasz.

tomaszslabon: I agree with that and changed the comment in https://github.com/keep-network/tbtc-v2/pull/13/commits/802c155d8e6b78bfe17a77572814367052f3b4f7.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: I added a TODO in: https://github.com/keep-network/tbtc-v2/pull/13/commits/802c155d8e6b78bfe17a77572814367052f3b4f7:

yearn/contracts/SaddleStrategy.sol
tomaszslabon: I agree, removed the TODO in https://github.com/keep-network/tbtc-v2/commit/802c155d8e6b78bfe17a77572814367052f3b4f7

yearn/contracts/SaddleStrategy.sol
tomaszslabon: @dimpar:
Should we make all the letters capital, like:
`keep` -> `KEEP`
`weth` - > `WETH` ?

Should the contract addresses stay unchanged:
`tbtcSaddlePoolSwap`
`tbtcSaddleLPRewards`
`uniswap`
?

dimpar: yes, only token names change to capital, like keep->KEEP, wbtc->WBTC etc..

pdyraga: I believe that in Solidity code we usually do `keepToken`, `wbtcToken`.

dimpar: ok, I'm fine with the version suggested by Piotr

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Done at https://github.com/keep-network/tbtc-v2/pull/13/commits/1d94f9943e965f45638341f1f2d941636caf3d53.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/13/commits/90c065d3d25770743b95530c90d9289f89522f74.

yearn/contracts/SaddleStrategy.sol
tomaszslabon: According to [Saddle-finances contract](https://github.com/saddle-finance/saddle-contract/blob/2b6c3c2a25b1b78d62cb048cbbc6bf0700e2f711/contracts/guarded/SwapGuarded.sol#L429) this parameter is ignored when the pool is not guarded:
"""
@param merkleProof data generated when constructing the allowlist merkle tree. Users can
get this data off chain. Even if the address is in the allowlist, users must include
a valid proof for this call to succeed. If the pool is no longer in the guarded release phase,
this parameter is ignored.
"""

And their [contract](https://etherscan.io/address/0x4f6A43Ad7cba042606dECaCA730d4CE0A57ac62e) is indeed not guarded, so I think we can ignore this parameter, I simply used:
'''"
new bytes32[](0) // ignored
"""

yearn/contracts/SaddleStrategy.sol
dimpar: withdrawAllAndUnwrap -> withdraw

tomaszslabon: Fixed in https://github.com/keep-network/tbtc-v2/pull/13/commits/34980e454124e22d92e144195e0a2062c4eb1b84.

dimpar:
The rest looks good to me.

yearn/contracts/SaddleStrategy.sol
```diff
@@ -0,0 +1,439 @@
...
+        // vault's underlying tokens.
+        uint256 wbtcBalance = IERC20(wbtcToken).balanceOf(address(this));
+        if (wbtcBalance > 0) {
+            IERC20(wbtcToken).safeIncreaseAllowance(
+                tbtcSaddlePoolSwap,
+                wbtcBalance
+            );
+
+            uint256[] memory amounts = new uint256[](4);
+            amounts[1] = wbtcBalance;
```
lukasz-zimnoch, line 345: Can we add a TODO here, just like we have in other contracts and in `ethToWant` methods:
```
// TODO: When the new curve pool with tBTC v2 is deployed, verify that
//       the index of wBTC in the array is correct.
```
?

tomaszslabon: Sure, done in https://github.com/keep-network/tbtc-v2/pull/13/commits/04cf745a5c63b0cdffc59de7c4a839b82191d66f.

lukasz-zimnoch:
Looks good! If there are no more things to do, I think we can make it ready and merge.

lukasz-zimnoch (approved):

PR #14, opened 2021-07-20
Yearn: Deployment script

lukasz-zimnoch:
Refs: #7 
Depends on: #10 

This pull request introduces deployment scripts for all strategies contracts. It also brings a README file that describes the `yearn` submodule and how to work with it. 

Worth flagging that all external contracts addresses are passed to the strategies contracts via Hardhat externals. All those external contracts are established on mainnet, as parts of Yearn and Curve ecosystems, and are not delivered via npm or another package registry. Local deployment would require deploying those externals manually which is rather impractical, especially given the fact we use system tests leveraging Hardhat mainnet forking feature.


yearn/README.adoc
```diff
@@ -0,0 +1,181 @@
...
+If contracts haven't been built yet or changes occurred, this task will build
+the contracts before running the deployment script. This command produces
+an `export.json` file containing contract deployment info.
+
+Please note that all strategies contracts make a call to their vault contract
+upon initialization. That means the deployment will fail if the vault address
+passed in the strategy constructor does not point to a contract compatible
+with the Yearn vault API.
+
+[[deployment-parametrization]]
```
dimpar, line 139: Was there a specific reason on why using addresses of contracts as env params? Do you think we can follow the path of deployment as we have in [coverage-pool](https://github.com/keep-network/coverage-pools/blob/main/README.adoc#predefined-artifacts) and having "./external/testnet" & "./external/main" pattern for the addresses? I think the current solution is fine, but I think we want to minimize a human factor which might result in a typo, copy&paste error or chosing a wrong contract during deployment.
If you'd agree, then the deployment scripts will need to be adjusted as well.
If there are no contracts available for "./external/testnet" then I guess we can use only "./external/main" and replace the addresses once tBTCv2 & new pools are deployed to mainnet.

dimpar:
I haven't gone through all the files, since they might change based on the decision regarding the comments above. But I think it's also worth to take a look at the [main network PR](https://github.com/keep-network/coverage-pools/pull/165/files) in the cov pool. They added [dependency](https://github.com/keep-network/hardhat-local-networks-config) for hiding sensitive data as well.

yearn/README.adoc
lukasz-zimnoch: The main cause was the fact that most of the external dependencies are not shipped as npm packages and we don't know their addresses until tBTC v2 is deployed. However, your proposition sounds good and we can use `Predefined artifacts` with zero addresses for now and leave TODOs to fill proper addresses once we know them.

yearn/README.adoc
```diff
@@ -0,0 +1,181 @@
...
+=== Deploy contracts
+
+To deploy all contracts on the given network, please run:
+```
+yarn deploy --network <network>
+```
+with required environment variables listed in the <<deployment-parametrization>>
+section.
+
+To deploy a single contract, use the same command with the `--tags` option
```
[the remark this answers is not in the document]
lukasz-zimnoch: Done as part of https://github.com/keep-network/tbtc-v2/pull/14/commits/0e577c4a3dcf2c3cb8cc20520e537b39d3c8d18e

yearn/README.adoc
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/14/commits/0e577c4a3dcf2c3cb8cc20520e537b39d3c8d18e

yearn/.eslintrc
```diff
@@ -1,15 +1,15 @@
...
-      "sourceType": "module"
-    },
-    "env": {
-      "es6": true,
-      "mocha": true
-    },
-    "rules": {
-      "new-cap": "off"
-    }
+  "extends": ["eslint-config-keep", "prettier"],
```
nkuba, line 2: `prettier` is already included in `eslint-config-keep`.
```suggestion
  "extends": ["eslint-config-keep"],
```

yearn/deploy/00_deploy_cvp_strategy.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  const TBTCCurvePoolGauge = await deployments.getOrNull("TBTCCurvePoolGauge")
+  if (!TBTCCurvePoolGauge || !helpers.address.isValid(TBTCCurvePoolGauge.address)) {
+    throw new Error("Address of TBTCCurvePoolGauge must be set")
+  }
+
+  const TBTCCurvePoolGaugeReward = await deployments.getOrNull("TBTCCurvePoolGaugeReward")
+  if (!TBTCCurvePoolGaugeReward || !helpers.address.isValid(TBTCCurvePoolGaugeReward.address)) {
+    TBTCCurvePoolGaugeReward.address = "0x0000000000000000000000000000000000000000"
```
nkuba, line 26: Why set it to zero instead of throwing an error?

yearn/deploy/00_deploy_cvp_strategy.ts
```diff
@@ -0,0 +1,49 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types"
+import { DeployFunction } from "hardhat-deploy/types"
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+  const { deployments, getNamedAccounts, helpers } = hre
+  const { deploy, log } = deployments
+  const { deployer } = await getNamedAccounts()
+
+  const TBTCCurveVault = await deployments.getOrNull("TBTCCurveVault")
+  if (!TBTCCurveVault || !helpers.address.isValid(TBTCCurveVault.address)) {
```
nkuba, line 10: What if we use `get` instead of `getOrNull`? `get` should throw an error if the deployment is not found.

```suggestion
  const TBTCCurveVault = await deployments.get("TBTCCurveVault")
  if (!helpers.address.isValid(TBTCCurveVault.address)) {
```

yearn/deploy/00_deploy_cvp_strategy.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const TBTCCurvePoolGaugeReward = await deployments.getOrNull("TBTCCurvePoolGaugeReward")
+  if (!TBTCCurvePoolGaugeReward || !helpers.address.isValid(TBTCCurvePoolGaugeReward.address)) {
+    TBTCCurvePoolGaugeReward.address = "0x0000000000000000000000000000000000000000"
+  }
+
+  log(`tbtcCurveVault: ${TBTCCurveVault.address}`)
+  log(`tbtcCurvePoolDepositor: ${TBTCCurvePoolDepositor.address}`)
+  log(`tbtcCurvePoolGauge: ${TBTCCurvePoolGauge.address}`)
+  log(`tbtcCurvePoolGaugeReward: ${TBTCCurvePoolGaugeReward.address}`)
```
nkuba, line 32:
```suggestion
  log(`TBTCCurveVault: ${TBTCCurveVault.address}`)
  log(`TBTCCurvePoolDepositor: ${TBTCCurvePoolDepositor.address}`)
  log(`TBTCCurvePoolGauge: ${TBTCCurvePoolGauge.address}`)
  log(`TBTCCurvePoolGaugeReward: ${TBTCCurvePoolGaugeReward.address}`)
```

yearn/deploy/00_deploy_cvp_strategy.ts
```diff
@@ -0,0 +1,49 @@
...
+import { DeployFunction } from "hardhat-deploy/types"
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+  const { deployments, getNamedAccounts, helpers } = hre
+  const { deploy, log } = deployments
+  const { deployer } = await getNamedAccounts()
+
+  const TBTCCurveVault = await deployments.getOrNull("TBTCCurveVault")
+  if (!TBTCCurveVault || !helpers.address.isValid(TBTCCurveVault.address)) {
+    throw new Error("Address of TBTCCurveVault must be set")
```
nkuba, line 11: Can we include `TBTCCurveVault.address` in the message?

yearn/deploy/01_deploy_convex_strategy.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  const TBTCConvexRewardPool = await deployments.getOrNull("TBTCConvexRewardPool")
+  if (!TBTCConvexRewardPool || !TBTCConvexRewardPool.linkedData.id) {
+    throw new Error("ID of TBTCConvexRewardPool must be set")
+  }
+
+  log(`tbtcCurveVault: ${TBTCCurveVault.address}`)
+  log(`tbtcCurvePoolDepositor: ${TBTCCurvePoolDepositor.address}`)
+  log(`tbtcConvexRewardPoolId: ${TBTCConvexRewardPool.linkedData.id}`)
```
nkuba, line 26: What is the story behind the id? Is it a property that can be pulled from a contract with a get function or some fixed value? 
I just want to make sure it doesn't get lost in the external artifact that we want to store and we're sure it will be there.

yearn/hardhat.config.ts
```diff
@@ -0,0 +1,76 @@
...
+          process.env.FORKING_BLOCK && parseInt(process.env.FORKING_BLOCK),
+      },
+      tags: ["local"]
+    },
+    development: {
+      url: "http://localhost:8545",
+      chainId: 1101,
+      tags: ["local"]
+    },
+    mainnet: {
```
nkuba, line 46: We will define these data in an external file, please see https://github.com/keep-network/coverage-pools/commit/5e899ef7cc596f2bdf087f647676d7e8cdbccb46

yearn/deploy/00_deploy_cvp_strategy.ts
lukasz-zimnoch: Because it's an optional value and zero address is a valid value. We don't know at this point if tBTC v2 Curve pool gauge will distribute KEEP rewards as the v1 Curve pool gauge does. We are prepared for both scenarios :)

yearn/deploy/00_deploy_cvp_strategy.ts
lukasz-zimnoch: Those are constructor parameters that are uncapitalized in the contract so this is why I display them that way to avoid confusion. I'd rather prefer to leave as is but arguments for making them capitalized are welcome :)

yearn/deploy/01_deploy_convex_strategy.ts
lukasz-zimnoch: This is a fixed value. Long story short, there is a Convex booster contract which is a kind of reward pool manager. Each registered reward pool gets its own ID used for further identification. It is simpler to operate knowing the pool ID than its address so this is why it is here. I know passing it using `linkedData` looks a bit strange at first sight. However, I think it's better than using an env variable and it is consistent with the way we deal with external dependencies.

yearn/hardhat.config.ts
lukasz-zimnoch: Ok, I will change. However, I'm curious about the motivations behind the approach you propose. I'm asking because creating a file from a template, filling it with data, and linking in the Hardhat config seems to be more cumbersome than just passing those things via env variables, as we used to do in other projects. Can you shed some light?

yearn/deploy/01_deploy_convex_strategy.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  const TBTCCurvePoolDepositor = await deployments.getOrNull("TBTCCurvePoolDepositor")
+  if (!TBTCCurvePoolDepositor || !helpers.address.isValid(TBTCCurvePoolDepositor.address)) {
+    throw new Error("Address of TBTCCurvePoolDepositor must be set")
+  }
+
+  const TBTCConvexRewardPool = await deployments.getOrNull("TBTCConvexRewardPool")
+  if (!TBTCConvexRewardPool || !TBTCConvexRewardPool.linkedData.id) {
+    throw new Error("ID of TBTCConvexRewardPool must be set")
```
dimpar, line 21: Do we know if `TBTCConvexRewardPool` ID will be actually `0`? If yes, then the error will be thrown, cause `!0` => `1` and we enter inside of the `if` statement.

dimpar:
The rest looks good to me.

yearn/deploy/01_deploy_convex_strategy.ts
lukasz-zimnoch: It won't be for sure.

yearn/.eslintrc
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/14/commits/52313e312cf5491832a64b83bef5c4db38432d0d

yearn/deploy/00_deploy_cvp_strategy.ts
lukasz-zimnoch: Good point! Please see: https://github.com/keep-network/tbtc-v2/pull/14/commits/96b6813293702d9bec6aaa5f0af1f75d6c057ad8

yearn/deploy/00_deploy_cvp_strategy.ts
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/14/commits/fd4c1aed9986be5c8003bf17f3cd4200d8ac13d5

yearn/hardhat.config.ts
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/14/commits/6d74161552b3754d66ef2ba6d90ac3365addf7a0

yearn/README.adoc
```diff
@@ -0,0 +1,171 @@
...
+```
+yarn deploy --network <network>
+```
+If applicable, use environment variables listed in the
+<<deployment-parametrization>> section.
+
+To deploy a single contract, use the same command with the `--tags` option
+followed by the contract name:
+```
+yarn deploy --network localhost --tags ConvexStrategy
```
nkuba, line 148:
```suggestion
yarn deploy --network development --tags ConvexStrategy
```

yearn/package.json
```diff
@@ -1,9 +1,25 @@
...
+    "lint:fix:js": "eslint . --fix",
+    "lint:fix:sol": "solhint 'contracts/**/*.sol' --fix && prettier --write '**/*.sol'",
+    "lint:js": "eslint . ",
+    "lint:sol": "solhint 'contracts/**/*.sol' && prettier --check '**/*.sol'",
+    "size-contracts": "hardhat compile && hardhat size-contracts",
+    "test": "hardhat test",
+    "test:system": "NODE_ENV=system-test hardhat test ./test/system/*.test.js"
+  },
   "dependencies": {
     "@keep-network/yearn-vaults": "git://github.com/keep-network/yearn-vaults.git#5f1e829",
```
nkuba, line 17: Can we use `github:` for consistency with other references to github packages?

```suggestion
    "@keep-network/yearn-vaults": "github:keep-network/yearn-vaults.git#5f1e829",
```

yearn/package.json
```diff
@@ -13,24 +29,16 @@
...
-    "lint": "npm run lint:js && npm run lint:sol",
-    "lint:fix:js": "eslint . --fix",
-    "lint:fix:sol": "solhint 'contracts/**/*.sol' --fix && prettier --write '**/*.sol'",
-    "lint:js": "eslint . ",
-    "lint:sol": "solhint 'contracts/**/*.sol' && prettier --check '**/*.sol'",
-    "size-contracts": "hardhat compile && hardhat size-contracts",
-    "test": "hardhat test",
-    "test:system": "NODE_ENV=system-test hardhat test ./test/system/*.test.js"
+  "engines": {
+    "node": ">= 12.0.0"
```
nkuba, line 42: Can we switch to node 14 as this is the current LTS version?

yearn/hardhat.config.ts
```diff
@@ -0,0 +1,72 @@
+import { HardhatUserConfig } from "hardhat/config"
+
+import "@keep-network/hardhat-helpers"
```
nkuba, line 3: To use `localNetworksConfig` property we need to add an import here:
```js
import "@keep-network/hardhat-local-networks-config"
```

And in `tsconfig.json` add `"node_modules/@keep-network/hardhat-local-networks-config/src/type-extensions.d.ts"` to `files` list.

yearn/deploy/00_deploy_cvp_strategy.ts
```diff
@@ -0,0 +1,64 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types"
+import { DeployFunction } from "hardhat-deploy/types"
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+  const { deployments, getNamedAccounts, helpers } = hre
+  const { deploy, log } = deployments
+  const { deployer } = await getNamedAccounts()
+
+  const TBTCCurveVault = await deployments.get("TBTCCurveVault")
```
nkuba, line 9: Can we have a way for `local` networks to test the deployment scripts? Maybe we could fork the mainnet in hardhat network? Or use stubs? Or use fake values? Any solution that would let us test the deployment scripts on local network would be fantastic. If it's not possible with `development` network `hardhat` network will be super fine.

If we already have this possibility can we describe it in documentation?

nkuba (commented):
I tried to run the script locally with some tweaks but got a failure:
```
Error: ERROR processing /Users/jakub/workspace/keep-network/tbtc-v2/yearn/deploy/00_deploy_cvp_strategy.ts:
Error: cannot estimate gas; transaction may fail or may require manual gas limit (...)
```

Is it related to the `GAS_LIMIT` I didn't set? I tried to use local geth, why it failed to calculate the estimate?

lukasz-zimnoch:
> I tried to run the script locally with some tweaks but got a failure:
> 
> ```
> Error: ERROR processing /Users/jakub/workspace/keep-network/tbtc-v2/yearn/deploy/00_deploy_cvp_strategy.ts:
> Error: cannot estimate gas; transaction may fail or may require manual gas limit (...)
> ```
> 
> Is it related to the `GAS_LIMIT` I didn't set? I tried to use local geth, why it failed to calculate the estimate?

I posted in the readme:
```
Please note that all strategies contracts make a call to their vault contract upon initialization. That means the deployment will fail if the vault address passed in the strategy constructor does not point to a contract compatible with the Yearn vault API.
```
This fail manifests as a problem with gas estimation.

yearn/README.adoc
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/14/commits/40b5281258764a3ab96d94c87255e7fd9148a70e

yearn/package.json
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/14/commits/86b04f0287765fbbce093787d9771e81a78320fb

yearn/package.json
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/14/commits/efddb833fde05f4f67c27d8b43e611854c182a7a

yearn/hardhat.config.ts
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/14/commits/436541e53f3fe038491632ce2639c9b6f44db132

yearn/deploy/00_deploy_cvp_strategy.ts
lukasz-zimnoch: I tried to do this with `development` network but wasn't able to make it work. There are so many dependencies to stub and I still received failures during deployment. I will try with `hardhat`.

yearn/hardhat.config.ts
nkuba: We used storing secret data in a file for our mainnet contracts migrations in the past. This protects better against secrets leaking during a screen share when you have to set env variables in a terminal when running a command.
It's also simpler to store the values in a file once and reuse it for other projects at different times. You can define the file in your home dir and point to this one file from all your hardhat configs across the projects.

yearn/package.json
nkuba: We also have to update it in README.md, as the node version is mentioned there too.

yearn/deploy/00_deploy_cvp_strategy.ts
nkuba: It would be awesome to handle it for `hardhat` network, as now we have no way to test the scripts, except deploying them on mainnet.

lukasz-zimnoch: Ok, I managed to run the script on `hardhat` network: https://github.com/keep-network/tbtc-v2/pull/14/commits/299f4985dbf6794e0bd337bba96bc3eabeddaa96. I used mainnet forking and leveraged Yearn, Curve and Convex contracts existing on mainnet for tBTC v1. The two first strategies (`CurveVoterProxyStrategy` and `ConvexStrategy`) are deployed without hassle. However the script doesn't pass for `SaddleStrategy`. This is because there is no Saddle vault for tBTC v1 which can be used here. I left a TODO and we'll make it work once Saddle vault for tBTC v2 is created. I don't see any simple workaround for that until then.

yearn/package.json
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/14/commits/3c1901a27a86d1587e0d3c5c2180944be678da45

dimpar:
Tested latest changes by running: `FORKING_URL=https://eth-mainnet.alchemyapi.io/v2/<id> FORKING_BLOCK=12786839 yarn deploy --network hardhat` Both 2 curve strategies deployed successfully on hardhat network.

nkuba (approved):

PR #15, opened 2021-07-22
Estimating amount of vault's underlying tokens ('want' tokens) received for ETH

dimpar:
Refs: #7 

In this PR we estimate vault's underlying token amount that can be received for ETH. It will cover the scenario when calling `harvest()` and paying for gas cost is higher than the profit.

As per [this comment](https://github.com/keep-network/tbtc-v2/pull/15#issuecomment-884842754) a specific 'Curve.fi tBTC/sbtcCRV' token was used for tBTC v1 vault. When the time comes for tBTC v2, the addresses will be changed and the underlying token will also change from 'Curve.fi tBTC/sbtcCRV' to a new one.

There are two steps to estimate LP token amount received for ETH. The first step is to call dex like Uniswap for tBTC-ETH rates and later call Curve `calc_token_amount()` to estimate received LP tokens for tBTC.


lukasz-zimnoch:
Just a remark about PR title - `tBTC/sbtcCRV` is specific for tBTC v1 vaults working with Curve. For tBTC v2 that may be another token. For tBTC v2 vault working with Saddle, there will be something else. I propose naming it more genereally as "vault's underlying token" or "vault's want token".

yearn/contracts/ConvexStrategy.sol
```diff
@@ -606,7 +613,23 @@ contract ConvexStrategy is BaseStrategy {
...
+        // profitability of calling 'harvest()' we need to workaround and see the
+        // rate between tBTC - ETH on dex like Uniswap. tBTC was chosen arbitrary
+        // since it is availbale on Uniswap and was built on the keep-network.
+        // amounts[0] -> ETH in wei
+        // amounts[1] -> tBTC
+        uint256[] memory amounts = IUniswapV2Router(uniswap).getAmountsOut(amtInWei, path);
+
+        // By knowing ETH->tBTC amounts it is possible to estimate the amount of
+        // 'Curve.fi tBTC/sbtcCrv' LP tokens to be received for tBTC tokens.
+        return ICurvePool(tbtcCurvePoolDepositor).calc_token_amount([amounts[1], 0, 0, 0], false);
```
lukasz-zimnoch, line 633: The logic seems correct but I'd change those comments a little bit. They are tightly tied to the tBTC v1 and its Curve pool internals. My proposition:
```
// Convert the ETH amount denominated in WEI into an amount denominated in one of the tokens accepted by 
// the tBTC v2 Curve pool using Uniswap. The tBTC v2 token was chosen arbitrarily since it is available on Uniswap
// and was built on the Keep Network.
uint256[] memory amounts = IUniswapV2Router(uniswap).getAmountsOut(amtInWei, path);

// Use the amount denominated in tBTC v2 to calculate the amount of LP token (vault's underlying token) that 
// could be obtained if that tBTC v2 amount was deposited in the tBTC v2 Curve pool. This way we obtain an
// estimated value of the original WEI amount represented in the vault's underlying token
return ICurvePool(tbtcCurvePoolDepositor).calc_token_amount([amounts[1], 0, 0, 0], false);
```
WDYT?

lukasz-zimnoch (changes requested):

lukasz-zimnoch:
@tomaszslabon You may be interested in this logic for #13.

yearn/contracts/ConvexStrategy.sol
dimpar: Sure, good points, I've realised that after your first comment regarding changing the tile of this PR. I've changed the original comment as you suggested with just some minor changes e247269

yearn/contracts/ConvexStrategy.sol
```diff
@@ -154,6 +158,9 @@ contract ConvexStrategy is BaseStrategy {
     // Address of the WBTC token contract.
     address public constant wbtc =
         address(0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599);
+    // Address of the tBTC token contract.
+    address public constant tbtc =
```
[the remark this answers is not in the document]
dimpar: I totally agree on adding TODOs but not sure about replacing the current approach by programmatically fetching tbtc(v1/v2) address. Once the tBTCv2 is deployed it stays there and I think we can safely hardcode it just like with other tokens. Adding code for fetching the address surely works, but it adds some code that can be avoided I think. Regarding the index, similar thoughts as for the token address. Once the pool is deployed it stays there and the indexes won't change just like we do for `wBTC` index. I guess we can assume the index will be `0` for tBTC but we also don't know for sure if Curve has any plans with updating their pool contracts and how the indexes will look like. Idk, I think the safest and the most simple way is to wait until all the contracts are deployed (tBTCv2 + curve pool with tBTCv2) and use the addresses and indexes accordingly.
If you feel strongly about fetching addresses programatically and assuming the index will still be `0`, I guess we can do that as well. 
Added TODOs here 45e577f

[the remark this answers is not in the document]
lukasz-zimnoch: Yeah those arguments sound reasonable. But, what do you think about using wBTC instead of tBTC as we planned at the beginning? This way we don't have to depend on tBTC address and we can rely on wBTC which is stable and already used in those contracts. The only concern we should remember about is checking if wBTC index in the v2 pool remains the same as in the current v1 pool. We can leave TODOs for that here (near `calc_token_amount` invocation) and next to `add_liquidity` calls in both strategies. What do you think?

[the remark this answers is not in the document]
dimpar: Yes! I like this idea, we can simplify existing approach by replacing `tBTC` with `wBTC`

yearn/contracts/CurveVoterProxyStrategy.sol
```diff
@@ -122,6 +126,9 @@ contract CurveVoterProxyStrategy is BaseStrategy {
     // Address of the WBTC token contract.
     address public constant wbtc =
         address(0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599);
+    // Address of the tBTC token contract.
+    address public constant tbtc =
```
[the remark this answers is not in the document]
dimpar: Please see 7a5691b

yearn/contracts/ConvexStrategy.sol
```diff
@@ -606,7 +617,34 @@ contract ConvexStrategy is BaseStrategy {
...
+        // Use the amount denominated in wBTC to calculate the amount of LP token
+        // (vault's underlying token) that could be obtained if that wBTC amount
+        // was deposited in the Curve pool that has tBTC v2 in it. This way we
+        // obtain an estimated value of the original WEI amount represented in
+        // the vault's underlying token.
+        // TODO: When the new curve pool with tBTC v2 is deployed, verify that
+        // the index of wBTC (amounts[1]) in the array is correct.
+        return
+            ICurvePool(tbtcCurvePoolDepositor).calc_token_amount(
+                [amounts[1], 0, 0, 0],
```
lukasz-zimnoch, line 646: Shouldn't we pass `amounts[1]` at index `2` like we do in `CurveVoterProxyStrategy`?

yearn/contracts/ConvexStrategy.sol
```diff
@@ -606,7 +617,34 @@ contract ConvexStrategy is BaseStrategy {
...
+            amtInWei,
+            path
+        );
+
+        // Use the amount denominated in wBTC to calculate the amount of LP token
+        // (vault's underlying token) that could be obtained if that wBTC amount
+        // was deposited in the Curve pool that has tBTC v2 in it. This way we
+        // obtain an estimated value of the original WEI amount represented in
+        // the vault's underlying token.
+        // TODO: When the new curve pool with tBTC v2 is deployed, verify that
```
lukasz-zimnoch, line 642: Nitpick: Can we add a blank line comment between the description and the TODO? It will make the TODO more visible.

yearn/contracts/CurveVoterProxyStrategy.sol
```diff
@@ -504,7 +516,34 @@ contract CurveVoterProxyStrategy is BaseStrategy {
...
+            amtInWei,
+            path
+        );
+
+        // Use the amount denominated in wBTC to calculate the amount of LP token
+        // (vault's underlying token) that could be obtained if that wBTC amount
+        // was deposited in the Curve pool that has tBTC v2 in it. This way we
+        // obtain an estimated value of the original WEI amount represented in
+        // the vault's underlying token.
+        // TODO: When the new curve pool with tBTC v2 is deployed, verify that
```
lukasz-zimnoch, line 541: Nitpick: Can we add a blank line comment between the description and the TODO? It will make the TODO more visible.

lukasz-zimnoch (commented):
Looks good! Just three comments regarding the code.

yearn/contracts/ConvexStrategy.sol
dimpar: Yes! I forgot to change it here, done in 1811272

yearn/contracts/ConvexStrategy.sol
dimpar: Sure, done in 1811272

yearn/contracts/CurveVoterProxyStrategy.sol
dimpar: Done in 1811272

yearn/test/system/convex-strategy.test.js
```diff
@@ -220,5 +221,24 @@ describeFn("System -- convex strategy", () => {
...
+      //
+      // TODO: When the new curve pool with tBTC v2 is deployed, the fork block
+      // number would need be adjusted acoordingly. In consequnce, the rates of
+      // tokens will be different and will need be adjusted as well.
+      it("should calculate LP tokens in exchange for ETH ", async () => {
+        const lpTokens = await strategy.ethToWant(to1e18(1000))
+
+        expect(lpTokens).to.be.closeTo(
+          to1ePrecision(633018, 14),
+          to1ePrecision(1, 14) // 0.0000001 precision
```
lukasz-zimnoch, line 239: Isn't it `0.0001` precision?

yearn/test/system/convex-strategy.test.js
```diff
@@ -220,5 +221,24 @@ describeFn("System -- convex strategy", () => {
       expect(amountWithdrawn.gt(curveStrategyFixture.vaultDepositAmount)).to.be
         .true
     })
+
+    describe("eth to LP token (want) conversion", () => {
+      // At block 12786839 (Jul-08-2021) the exchange rate between wBTC/ETH was
+      // ~1/15.65 Swapping 1000ETH on Uniswap gave ~63.9wBTC. It means that by
+      // depositing 63.9wBTC to Curve.fi LP Pool a depositor would get
+      // ~63,3018 LP tokens.
```
lukasz-zimnoch, line 229: Can we use a dot instead of a comma? I thought we're talking about thousands at the first sight.

yearn/test/system/curve-voter-proxy-strategy.test.js
```diff
@@ -215,4 +216,23 @@ describeFn("System -- curve voter proxy strategy", () => {
         .true
     })
   })
+
+  describe("eth to LP token (want) conversion", () => {
+    // At block 12786839 (Jul-08-2021) the exchange rate between wBTC/ETH was
+    // ~1/15.65 Swapping 1000ETH on Uniswap gave ~63.9wBTC. It means that by
+    // depositing 63.9wBTC to Curve.fi LP Pool a depositor would get
+    // ~63,3018 LP tokens.
```
lukasz-zimnoch, line 224: Same with the dot.

lukasz-zimnoch (commented):
Three last comments and I'm ready to merge!

yearn/test/system/convex-strategy.test.js
dimpar: Eh, yes.. done in e369c05

yearn/test/system/convex-strategy.test.js
dimpar: Sure e369c05

yearn/test/system/curve-voter-proxy-strategy.test.js
dimpar: e369c05

lukasz-zimnoch (approved):

PR #18, opened 2021-07-28
[System test] Saddle strategy migration

tomaszslabon:
This PR adds a system test for testing migrating Saddle strategies - funds (LP tokens and rewards) from one strategy should be transferred to the new one.


dimpar:
Do you think we can refactor a bit and add this scenario to the existing `saddle-strategy.test.js`?

tomaszslabon:
@dimpar 
> Do you think we can refactor a bit and add this scenario to the existing `saddle-strategy.test.js`?
We can, but it may clutter the already complicated existing scenario.
I can try to do that and see if this makes sense.

dimpar:
> @dimpar
> 
> > Do you think we can refactor a bit and add this scenario to the existing `saddle-strategy.test.js`?
> > We can, but it may clutter the already complicated existing scenario.
> > I can try to do that and see if this makes sense.

Ok, I noticed that we're duplicating test setup and wondered if we can avoid it. I guess we can try and see what we end up with just like you suggested.

lukasz-zimnoch:
> Do you think we can refactor a bit and add this scenario to the existing `saddle-strategy.test.js`?

I'd vote for keeping them separated. It will not be so easy to add a migration test to the existing test in a clean way. I think we can just rename the existing test to `saddle-strategy-harvest.test.js` and keep the migration scenario in `saddle-strategy-migration.test.js`. I know we will duplicate some code but I think we can pay this price for higher readability given the fact the contract logic is quite complex.

dimpar:
> > Do you think we can refactor a bit and add this scenario to the existing `saddle-strategy.test.js`?
> 
> I'd vote for keeping them separated. It will not be so easy to add a migration test to the existing test in a clean way. I think we can just rename the existing test to `saddle-strategy-harvest.test.js` and keep the migration scenario in `saddle-strategy-migration.test.js`. I know we will duplicate some code but I think we can pay this price for higher readability given the fact the contract logic is quite complex.

Ok, I played with the code myself and I see it's not as simple as I thought and indeed we will make a test even more complex by putting everything in one file. @tomaszslabon the test looks good

tomaszslabon:
> Ok, I played with the code myself and I see it's not as simple as I thought and indeed we will make a test even more complex by putting everything in one file. @tomaszslabon the test looks good

@dimpar 
Ok, Let's keep them separated then.

dimpar (approved):

PR #20, opened 2021-07-28
[System test] CurveVoterProxyStrategy and ConvexStrategy migration

tomaszslabon:
This PR adds system tests for migration for `CurveVoterProxyStrategy` and `ConvexStrategy`.
It checks whether a migration from one strategy to another is successful and funds locked in the old strategy are transferred to the new one.


yearn/contracts/CurveVoterProxyStrategy.sol
```diff
@@ -328,6 +328,26 @@ contract CurveVoterProxyStrategy is BaseStrategy {
             tbtcCurvePoolGauge,
             address(want)
         );
+        // Harvest the CRV rewards from the Curve pool's gauge and transfer
```
dimpar, line 331: Oh, that's a good one. We do transfer `crv` & `cvx` + extra rewards in convexStrategy. Perhaps we missed it here? I think Łukasz will shed more light on it.

dimpar:
Tests looks good to me. Also interested in the new change in Solidity for transferring rewards.

yearn/contracts/CurveVoterProxyStrategy.sol
lukasz-zimnoch: Yep indeed. Good catch Tomasz!

lukasz-zimnoch (approved):

PR #12, opened 2021-07-14
Add global prettier configuration and fix formatting

michalinacienciala:
We added repo-wide prettier configuration to handle all possible file
types.
With lint:js we are supporting `*.js` files.
With lint:sol we are supporting `*.sol` files.

We were missing handling for bunch of other file types, like `*.json`,
`*.sh`, `*.yaml` which prettier allows us to.

As part of this PR we're also fixing formatting errors reported by
prettier/linter.

Closes: #11

PR #21, opened 2021-07-28
Updated to global format configuration

nkuba:
This PR enhances https://github.com/keep-network/tbtc-v2/pull/12 with additional updates to packages code formatting.
In `solidity` and `yearn` packages we add `format` script that will execute linting for both `js` and `sol` file types, but also will run `prettier` for all supported files (e.g. `json`).

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Run tests
         run: yarn test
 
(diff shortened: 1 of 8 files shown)


michalinacienciala (approved):
Makes perfect sense. Leaving for @dimpar for review.

dimpar (approved):
Looks good to me!

PR #23, opened 2021-07-29
RFC 1: tBTCv2 Design

beaushinkle:
The intent of this document is to provide a living source of reference for the design for the tBTCv2 system and be useful both on an descripive level and a technical level.

as of right now, it is but a rough draft :)

tagging @pdyraga for early review - I want to make sure the direction that this is going is correct before I write more!

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -10,17 +10,14 @@ toc::[]
 
 === Minting
 
-Stakers periodically create https://en.bitcoin.it/wiki/BIP_0032[Hierarchical
-Deterministic Wallets] to hold frozen BTC assets to maintain the peg.
-Depositors send BTC funds to the most-recently-created-wallet at a secret
-derivation path. Then, the depositor hashes the combination of their secret
-derivation path and ethereum address and publishes that to the ethereum
-blockchain (a commitment). Once the commitment is published, the depositor
-reveals their secret derivation path and ethereum address. The tbtc system
-verifies that the reveal matches the commitment and that there UTXOs at the
-specified secret derivation path. If everything checks out, the bitcoin wallet
-moves the funds from the secret derivation path to the public derivation path
-and mints the equivalent amount of tBTCv2 at the revealed ethereum address.
+Stakers periodically create 51-of-100 ecdsa-backed wallets to hold frozen BTC
+assets to maintain the peg. Depositors send BTC funds to the
+most-recently-created-wallet by using pay-to-script-hash (P2SH) which contains
+hashed information about the depositor's minting ethereum address. Once
+the block is mined, the depositor reveals their desired ethereum minting
+address to the ethereum chain. The TBTC system listens for these sorts of
+messages and when it gets one, it checks the bitcoin network to make sure the
+funds line up. If everything checks out, then we mint!
 
 === Redemption
 
@@ -34,58 +31,38 @@ been released.
 
 === Minting
 
-==== Hierarchical Deterministic Wallet Derivation Paths Are A Secret
+==== Deposit + Commitment
 
-A huge part of understanding the technical side of the minting process is
-understanding how Hierarchical Deterministic Wallets work. The basic concept is
-that you have a root private key and public key that allows you to supply an
-index that creates another private key + public key. To experiment with this,
-check out https://iancoleman.io/bip39/[bip39]. Conventionally, a derivation path is written as
-`m / purpose' / coin_type' / account' / change / address_index`. A `'` denotes
-that the index is "hardened", which means that a base of `2147483648` has been
-added to the address (and that the private key is required to derive the
-address). The HD wallet standard is blockchain agnostic, and different
-standards have been adopted for each community. Bitcoin, for example, always
-uses a purpose of `44'`, a coin type of `0`, and a change of `0` (the change
-address denotes where leftover bitcoin goes after being sent from the address).
-This means that for our purposes, we can vary the `account` and `address_index`
-fields: `m/44'/0/x/0/y`. Each path has `2^31` available hardened addresses,
-which means there are `2^62` unique hardened paths.
+Once we know the active wallet's public key hash, the dApp can put together a
+pay-to-script-hash (P2SH) address to receive the funds. This script will be
+unique to each depositor and will look like:
+```
+if(verifySig(signingGroupPubKey) && hash160(ethereumAddress + blindingFactor)) {
+    // allow funds to be spent
+} else if(verifySig(refundPubKey) && daysSinceUTXOCreation > 30) {
+    // allow funds to be spent
+}
+```
 
-The main attack that we want to avoid is someone monitoring children of the
-active wallet for deposits, and then telling the ethereum chain that they
-deposited the bitcoin before the real depositor did. With `2^62` possible child
-addresses, it is currently computationally infeasible to churn through the
-addresses fast enough to accomplish this.
+// TODO: represent the above script in the bitcoin scripting language and prototype it
 
-==== Commit + Reveal Scheme
+Since each depositor has their own ethereum address and their own secret
+blinding factor (which is an additional security layer), each depositor's
+script will be unique, and the hash of each depositor's script will be unique.
 
-Now that we've establed that a particular derivation path like
-`m/44'/0/281'/0/1821912'` is a secret, then we can craft a commit+reveal scheme
-from it. First, the user deposits BTC to the public key at
-`m/44'/0/281'/0/1821912'`. Second, the user concatenates that secret path with their minting address:
-`m/44'/0/281'/0/1821912'+0x6EB149f6FBA44724cD2e7CAD13B18E5E8D2b709D`. This is their final secret.
+Once the deposit has been made, the active wallet will be unable to spend the
+funds until the ethereumAddress and blinding factor are revealed.
 
-They hash that secret and get something like `keep btc deposit:
-924deb62fcb9780220c04ecab100b97ee1dea25a30771552d483d1c7fc7ef04a` (using
-sha256) and submit that to the ethereum chain.
+After 30 days, the refund wallet (specified by the user) will be able to spend the funds.
 
-Then, once that commitment is on the blockchain, they reveal their secret in
-plain text:
-`m/44'/0/281'/0/1821912'+0x6EB149f6FBA44724cD2e7CAD13B18E5E8D2b709D` to the
-ethereum chain. Once the keep network sees the derivation path + minting
-address, it checks to see if there was a deposit at the supplied address and if
-the sha256 of the supplied data is an existing commitment. This prevents miners
-from reading the mempool and submitting their own message like
-`m/44'/0/281'/0/1821912'+0xMYADDRESS` to claim that the tBTC should be theirs.
-If everything checks out, then we're good!
+==== The Big Reveal
 
-==== Sweeping The Wallet
-
-When a user submits their reveal, the wallet should move the funds from the
-secret derivation path to the main wallet located at `m/44'/0/0/0/0`. After
-this step, the system is good to mint the equivalent amount of tBTC to the
-supplied ethereum address in the reveal.
+After the deposit transaction has been mined, the user is able to reveal their
+ethereum address and blinding factor to the ethereum chain. The TBTC system listens
+for these sorts of messages and when it sees one, is able to generate a script that
+can spend the funds. Once successful, wallet moves the funds from the P2SH
+address to the main wallet address, which allows the system to mint tBTC and
+disables the 30-day refund.
 
 ==== Automated Refunds
 
@@ -94,9 +71,57 @@ simple as "these funds can be spent by wallet 0xabc", or in our case, as
 complex as "these funds can be spent by wallet 0xabc but if they aren't spent
 within 30 days they can be spent by wallet 0x123". This gives us the ability to
 create deposits that automatically are refunded after 30 days if they aren't
-swept. Then, if a user misfunds or they get cold feet (for any reason), all
-they need to do is not submit their reveal (or commitment) and wait 30 days.
+swept. Thus, if a user misfunds or they get cold feet (for any reason), all
+they need to do is not submit their reveal and wait 30 days.
 
 === Redemption
 
-TODO
+To initiate a redemption, a user supplies an amount `x` of TBTC and a bitcoin
+address. Then, the system calculates the redemption fee `fee`, and releases an
+amount of bitcoin `y` such that `x = y + fee` to the supplied bitcoin address.
+`y` amount of TBTC is burned to maintain the peg. The remaining `fee` TBTC is
+sold by the system to buy back `T` tokens (more about this process in the fee
+section).
+
+Redemptions for BTC follow first-in-first-out principles. Reedemed bitcoin is
+sent from the oldest, live wallet, potentially carrying over until the full
+amount is covered. For example, if 100 BTC needs to be redeemed, and the 4
+oldest live wallets each have a 30 BTC balance, then 30 BTC would come from the
+oldest, 30 BTC would come from the second oldest, 30 BTC would come from the
+third oldest, and 10 BTC would come from the 4th oldest. Wallets that no longer
+have a balance and are not the active wallet for deposits are effectively archived.
+
+=== Wallet Lifecycle
+
+Wallets are periodically created, where the period length is a governable
+parameter. To create a new wallet, a group of 100 operators is selected from
+the pool of available operators (some operators may be selected twice if there
+are not enough) using a process called sortition. The probabiliy that a
+particular operator is chosen is based on their stake weight, which in turn is
+based on the number of `T` tokens they have invested in the staking contract.
+
+Once the 100 operators have been selected, they generate a 51-of-100 ecdsa
+signing group to handle the bitcoin key material. The distributed key
+generation process requires that all 100 participants are available, but future
+signing events (like minting and redemption) only require 51 of the 100.
+
+As time passes and operators drop out of the system, a wallet becomes at risk of
+being able to meet the 51-of-100 threshhold to produce signatures. Additionally,
+we want to avoid situations where operators are the custodians of a wallet for
+extended periods. To avoid these issues, we can set a max age of a wallet and a minimum
+liveness threshhold. Once a wallet is older than the max age, or if it drops below
+the liveness threshhold (say, below 60 on a heartbeat), we motion to transfer
+the funds to another randomly selected wallet.
+
+Once a wallet no longer has funds and is not the primary wallet for new
+deposits, it can be archived and operators are no longer required to maintain
+it.
+
+=== Heartbeats
+
+To make sure that older wallets are still accessible for redemption, we need to
+perform heartbeats. This can take the form of small, cheap work that can be
+used to verify that a node is still able to provide its part in a 51-of-100
+signing event.
+
+// FIXME: figure out how this works
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -22,10 +22,14 @@ If everything checks out, then we mint!
 === Redemption
 
 A user supplies tBTCv2 and a bitcoin address. The system burns the tBTCv2 and
-releases the equivalent amount of bitcoin to the user from the oldest
-outstanding bitcoin wallet. If the oldest does not have enough, the remaining
-balance carries over to the next oldest until an equivalent amount of BTC has
-been released.
+releases the equivalent amount of bitcoin to the user from a user-supplied bitcoin wallet.
+
+In the default case, the user-supplied wallet will be determined by the dApp,
+and will be the oldest wallet that is large enough to fulfill the redemption.
+
+The maximum redemption size is capped by the size of the largest wallet, and
+any redemption bigger than that will need to be split into multiple
+redemptions.
 
 === Continuous Fees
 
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-1.adoc
```diff
@@ -0,0 +1,102 @@
...
+from reading the mempool and submitting their own message like
+`m/44'/0/281'/0/1821912'+0xMYADDRESS` to claim that the tBTC should be theirs.
+If everything checks out, then we're good!
+
+==== Sweeping The Wallet
+
+When a user submits their reveal, the wallet should move the funds from the
+secret derivation path to the main wallet located at `m/44'/0/0/0/0`. After
+this step, the system is good to mint the equivalent amount of tBTC to the
+supplied ethereum address in the reveal.
```
pdyraga, line 88: Why do we need to sweep them? Can't we track UTXO and use them during redemption? Or maybe we do not want to track UTXOs and we do it for simplification? If we sweep, someone has to pay for sweeping transactions. Where is this cost included so that the peg can be maintained?

docs/rfc/rfc-1.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+Then, once that commitment is on the blockchain, they reveal their secret in
+plain text:
+`m/44'/0/281'/0/1821912'+0x6EB149f6FBA44724cD2e7CAD13B18E5E8D2b709D` to the
+ethereum chain. Once the keep network sees the derivation path + minting
+address, it checks to see if there was a deposit at the supplied address and if
+the sha256 of the supplied data is an existing commitment. This prevents miners
+from reading the mempool and submitting their own message like
+`m/44'/0/281'/0/1821912'+0xMYADDRESS` to claim that the tBTC should be theirs.
+If everything checks out, then we're good!
```
pdyraga, line 81: Why don't we increment account/index on-chain and instead of having commit transaction to Ethereum, have a transaction where the depositor would associate their Ehtereum address with the next available derivation path? This could simplify the logic in the client.

docs/rfc/rfc-1.adoc
```diff
@@ -0,0 +1,102 @@
...
+Deterministic Wallets] to hold frozen BTC assets to maintain the peg.
+Depositors send BTC funds to the most-recently-created-wallet at a secret
+derivation path. Then, the depositor hashes the combination of their secret
+derivation path and ethereum address and publishes that to the ethereum
+blockchain (a commitment). Once the commitment is published, the depositor
+reveals their secret derivation path and ethereum address. The tbtc system
+verifies that the reveal matches the commitment and that there UTXOs at the
+specified secret derivation path. If everything checks out, the bitcoin wallet
+moves the funds from the secret derivation path to the public derivation path
+and mints the equivalent amount of tBTCv2 at the revealed ethereum address.
```
pdyraga, line 23: To move funds from the address created from the derivation path, do I need to know the private key for the account under that path? If so, I'd assume I also need to know the root private key and we do not know it - it's owned by a threshold singing group. 🤔 Can I do signing without constructing the account's private key so that I can sweep/redeem from the given path?

docs/rfc/rfc-1.adoc
```diff
@@ -0,0 +1,143 @@
...
+days (specified as `locktime`).
+
+==== The Big Reveal
+
+After the deposit transaction has been mined, the user is able to reveal their
+ethereum address and blinding factor to the ethereum chain. The TBTC system listens
+for these sorts of messages and when it sees one, is able to generate a script that
+can spend the funds. Once successful, wallet moves the funds from the P2SH
+address to the main wallet address, which allows the system to mint tBTC and
+disables the 30-day refund.
```
pdyraga, line 81: When is TBTC minted? I think we agreed on minting it immediately once the script is revealed and then sweeping BTC periodically. For the first iteration, would it be enough to sweep immediately?

docs/rfc/rfc-1.adoc
```diff
@@ -0,0 +1,143 @@
...
+within 30 days they can be spent by wallet 0x123". This gives us the ability to
+create deposits that automatically are refunded after 30 days if they aren't
+swept. Thus, if a user misfunds or they get cold feet (for any reason), all
+they need to do is not submit their reveal and wait 30 days.
+
+=== Redemption
+
+To initiate a redemption, a user supplies an amount `x` of TBTC and a bitcoin
+address. Then, the system calculates the redemption fee `fee`, and releases an
+amount of bitcoin `y` such that `x = y + fee` to the supplied bitcoin address.
```
pdyraga, line 97: Hmm... I am not sure if we can assume `priceOf(1 TBTC) == priceOf(1 BTC)`. They should be ~ but what I think we observed for v1 is that TBTC price is a little higher given the deposit costs.

pdyraga: Also, the amount of BTC released will probably be less given the BTC transfer fee - the redeemer should pay for it, it's the simplest solution I think. A separate question is how the system establishes BTC transaction fee. For v1, it has to be done on-chain. Are we willing to trust the judgment of clients? Having it on-chain feels more secure and we could avoid BTC transaction fee consensus issues between client instances.

docs/rfc/rfc-1.adoc
```diff
@@ -0,0 +1,143 @@
...
+generation process requires that all 100 participants are available, but future
+signing events (like minting and redemption) only require 51 of the 100.
+
+As time passes and operators drop out of the system, a wallet becomes at risk of
+being able to meet the 51-of-100 threshhold to produce signatures. Additionally,
+we want to avoid situations where operators are the custodians of a wallet for
+extended periods. To avoid these issues, we can set a max age of a wallet and a minimum
+liveness threshhold. Once a wallet is older than the max age, or if it drops below
+the liveness threshhold (say, below 60 on a heartbeat), we motion to transfer
+the funds to another randomly selected wallet.
```
pdyraga, line 130: How are we selecting that random wallet? Wouldn't it be simpler to always transfer to the most recent wallet?

docs/rfc/rfc-1.adoc
```diff
@@ -0,0 +1,143 @@
...
+messages and when it gets one, it checks the bitcoin network to make sure the
+funds line up. If everything checks out, then we mint!
+
+=== Redemption
+
+A user supplies tBTCv2 and a bitcoin address. The system burns the tBTCv2 and
+releases the equivalent amount of bitcoin to the user from the oldest
+outstanding bitcoin wallet. If the oldest does not have enough, the remaining
+balance carries over to the next oldest until an equivalent amount of BTC has
+been released.
```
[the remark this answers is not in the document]
beaushinkle: oh whoops, yeah I think I updated this in another section but forgot to do it here

docs/rfc/rfc-1.adoc
beaushinkle: I'll clarify in the RFC, but the prices there aren't about the exchange rate - they're about maintaining the peg.

For example, if we're the custodians of 100 BTC, then we need to have minted 100 TBTC. If there's a 1% deposit fee, then 99 of TBTC go to the depositor, and 1 TBTC is sold off to buy T.

If someone comes back with 100 TBTC and there's a 1% redemption fee, then we'll burn 99 TBTC, release 99 BTC, and sell 1 TBTC to buy T.

Also, even though we're *releasing* 99 BTC in that case, that doesn't mean that 99 BTC lands in the hand of the redeemer, instead it's 99 minus the btc transaction fee.

> A separate question is how the system establishes BTC transaction fee. For v1, it has to be done on-chain.

Can you elaborate a bit here? I know that for liquidation recovery, we use the esplora fee API. How does v1 calculate the fee on-chain?

docs/rfc/rfc-1.adoc
beaushinkle: > How are we selecting that random wallet?

That's a good question - calling the random beacon might be too expensive. I think the recent-wallet strategy is the easiest and least costly (because the newest wallet should be the one with the highest expected lifespan), but it makes us a little more vulnerable to folks intentionally trying to gain control over the newest wallet and then trying to create a transfer from an older wallet to the wallet they now control.

docs/rfc/rfc-1.adoc
```diff
@@ -0,0 +1,143 @@
...
+
+:icons: font
+:numbered:
+toc::[]
+
+== Overview
+
+=== Minting
+
+Stakers periodically create 51-of-100 ecdsa-backed wallets to hold frozen BTC
```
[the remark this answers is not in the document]
beaushinkle: Just the initial assumption - do you know what sort of prototyping we'd need to do to commit to some numbers?

docs/rfc/rfc-1.adoc
```diff
@@ -0,0 +1,143 @@
...
+been released.
+
+=== Continuous Fees
+
+The system is able to collect fees in the form of tBTC when the BTC is
+deposited and when it is redeemed. Rather than paying operators at those two
+instances we would prefer to pay them continuously. Since we do not have the
+tBTC to pay them with until we collect the redemption fee (and can't mint it
+and maintain the peg), we must turn to another token. Instead, the system pays
+operators continuously in mintable T tokens and whenever it collects tBTC fees
```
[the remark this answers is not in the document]
beaushinkle: It either needs to be able to mint T or have a treasury that the DAO can fill. Not sure what else would work

docs/rfc/rfc-1.adoc
```diff
@@ -0,0 +1,143 @@
...
+  CHECKSIG
+ELSE
+  DUP HASH160 <refundPubkey> EQUALVERIFY
+  <locktime> CHECKLOCKTIMEVERIFY DROP
+  CHECKSIG
+ENDIF
+```
+
+Since each depositor has their own ethereum address and their own secret
+blinding factor (which is an additional security layer), each depositor's
```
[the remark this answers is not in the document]
beaushinkle: > I'd love to see an explanation of why the blinding factor is needed here. Is it because we need to strengthen the hash? 

Yeah - the basic idea is that if a bad actor gets in control of the wallet, they can take the btc funds without the user revealing their eth address by grinding on every active eth address. @mhluongo has a better handle on the attack he has in mind though!

docs/rfc/rfc-1.adoc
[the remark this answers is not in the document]
pdyraga: I think we need to analyze it from the point of a probability of having malicious members in the group and how much one wallet can secure given the expected TVL of a coverage pool.

docs/rfc/rfc-1.adoc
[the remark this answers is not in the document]
pdyraga: Yeah, some kind of a contract releasing rewards previously allocated by DAO makes sense. That's ~ how we release rewards in coverage pools.

docs/rfc/rfc-1.adoc
eth-r: Minting before sweeping is risky. If an attacker gains the ability to DoS a wallet, they can deposit a large amount, minting the TBTC (and selling it before the rest of the attack is revealed), deliberately fail to sweep it, and thus take the 30-day refund while also getting the same BTC on-chain.

docs/rfc/rfc-1.adoc
eth-r: Can't we let the redeemer choose what BTC transaction fee would be used? Probably with a governable minimum so people don't try to do zero-fee transactions, but they'd be otherwise free to decide how to split the amount of BTC that gets released.

docs/rfc/rfc-1.adoc
eth-r: Consider the following:

We have wallets A and B. Each holds 100 BTC and there are 200 TBTC outstanding. Mallory starts with 10 BTC and is able to selectively DoS wallet B.

Wallet A: 100 BTC
Wallet B: 100 BTC

Mallory: 10 BTC

Mallory deposits 10 BTC in wallet B, receiving 10 TBTC. Mallory redeems 10 TBTC. The situation is now:

Wallet A: 100 BTC (-10 awaiting redemption to Mallory)
Wallet B: 100 BTC (+10 awaiting sweep)

Mallory: 0 BTC (+10 pending redemption, +10 refundable if not swept)

Mallory prevents wallet B from sweeping the deposit, but wallet A completes Mallory's redemption:

Wallet A: 90 BTC
Wallet B: 100 BTC (+10 awaiting sweep)

Mallory: 10 BTC (+10 refundable)

Mallory can now repeat the attack by re-depositing the BTC and redeeming it again. Once wallet A is empty, Mallory either stops the attack, with 10 BTC at hand and 90 BTC refundable in 30 days, or keeps depositing BTC and lets wallet B complete the redemption transactions but not the deposit sweeps. Even if other members of wallet B don't generate redemption transactions before the pending sweeps complete, Mallory is able to loot all BTC held by non-targeted wallets with an almost arbitrarily small initial capital investment.

beaushinkle: Ah, and this is would be thwarted by sweeping before minting. 

Then, Mallory won't get 10 TBTC until Wallet B has 110 BTC. She's able to effectively move funds to a wallet that she can't DoS to a wallet that she can by paying deposit/redemption fees, but isn't able to de-peg.

eth-r: Yep. If depositing gives you a NFT representing the UTXO and sweeping mints to the holder of the NFT, we can keep the peg safe and provide a quickminting service which redeems unswept UTXO NFTs for TBTC without the holders having to wait for the sweep. Then if someone manages to DoS the sweeps they can only drain the quickminting service instead of the entirety of TBTC.

eth-r: [https://github.com/keep-network/tbtc-research/pull/3/files#diff-93ad61cca4bd3006ea9d4948930a3e30b95a52de58ae02d0246e034ef85a6f1b](Here's) my earlier prototype for the quickminting. Because ethereum nonces enforce ordering of transactions from the same sender, it's possible for the dapp to submit the quickminting transaction immediately after the deposit transaction, without having to wait for the deposit to be mined first.

beaushinkle: Ohhhhhhh, I think I get it.

When you deposit, you give the depositor a NFT representing their deposit.

The depositor can exchange their NFT for TBTC from the quickminter's treasury (if it has a large enough balance).

When we get around to sweeping the deposits, we mint TBTC to the quickminter's treasury.

The extra layer of indirection means that Mallory is only able to make unsweepable deposits (via DoS) to Wallet B and redeem from *the treasury* instead of just getting minted TBTC straight up. If the treasury has 30 TBTC, she could get all of that, but not everything.

> Because ethereum nonces enforce ordering of transactions from the same sender, it's possible for the dapp to submit the quickminting transaction immediately after the deposit transaction, without having to wait for the deposit to be mined first.

Can you elaborate here?

docs/rfc/rfc-1.adoc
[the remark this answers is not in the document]
pdyraga: ... so it's essentially what @eth-r did yesterday here: https://docs.google.com/spreadsheets/d/1mptuaB8RMxuQxBJTxODhNGSh6WCYmG-LoFwktFG5zWU/edit?usp=sharing

docs/rfc/rfc-1.adoc
pdyraga: Quickminter is an interesting concept but I think it requires two transactions: first one minting NFT, and second one proving the deposit has been swept and minting TBTC. There is also a question of how is the quickminter recharged - is it happening automatically with each successful sweep and if so, who is sending the transaction? Or maybe it is recharged manually by DAO and in this case, how often should it happen, and who should pay for it?

I think it could be worth taking a step back and agree on what we are trying to achieve by sweeping in batches instead of sweeping individually. I think the most important goal is to reduce the fees on Bitcoin transactions and the fact user has time to refund is just a nice side effect.

An alternative could be to let the user decide what is their refund timeout and allow users to claim their TBTC only when the deposit has been swept:
- Eve deposits 1 BTC and sets the refund time to T,
- Eve reveals her Ethereum address and blinding factor off-chain,
- When T comes, the wallet sweeps Eve's UTXO plus other pending UTXOs with similar T
- Eve proves the sweep happened on-chain and receives TBTC.

I think in this model, the amount of TBTC Eve receives could depend on the Bitcoin transaction fee and the dApp could suggest selecting the time T to reduce this cost. If Eve agrees to wait so that her sweep can be batched, she would cut less from her BTC on sweeping fees.

Disclaimer: I have not validated this yet, it's still just a loose idea.

docs/rfc/rfc-1.adoc
pdyraga: I now see we talk about two different things.

The RFC and @beaushinkle talk about redemption fee in TBTC, e.g. redeemer has to provide e.g. 1.1 TBTC to redeem 1 TBTC and 0.1 TBTC is used to buy back T and pay rewards to signers.

I, and @eth-r talk about Bitcoin transaction fee that needs to be paid for the redemption transaction and that needs to be chosen by someone. In tBTC v1 we let the redeemer to choose it and it could be increased by anyone (in practice, signers) in case the transaction could not be minted before the timeout, see [here](https://github.com/keep-network/tbtc/blob/fea1f4470544bc831cb789709eed6c027305cd48/solidity/contracts/deposit/Deposit.sol#L572).

I think I agree with @eth-r that it should be chosen by redeemer with some minimum. I think the minimum should be reasonable enough so that the wallet operators rarely have to increase it. Otherwise, we are requiring them to issue transaction just because the redeemer was too greedy and applied unrealistic constraints.

Side note, it could be a nice feature of the dApp to suggest the fee based on the current market conditions.

docs/rfc/rfc-1.adoc
eth-r: > Quickminter is an interesting concept but I think it requires two transactions: first one minting NFT, and second one proving the deposit has been swept and minting TBTC. There is also a question of how is the quickminter recharged - is it happening automatically with each successful sweep and if so, who is sending the transaction? Or maybe it is recharged manually by DAO and in this case, how often should it happen, and who should pay for it?

My prototype mints the TBTC for the current holder of the UTXO NFT automatically when the wallet sweeps (the wallet submits the second transaction proving that the UTXO has been swept), so the quickminter effectively buys the UTXO NFT from the depositor and receives the minted TBTC when the sweep is performed.

I haven't been able to figure out a way to do it securely without proving the sweep on ethereum so I've been taking the necessity of this second transaction for granted. That's also why I wanted to pay the operators a lump sum of rewards when the sweep transaction is submitted (with the submitter once again getting some extra).

eth-r: > The extra layer of indirection means that Mallory is only able to make unsweepable deposits (via DoS) to Wallet B and redeem from the treasury instead of just getting minted TBTC straight up. If the treasury has 30 TBTC, she could get all of that, but not everything, because she can't keep looping after the treasury runs out.

Yes. That is a critical security feature as it lets us limit our exposure to this attack 

> Can you elaborate here?

For the quickminting, we need 2 transactions: the first one proving the BTC deposit and creating the UTXO NFT, and the second one selling the UTXO to the quickminter. The second transaction needs to know the identifier of the first transaction, and it needs to be executed after the first one.

If we assign each UTXO NFT a GUID that is predictable in advance from a view function, the dapp is able to create both transactions without waiting for the first one to be mined first. The second transaction will simply use the ID received from the view function.

Because ethereum transactions have nonces, it is possible to submit both transactions immediately and the second transaction cannot be mined before the first one.

This gives better UX as the user doesn't need to return later, but can complete the entire operation in a single click.

docs/rfc/rfc-1.adoc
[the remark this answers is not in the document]
pdyraga: Also covered here: https://github.com/keep-network/tbtc-v2/pull/33

docs/rfc/rfc-1.adoc
beaushinkle: I'm sold - I'll add this to the RFC

docs/rfc/rfc-1.adoc
beaushinkle: Can we choose the next random wallet by seeding a RNG based on the hash of the next ethereum block? We don't need a *ton* of security here imo

beaushinkle:
Okay - I think this should be good for another look

docs/rfc/rfc-1.adoc
beaushinkle: Had a chat w/ Matt and Piotr Monday about this. One of the main concerns we want to solve earlier than later is making sure that we build in the freedom to create financial services other than straight minting from a deposit.

For instance, we would like to be able to deposit bitcoin and then use that as collateral to take out TBTC as a loan, rather than selling BTC directly for TBTC. In order to make this work, we'd like to have account balances instead of NFTs that represent UTXO's.

eth-r (approved):
Looks good for a starting point with further refinement in later RFCs

pdyraga (approved):
My personal wishlist for future iterations:
- What is the wallet group size, what is the threshold?
- Are we releasing to the user BTC amount equivalent to the provided TBTCv2 or do we subtract Bitcoin transaction fee?
- Better explanation of the purpose of the blinding factor.
- Do we still need a lock time with the new approach? Is it configurable per user? (I think yes/yes)
- How buying T tokens back from treasury using TBTC fees looks in practice?
- Clarify minting process (no NFT, bridge balances instead)
- Clarify how do we select the wallet to move the funds to in case another wallet is at risk of not providing a heartbeat.
- Spec out heartbeats - this deserves a separate RFC.

I think we will also need to open PRs to this RFC, or at least modify this RFC to refer to new RFCs, for example for heartbeats.

PR #25, opened 2021-08-10
Use MisfundRecovery from @thesis/solidity-contracts

pdyraga:
~~Depends on https://github.com/thesis/solidity-contracts/pull/9~~

We extracted `MisfundRecovery` to `@thesis/solidity-contracts` so that we
can reuse it from `keep-network/tbtc-v2` and `threshold-network/solidity-contracts`.

The functionality stays as-is.

Also, worth mentioning, `ERC20WithPermit` from the updated commit hash
does not allow to transfer tokens to the token address. Most misfund
cases we've seen were about transferring KEEP to KEEP token address. This
can be prevented with one `require` added to `ERC20WithPermit` in the
updated commit hash.


lukasz-zimnoch (dismissed):

solidity/contracts/token/TBTC.sol
```diff
@@ -3,7 +3,7 @@
 pragma solidity 0.8.4;
 
 import "@thesis/solidity-contracts/contracts/token/ERC20WithPermit.sol";
-import "./MisfundRecovery.sol";
+import "@thesis/solidity-contracts/contracts/token/MisfundRecovery.sol";
 
 contract TBTC is ERC20WithPermit, MisfundRecovery {
```
nkuba, line 8: Can we add a simple happy-path test case to check that misfund recovery is possible for `TBTC` contract? I know it's an external contract tested elsewhere but we should check that `TBTC` contract gives a basic functionality expected after extending the `MisfundRecovery` contract.

nkuba (changes requested):

solidity/contracts/token/TBTC.sol
pdyraga: Shouldn't we test ERC20 functions then?

pdyraga:
https://github.com/thesis/solidity-contracts/pull/9 got merged so I updated the dependency to commit hash from `main` and marked the PR as ready for review.

solidity/contracts/token/TBTC.sol
nkuba: Aren't we testing them? I think we should have very simple tests for them too.

pdyraga: I think we should rely on PR review for it, especially that post-deployment contracts are immutable. Green build with simple tests would not be enough - one could overwrite some functions from `ERC20WithPermit` and simple tests might not be able to catch the problem.

nkuba: Let's agree to disagree, I won't block this PR.

nkuba (approved):

PR #33, opened 2021-08-18
RFC 2: group selection and DKG

eth-r:
Specify a group selection protocol using a sortition pool, with the submitter of the expensive sortition transaction being reimbursed, and the frequency of wallet creation being set by the market based on the gas costs of sortition and the available reimbursement amounts.

For distributed key generation, use the protocol from `tss-lib` with a number of attempts over 7 days to avoid expensive re-sortition even if some operators are inactive. Misbehaviour accusations are resolved by governance.

### DIFF
--- a/docs/rfc/rfc-2.adoc
+++ b/docs/rfc/rfc-2.adoc
@@ -1,0 +1,300 @@
+:toc: macro
+
+= RFC 2: tBTCv2 Group Selection and Key Generation
+
+:icons: font
+:numbered:
+toc::[]
+
+== Overview
+
+The key generation uses the protocol from `tss-lib`.
+
+We select members from the sortition pool,
+and they perform one or more attempts to complete the key generation.
+
+Inactive members are removed from the DKG participants
+and the resulting wallet if the DKG is successful.
+
+If there is misbehaviour,
+the accusations will be resolved by governance
+as the protocol in `tss-lib` is not attributable.
+
+=== Parameter selection
+
+For tbtc2 wallets,
+we need to choose the group size _N_
+(how many members we select from the sortition pool)
+and the threshold _T_
+(how many members must cooperate to generate a threshold ECDSA signature.
+
+In addition, we need to choose the critical size _C_
+which is number of members required for a successful key generation.
+Some of the selected _N_ members may not be available for the key generation.
+To reduce sortition costs,
+we let some of the original _N_ be inactive
+as long as at least _C_ participants can be assembled to complete the DKG.
+After successful DKG completion,
+the created wallet will have a number of members
+that is between _C_ and _N_ (inclusive).
+The inactive members that did not participate in the successful DKG
+are removed from the wallet's members and are ineligible for rewards.
+
+If we cannot complete DKG successfully
+with at least _C_ participants
+within the allowed time,
+the failed wallet will not be used by TBTC.
+
+==== Group size and threshold
+
+Assuming we create groups every 24h and the lifespan of TBTC is 100 years,
+we would perform 36,500 sortitions in total.
+If an adversary has a 1:1,000,000 probability of controlling any given group,
+they'd have a roughly 3.6% probability of controlling one group at some point.
+
+Assuming 51/100 (N = 100, T = 51) would make an acceptably secure wallet,
+we can compare it to other parameter choices of similar security.
+
+It can be generally observed,
+that a slight increase in the T:N ratio
+compensates for a fairly large reduction in N (and T).
+A lower _N_ significantly reduces sortition costs.
+The downside of this is that the group has fewer redundant members
+which requires greater reliability from each individual node.
+In addition,
+a malicious actor with more than _N - T_ members
+is guaranteed to be able to DoS the group.
+In practice this is less significant
+as non-attributable threshold ECDSA protocols are vulnerable to DoS
+by an adversary controlling a significantly smaller number of members.
+
+From modeling we get the following security levels for different parameters
+against adversaries of various power:
+
+[%header,cols=7]
+|===
+|Parameters |Bits against 0.15 |0.2 |0.25 |0.3 |0.35 |0.4
+
+|76/150 |>54 |>54 |36.74 |23.64 |14.23 |7.66
+
+|51/100 |>54 |38.00 |25.74 |16.91 |10.49 |5.94
+
+|41/80 |>54 |31.19 |21.36 |14.21 |8.97 |5.23
+
+|44/80 |>54 |37.90 |26.78 |18.56 |12.38 |7.78
+
+|36/70 |40.30 | 27.82 |19.16 |12.85 |8.21 |4.87
+
+|40/70 |>54 |37.01 |26.66 |18.93 |13.03 |8.54
+
+|26/50 |29.95 |21.07 |14.75 |10.11 |6.66 |4.14
+
+|32/50 |>54 |36.07 |27.29 |20.53 |15.20 |10.95
+|===
+
+The probability of an adversary controlling the given fraction of operators
+controlling at least _T_ members of _N_ randomly selected operators
+equals `1 / 2^bits`.
+
+The given numbers are calculated using a hypergeometric distribution
+out of 10,000 operators;
+using a binomial distribution instead has insignificant effect.
+
+==== Critical size
+
+We don't expect all _N_ members to be available for key generation.
+However, to ensure sufficient availability
+we require at least _C_ members to complete key generation.
+
+A _C_ closer to _N_ prevents the creation of low-quality wallets
+which would be at risk of low availability
+at the cost of increasing the risk of key generation failing to complete.
+A lower _C_ increases the odds of successfully creating a wallet.
+
+In any case,
+_C_ should be sufficiently above _T_
+that created wallets are adequately reliable in use
+and the risk of losing a wallet in its entirety
+by the number of active operators falling below _T_
+is minimised.
+
+=== Group selection
+
+At any time,
+anyone can submit (and pay for) a transaction creating a new wallet.
+This transaction uses the sortition pool to select N members
+and creates a new wallet with the selected members,
+instructing them to begin key generation.
+The selected members may contain duplicate operators.
+The new wallet is added to the pending wallets.
+
+The submitter of this sortition transaction is rewarded.
+
+==== Sortition transaction reimbursement
+
+When TBTC is collected in deposit fees,
+a certain fraction (e.g. 10%) of the fee is placed in the sortition fee pool.
+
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,300 @@
...
+Instead, the group signs and submits an on-chain transaction
+announcing that key generation was aborted due to disqualification,
+and identifying the participants in the aborted attempt,
+the members who declared a disqualification abort
+and the members who were accused of sending invalid messages.
+All participants will also store their logs of the aborted attempt
+to facilitate later investigation.
+These logs must contain all messages sent and received by the participant,
+private keys the participant had established with all other participants,
+and all secret values the participant generated for the key generation attempt.
```
pdyraga, line 256: I would probably start with a more lightweight approach: do not submit an on-chain transaction and just log a warning in logs of all clients. This needs to be resolved off-chain anyway.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,300 @@
...
+
+When an attempt at key generation finishes successfully,
+a transaction announcing this result is submitted on-chain.
+The transaction contains the public key from the successful DKG attempt,
+the list of participants in that successful attempt,
+and signatures from a majority of participants.
+
+The on-chain contract verifies
+that all listed participants are among the originally selected members,
+and that all provided signatures are valid signatures from different members.
```
pdyraga, line 281: Do we expect signatures from all members not marked as inactive (previously validating this number >=  C)? Or are C signatures enough?

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,300 @@
...
+If the submission passes these checks,
+the newly created wallet is removed from the pending wallets
+and appended to the list of active wallets,
+its state is set to _Active_,
+and its activation time is recorded.
+
+If some participants were inactive in the key generation,
+it might be desirable to remove them from the sortition pool.
```
pdyraga, line 289: Agreed. Right now there is no way to kick someone out and I think having this possibility is desirable. My only worry is that it probably needs to happen in a loop with up to N-C steps. If N-C is small enough, we should be fine. If not, we'll have to figure out something else.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,300 @@
...
+|44/80 |>54 |37.90 |26.78 |18.56 |12.38 |7.78
+
+|36/70 |40.30 | 27.82 |19.16 |12.85 |8.21 |4.87
+
+|40/70 |>54 |37.01 |26.66 |18.93 |13.03 |8.54
+
+|26/50 |29.95 |21.07 |14.75 |10.11 |6.66 |4.14
+
+|32/50 |>54 |36.07 |27.29 |20.53 |15.20 |10.95
+|===
```
pdyraga, line 93: Could be worth comparing estimated sortition cost cc @beaushinkle 

Off-chain, we have much more flexibility. We have not tried with `tss-lib` yet but for the random beacon GJKR and the same broadcast channel implementation, we were able to generate 128 member groups with no problems.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,300 @@
...
+
+In this RFC, the word "majority" means at least _T_ out of _N_ operators,
+if _T_ is greater than _N/2+1_.
+
+==== Two-phase retry protocol
+
+The full key generation protocol consists of individual _attempts_.
+In each attempt,
+the specified participants try to perform DKG
+with the protocol from `tss-lib`.
```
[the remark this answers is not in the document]
eth-r: That sounds like a good idea

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,300 @@
...
+If the attempt fails due to one or more participants being inactive,
+the inactive participants are removed from the DKG participants
+and another attempt is made.
+
+If the number of DKG participants falls below the critical size _C_,
+no further attempt is made by the remaining participants.
+Instead, the group will clear the list of inactive participants
+and wait 12h before making another attempt with the entire group.
```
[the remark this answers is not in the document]
eth-r: The idea is that operators in different time zones have the opportunity to be alerted and respond; this was discussed in the thread and 12h was mentioned as the minimum time to cover timezones, so it was picked up as the timing for the hybrid retry system. If we don't have tight constraints for the ultimate timeout, 12h retries gives around 13 rounds of attempts in the 7 day period, and between each round there is a good chance that any long-term problem affecting a significant operator would be noticed and addressed.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,300 @@
...
+This repeats until 7 days have elapsed
+since the timestamp of the sortition transaction.
+If this ultimate timeout is reached without succesful key generation,
+the group dissolves and makes no further attempts.
```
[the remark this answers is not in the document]
eth-r: Yes, that's the idea and that's why we can afford the 7 day ultimate timeout. It creates some of the cross-cutting concerns earlier mentioned wrt the ordering of wallets and dealing with multiple key generations, but it isn't particularly difficult to address especially if we add some sanity checks (e.g. a maximum number of concurrently active DKGs so that someone can't spend lots of gas money to grief the system by creating too many pending wallets)

docs/rfc/rfc-2.adoc
eth-r: If there are on-chain repercussions I think we should have an on-chain signal, for both transparency (everyone can see when misbehaviour happens and there is no way to hide things behind the scenes) and the ability to limit the powers governance has (any governance-imposed punishments can only apply to operators named in an abort message signed by T members of a wallet)

docs/rfc/rfc-2.adoc
eth-r: T signatures is enough

docs/rfc/rfc-2.adoc
eth-r: Adding the ability for a sortition pool owner to kick/ban operators (kick just removes and should be used for inactivity, ban permanently stops them joining) should be enough and pretty simple.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,300 @@
...
+the amount released to the submitter is reduced,
+and the remainder is released as general rewards
+to other participants in the TBTC system.
+
+This reduction can be linear
+(the fraction released equals the fraction of 24h that has passed),
+quadratic/cubic
+(the fraction released equals the square/cube of the time passed)
+or exponential
+(for every hour before 24h has passed, the released fraction is halved).
```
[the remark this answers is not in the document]
eth-r: Exponential ensures that we never create new wallets massively more often than every 24h (unless someone wastes money) but it has a bit of sharpness (every hour the reward doubles, so you can get a rush of submissions trying to hit that specific block)

Polynomial (quadratic/cubic) is a bit weaker in preventing rapid wallet creation, and it's also nicely smooth

Linear is very simple and ensures that if TBTC gets really popular the amount of BTC in any individual wallet doesn't grow too much (but it still grows if the natural creation rate would get faster than 1/day) but it also provides the least reduction in wallet creation rate so we might get more concurrent key generations

I don't have a preference yet

docs/rfc/rfc-2.adoc
eth-r: I think sortition pools are fairly linear so if 150 members costs 3M gas, 100 members should be 2M gas, 70 members should be around 1.4M gas, and 50 members should be around 0.5M gas

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,300 @@
...
+which is number of members required for a successful key generation.
+Some of the selected _N_ members may not be available for the key generation.
+To reduce sortition costs,
+we let some of the original _N_ be inactive
+as long as at least _C_ participants can be assembled to complete the DKG.
+After successful DKG completion,
+the created wallet will have a number of members
+that is between _C_ and _N_ (inclusive).
+The inactive members that did not participate in the successful DKG
+are removed from the wallet's members and are ineligible for rewards.
```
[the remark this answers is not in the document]
eth-r: Modifying sortition pools to accommodate constant rewards would be quite a bit of work but it can be done. The penalty for inactive members would then come from being removed from the pool and having to rejoin and wait the initialisation period before receiving rewards.

Rewarding the members of a wallet when the wallet sweeps deposits/executes redemptions would be simpler and we could use similar logic as in the random beacon to add extra reward for transaction submitters, but it makes the rewards for any given operator a lot more random. A hybrid approach (some reward to all, some to members) is also possible.

docs/rfc/rfc-2.adoc
beaushinkle: Ran the tests with 100 iterations against the simplest possible staking contract and got:

```
150 members
mean	3208155
stddev	31427

100 members
mean	2150957
stddev	43262

70 members
mean	1508201
stddev	15485

50 members
mean	1085133
stddev	2532
```
linearity: confirmed

docs/rfc/rfc-2.adoc
eth-r: Given that banning an operator and removing them costs around 75k gas, I think we have room for some looping. Agreed that we should ensure that we can afford it in any possible circumstance.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
beaushinkle: Want to start with polynomial and make governance able to move between them?

beaushinkle:
Did a fully line-by-line here and didn't find anything that stuck out at me :tada:

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
beaushinkle: Any thoughts on how to pick one here so that the implementors can implement when the time comes?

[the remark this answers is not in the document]
eth-r: This doesn't need to be called often, so we could use a simple external contract like the [staking policies of KEEP grants](https://github.com/keep-network/keep-core/blob/main/solidity/contracts/GrantStakingPolicy.sol) and let governance switch the address it's pointing at

pdyraga (approved):
:+1: There are two more questions I'd like to ask but we can address them later:
- Should governance have a way to ban operators from the sortition pool?
- For modeling security levels with the given group size and threshold we assumed groups are created every 24h but section 1.2.1 says that new groups are created if enough T accumulates in a special pool. How much security levels are affected if we assume that for a month or two there are no new deposits in the network and no wallets are created?

Issue #38, opened 2021-08-30
Staker rewards


pdyraga:
In tBTC v2, signers will be rewarded continuously. 

@eth-r proposed to implement the following algorithm on the top of the sortition pool:
- we keep a monotonously increasing counter of weight-seconds of the pool `P`
- this is updated by multiplying the total weight of the pool with the number of seconds since the last update, whenever the pool weight changes
- we keep a monotonously increasing counter of total rewards paid into the pool `R`
- this is updated whenever rewards are paid in
- each operator tracks the amount of weight-seconds they've contributed to the pool `W`
- they also track the amount of total rewards that had accumulated in the pool when they last withdrew rewards `Rw`, and the amount of weight-seconds they had contributed at that time `Ww`, and the total weight-seconds in the pool at that time `Pw`

when an operator withdraws rewards, we take the amount of newly accumulated rewards in the pool since their last withdrawal `R - Rw`, calculate their contribution to the accumulated weight-seconds in that time `W - Ww / P - Pw`, and multiply the reward difference with the contribution fraction

then we update `Rw = R, Ww = W, Pw = P`

when an operator joins the pool, we probably need to do a synchronisation step where we add and withdraw imaginary rewards that equal the amount the operator would've received if they had been present in the pool right from the beginning:

T = 0:Alice joins the pool with 100 weight
T = 10: 20 rewards are paid in
T = 20: Bob joins the pool with 50 weight; the pool has 2,000 weight-seconds and 20 rewards accumulated, and Bob's weight is 50% of the current weight so we add 50% to both the weight-seconds and accumulated rewards, setting them to 3,000 and 30 respectively, and record that Bob "withdrew" when `P = 3,000; R = 30` having "contributed" `W = 1,000`
T= 30: Alice withdraws rewards, `P = 4,500 and Alice's W = 3,000`, so Alice gets 2/3 of the "accumulated" 30 rewards = 20 rewards are paid out to Alice
T = 40: 30 rewards are paid in, `R = 60`
T = 50: Bob withdraws rewards, `P = 7,500` and Bob's `W = 2,500`; because Bob's `Ww = 1,000; Rw = 30, Pw = 3,000` we get a difference of 1,500 contributed weight-seconds, 30new reward accumulation, and 4,500accumulated weight-seconds poolwide, thus Bob withdraws `(1,500 / 4,500) * 30 = 10` rewards

[...]

PR #42, opened 2021-08-30
RFC 3: Rewards

eth-r:
Specify a broad outline of how rewards are paid to sortition pool members, to wallet members for performing specific actions, and to accounts performing necessary maintenance tasks. #38

### DIFF
--- a/docs/rfc/rfc-3.adoc
+++ b/docs/rfc/rfc-3.adoc
@@ -56,20 +56,13 @@ and setting their `rewardAvailable = 0`.
 Whenever a wallet is required to perform some transaction on-chain,
 a _transaction reward_ is reserved for the wallet.
 When the transaction is correctly submitted,
-the transaction reward is divided
-between the submitter and the rest of the wallet members.
+the submitter of the transaction receives a fraction of the transaction reward.
 If the transaction is submitted immediately,
-the full transaction reward is paid out
-and the non-submitter members receive a larger share.
-If the transaction is submitted later,
-the submitter gets a larger reward,
-the other members receive less,
-and some of the reward is withheld from the wallet members altogether.
-The withheld amount is either paid to the sortition pool,
-or kept for future transaction rewards.
-
-For a precise calculation,
-we can use e.g. the Random Beacon entry submission reward rules.
+the size of the reward is zero.
+The transaction reward grows until the _maximum reward time_ has elapsed.
+From that point to the applicable timeout,
+the submitter of the transaction would receive the entire transaction reward.
+Any leftover reward is paid to the sortition pool.
 
 ==== Earmarking
 
--- a/docs/rfc/rfc-3.adoc
+++ b/docs/rfc/rfc-3.adoc
@@ -20,6 +20,8 @@ as a member of an active wallet.
 In addition, any Ethereum account may be rewarded
 for performing various necessary maintenance tasks.
 
+All reward allocations are governable.
+
 === Sortition pool rewards
 
 All operators present in the sortition pool
--- a/docs/rfc/rfc-3.adoc
+++ b/docs/rfc/rfc-3.adoc
@@ -1,0 +1,119 @@
+:toc: macro
+
+= RFC 3: Rewards
+
+:icons: font
+:numbered:
+toc::[]
+
+== Overview
+
+When various actions are taken in the tBTC system,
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-3.adoc
```diff
@@ -0,0 +1,119 @@
...
+If someone were to deposit 1 BTC,
+the fee would be 0.003 BTC
+of which 0.0005 BTC would be earmarked for sweeping
+and another 0.0005 BTC for redemption.
+These fees could be collected in the form of TBTC tokens
+immediately upon sweeping the deposit transaction
+and either paid out (sweeping reward)
+or kept in the wallet for future use (redemption reward).
+The remaining 0.002 TBTC
+would be used for sortition pool and maintenance rewards.
```
beaushinkle, line 96: Do we want these allocations to be governable?

docs/rfc/rfc-3.adoc
```diff
@@ -0,0 +1,119 @@
...
+with some fraction (or all) of the maintenance reward pool for the task,
+according to the reward rules of that particular maintenance task.
+
+If a part of the maintenance reward is withheld
+the withheld amount should be paid out as a sortition pool reward.
+This is distinct from paying out a fraction;
+e.g. the maintenance reward pool might pay 20% of its contents each time,
+but in some circumstances 50% of this might be withheld,
+resulting in 10% being paid to the submitter of the transaction
+and another 10% being withheld and paid to the sortition pool instead.
```
beaushinkle, line 119: I think it's worth writing out the incentive behind the submitter/sortition pool split (and how that helps the market determine how often maintenance needs to be done).

docs/rfc/rfc-3.adoc
eth-r: Yes, I think. I can add a more thorough specification for the fees.

docs/rfc/rfc-3.adoc
```diff
@@ -0,0 +1,119 @@
...
+In addition, each operator's `weight` can be queried.
+Whenever an operator's weight changes from `oldWeight` to `newWeight`,
+the rewards earned with the old weight are collected:
+`rewardAvailable += (accumulatedWeightReward - weightRewardCollected) * oldWeight`,
+`weightRewardCollected = accumulatedWeightReward`
+and `weight = newWeight`.
+
+At any time, an operator can withdraw their collected rewards,
+transferring `rewardAvailable` T tokens to them,
+and setting their `rewardAvailable = 0`.
```
pdyraga, line 52: In [RFC 1](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#continuous-fees) we say that:

> The system is able to collect fees in the form of tBTC when the BTC is deposited and when it is redeemed. Rather than paying operators at those two instances we would prefer to pay them continuously. Since we do not have the tBTC to pay them with until we collect the redemption fee (and can’t mint it and maintain the peg), we must turn to another token. Instead, the system pays operators in T tokens from a treasury funded by the T-token DAO and whenever it collects tBTC fees it uses those fees to buy T tokens back for its treasury.

Can we explain in this RFC how 

> whenever it collects tBTC fees it uses those fees to buy T tokens back for its treasury

should work?

docs/rfc/rfc-3.adoc
```diff
@@ -0,0 +1,119 @@
...
+
+=== Wallet action rewards
+
+Whenever a wallet is required to perform some transaction on-chain,
+a _transaction reward_ is reserved for the wallet.
+When the transaction is correctly submitted,
+the transaction reward is divided
+between the submitter and the rest of the wallet members.
```
pdyraga, line 60: How do we divide the reward between wallet members? It sounds similar to per-group beacon rewards that we know turned to be problematic because of withdrawal costs. Can we just reward the submitter and if there is anything left (because of a delay penalty, for example), return it to pool-wide rewards pool? 

(similar approach was proposed in https://github.com/keep-network/keep-core/pull/2585)

docs/rfc/rfc-3.adoc
pdyraga: I think we should mark it explicitly in the RFC they are expected to be governable so that we do not lose it during the implementation.

beaushinkle:
I'm going to close this, and if we ever want to re-open it, it'll be here

Issue #36, opened 2021-08-30
Calling the random beacon


pdyraga:
In tBTC v1, random beacon provided a new relay entry asynchronously, after sortition pool selected group members, and after new bonded ECDSA keep was created. Random beacon was getting called for each new deposit and in case it did not provide an entry, new deposits could get created anyway.

We need to define when and how the random beacon is called for tBTC v2.

Questions we need to answer (and there is more for sure):
- When is the random beacon called?
- Can a new entry be provided after the group selection?
- What happens if the beacon is busy?

eth-r:
If we leave group selection up to the market to do based on the reward:gas ratio, I think the best way to do it would be something approximately as follows:

- anybody can call the "start new group selection" function on tBTC, which:
  1. requires that a governance-defined minimum payment is provided
  2. calls the random beacon for a new entry passing the payment onward along with gas limit etc.
  3. if the beacon is not busy, records who called this function for future use and blocks the "start new group selection" function from being called
- now the random beacon produces an entry, and once it is done it calls tBTC with the new entry:
  1. tBTC performs the actual sortition
  2. a new wallet is formed
  3. the original caller is reimbursed if everything is successful and "start new group selection" becomes available again

This is able to handle busy beacon, unsuccessful sortition, etc. and retry gracefully without being too vulnerable to griefing by e.g. providing insufficient gas money to the beacon. Using the random beacon entry immediately is also better for preventing sortition pool manipulation and other shenanigans.

PR #43, opened 2021-08-30
RFC 4: Random Beacon

eth-r:
We can use the random beacon as follows:

- anybody can call the "start new group selection" function on tBTC, which:
  1. requires that a governance-defined minimum payment is provided
  2. calls the random beacon for a new entry passing the payment onward along with gas limit etc.
  3. if the beacon is not busy, records who called this function for future use and blocks the "start new group selection" function from being called
- now the random beacon produces an entry, and once it is done it calls tBTC with the new entry:
  1. tBTC performs the actual sortition
  2. a new wallet is formed
  3. the original caller is reimbursed if everything is successful and "start new group selection" becomes available again

This is able to handle busy beacon, unsuccessful sortition, etc. and retry gracefully without being too vulnerable to griefing by e.g. providing insufficient gas money to the beacon. Using the random beacon entry immediately is also better for preventing sortition pool manipulation and other shenanigans. #36


beaushinkle:
I'm good to go here :+1:

beaushinkle (approved):

docs/rfc/rfc-4.adoc
```diff
@@ -0,0 +1,44 @@
...
+If the beacon is busy,
+or the provided payment is insufficient,
+the transaction reverts.
+
+Once the random beacon has generated an entry,
+it calls `selectGroup(newEntry)` with a high gas limit.
+This performs sortition from the sortition pool,
+and if successful, creates a new wallet with the selected members
+and pays the appropriate maintenance reward to Alice.
+In any case, `currentWalletCreator` is cleared.
```
pdyraga, line 44: In https://github.com/keep-network/keep-core/pull/2585 it is proposed to change the beacon callback mechanism so that the given application's maintainers need to submit a transaction using an entry. If we agree on this approach, can we rework this section to say Alice is calling the second transaction and to specify if there is any expiration period for an entry delivered in the previous transaction? Also, it feels that we will have to lock the sortition pool on the first transaction from Alice, given the most recent idea from @eth-r about optimistic selection of members.

pdyraga (approved):

Issue #35, opened 2021-08-30
Moving funds between wallets


pdyraga:
We need to decide when and how funds are moved between wallets. In #34 we defined a heartbeat protocol that may fail completely or that may identify that the number of active members is below a certain threshold. Additionally, new wallets are being created and old wallets are getting older. In some circumstances, funds should be moved between wallets. What are those circumstances and how the fund move operation is working?

Questions we need to answer (and there is more for sure):
- When do we move funds between wallets?
- How the on-chain transaction looks like?
- What transaction is submitted to the chain and who is submitting it? Is the submitter reimbursed? If so, how?
- What is happening with not-yet swept funds? 
- How the wallet proves funds were moved?
- What is happening if the wallet proved funds are moved?
- What is happening if the wallet did not provide proof? 
- What is the timeout? What is happening if it is exceeded?

PR #44, opened 2021-08-30
RFC 5: Emptying and closing wallets

eth-r:
When a wallet reaches a maximum age, falls below a minimum amount of held BTC, falls below a minimum number of operators, or fails to complete any operation within the allowed timeout, it is to be emptied out with any remaining BTC transferred out.

The transaction emptying out a wallet is treated like a BTC redemption. #35

### DIFF
--- a/docs/rfc/rfc-5.adoc
+++ b/docs/rfc/rfc-5.adoc
@@ -23,7 +23,8 @@ and must produce a transaction moving its remaining BTC to a newer wallet.
 
 === Emptying wallets
 
-When a wallet has began `Emptying`
+When a wallet has began `Emptying`,
+either from its on-chain state being updated or from a timeout being reached,
 the following things change:
 
 - If the wallet was previously a valid deposit wallet,
--- a/docs/rfc/rfc-5.adoc
+++ b/docs/rfc/rfc-5.adoc
@@ -99,6 +99,8 @@ the wallet`s status is changed on-chain to `Emptying`.
 If a wallet's heartbeat is signed
 by less than the specified minimum number of active members,
 the wallet's status is changed on-chain to `Emptying`.
+Undelegating members are also counted as inactive
+for the purposes of determining a weak heartbeat.
 
 ==== Failed heartbeat or other operation
 
--- a/docs/rfc/rfc-5.adoc
+++ b/docs/rfc/rfc-5.adoc
@@ -1,0 +1,121 @@
+:toc: macro
+
+= RFC 5: Emptying and closing wallets
+
+:icons: font
+:numbered:
+toc::[]
+
+== Overview
+
+If a wallet still containing BTC is to be closed,
+the remaining BTC must be transferred to a new wallet if possible.
+
+This applies in the following situations:
+
+- a wallet has fallen below a specified minimum BTC amount
+- a wallet has failed a heartbeat due to losing too many operators
+- a wallet has reached a specified maximum lifetime
+
+When one of these _closure criteria_ is met,
+the wallet changes state to `Emptying`
+and must produce a transaction moving its remaining BTC to a newer wallet.
+
+=== Emptying wallets
+
+When a wallet has began `Emptying`
+the following things change:
+
+- If the wallet was previously a valid deposit wallet,
+it stops being a valid deposit wallet.
+- No redemptions can be made from the wallet.
+- The wallet has 24h to sweep any remaining unswept deposits
+and 48h to transfer its remaining BTC to a newer wallet.
+- The time the wallet began emptying may be recorded,
+or it may be calculated from the time the wallet reached its maximum age
+or reached a timeout in some operation.
+
+==== Sweeping and rejecting deposits
+
+In the scenario where a previously valid deposit wallet begins emptying,
+it stops receiving deposits.
+Any deposits erroneously made to an emptying wallet after it began emptying
+will be rejected and not add to the depositor's BTC balance,
+and must be allowed to time out.
+
+Any deposits made before the wallet began emptying,
+if the wallet was a valid deposit wallet when the deposit was made,
+must be swept within 24h of the wallet beginning to empty.
+This overrides any other deposit sweeping timeouts
+that would give the wallet more time to complete the sweep.
+If the deposit had an earlier timeout,
+the original timeout is unchanged.
+The submitter/s of transaction/s proving the sweeps are rewarded as usual.
+
+Failure to sweep a remaining deposit within the allowed 24h
+is treated like any other failure to sweep.
+
+==== Transferring remaining BTC
+
+An emptying wallet has 48h to move its remaining BTC to a newer wallet.
+The recipient wallet can either be the most recent valid deposit wallet,
+or it can be chosen pseudorandomly
+from some number of most recently created wallets.
+
+This is done in the form of a simple transfer from the emptying wallet
+to the recipient wallet.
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-5.adoc
```diff
@@ -0,0 +1,121 @@
...
+=== Emptying wallets
+
+When a wallet has began `Emptying`
+the following things change:
+
+- If the wallet was previously a valid deposit wallet,
+it stops being a valid deposit wallet.
+- No redemptions can be made from the wallet.
+- The wallet has 24h to sweep any remaining unswept deposits
+and 48h to transfer its remaining BTC to a newer wallet.
```
beaushinkle, line 33: Any specific reason why you chose the 48h number here?

docs/rfc/rfc-5.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+== Overview
+
+If a wallet still containing BTC is to be closed,
+the remaining BTC must be transferred to a new wallet if possible.
+
+This applies in the following situations:
+
+- a wallet has fallen below a specified minimum BTC amount
+- a wallet has failed a heartbeat due to losing too many operators
```
pdyraga, line 17: Has it completely failed a heartbeat or has it provided the heartbeat but with a number of active members below the critical number?

To clarify, I think we have two "critical" numbers. One is the minimum number of active members during the DKG so that the result can be accepted, let's call it C_dkg. The second number is the minimum number of wallet members actively signing the heartbeat, let's call it C_h. 

T < C_h < C_dkg < N. 

If the number of members signing the heartbeat falls below C_h, we can consider the wallet at risk and start moving the funds before T is reached. It is especially important for how do we want to implement undelegations.

docs/rfc/rfc-5.adoc
```diff
@@ -0,0 +1,121 @@
...
+- a wallet has reached a specified maximum lifetime
```
pdyraga, line 18: ...which should always be shorter than the undelegation time, right? 

Say we are at C_h + 1 and one whale undelegates and that whale was lucky enough to have more than C_h - T members in the group and the wallet is now under T and can neither provide a heartbeat nor the transaction to move funds.

docs/rfc/rfc-5.adoc
```diff
@@ -0,0 +1,121 @@
...
+if the wallet was a valid deposit wallet when the deposit was made,
+must be swept within 24h of the wallet beginning to empty.
+This overrides any other deposit sweeping timeouts
+that would give the wallet more time to complete the sweep.
+If the deposit had an earlier timeout,
+the original timeout is unchanged.
+The submitter/s of transaction/s proving the sweeps are rewarded as usual.
+
+Failure to sweep a remaining deposit within the allowed 24h
+is treated like any other failure to sweep.
```
pdyraga, line 56: We do nothing because depositors can use the refund option from their BTC script, correct?

docs/rfc/rfc-5.adoc
```diff
@@ -0,0 +1,121 @@
...
+- a wallet has reached a specified maximum lifetime
+
+When one of these _closure criteria_ is met,
+the wallet changes state to `Emptying`
+and must produce a transaction moving its remaining BTC to a newer wallet.
+
+=== Emptying wallets
+
+When a wallet has began `Emptying`
+the following things change:
```
pdyraga, line 27: How is the wallet entering the state? If there a transaction to Ethereum submitted? If so, who is submitting it and what should it contain? Is the submitter receiving maintenance reward?

docs/rfc/rfc-5.adoc
pdyraga: OK, I see it's described in section 1.2.2. Weak heartbeat and 1.2.3. Failed heartbeat or other operation.

docs/rfc/rfc-5.adoc
```diff
@@ -0,0 +1,121 @@
...
+If a wallet's heartbeat is signed
+by less than the specified minimum number of active members,
+the wallet's status is changed on-chain to `Emptying`.
+
+==== Failed heartbeat or other operation
+
+If the wallet fails to submit a valid heartbeat within the allowed timeout
+the wallet counts as `Emptying` without a change in on-chain status,
+and is considered to have started `Emptying`
+when the heartbeat timeout was reached.
```
pdyraga, line 107: Does it mean heartbeats need to be submitted to Ethereum periodically?

If anything should happen on Ethereum side in case of a failure to complete a BTC redemption, it seems that the wallet's health status needs to be known there.

beaushinkle:
Clean and concise :tada:

docs/rfc/rfc-5.adoc
beaushinkle: I'm not sure if it belongs in this RFC specifically, but I think we talked about counting pending undelegating operators as inactive for determining weak heartbeats (so they'd make the heartbeat test fail but still be able to sign the transaction to move the funds.

This means that in the scenario where, say, the undelegation time was 2 months and the max wallet time was 1 year, I think we're still okay:

```
month 1: wallet gets created with a whale that has 40% of the stake
month 6: whale undelegates
month 6: weak heartbeat detected because whale's stake is marked as inactive
month 6: underlying btc is able to get transferred to a different wallet because the whale can still sign messages
```

docs/rfc/rfc-5.adoc
beaushinkle: We slash the operators, I believe. Normally they have more time before slashing starts happing, but in this case we're saying that slashing happens sooner.

docs/rfc/rfc-5.adoc
eth-r: 24h to sweep deposits, another 24h to complete the redemption. If we have a different standard timeout we can update this to be 1x and 2x the standard timeout to match.

docs/rfc/rfc-5.adoc
pdyraga: That makes sense. Undelegating operators are not participating in the heartbeat but they will sign any redemption transaction. I think it is worth noting in this RFC to avoid confusion.

docs/rfc/rfc-5.adoc
pdyraga: Do we have slashing rules listed somewhere? I think it could be a good candidate for another RFC.

eth-r: Slashing is upcoming

docs/rfc/rfc-5.adoc
eth-r: If a BTC redemption fails, you already know that the required number of operators weren't able to complete the redemption in time, which implies that they wouldn't be able to complete a heartbeat either.

We don't necessarily need to submit heartbeats periodically, we can have optimistic heartbeats where submission is only necessary if the group is challenged; someone claims that the group has fallen below `C_h` in off-chain signing, and as a consequence the group is expected to either:

- produce a strong heartbeat (above `C_h`) within the allowed timeout period, continuing as normal
- produce a weak heartbeat (below `C_h` but able to produce signatures) within the timeout, resulting to the wallet being emptied to a new one
- fail to produce a heartbeat, resulting in the wallet being emptied if possible but probably losing the funds

Where appropriate, we can reward the operators/submitters for these

beaushinkle:
Closing this, since this work has been fleshed out by https://github.com/keep-network/tbtc-v2/pull/127

PR #45, opened 2021-09-01
Groupsize Seeding

beaushinkle:
This PR attempts to formalize some of the conversations we've had around how to pick the initial parameters for group selection.

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -155,7 +155,8 @@ days (specified as `locktime`).
 After the deposit transaction has been mined, the user is able to reveal their
 ethereum address and blinding factor to the ethereum chain. The bridge listens
 for these sorts of messages and when it sees one, is able to generate a script that
-can spend the funds. Once successful, we increase the account's unswept balance.
+can spend the funds. Once successful, we increase the account's unswept balance
+and charge a deposit fee.
 
 Second, we schedule an operation that batches all outstanding known-refundable
 transactions together to be combined with the existing wallet output into a
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -46,6 +46,7 @@ collects tBTC fees it uses those fees to buy T tokens back for its treasury.
 [[group-size]]
 === Group Size
 There are three main governable parameters that we need to choose:
+
 * The number of signers required to sign messages `threshold`
 * The number of group members required for a <<heartbeat,heartbeat>> to be considered "strong". `heartbeat`
 * The minimum number of members we're allowed to drop down to during the DKG
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -67,14 +67,15 @@ There are three main governable parameters that we need to choose:
 `threshold < heartbeat < dkg < groupSize`
 
 The total gas cost for sortiton increases linearly with group size by at least
-20k gas per member. Requiring more signers to sign messages makes it harder for
-adversaries to take over the group, but makes it so that the pool is less
-resilient to undelegation or operators going offline. Raising the
-strong-heartbeat threshold decreases the chance that you'll not have enough
-signers to sign transactions, but decreases a wallets lifespan and increases
-overhead. Lowering the dkg threshold increases the chance that we're able to
-create a wallet successfully but gives us a wallet with potentially fewer
-operators and a lower lifespan.
+20k gas per member (though, we think we can get this down to roughly 6k gas per
+member in the happy path with some upcoming optimizations). Requiring more
+signers to sign messages makes it harder for adversaries to take over the
+group, but makes it so that the pool is less resilient to undelegation or
+operators going offline. Raising the strong-heartbeat threshold decreases the
+chance that you'll not have enough signers to sign transactions, but decreases
+a wallets lifespan and increases overhead. Lowering the dkg threshold increases
+the chance that we're able to create a wallet successfully but gives us a
+wallet with potentially fewer operators and a lower lifespan.
 
 Also important is optics - even though a group size of 70 with 40 signers
 required to sign messages equally secure against being controlled by an
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -10,7 +10,7 @@ toc::[]
 
 === Minting
 
-Stakers periodically create 51-of-100 (subject to change) ecdsa-backed wallets
+Stakers periodically create <<group-size,51-of-100>> ecdsa-backed wallets
 to hold frozen BTC assets to maintain the peg. Depositors send BTC funds to the
 most-recently-created-wallet by using pay-to-script-hash (P2SH) which contains
 hashed information about the depositor's minting ethereum address. Once the
@@ -42,6 +42,34 @@ operators in T tokens from a treasury funded by the T-token DAO and whenever it
 collects tBTC fees it uses those fees to buy T tokens back for its treasury.
 
 == In Depth
+
+[[group-size]]
+=== Group Size
+There are three main governable parameters that we need to choose:
+* The total size of the group
+* The number of signers required to sign messages
+* The number of group members required for a <<heartbeat,heartbeat>> to be considered "strong".
+
+The total gas cost for sortiton increases linearly with group size by at least
+20k gas per member. Requiring more signers to sign messages makes it harder for
+adversaries to take over the group, but makes it so that the pool is less
+resilient to undelegation or operators going offline. Raising the
+strong-heartbeat threshold decreases the chance that you'll not have enough
+signers to sign transactions, but decreases a wallets lifespan and increases
+overhead.
+
+Another important thing to keep in mind is optics - even though a group size of
+70 with 40 signers required to sign messages might have all of the same
+security properties as a group size of 100 with 51 signers required to sign
+messages, it might not _feel_ as secure. That perception may cause a decreased
+willingness to invest capital in the system leading to less money going over
+the bridge, even though the fees are less overall.
+
+Since we don't have any workable data that allows us to reasonably estimate
+costs (especially with regard to optics), I suggest we start the system out
+at decent https://en.wikipedia.org/wiki/Focal_point_(game_theory)[schelling
+point] of groupsize = 100, signers = 51, strong heartbeat = 70, and then let
+governance adjust from there once we've gathered data from the system being used.
 
 === Minting
 
@@ -176,13 +204,14 @@ being able to meet the 51-of-100 threshhold to produce signatures. Additionally,
 we want to avoid situations where operators are the custodians of a wallet for
 extended periods. To avoid these issues, we can set a max age of a wallet and a minimum
 liveness threshhold. Once a wallet is older than the max age, or if it drops below
-the liveness threshhold (say, below 60 on a heartbeat), we motion to transfer
+the liveness threshhold (say, below 60 on a <<heartbeat,heartbeat>>), we motion to transfer
 the funds to another randomly selected wallet.
 
 Once a wallet no longer has funds and is not the primary wallet for new
 deposits, it can be closed and operators are no longer required to maintain
 it.
 
+[[heartbeat]]
 === Heartbeats
 
 To make sure that older wallets are still accessible for redemption, we need to
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -26,8 +26,16 @@ A user with an account balance supplies a bitcoin address. The system decreases
 their account balance and releases the equivalent amount of bitcoin to the user
 from a user-supplied bitcoin wallet.
 
-In the default case, the user-supplied wallet will be determined by the dApp,
-and will be the oldest wallet that is large enough to fulfill the redemption.
+The redeeming wallet will be determined by the system on chain via an upgradable
+(by governance) contract. The initial implementation for this contract should
+select the oldest wallet. Redemptions should be batched (details to be
+determined) to both decrease fees and processing time.
+
+One possible implementation for efficient on-chain calculation of the oldest wallet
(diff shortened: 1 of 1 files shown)


beaushinkle:
How do I link to another RFC? I'd love to link to RFC 2 in the group size section!

pdyraga:
> How do I link to another RFC? I'd love to link to RFC 2 in the group size section!

See here: https://github.com/keep-network/keep-core/pull/2585/commits/87f48a7914eb3c455cb3c6ca62cdd75a7989e029

docs/rfc/rfc-1.adoc
```diff
@@ -43,6 +43,34 @@ collects tBTC fees it uses those fees to buy T tokens back for its treasury.
 
 == In Depth
 
+[[group-size]]
+=== Group Size
+There are three main governable parameters that we need to choose:
+* The total size of the group
+* The number of signers required to sign messages
+* The number of group members required for a <<heartbeat,heartbeat>> to be considered "strong".
```
pdyraga, line 51: We also need to determine the minimal number of active group members during the key generation needed so that the key generation result can be accepted. See [this](https://github.com/keep-network/tbtc-v2/pull/44#discussion_r700222208) comment.

docs/rfc/rfc-1.adoc
```diff
@@ -43,6 +43,34 @@ collects tBTC fees it uses those fees to buy T tokens back for its treasury.
...
+security properties as a group size of 100 with 51 signers required to sign
+messages, it might not _feel_ as secure. That perception may cause a decreased
+willingness to invest capital in the system leading to less money going over
+the bridge, even though the fees are less overall.
+
+Since we don't have any workable data that allows us to reasonably estimate
+costs (especially with regard to optics), I suggest we start the system out
+at decent https://en.wikipedia.org/wiki/Focal_point_(game_theory)[schelling
+point] of groupsize = 100, signers = 51, strong heartbeat = 70, and then let
+governance adjust from there once we've gathered data from the system being used.
```
pdyraga, line 72: Not in this PR but we need to figure out how heartbeats actually work cc @eth-r. That is:
- Are we signing every block or every nth block? 
- What happens if one signature failed? Is it enough to fail a heartbeat? Do we retry?
- Is weak heartbeat reported on-chain. If so, who is reporting it? 
- Is failed heartbeat reported on-chain. If so, who is reporting it?
- Is successful heartbeat reported on-chain? Is it every successful one or maybe every nth successful?

pdyraga (commented):
Just one comment on the number of active group members during the DKG, the rest looks good to me.

docs/rfc/rfc-1.adoc
beaushinkle: Yup!, So something like: We want a 51-of-100 wallet, and we want to start moving the funds if we ever dip below 70 active operators and we want to make sure we don't even create the wallet in the first place if we don't have 90 of the operators for the original DKG (so the worst we can have is 51-of-90).

docs/rfc/rfc-1.adoc
```diff
@@ -43,6 +43,42 @@ collects tBTC fees it uses those fees to buy T tokens back for its treasury.
...
 == In Depth
 
+[[group-size]]
+=== Group Size
+There are three main governable parameters that we need to choose:
+* The number of signers required to sign messages `threshold`
+* The number of group members required for a <<heartbeat,heartbeat>> to be considered "strong". `heartbeat`
+* The minimum number of members we're allowed to drop down to during the DKG
+  group formation re-try period. `dkg`
+* The total size of the group `groupSize`
```
pdyraga, line 53: This is not rendered as a list, see the [preview](https://github.com/keep-network/tbtc-v2/blob/bed5cae0d1970924a658f6e3f47fadefd382aa39/docs/rfc/rfc-1.adoc#group-size)

docs/rfc/rfc-1.adoc
```diff
@@ -43,6 +43,42 @@ collects tBTC fees it uses those fees to buy T tokens back for its treasury.
 
 == In Depth
 
+[[group-size]]
+=== Group Size
+There are three main governable parameters that we need to choose:
+* The number of signers required to sign messages `threshold`
```
pdyraga, line 49: Worth confirming in tss-lib and the original paper if `threshold+1` are required to sign a message and up to `threshold` can be corrupted so that the key is not in danger. Not something blocking this PR but stating it precisely could avoid confusion during the implementation.

docs/rfc/rfc-1.adoc
beaushinkle: Should be good now:
image

8bd92e7

docs/rfc/rfc-1.adoc
beaushinkle: In v1, we set our `threshold`param to `2` so that we require 3-of-3, so I think it makes sense to follow that same definition! Updated in 682e5cb

docs/rfc/rfc-1.adoc
```diff
@@ -78,57 +153,15 @@ days (specified as `locktime`).
...
-would only be able to drain the quickminter's treasury rather than the whole
-system.
-
-From a UX perspective, the quickminter performs like the
-sweep-in-batches-mint-afterward strategy when the treasury is empty, and
-performs like sweep-in-batches-mint-immediately when the treasury is
-sufficient.
+can spend the funds. Once successful, we increase the account's unswept balance.
+
+Second, we schedule an operation that batces all outstanding known-refundable
```
michalinacienciala, line 160:
```suggestion
Second, we schedule an operation that batches all outstanding known-refundable
```

docs/rfc/rfc-1.adoc
```diff
@@ -8,21 +8,23 @@ toc::[]
...
+the funds line up. If everything checks out, then we update your ethereum-side
+account balance.
 
 === Redemption
 
-A user supplies tBTCv2 and a bitcoin address. The system burns the tBTCv2 and
-releases the equivalent amount of bitcoin to the user from a user-supplied bitcoin wallet.
+A user with an account balance supplies a bitcoin address. The system decreases
+their account balance releases the equivalent amount of bitcoin to the user
+from a user-supplied bitcoin wallet.
```
eth-r, line 27: The wallet should probably be assigned by the bridge, not supplied by the user

docs/rfc/rfc-1.adoc
```diff
@@ -41,9 +44,81 @@ and maintain the peg), we must turn to another token. Instead, the system pays
...
+* The number of signers that can be controlled by the adversary before the key is in danger `threshold`
+* The number of group members required for a <<heartbeat,heartbeat>> to be considered "strong". `heartbeat`
+* The minimum number of members we're allowed to drop down to during the DKG
+  group formation re-try period. `dkg`
+* The total size of the group `groupSize`
+
+`threshold < heartbeat < dkg < groupSize`
+
+The total gas cost for sortiton increases linearly with group size by at least
+20k gas per member. Requiring more signers to sign messages makes it harder for
```
eth-r, line 70: With optimistic sortition we don't need to spend 20k gas per member in the happy path. If we use 32-bit operator IDs and pack them densely, the marginal cost per member would be an average of 2.5k gas from storing the ID. Additionally, we need at least `threshold + 1` signatures for the DKG result submission, which should be 3k gas per member, for a total of around 6k+change.

docs/rfc/rfc-1.adoc
```diff
@@ -41,9 +44,81 @@ and maintain the peg), we must turn to another token. Instead, the system pays
...
+adversaries to take over the group, but makes it so that the pool is less
+resilient to undelegation or operators going offline. Raising the
+strong-heartbeat threshold decreases the chance that you'll not have enough
+signers to sign transactions, but decreases a wallets lifespan and increases
+overhead. Lowering the dkg threshold increases the chance that we're able to
+create a wallet successfully but gives us a wallet with potentially fewer
+operators and a lower lifespan.
+
+Also important is optics - even though a group size of 70 with 40 signers
+required to sign messages might have all of the same security properties as a
```
eth-r, line 80: Not entirely; a 40/70 might be equally secure against being controlled by an adversary of around 25% power as 51/100, but it won't be as secure against becoming unavailable due to operators dropping out or the adversary trying to DoS it.

docs/rfc/rfc-1.adoc
```diff
@@ -78,57 +153,15 @@ days (specified as `locktime`).
...
-From a UX perspective, the quickminter performs like the
-sweep-in-batches-mint-afterward strategy when the treasury is empty, and
-performs like sweep-in-batches-mint-immediately when the treasury is
-sufficient.
+can spend the funds. Once successful, we increase the account's unswept balance.
+
+Second, we schedule an operation that batces all outstanding known-refundable
+transactions together to be combined with the existing wallet output into a
+single output. The frequency of this operation is a governable parameter. When
+this sweep occurs, we decrease the relevant account's unswept balances and
```
eth-r, line 163: `relevant accounts' unswept balances`

docs/rfc/rfc-1.adoc
```diff
@@ -78,57 +153,15 @@ days (specified as `locktime`).
...
-
-By minting to the quickminter instead of directly to the depositor, Mallory
-would only be able to drain the quickminter's treasury rather than the whole
-system.
-
-From a UX perspective, the quickminter performs like the
-sweep-in-batches-mint-afterward strategy when the treasury is empty, and
-performs like sweep-in-batches-mint-immediately when the treasury is
-sufficient.
+can spend the funds. Once successful, we increase the account's unswept balance.
```
eth-r, line 158: We should specify whether we take deposit fees or not

docs/rfc/rfc-1.adoc
```diff
@@ -8,21 +8,23 @@ toc::[]
...
+of messages and when it gets one, it checks the bitcoin network to make sure
+the funds line up. If everything checks out, then we update your ethereum-side
+account balance.
 
 === Redemption
 
-A user supplies tBTCv2 and a bitcoin address. The system burns the tBTCv2 and
-releases the equivalent amount of bitcoin to the user from a user-supplied bitcoin wallet.
+A user with an account balance supplies a bitcoin address. The system decreases
+their account balance releases the equivalent amount of bitcoin to the user
```
michalinacienciala, line 26:
```suggestion
their account balance and releases the equivalent amount of bitcoin to the user
```

docs/rfc/rfc-1.adoc
beaushinkle: Good point - I'll be more specific here

docs/rfc/rfc-1.adoc
beaushinkle: resolved in d8a82d698e1ebdc692d12b35cc7a83d309c2a846

docs/rfc/rfc-1.adoc
beaushinkle: resolved in d8a82d698e1ebdc692d12b35cc7a83d309c2a846

docs/rfc/rfc-1.adoc
beaushinkle: Awesome! Is that documented somewhere so I can edit this into the RFC and point to it?

docs/rfc/rfc-1.adoc
beaushinkle: Can you elaborate more here?

docs/rfc/rfc-1.adoc
beaushinkle: Alrighty - rewrote this in b1b3c8d

docs/rfc/rfc-1.adoc
beaushinkle: alrighty - gave the deposit fee a shout out: f06c70e

docs/rfc/rfc-1.adoc
eth-r: Not yet, all the discussion here has been kind of eating up my time so I haven't had the focus to write up the sortition pool v2 spec

docs/rfc/rfc-1.adoc
eth-r: Instead of `releases the equivalent amount of bitcoin to the user from a user-supplied bitcoin wallet` the bridge should determine the correct wallet to release the BTC from, and release them _to_ the BTC address specified by the user.

beaushinkle: If you mean on-chain, I think that can get expensive since we need to find a wallet appropriate for the size of their redemption. Is there an attack that you have in mind? Otherwise it's simpler/cheaper to let the user/dApp specify the wallet, I think

docs/rfc/rfc-1.adoc
beaushinkle: Added a note in 0292903

docs/rfc/rfc-1.adoc
eth-r: why wouldn't "the oldest wallet with enough BTC to cover the redemption" work? if we have a minimum BTC amount per wallet and close any wallet falling below that number, there shouldn't be too many small wallets to iterate through (and the iterations are storage reads, meaning that we'd be able to go through 25 wallets before the cost of the reads exceeds the base cost of a transaction)

it would also be possible to store the wallet information in a tree not unlike what we use in sortition pools, ensuring that no matter how many wallets we have active the cost of finding a wallet is always logarithmic in the maximum number of wallets over the system's lifetime

in the case of user-provided wallets, we could run into things like race conditions where two redemptions specify the same wallet but it doesn't have enough to cover both of them; in that case we'd either need to reject one redemption and force the user to try it again, or find a suitable wallet anyway

beaushinkle: I'm sold. @pdyraga @mhluongo ?

beaushinkle: Updated in https://github.com/keep-network/tbtc-v2/pull/45/commits/52beba8c7448d2161b6bfdfa071a85b1c3c274bb

beaushinkle:
Okay - I think this is good for another look!

Issue #37, opened 2021-08-30
When is TBTC minted?


pdyraga:
In #23 we discussed several options for minting TBTC. Four the most popular are:
- minting in the same transaction as Ethereum address and blinding factor reveal,
- minting after the sweep is done and proved, 
- using a minting buffer to mint earlier in the process but only if the threshold of not yet unswept transactions is not exceeded,
- outsource the risk to a quickminter pool of separate underwriters.

We need to decide which option do we want to go for the first release and spec the details.

Questions that should also be answered:
- When does the sweep happen and who is proving it? If wallet - is this refundable?
- Do we allow for configurable refund timeouts? How do they affect the sweep schedule?
- What happens if the sweep did not happen?

pdyraga:
In the first version we will:
- mint in the same transaction in which the sweep is proved, if the user specified it wants their balance to go to TBTC v2 application,
- allow anyone to mint at any moment based on their Bitcoin bank balance.

PR #46, opened 2021-09-02
Account Balances

beaushinkle:
This PR rewrites RFC1 from an account balances perspective. 

There's still an open FIXME touching on #37 of how we want to handle minting TBTC from unswept balances. I'm down to leave that a `?` until we get that issue resolved separately and then go back and edit.


pdyraga (approved):
🚀

PR #47, opened 2021-09-06
Validate `_keepCRV` parameter

lukasz-zimnoch:
Method `setKeepCRV` should validate their input parameter to be less or equal to the `DENOMINATOR` value.


tomaszslabon (approved):
We can merge it.

dimpar (approved):

PR #29, opened 2021-08-17
Migrate contracts and publish package

michalinacienciala:
We're adding jobs which will deploy and publish `tbtc-v2` contracts.
Published artifacts will be used in the Continuous Integration proccess.
`NPM` workflow will build & publish contracts every time there is a
change in contracts or in the files describing package dependencies.
The `-dev` suffix will be used in the name of the published package.
`Solidity` workflow will deploy and publish contracts every time
worklfow is triggered by the `workflow_dispatch` event. Published
package name will point to name of the environment on which the
contracts were deployed.

TODO:

- [x] remove `--dry-run` from jobs configuration
- [x] prepare PR modifying `config.json` in the `ci` project (https://github.com/keep-network/ci/pull/16)
- [x] fix problem with the `prepare-artifacts.sh` script not allowing for publish of contracts built on `development` env (https://github.com/keep-network/tbtc-v2/pull/29/commits/943ce01eb2cf5ad79ace7707b9af806416a22833)

TODO **after** merging and publishing new NPM package:
- [ ] Update PR https://github.com/keep-network/keep-core/pull/2609 (update `yarn.lock` to contain dependency to the right NPM package with `tbtc-v2` contracts)

Refs:
https://github.com/keep-network/ci/pull/16

PR #52, opened 2021-09-17
Verify contracts on Etherscan

michalinacienciala:
We're adding a step verifying the migrated contracts on Etherscan. We're
using `etherscan-verify` plugin. In order for the plugin not to pick up
wrong files for verification we have to remove contracts of the
dependent projects stored in the `node_modules`.
We're also adding a step which executes in case of failures during the
contracts verifications. The step adds a warning annotation to the
GitHub page with the workflow run details.
Some info regarding configuration of `annotations.json` and
`github-action-create-annotations` action:
1. Values of `paths`, `start_line`, `end_line`, `start_column` and
   `end_column` don't seem to have much effect on how the annotation is
   displayed for workflows triggered by the `workflow_dispatch` event
   (they matter more for workflows triggered by `pull_request`, but we
   only annotate for `workflow_dispatch`).
2. We were getting errors when `end_line` was configured to different
   value than `start_line`.
3. The value of the `check-name` parameter of the action needs to be the
   same as the name of the job we want to annotate. Otherwise the
   annotation may get assigned to the random job (even in different
   workflow).

Refs:
https://github.com/keep-network/tbtc-v2/pull/29

https://github.com/keep-network/keep-core/pull/2598
https://github.com/keep-network/keep-ecdsa/pull/913
https://github.com/keep-network/tbtc/pull/834

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -148,6 +148,29 @@ jobs:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: npm publish --access=public --tag ${{ github.event.inputs.environment }} --network=${{ github.event.inputs.environment }}
 
+      # If we don't remove the `tbtc` contracts from `node-modules`, the
+      # `etherscan-verify` plugins tries to verify them, which is not desired.
+      - name: Prepare for verification on Etherscan
+        run: rm -rf ./node_modules/@keep-network/tbtc
+
+      - name: Verify contracts on Etherscan
+        continue-on-error: true
+        id: verify-contracts
+        env:
+          ETHERSCAN_API_KEY: ${{ secrets.ETHERSCAN_API_KEY }}
+          CHAIN_API_URL: ${{ secrets.KEEP_TEST_ETH_HOSTNAME_HTTP }}
+        run: yarn run hardhat --network ${{ github.event.inputs.environment }} etherscan-verify --license MIT
+
+      - name: Annotate in case of verification failure
+        if: steps.verify-contracts.outcome == 'failure'
+        continue-on-error: true
+        uses: thesis/github-action-create-annotations@v0.2.0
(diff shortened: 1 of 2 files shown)


michalinacienciala:
Workflow run proving successful contracts verification: https://github.com/keep-network/tbtc-v2/runs/3632863551?check_suite_focus=true (executed on a stripped workflow yml)
Workflow run showing annotations added in case of verification failure: https://github.com/keep-network/tbtc-v2/actions/runs/1245161578

.github/workflows/contracts.yml
```diff
@@ -148,6 +148,29 @@ jobs:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: npm publish --access=public --tag ${{ github.event.inputs.environment }} --network=${{ github.event.inputs.environment }}
 
+      # If we don't remove the `tbtc` contracts from `node-modules`, the
+      # `etherscan-verify` plugins tries to verify them, which is not desired.
+      - name: Prepare for verification on Etherscan
+        run: rm -rf ./node_modules/@keep-network/tbtc
+
+      - name: Verify contracts on Etherscan
```
nkuba, line 156: What if we extract `Verify contracts on Etherscan` to a separate job that would depend on `contracts-deployment-testnet` and use upload-artifacts/download-artifacts action to transfer data required for publication? With this apporach, we could avoid using annotations.

michalinacienciala:
Tested (on a code with NPM publish commented out) here: https://github.com/keep-network/tbtc-v2/runs/3768470754?check_suite_focus=true.

.github/workflows/contracts.yml
michalinacienciala: Moved, 669c4d003f5c6b5aabe1ea730dd1342105fff24d.

nkuba (approved):

PR #54, opened 2021-10-06
RFC 1: Batched Sweeping

beaushinkle:
This PR fleshes out a proposal for sweeping!

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -54,11 +54,10 @@ collects tBTC fees it uses those fees to buy T tokens back for its treasury.
 
 === Minting And Unminting
 
-A user with an account balance can drain their account balance to mint TBTC.
+A user with a swept account balance can drain their account balance to mint TBTC.
 Similarly, a user can burn TBTC to refill their account balance.
 
-// FIXME: fill in this section with how much TBTC they can mint with regards to
-// their unswept balance
+A user with a swept account balance can drain their account balance to redeem BTC.
 
 == In Depth
 
@@ -170,7 +169,7 @@ and charge a deposit fee.
 Second, we schedule an operation that batches all outstanding known-refundable
 transactions together to be combined with the existing wallet output into a
 single output. The frequency of this operation is a governable parameter. When
-this sweep occurs, we decrease the relevant accounts' unswept balances and
+this <<sweeping,sweep>> occurs, we decrease the relevant accounts' unswept balances and
 increase their balances. This disables any outstanding 30-day refunds.
 
 ==== Automated Refunds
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-1.adoc
```diff
@@ -180,12 +179,48 @@ simple as "these funds can be spent by wallet 0xabc", or in our case, as
...
+consolidates the UTXO set and it disables the refund.
+
+This process is called a "sweep", and occurs based on a governable frequency.
+Any deposit below a governable threshold is ignored.
+
+After submitting the transaction, it will cost some amount of bitcoin based on
+what miners are charging for the bitcoin fee. This fee comes out of the unswept
+deposits. When the transaction clears, the users unswept balances are
+decreased, and their swept balances are increased (after paying their share of
+the mining fee).
```
pdyraga, line 199: The wallet needs to submit a transaction proving the sweep happened and set/increase users' swept balances. This transaction is expensive. How do we select which operator from the wallet is responsible for submitting it? How do we incentivize it? 

From our earlier discussions, we had an idea of *maintainers* who do not necessarily need to have any stake and their responsibility is just to execute transactions on-chain and get rewards for that. This can even be an additional mode of keep client to make it simple - an optional module one needs to enable/disable with a flag. Then, regular operators do not even need to have any ETH for a day-to-day work.

Still, there is a question of how these maintainers should work. How they are paid and if there is any order in which they are expected to act so that they do not have to race and waste gas on reverted transactions because they compete with each other and because some MEV bot keeps frontrunning everyone.

It could be a good idea to see how Keep3rs works in Yearn and thinking if we can maybe implement a similar ordering mechanism as for relay entry / DKG result submission in random beacon v2, see [here](https://github.com/keep-network/keep-core/blob/main/docs/rfc/rfc-19-random-beacon-v2.adoc#fees-and-rewards). E.g. there is a pool of maintainers, they are getting rewards only for successful transactions, and if the given maintainer missed their chance, they are removed from the pool for a long time. Those who are not in the pool are not eligible to do the maintainer's work. The pool has some initialization period and/or locks some T deposit (or uses stake) to prevent griefing.

docs/rfc/rfc-1.adoc
```diff
@@ -180,12 +179,48 @@ simple as "these funds can be spent by wallet 0xabc", or in our case, as
...
+product to compete with the centralized ones.
+
+2) We allow users to request that their TBTC is minted as soon as they have a
+swept account balance. This makes it so they don't have to wait, check, and
+come back later and mint.
+
+Combining these ideas, a user would deposit some BTC, reveal their eth address
+and blinding factor, and then request that TBTC gets minted ASAP. Checking the
+dApp, they can see that they should expect TBTC in their provided wallet
+address in 3 hours with no further interaction.
```
pdyraga, line 220: There is one more thing that came up in earlier discussions and is worth thinking about: `CHECKLOCKTIMEVERIFY` requirements. One may reveal their ETH address and blinding factor just a moment before BTC script unlocks the funds for refund. We should have a requirement regarding the time left for the wallet to execute the sweep. If that requirement is not met deposit is ignored.

Now, there is a question where and how this requirement is checked. Is it possible to check it during the reveal transaction? (I doubt it). Should the wallet check it when sweeping and not even try to sweep such a deposit?

pdyraga (approved):
This looks good to me, I think it's an improvement to the RFC so it is worth merging. I left some comments about areas that require clarification and more protocol design work. Flagging @mhluongo and @eth-r to take a look as well.

docs/rfc/rfc-1.adoc
```diff
@@ -180,12 +179,48 @@ simple as "these funds can be spent by wallet 0xabc", or in our case, as
...
+reason), all they need to do is not submit their reveal and wait 30 days.
+
+[[sweeping]]
+==== Sweeping
+The operators sign a transaction that unlocks all revealed deposits, combines
+them into a single UTXO with the existing UTXO, and relocks that transactions
+without a 30-day refund clause to same wallet. This has two main effects: it
+consolidates the UTXO set and it disables the refund.
+
+This process is called a "sweep", and occurs based on a governable frequency.
```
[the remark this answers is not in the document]
beaushinkle: good catch - i'll add in the piece about governable accumulated deposits

docs/rfc/rfc-1.adoc
beaushinkle: > There is one more thing that came up in earlier discussions and is worth thinking about: CHECKLOCKTIMEVERIFY requirements. One may reveal their ETH address and blinding factor just a moment before BTC script unlocks the funds for refund. We should have a requirement regarding the time left for the wallet to execute the sweep. If that requirement is not met deposit is ignored.

Maybe I'm missing something obvious. Say that they do this, and they reveal. Now, they *could* refund their BTC, but we can also still sweep. If they refund their BTC, we can't sweep. They'd have to mine the transaction in between when we start the signing process and when we go to submit the transaction, and for what gain? (I don't actually know what happens if we submit a big batch transaction and one of the inputs is already spent)

If they don't refund, then we can proceed as normal!

You're totally right in that if we put in a time limit like "we'll only process stuff that has >1 day left on the cltv" then we don't run into this nonsense, so it might be worth just nipping in the bud

PR #57, opened 2021-11-05
Initial draft of Bank, Vault, and Bridge

pdyraga:
This is an attempt to draft the initial components of tBTC v2 and define how they will interact with each other.

This is not a final version by any means.

# Bank

`Bank` is a non-upgradeable contract holding BTC account balances. Balances can be transferred between holders just like ERC20 tokens, with `transfer` and `transferFrom` functions. One can give an approval to use their balance to someone else by calling `approve` function. In the future, we may also consider adding support for EIP2612 approvals.

The only contract allowed to increase or decrease balances in `Bank` is `Bridge`. `Bridge` is upgradeable by `Bank` contract owner which eventually should be the DAO. `Bank` is currently only interested in confirmed balances but we might want to add tracking of unconfirmed balances as well.

All components of the system except `TBTC` token and `Bank` will be upgradeable in the early days. Upgrading functionality is easy, upgrading state is difficult. `Bank` is just about the global state of balances and its functionality is reduced to an absolute minimum.

# Vault

`Vault` is a base contract for all applications (products of Bank) allowed to lock and unlock user's balances. Every `Vault` has its balance in `Bank`, just like any other account. To lock balance, `Vault` needs to have a balance `approve`d and transferred from holder's to `Vault`'s address. `Vault` base class allows keeping track of locked balances that can be later unlocked and transferred back to their original holders.

Anyone can deploy any `Vault` and anyone can approve any `Vault` to use their balances. This approach gives us extendability and draws a clear separation line between `Bank` and `Bridge` which manage balances and `Vault` that can use only balances it was `approve`d for. `Vault` can not change balances of other `Vault`s or print balances in `Bank`. Every `Vault` is clearly separated from the balances of other `Vault`s and holders in `Bank`.

One implementation of a `Vault` is `TBTCVault` minting and burning TBTC tokens based on locked balances.

# Bridge

Bridge manages BTC deposit and redemption and is the only component being able to increase and decrease the balance of the given holder (including `Vault`s) in `Bank` contract. `Bridge` has a shortcut path to `Vault` in case the depositor wants to lock their balance in the given `Vault` as soon as their UTXO gets swept. One example could be the auto-minting of TBTC token. Depositor can supply an optional `Vault` address when revealing their UTXO and then, once the given UTXO is swept, vault implementing `IQuickLock` interface can do whatever is needed, for example, mint TBTC.

# Open questions

We were considering allowing to ming TBTC based on yet unconfirmed balance. This option needs to be thought through and tracking of unconfirmed balances might need be added to `Bank`. This is nice functionality but complicates everything. For example, should unconfirmed balances be transferrable? What happens if they could not be swept etc. Right now, it's `Bridge`'s concern to keep track and manage them and `Bank` is only interested in confirmed balances that can be used right away.

### DIFF
--- a/solidity/contracts/bank/Bank.sol
+++ b/solidity/contracts/bank/Bank.sol
@@ -1,0 +1,117 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.4;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "./IVault.sol";
+import "../bridge/IBridge.sol";
+
+contract Bank is Ownable {
+    mapping(address => uint256) public balance;
+    mapping(address => mapping(address => uint256)) public allowance;
+
+    IBridge public bridge;
+
+    event Approval(
+        address indexed owner,
+        address indexed spender,
+        uint256 value
+    );
+    event Transfer(address indexed from, address indexed to, uint256 value);
+
+    modifier onlyBridge() {
+        require(msg.sender == address(bridge), "Caller is not the bridge");
+        _;
+    }
+
+    function setBridge(IBridge newBridge) external onlyOwner {
+        bridge = newBridge;
+    }
+
+    function increaseBalance(address owner, uint256 amount)
+        external
+        onlyBridge
+    {
+        balance[owner] += amount;
+    }
+
+    function quickLock(
+        address owner,
+        uint256 amount,
+        address vault
+    ) external onlyBridge {
+        balance[owner] += amount;
+        _approve(owner, vault, amount);
+        IQuickLock(vault).quickLock(owner, amount);
+    }
+
+    function decreaseBalance(address owner, uint256 amount)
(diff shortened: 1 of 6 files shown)


beaushinkle:
I did a pass of the description and then skimmed the solidity 😈.

I love the separation of concerns, especially with the `Bank <-> Vault -> TBTCVault`  split. I also like that the vaults are never actually burning balances but are just locking them!

Seems solid so far, and I'll do another pass tomorrow for solidity, but wanted to get this out before I moved on to other stuff

pdyraga:
Closing without merging, I will open separate PRs for components. The idea will _mostly_ stay the same.

pdyraga:
`Bank`: https://github.com/keep-network/tbtc-v2/pull/84

pdyraga:
`Vault` and `TBTCVault`: #87

PR #58, opened 2021-11-05
RFC 1: Wallet Lifecycle Cleanup

beaushinkle:
- Added a governable parameters header
- Added information about the wallet dust threshold

tagging @pdyraga for review

What do you think about the governable parameters header at the beginning of the section? I think it looks snazzy and if you agree, I can go through the rest of the document and make the rest of the sections follow the same structure.


pdyraga:
> What do you think about the governable parameters header at the beginning of the section? I think it looks snazzy and if you agree, I can go through the rest of the document and make the rest of the sections follow the same structure.

💯 👍🏻

docs/rfc/rfc-1.adoc
```diff
@@ -253,12 +253,26 @@ wallet and closes.
...
-parameter. To create a new wallet, a group of 100 operators is selected from
-the pool of available operators (some operators may be selected twice if there
-are not enough) using a process called sortition. The probabiliy that a
-particular operator is chosen is based on their stake weight, which in turn is
-based on the number of `T` tokens they have invested in the staking contract.
+Governable Parameters:
+
+- `wallet_dust_threshold` the smallest amount of btc a wallet can hold before we
+  attempt to close the wallet and transfer the funds to a randomly selected
+  wallet
```
pdyraga, line 260: This is only for the currently active wallet, right? If yes, worth being explicit about it.

pdyraga: And if yes, why not `wallet_min_btc` to keep it aligned with `wallet_max_btc`? 🤔

docs/rfc/rfc-1.adoc
```diff
@@ -253,12 +253,26 @@ wallet and closes.
...
+  early creation of a new wallet
+- `max_wallet_age` the oldest we allow a wallet to become before we transfer the funds
+  to a randomly selected wallet
+
+A new wallet is created when either enough time has passed as defined in
+`wallet_creation_period` *AND* the wallet contains at least
+`wallet_dust_threshold` btc *OR* the wallet currently contains more btc than
+`wallet_max_btc`, whichever happens first. To create a new wallet, a group of
+100 operators is selected from the pool of available operators (some operators
+may be selected twice if there are not enough) using a process called
```
pdyraga, line 272: > some operators may be selected twice if there are not enough

It is not entirely true. Some operators can be selected twice or more even if there are enough operators in the pool. The probability depends on the stake weight and nothing else.

docs/rfc/rfc-1.adoc
```diff
@@ -253,12 +253,26 @@ wallet and closes.
...
+sortition. The probabiliy that a particular operator is chosen is based on
+their stake weight, which in turn is based on the number of `T` tokens they
+have invested in the staking contract.
```
pdyraga, line 275: What is triggering wallet creation? New deposit? Or is it some public function that will be called by maintainers (lucky or unlucky)?

docs/rfc/rfc-1.adoc
beaushinkle: I was originally thinking that param would apply for any wallets and would be multi-purpose. For non-active wallets, if there was ever a redemption that puts us below `wallet_min_btc`, that's when we start the process to transfer the funds and close the wallet. 

Applied to the active wallet, it's what stops us from needlessly creating new wallets. I'm also okay with having two separate parameters so that we can control those two numbers separately (why not?).

docs/rfc/rfc-1.adoc
beaushinkle: A new wallet needs to be created either when enough time has passed since the last one was created or when there's too much bitcoin in the current wallet, whichever comes first. We also need to kick that process off on-chain somehow (and will need to call the sortition pool), so that we can penalize folks if it doesn't start happening.

I think this is a great candidate for the mechanism where we offer an increasing reward for submitting these transactions until someone's client wants to do it since I don't think we want to slow wallet creation on the whole network down waiting for one operator to get their act together. Same deal for paying the gas to have the sortition pool pick the 100 folks for the wallet (since I imagine that'll be expensive).

docs/rfc/rfc-1.adoc
beaushinkle: updated in f348255

docs/rfc/rfc-1.adoc
beaushinkle: updated in 75a7457

docs/rfc/rfc-1.adoc
pdyraga: For sortition pool selection, we can use the same pattern as for the random beacon: 

> New groups should be created with a fixed, governable frequency of relay requests. Instead of a ticket-based approach for group selection, we should use sortition pools. Group creation start transaction should be embedded into relay request transaction and should lock a sortition pool. From this moment, no operator can enter or leave the pool. Once a new relay entry appears on the chain, all off-chain clients should perform group selection optimistically, using the new entry value and a view sortition pool function call. After determining group members, clients should perform DKG as usual. Once the group submits the result to the chain, a challenge period starts.
>
> During the challenge period, anyone can notify that the submitted DKG result contains group members not selected by the pool. This leads to slashing all members who signed the result and the notifier receives 5% from the slashed amount. The malicious result is immediatelly discarded. The length of the challenge period and slashing amount are governable parameters.
>
> Once the challenge period passes, anyone can unlock the sortition pool and mark the DKG result as accepted. In this transaction, the DKG result submitter receives a reward, as described in Fees and Rewards section.
>
> There is a timeout before which DKG result should be submitted. The timeout equals the group size multiplied by the number of blocks for a member to become eligible to submit DKG result. The timer starts at the moment when the first member becomes eligible.
>
> In case the DKG result was not submitted before a timeout, anyone can notify DKG timed out and receive a reward, as described in Fees and Rewards section. DKG timeout includes the situation when no new relay entry was produced and sortition could not be performed.

https://github.com/keep-network/keep-core/blob/main/docs/rfc/rfc-19-random-beacon-v2.adoc#group-creation

So in this case, that would be:
1. Someone notifies the contract to create a new wallet.
2. Sortition pool is locked, relay request is published.
3. Beacon produces relay entry, publishes it to the chain.
4. Clients perform a selection from the sortition pool off-chain calling sortition function.
5. Clients execute tECDSA key generation and publish the result.
6. Challenge period starts, anyone can challenge the result, especially group members in the published result.
7. If no challenge was published, anyone can call a transaction to approve the result. Usually, it's the original result submitter - they are incentivized to do so because in this transaction we release their reward. If they do not do that until some timeout, anyone can call the approve transaction and receive the submitter's reward.

In this model, we need three types of rewards:
- reward for one notifying a new wallet should be created
- reward for DKG result submitter
- reward for notifier of DKG timeout (pool needs to be unlocked)

For the first reward we can use the mechanism you described, for two other rewards, we can use just a plain token transfer of some governable amounts from some pool.

What I like about it is that we can reuse some code from the random beacon.

What is problematic though - and I think it deserves a separate PR - is how these rewards are funded. Who, when, and why. The easiest option would be - DAO from inflation, and maybe that's enough for v1 🤷🏻

docs/rfc/rfc-1.adoc
```diff
@@ -253,12 +253,25 @@ wallet and closes.
...
+  wallet.
+- `wallet_max_btc`: The amount of btc a wallet can hold before we trigger the
+  early creation of a new wallet.
+- `wallet_max_age`: The oldest we allow a wallet to become before we transfer the funds
+  to a randomly selected wallet.
+
+A new wallet is created when either enough time has passed as defined in
+`wallet_creation_period` *AND* the wallet contains at least
+`wallet_min_btc` btc *OR* the wallet currently contains more btc than
+`wallet_max_btc`, whichever happens first. To create a new wallet, a group of
```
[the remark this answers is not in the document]
beaushinkle: `wallet_max_age` is about closing wallets, not about creating new ones! So for instance, if we have a wallet_max_age of 4 months, then soon as the first wallet is 4 months old (and there should have been ~16 new wallets created in that time), if it hasn't been closed for other reasons (failed heartbeats, out of btc from redemptions), then we close it anyway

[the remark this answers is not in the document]
pdyraga: Makes sense!

pdyraga (approved):

PR #59, opened 2021-11-05
RFC 1: make the heartbeat approach more concrete

beaushinkle:
This PR re-writes the heartbeat approach. Previously, there wasn't a great way for the signers to read bitcoin blocks and slashing was too harsh.

The new approach signs ethereum blocks, collects which operators aren't around if a heartbeat fails, and then submits that on chain. This opens us up to an attack where the wallet is controlled by the adversary and now they can kick honest operators out of the sortition pool for a few weeks, but honestly that's not the worst they can do 🤷 

tagging @pdyraga for review

open questions for a different PR:

* who submits the transaction to the chain? (an unlucky alive operator?)
* who submits the (very expensive) transaction to remove the inactive operators from the sortition pool? I think I heard something about a gas station idea

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -281,15 +281,22 @@ it.
 [[heartbeat]]
 === Heartbeats
 
-To make sure that older wallets are still accessible for redemption, we need to
-perform heartbeats. The signing group signs each bitcoin block and then does _not_
-publish the result. If a signer suspects other signers are not online, they can
-issue an on-chain challenge to publish a specified signed bitcoin block (with a
-maximum block age). Since publishing this information costs the signers gas, the
-challenger must pay a deposit to be distributed to the signers if they pass.
+Governable Parameters:
 
-Any signer unable to publish the signed block within a specified amount of time
-will begin to be slashed and the challenger will be rewarded.
+- `failed_heartbeat_sortition_timeout`: The amount of time an operator is
+  removed from the sortition pool after failing a heartbeat.
+- `heartbeat`: The number of group members required for a heartbeat to successful.
+
+To make sure that older wallets are still accessible for redemption, we need to
+perform heartbeats. The signing group signs every 40th ethereum block and then
+does _not_ publish the result. If there are ever less than `heartbeat`
+operators that participate in the heartbeat, the ones that did can create and
+sign a new transaction that lists the _inactive_ operators. Once this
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-1.adoc
```diff
@@ -281,15 +281,22 @@ it.
...
 To make sure that older wallets are still accessible for redemption, we need to
-perform heartbeats. The signing group signs each bitcoin block and then does _not_
-publish the result. If a signer suspects other signers are not online, they can
-issue an on-chain challenge to publish a specified signed bitcoin block (with a
-maximum block age). Since publishing this information costs the signers gas, the
-challenger must pay a deposit to be distributed to the signers if they pass.
-
-Any signer unable to publish the signed block within a specified amount of time
-will begin to be slashed and the challenger will be rewarded.
+perform heartbeats. The signing group signs every 40th ethereum block and then
```
pdyraga, line 291: I think 40th block should also be a governable parameter. Every 40th block means that they are singing every 10 minutes. I think it could make sense to make this time a little longer in some scenarios.

docs/rfc/rfc-1.adoc
```diff
@@ -281,15 +281,22 @@ it.
...
+does _not_ publish the result. If there are ever less than `heartbeat`
+operators that participate in the heartbeat, the ones that did can create and
+sign a new transaction that lists the _inactive_ operators. Once this
+transaction is posted to the ethereum chain, we can iterate through the
+inactive operators, remove them from the sortition pool for
+`failed_heartbeat_sortition_timeout` amount of time (which disables their `T`
+rewards), transfer the from the wallet to another random wallet and close this
+wallet.
```
pdyraga, line 299: See this discussion please: https://github.com/keep-network/sortition-pools/pull/130#pullrequestreview-797630853. We can't remove them from sortition pool because there is a chance the pool is locked at that moment. What we probably want to do instead, is to remove them from rewards.

pdyraga (commented):
This is a change in a good direction so merging regardless of the two small comments I've made.

docs/rfc/rfc-1.adoc
beaushinkle: addressed in a090772

docs/rfc/rfc-1.adoc
beaushinkle: addressed in 3920f67

PR #60, opened 2021-11-09
Rewrite tests so that they use typescript

tomaszslabon:
This PR modifies tests so that they use typescript.


solidity/.tsconfig-eslint.json
```diff
@@ -0,0 +1,5 @@
+{
+  "extends": "./tsconfig.json",
+  "include": [".eslintrc.js", "**/*.js", "**/*.ts", "**/*.tsx"],
+  "exclude": ["node_modules"]
+}
```
pdyraga, line 5: Shouldn't we symlink to `./node_modules/@thesis-co/eslint-config/.tsconfig-eslint.json` just like we do [here](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/.tsconfig-eslint.json)?

solidity/hardhat.config.ts
```diff
@@ -30,7 +31,7 @@ const config: HardhatUserConfig = {
         url: process.env.FORKING_URL || "",
         // latest block is taken if FORKING_BLOCK env is not provided
         blockNumber:
-          process.env.FORKING_BLOCK && parseInt(process.env.FORKING_BLOCK),
+          process.env.FORKING_BLOCK && parseInt(process.env.FORKING_BLOCK, 10),
```
pdyraga, line 34: Why do we need that `10`? I am looking at example [in coverage pools](https://github.com/keep-network/coverage-pools/blob/e85b7d624368eaf38c0405ae6b225e3be3bcc4d9/hardhat.config.ts#L31-L33) and in [threshold network contracts](https://github.com/threshold-network/solidity-contracts/pull/22/files#diff-57cf5378f8cab20b02b279097ba6fad08eb53f747d81770045dda9c5f95effe3R29-R31) and I do not see anything similar.

solidity/test/bridge/VendingMachine.test.ts
```diff
@@ -1,25 +1,33 @@
-const { expect } = require("chai")
-const {
-  increaseTime,
+import { ethers, waffle, helpers } from "hardhat"
+import { expect } from "chai"
+import type { Signer } from "ethers"
+import { vendingMachineDeployment, constants } from "../fixtures"
+// eslint-disable-next-line import/extensions
```
pdyraga, line 5: We do not need to do that in [random beacon tests](https://github.com/keep-network/keep-core/blob/874e247e8f21cec976f93b38604c37a1f77b8872/solidity/random-beacon/test/BLS.test.ts#L5). Did we miss some config rule?

solidity/test/helpers/contract-test-helpers.ts
```diff
@@ -0,0 +1,18 @@
...
+export function to1e18(n: number): BigNumber {
+  const decimalMultiplier = ethers.BigNumber.from(10).pow(18)
+  return ethers.BigNumber.from(n).mul(decimalMultiplier)
+}
+
+export async function getBlockTime(blockNumber: number): Promise<number> {
+  return (await ethers.provider.getBlock(blockNumber)).timestamp
+}
+
+export const ZERO_ADDRESS = "0x0000000000000000000000000000000000000000"
```
pdyraga, line 18: Not in this PR, but are functions from this file something we can abstract out to https://github.com/keep-network/hardhat-helpers ? Worth opening a PR, to `hardhat-helpers`, we are duplicating it in every repo now. 

cc @nkuba

solidity/test/bridge/VendingMachine.test.ts
```diff
@@ -1,25 +1,33 @@
...
+import type { TestERC20, TBTC, VendingMachine } from "../../typechain"
+
+import {
   to1e18,
   to1ePrecision,
   ZERO_ADDRESS,
```
pdyraga, line 11: We can now use `const ZERO_ADDRESS = ethers.constants.AddressZero`, see [here](https://github.com/keep-network/keep-core/blob/874e247e8f21cec976f93b38604c37a1f77b8872/solidity/random-beacon/test/RandomBeacon.Relay.test.ts#L24)

solidity/test/bridge/VendingMachine.test.ts
```diff
@@ -28,33 +36,28 @@ describe("VendingMachine", () => {
...
+    await vendingMachine
+      .connect(deployer)
+      .transferUnmintFeeUpdateInitiatorRole(
+        await unmintFeeUpdateInitiator.getAddress()
+      )
     await vendingMachine
       .connect(deployer)
       .transferVendingMachineUpgradeInitiatorRole(
-        vendingMachineUpgradeInitiator.address
+        await vendingMachineUpgradeInitiator.getAddress()
```
pdyraga, line 60: Would be good to agree on one solution across all repositories cc @dimpar @lukasz-zimnoch @nkuba. Another approach is [here](https://github.com/keep-network/keep-core/blob/874e247e8f21cec976f93b38604c37a1f77b8872/solidity/random-beacon/test/RandomBeacon.Relay.test.ts#L6). I think the one presented by @tomaszslabon in this PR is "cleaner" because we do not have to dig too deep into libraries used under the hood although it does require some boilerplate code.

solidity/.tsconfig-eslint.json
tomaszslabon: Done.

solidity/hardhat.config.ts
tomaszslabon: On the other we used it in https://github.com/keep-network/keep-core/blob/874e247e8f21cec976f93b38604c37a1f77b8872/solidity/random-beacon/hardhat.config.ts#L38.

solidity/test/bridge/VendingMachine.test.ts
tomaszslabon: Done.

solidity/test/helpers/contract-test-helpers.ts
tomaszslabon: I got rid of it. I suggest we use:
```
const ZERO_ADDRESS = ethers.constants.AddressZero
```
whenever we need a zero address.

solidity/test/bridge/VendingMachine.test.ts
tomaszslabon: I suspect that in case of `random-beacon` the `lint` check is run after contracts are compiled and there are files generated in the ` typechain` folder. While here it is run before we generate `typechain`.

If you checkout a fresh `random-beacon` repo and call `yarn lint` there, you will get the same error.

solidity/package.json
```diff
@@ -28,7 +29,7 @@
     "@keep-network/tbtc": ">1.1.2-dev <1.1.2-ropsten",
     "@openzeppelin/contracts": "^4.1.0",
     "@tenderly/hardhat-tenderly": "^1.0.12",
-    "@thesis/solidity-contracts": "github:thesis/solidity-contracts#507c647"
+    "@thesis/solidity-contracts": "github:thesis/solidity-contracts#4985bcf"
```
pdyraga, line 32: Non-blocking but it should go in a separate commit. "Test cleanup" is not a good commit message for such an important version bump :)

pdyraga (approved):

PR #61, opened 2021-11-09
RFC 1: Heartbeat Cleanup

beaushinkle:
This PR addresses some of the review from https://github.com/keep-network/tbtc-v2/pull/59

- We make the number of blocks until the next heartbeat a governable parameter
- We specify that we're not removing operators from the sorition pool when they fail a heartbeat. Rather, they stay in the pool but get removed from reward eligibility.

tagging @pdyraga for review


docs/rfc/rfc-1.adoc
```diff
@@ -283,20 +283,21 @@ it.
 
 Governable Parameters:
 
-- `failed_heartbeat_sortition_timeout`: The amount of time an operator is
-  removed from the sortition pool after failing a heartbeat.
+- `failed_heartbeat_reward_timeout`: The amount of time an operator is
```
pdyraga, line 286: Is `_timeout` correct word? I am not going to argue with a native speaker 😁 but after reading it the first time, I thought this parameter is about a timeout to provide a heartbeat. Isn't it more something like a... reward ban period? 

This parameter should say for how long the operator does not receive future rewards although they are in the sortition pool.

beaushinkle: hah, please argue! 

There are only two hard things in Computer Science: cache invalidation and naming things.

-- Phil Karlton

In my head, I was thinking about when my mom would put me in "time out" when I was bad 👿 . Bad naming; ambiguous, I'll put myself in time out.

address in 8cce32e

pdyraga (approved):

PR #62, opened 2021-11-09
RFC 1: Governable Parameters!

beaushinkle:
This PR takes out a lot of the fuzzy or conversational language and replaces it with code-formatted, named governable parameters.

Tagging @pdyraga for review


docs/rfc/rfc-1.adoc
```diff
@@ -184,22 +184,33 @@ reason), all they need to do is not submit their reveal and wait 30 days.
 
 [[sweeping]]
 ==== Sweeping
+
+Governable Parameters
+
+- `sweeping_refund_safety_time`: The amount of time prior to when a UTXO
+  becomes eligible where we will not include it in a sweeping transaction.
+- `sweep_period`: The amount of time we wait between sweeps.
+- `sweep_max_btc`: The amount of total bitcoin that will trigger an early sweep.
```
michalinacienciala, line 193: What do you mean by the `total bitcoin` here? The sum of all `unsweptBalance`s on a given wallet?
Would we trigger the sweeping in the situation where `sweep_max_btc` is reached collectively by multiple depositors, but, let's say, only one depositor declared `max_fee` high enough to be included in the sweep?

docs/rfc/rfc-1.adoc
```diff
@@ -184,22 +184,33 @@ reason), all they need to do is not submit their reveal and wait 30 days.
...
+- `dust_threshold`: The minimum bitcoin deposit amount for the transaction to
+  be included in a sweep.
+
 The operators sign a transaction that unlocks all revealed deposits, combines
```
michalinacienciala, line 197: It would be good to update this sentence at some point (not necessarily in this PR, maybe in #56?), as it is no longer true - not all revealed deposits will be consolidated into one UTXO, only those who qualify based on the declared `max_fee`.

docs/rfc/rfc-1.adoc
```diff
@@ -184,22 +184,33 @@ reason), all they need to do is not submit their reveal and wait 30 days.
...
 ==== Sweeping
+
+Governable Parameters
+
+- `sweeping_refund_safety_time`: The amount of time prior to when a UTXO
+  becomes eligible where we will not include it in a sweeping transaction.
+- `sweep_period`: The amount of time we wait between sweeps.
+- `sweep_max_btc`: The amount of total bitcoin that will trigger an early sweep.
+- `dust_threshold`: The minimum bitcoin deposit amount for the transaction to
+  be included in a sweep.
```
michalinacienciala, line 195: Should we also include as a governable param the amount of time after which we start bumping up the fee for the Bitcoin sweeping transaction?

docs/rfc/rfc-1.adoc
beaushinkle: > What do you mean by the total bitcoin here? 

The sum of unswept deposits where each unswept deposit is itself higher than the dust threshold. (which only matters to exclude 1 million deposits of 0.0000001 btc each from triggering a sweep). I'll clarify in the RFC!

> Would we trigger the sweeping in the situation where sweep_max_btc is reached collectively by multiple depositors, but, let's say, only one depositor declared max_fee high enough to be included in the sweep?

My gut instinct is yes, but we have room to iterate because this happens off-chain. We don't know which of the deposits is "too low" until we do the bin-packing and then at that point, we're already mid-sweep.

Though, we *could* simulate a bin-packing with an off-chain fee and see if we have enough deposits with high enough fees to make a sweep worth it before we start one 🤔. Then, if we do, we propose that fee on-chain, and that starts the actual bin-packing (which should be at least as good as the simulated bin-packing), and would save the network fees on sweeping overall :tada:

docs/rfc/rfc-1.adoc
beaushinkle: deal!

docs/rfc/rfc-1.adoc
```diff
@@ -184,22 +184,33 @@ reason), all they need to do is not submit their reveal and wait 30 days.
 
 [[sweeping]]
 ==== Sweeping
+
+Governable Parameters
+
+- `sweeping_refund_safety_time`: The amount of time prior to when a UTXO
+  becomes eligible where we will not include it in a sweeping transaction.
```
pdyraga, line 191: prior to? 🤔 

My understanding is that:
- UTXO becomes eligible for sweeping after some time (protection against forks)
- UTXO is eligible for sweeping
- UTXO becomes not eligible for sweeping because `sweeping_refund_safety_time` is reached

docs/rfc/rfc-1.adoc
pdyraga: I think these parameters are just a suggestion to the off-chain wallet, at least in v1. If clients in the wallet decide to wait for whatever reason when `sweep_max_btc` is reached, I think this is acceptable.

beaushinkle:
> I think these parameters are just a suggestion to the off-chain wallet, at least in v1. If clients in the wallet decide to wait for whatever reason when sweep_max_btc is reached, I think this is acceptable.

not sure why I couldn't reply in-line up there, but yeah that's correct. It's a way to sync up the clients to all read the same configuration

docs/rfc/rfc-1.adoc
beaushinkle: yup! It's a way to sync up the clients to all read the same configuration

docs/rfc/rfc-1.adoc
beaushinkle: oops, forgot a word

docs/rfc/rfc-1.adoc
beaushinkle: addressed in 2112b02952cd7de5acbd614eb11f53b07c44f01f

docs/rfc/rfc-1.adoc
beaushinkle: addressed in 0a50ade

docs/rfc/rfc-1.adoc
beaushinkle: Yeah - I'll include these params, but in a different PR since the current sweeping section doesn't reference fee increases and I want to scope this PR to just the formatting refactor

beaushinkle: alrighty - included in https://github.com/keep-network/tbtc-v2/pull/65

docs/rfc/rfc-1.adoc
```diff
@@ -63,15 +63,15 @@ A user with a swept account balance can drain their account balance to redeem BT
...
 
-* The number of signers that can be controlled by the adversary before the key is in danger `threshold`
-* The number of group members required for a <<heartbeat,heartbeat>> to be considered "strong". `heartbeat`
-* The minimum number of members we're allowed to drop down to during the DKG
-  group formation re-try period. `dkg`
-* The total size of the group `groupSize`
+Governable Parameters:
+
+- `threshold`: The number of signers that can be controlled by the adversary before the key is in danger.
+- `heartbeat`: The number of group members required for a <<heartbeat,heartbeat>> to successful.
```
michalinacienciala, line 70:
```suggestion
- `heartbeat`: The number of group members required for a <<heartbeat,heartbeat>> to be successful.
```

docs/rfc/rfc-1.adoc
michalinacienciala: Could you also add that clarification about "total bitcoin" referring to the sum of the unswept deposits higher than the dust?

docs/rfc/rfc-1.adoc
michalinacienciala: I think what @pdyraga was asking about here (Piotr, correct me if I'm wrong) was whether the `sweeping_refund_safety_time` is a value that's being added to the deposit creation time or being subtracted from the 30 days refund threshold...

docs/rfc/rfc-1.adoc
```diff
@@ -63,15 +63,15 @@ A user with a swept account balance can drain their account balance to redeem BT
...
+- `dkg`: The minimum number of members we're allowed to drop down to during the DKG group formation re-try period.
+- `group_size`: The total size of the signing group.
 
-`threshold < heartbeat < dkg < groupSize`
+`threshold < heartbeat < dkg < group_size`
 
 The total gas cost for sortiton increases linearly with group size by at least
```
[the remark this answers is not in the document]
michalinacienciala: I mean here the paragraph starting in line 76 ("The total gas cost for sortition..."). I couldn't add a comment directly to the lines with the phrases.

docs/rfc/rfc-1.adoc
beaushinkle: I clarified that dust is ignored in e703515 and that we're summing up unswept in a9eab39

docs/rfc/rfc-1.adoc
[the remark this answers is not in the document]
beaushinkle: fixed in 097e11b

docs/rfc/rfc-1.adoc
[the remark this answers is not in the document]
beaushinkle: updated in 369e212 and b4812a0

docs/rfc/rfc-1.adoc
```diff
@@ -184,22 +187,48 @@ reason), all they need to do is not submit their reveal and wait 30 days.
...
+stops that from happening.  Once a deposit crosses that
+`sweeping_refund_safety_time` threshold, the depositor should refund their
 deposit.
 
-This process is called a "sweep", and occurs based on a governable frequency or
-if enough deposits have accumulated (also governable), whichever comes first.
-Any deposit below a governable threshold is ignored.
+This process is called a "sweep", and occurs after `sweep_period` has passed or
+if enough deposits have accumulated to exceed `sweep_max_btc`, whichever comes
+first. Any deposit below `dust_threshold` is ignored, for both for triggering a
```
michalinacienciala, line 218: One redundant "for"?

```suggestion
first. Any deposit below `dust_threshold` is ignored, both for triggering a
```

michalinacienciala (commented):
Left one small comment. Other than that looking good to me.

We have also some governable parameters relating to the "Sweeping fee" section, which we could list in dedicated "Governable Parameters:" section, but it does not necessarily need to be done in this PR.

michalinacienciala (dismissed):
Looking good to me. Leaving for @pdyraga for final approval.

docs/rfc/rfc-1.adoc
```diff
@@ -184,22 +187,48 @@ reason), all they need to do is not submit their reveal and wait 30 days.
...
-the mining pool. Giving ourselves 8 hours of leeway stops this from happening.
-Once a deposit crosses that 8-hour threshold, the depositor should refund their
+Caveat: We only include deposits in batches that have at least
+`sweeping_refund_safety_time` their refund window. This prevents potential
+attacks or corner cases where we create a transaction with a valid, unspent
+input, but by the time we have signed that transaction, the depositor has
+already submitted a refund to the mining pool. Giving ourselves this leeway
+stops that from happening.  Once a deposit crosses that
+`sweeping_refund_safety_time` threshold, the depositor should refund their
 deposit.
```
michalinacienciala, line 214: One more thing that I noticed - this sentence is a bit misleading. The depositor will not be able to process the refund directly after `sweeping_refund_safety_time` threshold is crossed. Threshold is crossed at `deposit_creation_time + 30 days - sweeping_refund_safety_time` and depositor can process a refund at `deposit_creation_time + 30 days`.

beaushinkle: addressed in 94e76fd

docs/rfc/rfc-1.adoc
```diff
@@ -63,38 +63,41 @@ A user with a swept account balance can drain their account balance to redeem BT
...
-* The total size of the group `groupSize`
+Governable Parameters:
+
+- `threshold`: The number of signers that can be controlled by the adversary
+  before the key is in danger.
+- `heartbeat`: The number of group members required for a
+  <<heartbeat,heartbeat>> to be successful.
+- `dkg`: The minimum number of members we're allowed to drop down to during the
+  DKG group formation re-try period.
+- `group_size`: The total size of the signing group.
```
pdyraga, line 75: I am thinking whether `threshold` and `group_size` shouldn't be non-governable parameters requiring a hard-fork event to update. They should definitely be listed as parameters here, just unsure if governable. Or if governable, we could add a note that the update of these two parameters might be a hard-fork event (bridge component update).

Changing threshold/group size has a lot of implications for clients. For example, broadcast channel may be fine with 100 group members but not necessarily with 1000 group members.

pdyraga (approved):

docs/rfc/rfc-1.adoc
beaushinkle: Can the contract allow governance to change them within hard-coded ranges? As in, we can change `group_size` from 50 to 150 and `threshold` from `group_size*.25` to `group_size*.75`?

PR #63, opened 2021-11-10
RFC 1: Deposit Fraud Proof

beaushinkle:
This PR adds a Fraud Proof section for deposits!

Tagging @pdyraga for review


docs/rfc/rfc-1.adoc
```diff
@@ -234,6 +234,22 @@ and blinding factor, and then request that TBTC gets minted ASAP. Checking the
...
+inputs are unlocked using the `signingGroupPubkey` and locked under a single
+UTXO using the same `signingGroupPubkey`. This is a normal non-fraudulent sweep.
+
+2) After the 30-day refund time has passed, a particular UTXO is unlocked using
+the `refundPubkey` and then locked with whatever script the refunder wants. This
+is a normal refund.
+
+Any transaction that unlocks a UTXO using the `signingGroupPubkey` and then
+locks it using any other script other than to a single UTXO to any other public
+key than the same `signingGroupPubkey` is fraud.
```
pdyraga, line 251: What about redemption signatures?

docs/rfc/rfc-1.adoc
```diff
@@ -234,6 +234,22 @@ and blinding factor, and then request that TBTC gets minted ASAP. Checking the
 dApp, they can see that they should expect TBTC in their provided wallet
 address in 3 hours with no further interaction.
 
+==== Fraud Proof
+
+There are only two valid types of transactions for a sweep:
+
+1) A collection of deposit UTXO's plus the main UTXO for the wallet are the
+inputs are unlocked using the `signingGroupPubkey` and locked under a single
+UTXO using the same `signingGroupPubkey`. This is a normal non-fraudulent sweep.
```
pdyraga, line 243: Diving a bit deeper, is it fine to say: as long as UTXO uses `signingGroupPubKey`, this is not a fraud? I think this could simplify the implementation because we don't need to track UTXO, just extract `signingGroupPubKey` from the fraud proof and compare.

docs/rfc/rfc-1.adoc
pdyraga: Ah, this is only about deposit. Redemptions are covered in #64.

docs/rfc/rfc-1.adoc
beaushinkle: It's slightly more complicated than that. If we go `signingGroupPubKey` => `signingGroupPubKey` then it's fine. If we go `refundPubkey` => `anything` then it's fine. If we go `signingGroupPubKey` => `anything else` then it's fraud.

The explanation on line 249 is the summary, and the stuff from line 241-247 is the technical explanation for why that summary is true!

docs/rfc/rfc-1.adoc
```diff
@@ -234,6 +234,22 @@ and blinding factor, and then request that TBTC gets minted ASAP. Checking the
 dApp, they can see that they should expect TBTC in their provided wallet
 address in 3 hours with no further interaction.
 
+==== Fraud Proof
+
+There are only two valid types of transactions for a sweep:
```
michalinacienciala, line 239: Sorry if I ask something that was once already explained to me...
Are we talking here about types of Bitcoin transactions that, if included as an argument of `notifySweep` Ethereum transaction, will be considered as non-fraudulent and will result in update of the `unsweptBalance` and `balance`?
If so, how is it that the second mentioned type of the transaction (refund) can be considered as a "valid type of transaction for a sweep"?

docs/rfc/rfc-1.adoc
michalinacienciala, line 251: Do we already know if/how such fraud attempts will be punished?

beaushinkle: nothing final, but my working solution is that we slash as many T tokens from their staking contract as we're allowed to slash

docs/rfc/rfc-1.adoc
beaushinkle: Ah, good catch. I'll adjust the verbiage! It should say something along the lines of "there are only two valid types of transactions for an unswept deposit"

beaushinkle: addressed in 3077079

michalinacienciala (dismissed):

beaushinkle:
Hey @pdyraga anything else you want to see here?

pdyraga:
> Hey @pdyraga anything else you want to see here?

I have some more thoughts but I'll capture them in a follow-up issue. This PR is a great improvement so merging it as-is.

pdyraga (approved):

PR #64, opened 2021-11-10
RFC 1: Redemption Fraud Proof

beaushinkle:
This PR talks about how to implement redemption fraud detection at a high level. Tagging @pdyraga for review


docs/rfc/rfc-1.adoc
```diff
@@ -251,6 +251,13 @@ to submit multiple redemptions. After a redemption, if a wallet has under a
 governable threshold of BTC remaining, it transfers that BTC to the active
 wallet and closes.
 
+==== Fraud Proof
+
+When a redemption is requested, the redeeming bitcoin public key and amount are
+known on the ethereum chain. Any bitcoin transactions with the main wallet
+wallet UTXO as the input must have outputs that match those known redemption
+requests, otherwise the transaction was fraudulent.
```
pdyraga, line 259: It means that the increase redemption fee needs to be submitted to Ethereum, just like in v1, right? If so, I think it's worth being explicit about it here. 

Are there any timeouts for providing redemption signature? How are potential redemption fee increase transactions affecting timeouts? In v1, this was a serious PITA for signers because they were responsible for providing redemption proofs to not get liquidated and redemeers just kept ignoring it because they wanted someone else to pay.

pdyraga: I do not consider timeouts as something we need to solve in this PR, just flagging it.

beaushinkle: trying to make sure we're on the same page - are you saying that if we increase the fees related to redemption, then that action needs to be on chain? yeah for sure!

There are also other sorts of fraud, like where we charge the redeemer a different fee than they should have been charged (colluding with friends or stealing money from strangers), but those aren't mentioned in this PR. Mind if I include those in a follow-up?

pdyraga (approved):

PR #65, opened 2021-11-17
RFC 1: Incrementing Fee

beaushinkle:
This PR adds a blurb that talks about our strategy for when we submit a transaction with too low of a bitcoin fee. Tagging @pdyraga and @michalinacienciala for review

follows up on the comments from https://github.com/keep-network/tbtc-v2/pull/62#discussion_r749074475

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -205,9 +205,10 @@ Governable Parameters
   deposits to sweep.
 
 
-The operators sign a transaction that unlocks all revealed deposits, combines
-them into a single UTXO with the existing UTXO, and relocks that transactions
-without a 30-day refund clause to same wallet. This has two main effects: it
+The operators sign a transaction that unlocks some of revealed deposits based
+on the <<bitcoin-sweeping-fee,bitcoin sweeping fee>>, combines them into a
+single UTXO with the existing UTXO, and relocks that transactions without a
+30-day refund clause to same wallet.  This has two main effects: it
 consolidates the UTXO set and it disables the refund.
 
 Caveat: We only include deposits in batches that have at least
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -194,6 +194,16 @@ Governable Parameters
 - `sweep_max_btc`: The amount of total bitcoin that will trigger an early sweep.
 - `dust_threshold`: The minimum bitcoin deposit amount for the transaction to
   be included in a sweep.
+- `sweeping_fee_bump_period`: The amount of time we wait to see if a sweeping
+  tranaction is mined before increasing the fee.
+- `sweeping_fee_multiplier_increment`: The amount we add to the sweeping fee
+  multiplier each time a sweeping transaction is not mined within the
+  `sweeping_fee_bump_period`. For example, if this param is set to 0.2 and we
+  are currently at 1.6x, then the next time we would try 1.8x.
+- `sweeping_fee_max_multiplier`: The highest we will try to increment the fee
+  multiplier to before giving up and picking a new base fee and different
+  deposits to sweep.
+
 
 The operators sign a transaction that unlocks all revealed deposits, combines
 them into a single UTXO with the existing UTXO, and relocks that transactions
@@ -213,7 +223,6 @@ This process is called a "sweep", and occurs after `sweep_period` has passed or
 if enough deposits have accumulated to exceed `sweep_max_btc`, whichever comes
 first. Any deposit below `dust_threshold` is ignored.
 
-
 Caveat: Checking to see if enough deposits have accumulated to exceed
 `sweep_max_btc` is complex. Since whether or not we pick a deposit to include
 in a sweep is based on their associated fee and the bitcoin fee market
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-1.adoc
```diff
@@ -225,9 +234,17 @@ conditions for the btc fee and then off-chain see which deposits *would be*
 swept if we *were* to sweep. If the sum of those deposits exceeds the
 `sweep_max_btc` parameter, then we should initiate a sweep.
 
-After submitting the transaction, it will cost some amount of bitcoin based on
-what miners are charging for the bitcoin fee. This fee comes out of the unswept
-deposits. When the transaction clears, and the information has made its way
+The sweeping transaction will cost some amount of bitcoin based on what miners
+are charging for the bitcoin fee in the current market conditions. The fee is
+paid proportionally from the unswept deposits. Once the transaction is
```
michalinacienciala, line 239: Proportionally to the deposit amount? Meaning the bigger deposit, the bigger fee? Do we want that?
My understanding is that the fee depends on number of UTXOs that are being swept (and also market conditions), so it seems fair to me to split the fee equally between all UTXOs.

docs/rfc/rfc-1.adoc
```diff
@@ -225,9 +234,17 @@ conditions for the btc fee and then off-chain see which deposits *would be*
...
+submitted to the bitcoin mempool, the miners will either include it in a block
+within `sweeping_fee_bump_period` or not. If they don't, then we increment a
+fee multiplier: `fee_multiplier = fee_multiplier +
+sweeping_fee_multiplier_increment` and then calculate the new fee: `fee =
+base_fee * fee_multiplier`. We repeat until either the transaction posts or
+`sweeping_fee_multiplier_increment` exceeds `sweeping_fee_max_multiplier`.
```
michalinacienciala, line 245: Well, I think that there may be situations when we should stop multiplication earlier, because wee need to also take under consideration what was the `max_fee` declared by the owners of the deposits.

This relates to the Q20 which I asked on Discord:
> 20. Will we be able to bump up the fee on the sweeping tx only until we hit the `max_fee` of one of the depositors selected to the sweep?
Let’s say there were four revealed deposits:
utxo_1, max_fee: 100 satoshi
utxo_2, max_fee: 50 satoshi
utxo_3, max_fee: 21 satoshi
utxo_4, max_fee: 10 satoshi
Let’s say algorithm determined that utxos 1-3 should be included in the sweep operation and that 60 satoshi should be a starting fee for the sweeping operation (that gives 20 satoshi per deposit, which is lower that `max_fee` of each depositor).
Now let’s assume that the tx wasn’t mined and it was time to increase the offered fee by 10%, making the fee equal to 66 satoshi (22 satoshi per depositor). The new fee per depositor would be now higher than `max_fee` of the third deposit. What we do in such scenario?
A) Replace old sweeping tx (fee: 60) with a new one (fee: 66), using same utxos, and if it gets mined, charge depositors this way:
>    depositor 3: 21 satoshi
>    depositors 1 and 2: (66-21)/2 satoshi
> B) Replace old sweeping tx (fee: 60) with a new one, involving just utxo_1 and utxo_2 (let’s say that with just 2 deposits fee would be estimated to 50). // Not sure if we can do such replacement in one transaction...
C) Stop bumping up fee and wait patiently until the original tx gets mined
D) ... something I haven’t thought of?

To avoid this problem we could do the packing of the UTXOs more cautiously and choose only those UTXOs that leave us a margin for bumping of the fee to the max `fee_multiplier`. But this way less UTXOs would qualify for the sweep...

docs/rfc/rfc-1.adoc
beaushinkle: extremely good catch

docs/rfc/rfc-1.adoc
beaushinkle: > To avoid this problem we could do the packing of the UTXOs more cautiously and choose only those UTXOs that leave us a margin for bumping of the fee to the max fee_multiplier. But this way less UTXOs would qualify for the sweep...

the current algorithm assumes that all of the included UTXO's have a`max_fee` >= `sweeping_fee_max_multiplier * base_fee` otherwise you run into exactly what you're talking about above

also, maybe for some expectation setting

https://privacypros.io/tools/bitcoin-fee-estimator/

the current bitcoin fee for a normal segwit transaction is $0.73 !

beaushinkle: Oh - before I forget: I know the information is fragmented, but this is mentioned on [L388-390](https://github.com/keep-network/tbtc-v2/pull/65/files#diff-1044c578112615d9cb745f17e8203a444e72ddb3c5d8d5fd97d97038952c77b1R388-R390)

michalinacienciala: Oh, right, I missed that.

docs/rfc/rfc-1.adoc
```diff
@@ -194,6 +194,16 @@ Governable Parameters
...
+- `sweeping_fee_multiplier_increment`: The amount we add to the sweeping fee
+  multiplier each time a sweeping transaction is not mined within the
+  `sweeping_fee_bump_period`. For example, if this param is set to 0.2 and we
+  are currently at 1.6x, then the next time we would try 1.8x.
+- `sweeping_fee_max_multiplier`: The highest we will try to increment the fee
+  multiplier to before giving up and picking a new base fee and different
+  deposits to sweep.
+
 
 The operators sign a transaction that unlocks all revealed deposits, combines
```
michalinacienciala, line 208: Maybe this PR would be good to update this sentence (see my comment https://github.com/keep-network/tbtc-v2/pull/62#discussion_r748040498)? We could simply mention here that only some of the deposits will be combined and link to `bitcoin-sweeping-fee` section for more info.

beaushinkle: addressed in 6b3d7d4

michalinacienciala (approved):

pdyraga (approved):

PR #67, opened 2021-11-30
RFC 1: Miscellaneous Cleanup

beaushinkle:
Quick & Simple PR: I noticed some miscellaneous pieces that were individually wrong but weren't big enough to create cards for, so they got lumped together to create a PR.

tagging @michalinacienciala for review

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -382,6 +382,10 @@ stub!
 [[bitcoin-sweeping-fee]]
 === Bitcoin Sweeping Fee
 
+Governable Parameters
+- `bridge_deposit_percentage_fee`: The percentage fee for depositing bitcoin.
+- `bridge_deposit_flat_fee`: The flat fee for depositing bitcoin.
+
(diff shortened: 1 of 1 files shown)


michalinacienciala:
Looking good. I see that PR is in draft state, so if you're still working on this, I've got one more thing that could be added: listing of governable parameters in the "Bitcoin Sweeping Fee" section - we mention there that the fees are governable, but don't give those params names and don't list them.

beaushinkle:
Definitely down to add that! This PR (and the other ones) are all in draft state so that they don't get merged before the other ones (because they point at an integration branch). Once the earlier PRs get merged I can point this stuff at the main branch otherwise I end up in merge conflict hell!

beaushinkle:
addressed in f9a20cc

michalinacienciala (dismissed):

beaushinkle:
@michalinacienciala now that enough of the original PRs have been merged, i pointed this at main and got the commits resolved. the diff should be identical to when you approved it, so if you're still good feel free to merge

michalinacienciala (approved):

PR #68, opened 2021-11-30
RFC 1: Update Unswept Balances With Refunds Too

beaushinkle:
Previously, the transaction that updates the ethereum account balances only collected information about the most recent sweep. Now, we *also* lump in the deposits that have expired due to nearing their refund time (and thus should have their associated account's unswept balance decrease).

Tagging @pdyraga for review.

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -254,7 +254,9 @@ When the transaction clears, and the information has made its way
 over the relay maintainer, then another transaction needs to be created to on
 the ethereum side to update the account balances. The users unswept balances
 are decreased, and their swept balances are increased (after paying their share
-of the <<bitcoin-sweeping-fee,bitcoin sweeping fee>>).
+of the <<bitcoin-sweeping-fee,bitcoin sweeping fee>>). Additionally, we also
+include any `unsweptBalances` that should be *decreased* due to not being swept
+before the 30-day refund period (from not having a high enough bitcoin fee).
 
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-1.adoc
```diff
@@ -258,7 +258,9 @@ When the transaction clears, and the information has made its way
 over the relay maintainer, then another transaction needs to be created to on
 the ethereum side to update the account balances. The users unswept balances
 are decreased, and their swept balances are increased (after paying their share
-of the <<bitcoin-sweeping-fee,bitcoin sweeping fee>>).
+of the <<bitcoin-sweeping-fee,bitcoin sweeping fee>>). Additionally, we also
+include any `unsweptBalances` that should be *decreased* due to not being swept
+before the 30-day refund period (from not having a high enough bitcoin fee).
```
pdyraga, line 263: This could increase Ethereum transaction fee significantly and I think could be even used to DoS the bridge. For example, I am doing a lot of BTC deposits with unrealistic fees because I know the bridge will not be able to sweep them. I can unlock them after 30 days but one unlucky client needs to pay for the transaction updating unswept balances.

I wonder if it would be possible to move this responsibility to the user. That is, update their unswept balances not confirmed within 30 days in the next reveal TX they are doing. 

It does not help if that account was abandoned though.

beaushinkle: good point - closing the loop on this until I can revisit the design

PR #71, opened 2021-12-03
RFC 1: Dont Reset The Scheddy

beaushinkle:
I'd love to include the letterkenny reference that inspired the title here, but after watching the clip, it has way more language/skin than I remembered, so I left it out and put this in instead. Still a great scene though!

---

This PR adds a section that provides clarity on how other sorts of triggered sweeps interact with sweeps from `sweep_period`. Namely that they don't!

tagging @pdyraga for review


docs/rfc/rfc-1.adoc
```diff
@@ -285,6 +285,36 @@ and blinding factor, and then request that TBTC gets minted ASAP. Checking the
...
+pushing back the next scheduled sweep to 23:00, it remains at 21:00. If by 21:00
+there are any deposits with a high enough <<bitcoin-sweeping-fee,sweeping fee>>
+to be included in a sweep, we do it. Otherwise, we schedule the next sweep for
+05:00 the next day. The process repeats.
+
+Here, we're making the tradeoff between reducing fees (having less frequent
+batches) and increasing reliability from a user experience standpoint. This
+setup allows us to have two active wallets that are always staggered from one
+another, so the worst case scenario for the next sweep is always half of
+`sweep_period`.
```
michalinacienciala, line 316: Not sure I understand this staggering correctly. Do you mean here that if let's say the `sweep_period = 8 hours`, and one one wallet sweeps at 2:00, 10:00, 18:00, ..., then the sweeeping on other will be set to occur at 6:00, 14:00, 22:00, ...?

beaushinkle: Yeah exactly!

I'll add an example in the text

beaushinkle: I'm breaking my own rules a little bit (I probably should have referenced a stub in this PR and then fleshed out the stub in a follow-up), but regardless, addressed in 14fe56c

docs/rfc/rfc-1.adoc
```diff
@@ -312,10 +312,43 @@ to be included in a sweep, we do it. Otherwise, we schedule the next sweep for
...
+12:01, 20:01, 4:01, 16:01, 00:01, and 08:01. Those deposits would be swept at
+16:00, 24:00, 8:00, 20:00, 04:00, and 12:00 respectively, and the amount of
+time to wait is 3h59m.
+
+In the "Bad Example", there are 3 worst case scenarios: a user deposits at
+13:01, 21:01, and 05:01. Those deposits would be swept at 20:00, 04:00, and
+12:00 respectively, and the amount of time to wait is 6h59m.
+
+By intentionally offsetting the sweep schedule, we're able to reduce the worst
+case for a deposit significantly!
```
michalinacienciala, line 373: Let's say that `wallet_1` is the latest wallet. It will mean that all new deposits created from the Dashboard will be directed to that wallet, right? So I think in the worst case scenario (deposits created at 12:01, 20:01, 04:01) the user will need to wait almost 8h for the sweep to happen.
Or am I missing something?
If I'm right, then I don't see a gain from having sweeps on `wallet_1` and `wallet_2` executed at different times and think that having them executed the same time would be clearer for users (especially if in the Dashboard we'll be showing the next sweeping time for the primary wallet).

beaushinkle: This combines a couple of scattered ideas:

- The youngest two wallets are "active" wallets, meaning that they're responsible for sweeping
- The dApp gets to choose which wallet to direct deposits to
- Having more than one active wallet is good for redundancy / DoS protection

The dApp (and non-dApp users) can point to whichever wallet will sweep next rather than the youngest wallet, which should decrease wait times if we do it well

---

also, after typing this out, I'm wondering if having more than two active wallets is worth investigating

michalinacienciala: Oh, I see, I was missing this piece of info:
> The dApp (and non-dApp users) can point to whichever wallet will sweep next rather than the youngest wallet.
With this, it makes sense to do the staggering.

I previously thought that we did the sweeping of the second (older) wallet only to allow for sweeping of the deposits which were assigned to that wallet just before that wallet stopped being the newest one.

I wonder now, as it will be possible to choose in the dapp which of the "active" wallets user wants to deposit to (and user may select a wallet which is soon going to become not-"active"), if we should do something to allow for sweeping of the deposits performed to the wallet that becomes not-"active" before user reveals the deposit. If we don't do anything, user may end up with his funds blocked for 30 days.
For example we could:
- not allow for choosing for a deposit a wallet that is close to become "not-active"
- show a warning next to the problematic wallet informing about potential consequences of choosing it for deposit (or indicate in some other way that the wallet may soon become not-"active").
First option seems safer to me...

beaushinkle: Yeah - I think how the dApp wants to handle it could end up being complex (and ultimately really cool). I talked it over w/ @Battenfield  during the standup today and he seemed excited about it!

Anyway - is there anything specific you want to see in this PR or in a follow up? Super down to elaborate!

michalinacienciala: It would be good to show somewhere in the doc an information about the possibility to choose one of the active wallets when depositing.
In the Overview we have following sentence, which suggests that users can deposit to only one address:
> Depositors send BTC funds to the most-recently-created-wallet by using pay-to-script-hash (P2SH) which contains hashed information about the depositor’s minting ethereum address."

We don't have to do that in this PR, though. I know that there's a discussion regarding how many active wallets there should be (and if they should be created periodically), so we may do the update once we have a final decision on that.

michalinacienciala: It would be good to show somewhere in the doc an information about the possibility to choose one of the active wallets when depositing.
In the Overview we have following sentence, which suggests that users can deposit to only one address:

Depositors send BTC funds to the most-recently-created-wallet by using pay-to-script-hash (P2SH) which contains hashed information about the depositor’s minting ethereum address."

We don't have to do that in this PR, though. I know that there's a discussion regarding how many active wallets there should be (and if they should be created periodically), so we may do the update once we have a final decision on that.

michalinacienciala (approved):
Approving. Leaving for @pdyraga for final review and merge.

docs/rfc/rfc-1.adoc
pdyraga: I will capture it in a follow-up issue and merge this PR as-is.

pdyraga (approved):

PR #73, opened 2021-12-08
RFC 1: Mint From Unswept

beaushinkle:
This PR adds a section about Minting TBTC from unswept deposits!

This one was of the last huge puzzle pieces; big weight off my shoulders 🕺

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -281,6 +281,58 @@ and blinding factor, and then request that TBTC gets minted ASAP. Checking the
 dApp, they can see that they should expect TBTC in their provided wallet
 address in 3 hours with no further interaction.
 
+==== Minting From Unswept Balances
+
+Governable Parameters
+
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-1.adoc
```diff
@@ -285,6 +285,58 @@ and blinding factor, and then request that TBTC gets minted ASAP. Checking the
 dApp, they can see that they should expect TBTC in their provided wallet
 address in 3 hours with no further interaction.
 
+==== Minting From Unswept Balances
```
michalinacienciala, line 288: Here we discuss minting from unswept balances. We should update the 'Minting And Unminting' chapter which suggests that minting is done only from swept balances. We could also consider mentioning how `unsweptBalance` gets updated in case of immediate mint in the `Account Balances` section.

beaushinkle:
shelving this as we rework / rethink the approach for minting from unswept!

PR #74, opened 2021-12-10
RFC 1: Governable Parameter Index!

beaushinkle:
Simple formatting PR - we take all of the existing governable parameters and collect them into an index.

tagging @michalinacienciala for review (feel free to merge if it looks fine to you).

PR #75, opened 2021-12-10
RFC 1: Specify that `threshold` and `group_size` are not Governable

beaushinkle:
Since these two params hold so much weight and influence so many other things, they get special treatment. Rather than letting governance change them, instead they get baked into the bridge contract as a hard-coded magic number. If we want this magic number change, governance needs to vote to have the bridge contract upgraded.

@pdyraga this PR is ready for review - it's marked as a draft since I branched off of `rfc-1-governable-parameters`


michalinacienciala:
I think we can now move the PR to Open state. And I'm not sure how GH will handle the deletion of line 481 after merging of the latest changes from `main` - if it will mark the conflict or not... Before merging we should check if line added in https://github.com/keep-network/tbtc-v2/pull/74/commits/4bb3428120240e85031f0010a674f49a57d7096b gets deleted.

beaushinkle:
done!

pdyraga (approved):

PR #76, opened 2021-12-13
RFC 1: Deposit Overflows

beaushinkle:
This PR specs out the logic for how to handle one wallet receiving a risky amount of btc!

The basic idea is that we add another parameter above `wallet_max_btc`, and any amount of btc above this parameter gets immediately refunded to the depositor's 30-day refund address.

tagging @pdyraga for review

PR #77, opened 2021-12-15
RFC 1: Remove References to `wallet_max_btc`

beaushinkle:
This solidifies the design as having no cap on the amount of BTC any
particular wallet can contain. This simplifies the design, massivly
helps us scale, and helps us deal with whale-deposits. On the downside,
it exposes us to black-swan risk where if the adversary manages to steal
all of the BTC in a big wallet, the system can get absolutely blown out
since a wallet can be bigger than the whole covpool.

this PR is why #76 was closed

tagging @pdyraga for review


docs/rfc/rfc-1.adoc
```diff
@@ -346,19 +346,15 @@ Governable Parameters:
...
 
 A new wallet is created when either enough time has passed as defined in
-`wallet_creation_period` *AND* the wallet contains at least
-`wallet_min_btc` btc *OR* the wallet currently contains more btc than
-`wallet_max_btc`, whichever happens first. To create a new wallet, a group of
-100 operators is selected from the pool of available operators using a process
-called sortition. The probabiliy that a particular operator is chosen is based
-on their stake weight, which in turn is based on the number of `T` tokens they
-have invested in the staking contract.
+`wallet_creation_period` *AND* the wallet contains at least `wallet_min_btc`
```
michalinacienciala, line 353:
```suggestion
`wallet_creation_period` *OR* the wallet contains at least `wallet_min_btc`
```

docs/rfc/rfc-1.adoc
```diff
@@ -346,19 +346,15 @@ Governable Parameters:
...
+btc. To create a new wallet, a group of 100 operators is selected from the pool
+of available operators using a process called sortition. The probabiliy that a
```
michalinacienciala, line 355:
```suggestion
of available operators using a process called sortition. The probability that a
```

michalinacienciala (commented):
Two typos found, other than that looking good.

michalinacienciala:
Oh, also, there's a conflict that needs to be resolved.

docs/rfc/rfc-1.adoc
beaushinkle: This actually is correct at "and". We only want to make a new wallet if the current one has >= `wallet_min_btc` after a week, otherwise, we wait until next week and repeat!

beaushinkle: This is actually correct at *AND*. We create a new wallet every ~week (the `wallet_creation_period`), but only if the wallet has >= `wallet_min_btc`. If it doesn't, we skip wallet creation that week and repeat. Maybe next week it will have accumulated some more BTC and we will create a new wallet when we check again.

michalinacienciala: Oh right, I see now that `AND` should indeed be there. In that case I think we should remove `either` in line 354/352.

docs/rfc/rfc-1.adoc
```diff
@@ -483,8 +479,7 @@ Alphabetized list of Governable Parameters with additional notes.
 - `wallet_creation_period`: How frequently we attempt to create new wallets.
 - `wallet_max_age`: The oldest we allow a wallet to become before we transfer the funds
   to a randomly selected wallet.
-- `wallet_max_btc`: The amount of BTC a wallet can hold before we trigger the
-  early creation of a new wallet.
-- `wallet_min_btc`: The smallest amount of BTC a wallet can hold before we
+- `wallet_min_btc`: The smallest amount of btc a wallet can hold before we
+
```
michalinacienciala, line 483: Obsolete empty line

pdyraga: https://github.com/keep-network/tbtc-v2/pull/77/commits/8d224b523e22ac5c4f989e3be312933d566f055e

docs/rfc/rfc-1.adoc
pdyraga: @michalinacienciala Please see https://github.com/keep-network/tbtc-v2/pull/77/commits/31d5d3402c8ddf946aea18944f1d825f6239a879

pdyraga (approved):
Leaving the final approval to @michalinacienciala given I did two small tweaks in this PR.

michalinacienciala (approved):

PR #80, opened 2021-12-17
Backbone for `tbtc-ts` submodule

lukasz-zimnoch:
Here we introduce a backbone for the future `tbtc-ts` library allowing to interact with tBTC v2 system via CLI or dApp. The whole idea is based on the original `tbtc.js` existing for v1. This module, unlike `tbtc.js` is not an independent repository but a submodule of the `tbtc-v2` monorepo. Worth noting the name is `tbtc-ts` not `tbtc.ts` to avoid confusion with TS files and problems in other places like linting regexp.


tomaszslabon (approved):

tbtc-ts/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "version": "0.1.0",
+  "license": "MIT",
+  "main": "dist/src/index.js",
+  "types": "dist/src/index.d.ts",
+  "scripts": {
+    "format": "yarn run lint && prettier --check .",
+    "format:fix": "yarn run lint:fix && prettier --write .",
+    "lint": "eslint . --ext .js,.ts",
+    "lint:fix": "eslint . --ext .js,.ts --fix",
+    "test": "mocha --exit --recursive 'test/**/*.test.ts'",
```
tomaszslabon, line 12: I created a test folder and a simple test and it didn't work:
```
import { expect } from "chai"

describe("Test", async () => {
  it("test", async () => {
    expect(1).to.equal(1)
  })
})
```

The error was: `SyntaxError: Cannot use import statement outside a module`.

Replacing
```
"test": "mocha --exit --recursive 'test/**/*.test.ts'"
```
with
```
"test": "mocha --require ts-node/register --exit --recursive 'test/**/*.test.ts'",
```
seems to fix the issue

lukasz-zimnoch: Good catch! Fixed it in https://github.com/keep-network/tbtc-v2/pull/80/commits/efb3087e408bfc207cf450e5a6ae95557605dab2 by adding an appropriate `.mocharc.json` file.

tomaszslabon (approved):

Issue #79, opened 2021-12-16
Deposit reveal and sweep proof


pdyraga:
Expected flow:
1. Depositor is assembling P2SH transaction using their ETH address, blinding factor, and signing group public key.
2. Depositor is revealing their deposit on Ethereum providing ETH address, blinding factor, and *something* about UTXO.
3. Wallet is executing a BTC transaction sweeping some revealed deposits. Wallet is free to chose which ones.
4. Wallet is informing Ethereum about the sweep providing SPV proof of the transaction and updates balances in the bank.

Requirements:
- TS code needs to be implemented in such a way that it is easy to extract it later to tbtc.ts library.
- We are not integrating any quickminter logic into this code but we need to keep in mind to implement the code in such a way that adding quickminter later is possible.

Open questions:
- What are we doing with UTXOs that can not be unblocked? Should the wallet be able to remove them from Ethereum and reduce unswept balances? Why would it pay for it? Or should we ignore them and filter them out in the client?
- How fees should work? Is the first UTXO, the last UTXO charged? Or are they all charged evenly on BTC side? How should we validate max fee set by depositors? It feels that we should use the same fee amount for everyone in the batch and just ensure that it is not higher than the individual max fees set. Is it possible to validate individual max fees at all?
- What is happening if one of the transactions in the swept batch does not meet some requirements, for example the max fee? Do we allow to confirm other UTXOs and leaving this one? Sweep transaction already happened.

TODO:
- [x] TS code for assembling P2SH transaction (https://github.com/keep-network/tbtc-v2/pull/81)
- [ ] TS code for revealing P2SH transaction
- [x] Solidity code for revealing deposits (protect against overwriting by malicious actors) (https://github.com/keep-network/tbtc-v2/pull/82 and https://github.com/keep-network/tbtc-v2/pull/90)
- [x] TS code for assembling BTC sweep transaction (no logic for filtering out transactions for sweeping) (https://github.com/keep-network/tbtc-v2/pull/96)
- [ ] TS code for assembling SPV proof (https://github.com/keep-network/tbtc-v2/pull/86)
- [x] Solidity code for validating SPV proof (https://github.com/keep-network/tbtc-v2/pull/92 or https://github.com/keep-network/tbtc-v2/pull/106)
- [x] Draft code for updating Bank swept balances (https://github.com/keep-network/tbtc-v2/pull/92 or https://github.com/keep-network/tbtc-v2/pull/106)
[...]

PR #81, opened 2021-12-21
Assemble P2SH deposit transaction

lukasz-zimnoch:
Refs: https://github.com/keep-network/tbtc-v2/issues/79

This pull request introduces tooling meant to be used for assembling and broadcasting Bitcoin P2SH deposit transactions. Specifically, it extends the `TBTC` interface by adding the following methods:
- `makeDeposit`: Function which can be used to perform an end-to-end deposit flow. Takes deposit parameters, depositor private key, and Bitcoin client as arguments. It scans the depositor account for UTXOs, constructs the P2SH transaction according to given parameters, and broadcasts the transaction over the network. The main purpose of that function are future e2e tests of the whole system.
- `createDepositTransaction`: Function which assembles the P2SH transaction based on given deposit parameters, depositor private key, and UTXOs. It returns the transaction in a raw format. Such a transaction can be used for testing or can be broadcast out-of-band.
- `createDepositScript`: Creates the Bitcoin script used as a lock of the P2SH deposit transaction output. It implements the script described in [RFC-1](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#23-depositing). Returns the serialized script in a raw format for further processing.
- `createDepositScriptHash`:  Function that computes the hash of the script returned by `createDepositScript`. This hash has several use cases. It is used during P2SH transaction creation and while deriving deposit target address.
- `createDepositAddress`: Function that computes the deposit target address for the given Bitcoin network using the output of `createDepositScriptHash`. That address can be passed to the depositor who wants to perform the Bitcoin P2SH transaction on their own.

### DIFF
--- a/tbtc-ts/src/deposit.ts
+++ b/tbtc-ts/src/deposit.ts
@@ -5,7 +5,6 @@ import { opcodes } from "bcoin/lib/script/common"
 // @ts-ignore
 import wif from "wif"
 import { BigNumber } from "ethers"
-import { randomBytes } from "crypto"
 import {
   Client as BitcoinClient,
   isCompressedPublicKey,
@@ -13,11 +12,31 @@ import {
   UnspentTransactionOutput,
 } from "./bitcoin"
 
-// TODO: Documentation
+/**
+ * Contains deposit data.
+ */
 export interface DepositData {
+  /**
+   * Ethereum address prefixed with '0x' that should be used for TBTC accounting.
+   */
   ethereumAddress: string
+
+  /**
+   * Deposit amount in sathoshis.
+   */
   amount: BigNumber
+
+  /**
+   * Compressed (33 bytes long with 02 or 03 prefix) Bitcoin public key that
+   * is meant to be used during deposit refund after the locktime passes.
+   */
   refundPublicKey: string
+
+  /**
+   * An 8 bytes number. Must be unique for given Ethereum address, signing group
+   * public key and refund public key.
+   */
+  blindingFactor: BigNumber
 }
 
 // TODO: Documentation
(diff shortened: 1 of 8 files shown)


pdyraga:
What if we rename `tbtc-ts` to `typescript`? We have `solidity` and TS code is a part of `tbtc-v2`, it does not have to be a separate module. We can do it later, once PRs are merged so that we do not run into conflicts.

tbtc-ts/src/deposit.ts
```diff
@@ -1,7 +1,248 @@
...
+  const depositorKeyRing = createKeyRing(depositorPrivateKey)
+  const depositorAddress = depositorKeyRing.getAddress("string")
+
+  const utxos = await bitcoinClient.findAllUnspentTransactionOutputs(
+    depositorAddress
+  )
+
+  const utxosWithRaw: (UnspentTransactionOutput & RawTransaction)[] = []
+  for (const utxo of utxos) {
+    const rawTransaction = await bitcoinClient.getRawTransaction(
```
tomaszslabon, line 69: We will have to add error handling for communication with bitcoin, as it may always fail (I suggest we do that in another PR).

tomaszslabon (approved):

tbtc-ts/src/deposit.ts
lukasz-zimnoch: Yep for sure but that's the matter of the implementation, not the `BitcoinClient` interface. Agree we should address it within the PR introducing Electrum client.

lukasz-zimnoch:
> What if we rename `tbtc-ts` to `typescript`? We have `solidity` and TS code is a part of `tbtc-v2`, it does not have to be a separate module. We can do it later, once PRs are merged so that we do not run into conflicts.

Personally, I'm not a fan of naming those subdirectories by language names but I'd rather name them feature-wise. In my opinion, names like `solidity` or `typescript` give no clue about what that directory contains and should contain. For example, `solidity` directory contains typescript tests and deployment scripts as well. Such a directories also tend to be just bags for code written in a given language, not necessarily connected functionally. However, I know it's the convention we used in our projects so far and that requires more discussion to be changed. I totally agree we should address it in a separate PR/issue once we merge some code and see the direction things actually go.

PR #83, opened 2021-12-21
Assemble sweep transaction

tomaszslabon:
Refs: https://github.com/keep-network/tbtc-v2/issues/79

This pull request introduces tooling meant to be used for assembling and broadcasting Bitcoin P2WPKH sweep transactions. Specifically, it extends the TBTC interface by adding the following methods:
* `sweepUtxos`: Function that can be used to perform the end-to-end sweep transaction. It combines the provided UTXOs, creates a transaction with one output with value equal to the sum of all UTXOs values minus fee. The recipient of the created output is the wallet itself. NOTE: The caller must ensure the provided UTXOs are correctly formed, the wallet can spend them and their total value is bigger than the provided fee. The main purpose of that function are future e2e tests of the whole system.
* `createSweepTransaction`: Function that assembles the P2WPKH sweep transaction based on the provided UTXOs, fee and the private key of the wallet. It returns the transaction in a raw format. Such a transaction can be used for testing or can be broadcast out-of-band.


pdyraga:
What if we rename `tbtc-ts` to `typescript`? We have `solidity` and TS code is a part of `tbtc-v2`, it does not have to be a separate module. We can do it later, once PRs are merged so that we do not run into conflicts.

tomaszslabon:
> What if we rename `tbtc-ts` to `typescript`? We have `solidity` and TS code is a part of `tbtc-v2`, it does not have to be a separate module. We can do it later, once PRs are merged so that we do not run into conflicts.

@pdyraga We can do that, if @lukasz-zimnoch is ok with this idea.

tbtc-ts/src/index.ts
```diff
@@ -76,6 +78,40 @@ export interface TBTC {
 
   // TODO: Implementation and documentation.
   revealDeposit(): Promise<void>
+
+  /**
+   * Creates a Bitcoin P2WPKH sweep transaction.
```
lukasz-zimnoch, line 83: We probably want to keep it P2WPKH eventually but given our current struggle with P2SH and P2WSH, I just keep a note here this can be changed in the future.

tbtc-ts/src/sweep.ts
```diff
@@ -0,0 +1,107 @@
...
+    value: totalInputValue.toNumber(),
+  })
+
+  await transaction.fund(inputCoins, {
+    changeAddress: walletAddress,
+    hardFee: fee.toNumber(),
+    subtractFee: true,
+  })
+
+  transaction.sign(walletKeyRing)
```
lukasz-zimnoch, line 102: That won't work. We have P2(W)SH outputs taken as inputs here. To unblock such inputs you must assemble an appropriate unlocking script for each input. 

For P2SH the locking script is `OP_HASH160 <redeemScript hash> OP_EQUAL` so the unlocking script must be `<something> <redeemScript>` where `something` is used during script computation and depends on the script content. The `redeemScript` is the serialized script used to lock the P2SH output which can be obtained using `createDepositScript` function.

For P2WSH, the locking script is `OP_0 <redeemScript hash>` and the unlocking script should look similar as in P2SH.

tbtc-ts/src/sweep.ts
```diff
@@ -0,0 +1,107 @@
...
+ */
+export async function createSweepTransaction(
+  utxos: (UnspentTransactionOutput & RawTransaction)[],
+  fee: BigNumber,
+  walletPrivateKey: string
+): Promise<RawTransaction> {
+  const decodedWalletPrivateKey = wif.decode(walletPrivateKey)
+
+  const walletKeyRing = new bcoin.KeyRing({
+    witness: true,
```
lukasz-zimnoch, line 68: Have we confirmed that actually produces a P2WPKH transaction? Shouldn't we set something more in the `transaction.addOutput` or `transaction.fund` lines?

tbtc-ts/src/sweep.ts
```diff
@@ -0,0 +1,107 @@
...
+  Client as BitcoinClient,
+} from "./bitcoin"
+
+/**
+ * Sweeps UTXOs by combining all the provided UTXOs and broadcasting a Bitcoin
+ * P2WPKH sweep transaction.
+ * @dev The caller is responsible for ensuring the provided UTXOs are correctly
+ *      formed, can be spent by the wallet and their combined value is greater
+ *      then the fee.
+ * @param utxos - UTXOs to be combined into one output.
```
lukasz-zimnoch, line 18: There should be also the UTXO from the previous sweep. Worth noting that in all TSDocs.

lukasz-zimnoch:
Was this closed in favor of https://github.com/keep-network/tbtc-v2/pull/96 ?

Issue #40, opened 2021-08-30
Earn interest without minting


pdyraga:
We want to enable someone to:
- Deposit 1 BTC via the Bitcoin network,
- Get a non-transferable balance of 1 BTC in the bridge,
- Allow that balance to be borrowed against to mint tBTC, including an origination fee and a charged APR,
- If the fees aren't paid over time, allow that balance to be seized / liquidated.

If someone doesn't care about all that noise, they can just use an "abandonBalanceAndMint" to pay a one-time fee and get their tBTC. But if they do care, we now have a bunch more tools for later.  For example, we can now have people deposit BTC, not mint, and start earning interest. It also opens up lending other assets against that balance (borrow DAI against your BTC balance) and works like a BTC bank. It's similar to Maker CDPs / vaults, though it can be far simpler.

This issue aims at researching the API needed to make it happen and writing a formal specification.

Issue #39, opened 2021-08-30
Upgradeability


pdyraga:
The first iteration of tBTC v2 may not contain all features we would like to implement to make depositors' and signers' life better. One example could be improvements to TBTC minting flow to safely mint the token before sweep operation completes (see https://github.com/keep-network/tbtc-v2/issues/37). 

We need to design the system in upgradeable components that can be easily upgraded by the governance with no risk to the user's funds.

This issue aims at defining an upgradeability scheme for the system and documenting it. Upgradeability of vaults should also be included.

It is an epic for all upgradeability-related issues.

PR #84, opened 2021-12-28
The Bitcoin Bank

pdyraga:
Refs https://github.com/keep-network/tbtc-v2/issues/40

`Bank` contract is a central component tracking Bitcoin balances. Balances can be transferred between holders and holders can approve their balances to be spent by others. Balances in the Bank are updated for depositors who deposit their Bitcoin into the Bridge and only the Bridge can increase balances.

`Bank` contract is a governable contract and the Governance can upgrade the Bridge address.

### DIFF
--- a/solidity/contracts/bank/Bank.sol
+++ b/solidity/contracts/bank/Bank.sol
@@ -19,20 +19,38 @@ contract Bank is Ownable {
     }
 
     function increaseBalance(address recipient, uint256 amount)
-        external
+        public
         onlyBridge
     {
         require(
             recipient != address(this),
-            "Bank itself can not have a balance"
+            "Can not increase balance for Bank"
         );
         balanceOf[recipient] += amount;
     }
 
+    function increaseBalances(
+        address[] calldata recipients,
+        uint256[] calldata amounts
+    ) external {
+        for (uint256 i = 0; i < recipients.length; i++) {
+            increaseBalance(recipients[i], amounts[i]);
+        }
+    }
+
     function decreaseBalance(address spender, uint256 amount)
-        external
+        public
         onlyBridge
     {
         balanceOf[spender] -= amount;
+    }
+
+    function decreaseBalances(
+        address[] calldata spenders,
+        uint256[] calldata amounts
+    ) external {
+        for (uint256 i = 0; i < spenders.length; i++) {
+            decreaseBalance(spenders[i], amounts[i]);
+        }
     }
 }
--- a/solidity/contracts/bank/Bank.sol
+++ b/solidity/contracts/bank/Bank.sol
@@ -236,6 +236,10 @@ contract Bank is Ownable {
         address[] calldata recipients,
         uint256[] calldata amounts
     ) external onlyBridge {
+        require(
+            recipients.length == amounts.length,
+            "Arrays must have the same length"
+        );
         for (uint256 i = 0; i < recipients.length; i++) {
             _increaseBalance(recipients[i], amounts[i]);
         }
--- a/solidity/contracts/bank/Bank.sol
+++ b/solidity/contracts/bank/Bank.sol
@@ -6,12 +6,23 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 contract Bank is Ownable {
     address public bridge;
 
-    /// @notice The balance of the given account in the Bank.
+    /// @notice The balance of the given account in the Bank. Zero by default.
     mapping(address => uint256) public balanceOf;
+
+    /// @notice The remaining amount of balance that spender will be
+    ///         allowed to transfer on behalf of owner through `transferFrom`.
+    ///         Zero by default.
+    mapping(address => mapping(address => uint256)) public allowance;
 
     event BalanceTransferred(
         address indexed from,
         address indexed to,
+        uint256 amount
+    );
+
+    event BalanceApproved(
+        address indexed owner,
+        address indexed spender,
         uint256 amount
     );
 
@@ -26,6 +37,26 @@ contract Bank is Ownable {
 
     function transferBalance(address recipient, uint256 amount) external {
         _transferBalance(msg.sender, recipient, amount);
+    }
+
+    function approveBalance(address spender, uint256 amount) external {
+        _approveBalance(msg.sender, spender, amount);
+    }
+
+    function transferBalanceFrom(
+        address spender,
+        address recipient,
+        uint256 amount
+    ) external {
+        uint256 currentAllowance = allowance[spender][msg.sender];
+        if (currentAllowance != type(uint256).max) {
+            require(
+                currentAllowance >= amount,
+                "Transfer amount exceeds allowance"
+            );
+            _approveBalance(spender, msg.sender, currentAllowance - amount);
+        }
+        _transferBalance(spender, recipient, amount);
     }
 
     function increaseBalances(
@@ -83,5 +114,15 @@ contract Bank is Ownable {
         balanceOf[spender] = spenderBalance - amount;
         balanceOf[recipient] += amount;
         emit BalanceTransferred(spender, recipient, amount);
+    }
+
+    function _approveBalance(
+        address owner,
+        address spender,
+        uint256 amount
+    ) private {
+        require(spender != address(0), "Can not approve to the zero address");
+        allowance[owner][spender] = amount;
+        emit BalanceApproved(owner, spender, amount);
     }
 }
--- a/solidity/contracts/bank/Bank.sol
+++ b/solidity/contracts/bank/Bank.sol
@@ -62,6 +62,10 @@ contract Bank is Ownable {
         uint256 amount
     );
 
+    event BalanceIncreased(address indexed owner, uint256 amount);
+
+    event BalanceDecreased(address indexed owner, uint256 amount);
+
     modifier onlyBridge() {
         require(msg.sender == address(bridge), "Caller is not the bridge");
         _;
@@ -202,6 +206,7 @@ contract Bank is Ownable {
     ///         you really know what you are doing!
     function decreaseBalance(uint256 amount) external {
         balanceOf[msg.sender] -= amount;
+        emit BalanceDecreased(msg.sender, amount);
     }
 
     /// @notice Increases balance of the provided `recipient` by the provided
@@ -215,6 +220,7 @@ contract Bank is Ownable {
             "Can not increase balance for Bank"
         );
         balanceOf[recipient] += amount;
+        emit BalanceIncreased(recipient, amount);
     }
 
     /// @notice Returns hash of EIP712 Domain struct with `TBTC Bank` as
--- a/solidity/contracts/bank/Bank.sol
+++ b/solidity/contracts/bank/Bank.sol
@@ -18,6 +18,24 @@ contract Bank is Ownable {
         bridge = _bridge;
     }
 
+    function increaseBalances(
+        address[] calldata recipients,
+        uint256[] calldata amounts
+    ) external {
+        for (uint256 i = 0; i < recipients.length; i++) {
+            increaseBalance(recipients[i], amounts[i]);
+        }
+    }
+
+    function decreaseBalances(
+        address[] calldata spenders,
+        uint256[] calldata amounts
+    ) external {
+        for (uint256 i = 0; i < spenders.length; i++) {
+            decreaseBalance(spenders[i], amounts[i]);
+        }
+    }
+
     function increaseBalance(address recipient, uint256 amount)
         public
         onlyBridge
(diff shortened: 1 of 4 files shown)


solidity/contracts/bank/Bank.sol
```diff
@@ -0,0 +1,291 @@
...
+
+    constructor() {
+        cachedChainId = block.chainid;
+        cachedDomainSeparator = buildDomainSeparator();
+    }
+
+    /// @notice Allows the Governance to upgrade the Bridge address.
+    /// @dev The function does not implement any governance delay and does not
+    ///      check the status of the Bridge. The Governance implementation needs
+    ///      to ensure all requirements for the upgrade are satisfied before
```
mhluongo, line 82: In particular, you mean this owner should probably be a `Timelock`, right?

solidity/contracts/bank/Bank.sol
```diff
@@ -0,0 +1,291 @@
...
+        address recoveredAddress = ecrecover(digest, v, r, s);
+        require(
+            recoveredAddress != address(0) && recoveredAddress == owner,
+            "Invalid signature"
+        );
+        _approveBalance(owner, spender, amount);
+    }
+
+    /// @notice Increases balances of the provided `recipients` by the provided
+    ///         `amounts`. Can only be called by the Bridge.
```
mhluongo, line 194: Anyone can call this function :)

mhluongo (commented):
Direction looks good. Left a couple comments, will give it another look soon

solidity/contracts/bank/Bank.sol
pdyraga: Yes, plus I imagine there has to be a functionality pausing reveals in the Bridge and giving the wallet a chance to sweep deposits before the upgrade is finalized. That's why I left it unspecified in the doc, just "ensure all requirements are satisfied".

solidity/contracts/bank/Bank.sol
pdyraga: No because `increaseBalance` called by this function is `onlyBridge` :) This is tested [here](https://github.com/keep-network/tbtc-v2/pull/84/files#diff-2f96581e8d80fd83eef0ed8eed414e6445fe9fb7d9f0a673657250cc4539ee25R986-R997). I was thinking about adding `onlyBridge` here as well to make it explicit but my greed for gas savings won.

solidity/contracts/bank/Bank.sol
```diff
@@ -0,0 +1,291 @@
...
+    ) external {
+        for (uint256 i = 0; i < recipients.length; i++) {
+            increaseBalance(recipients[i], amounts[i]);
+        }
+    }
+
+    /// @notice Decreases caller's balance by the provided `amount`. There is no
+    ///         way to restore the balance so do not call this function unless
+    ///         you really know what you are doing!
+    function decreaseBalance(uint256 amount) external {
```
tomaszslabon, line 208: So here we will be able to force `msg.sender` to decrease their balance when this is needed, right?

pdyraga: Correct. This function should be called by the Bridge most of the time. Initially, I gave the Bridge option to decrease anyone's balance but changed it in https://github.com/keep-network/tbtc-v2/pull/84/commits/cfc25677d26897ec7d5393a476dc728286da3830. I think this option is a bit safer. During the redemption, depositor will first transfer their Bitcoin balance to the Bridge, and once the Bitcoin is redeemed, Bridge will decrease its own balance. Of course, it does not fully protect against compromised Bridge because it can still increase balances but at least the decrease path is covered.

solidity/contracts/bank/Bank.sol
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    /// @notice Allows the Governance to upgrade the Bridge address.
+    /// @dev The function does not implement any governance delay and does not
+    ///      check the status of the Bridge. The Governance implementation needs
+    ///      to ensure all requirements for the upgrade are satisfied before
+    ///      executing this function.
+    function updateBridge(address _bridge) external onlyOwner {
+        require(_bridge != address(0), "Bridge address must not be 0x0");
+        bridge = _bridge;
```
tomaszslabon, line 86: Would emitting an event make sense here?

solidity/test/bank/Bank.test.ts
```diff
@@ -0,0 +1,1109 @@
...
+
+      it("should emit the BalanceTransferred event", async () => {
+        await expect(tx)
+          .to.emit(bank, "BalanceTransferred")
+          .withArgs(spender.address, recipient, 0)
+      })
+    })
+  })
+
+  describe("approve", () => {
```
tomaszslabon, line 249: It should be named "approveBalance".

solidity/contracts/bank/Bank.sol
pdyraga: Yes! https://github.com/keep-network/tbtc-v2/pull/84/commits/2f16c5d154bf93c8db3d30a1438f756538f779ea

solidity/test/bank/Bank.test.ts
pdyraga: 😶 https://github.com/keep-network/tbtc-v2/pull/84/commits/2a97d82569ac798d9b423f10b3023a6744ac4283

solidity/contracts/bank/Bank.sol
mhluongo: 100% should be mentioned in `@dev` but I'm all for those sweet gas savings

pdyraga: Clarified in https://github.com/keep-network/tbtc-v2/pull/84/commits/dbbb8d81038dbf781f2222c888f26d3a52d9ef0c and then further optimized in https://github.com/keep-network/tbtc-v2/pull/84/commits/734ba437ef6fe899f8728cd7fcdc787aefc61252.

solidity/contracts/bank/Bank.sol
```diff
@@ -0,0 +1,299 @@
...
+        address recipient,
+        uint256 amount
+    ) external {
+        uint256 currentAllowance = allowance[spender][msg.sender];
+        if (currentAllowance != type(uint256).max) {
+            require(
+                currentAllowance >= amount,
+                "Transfer amount exceeds allowance"
+            );
+            _approveBalance(spender, msg.sender, currentAllowance - amount);
```
lukasz-zimnoch, line 134: Same comment regarding `unchecked`.

solidity/contracts/bank/Bank.sol
```diff
@@ -0,0 +1,299 @@
...
+    ///         balance.
+    /// @dev If the `amount` is set to `type(uint256).max` then
+    ///      `transferBalanceFrom` will not reduce an allowance.
+    ///      Beware that changing an allowance with this function brings the
+    ///      risk that someone may use both the old and the new allowance by
+    ///      unfortunate transaction ordering. One possible solution to mitigate
+    ///      this race condition is to first reduce the spender's allowance to 0
+    ///      and set the desired value afterwards:
+    ///      https://github.com/ethereum/EIPs/issues/20#issuecomment-263524729
+    function approveBalance(address spender, uint256 amount) external {
```
lukasz-zimnoch, line 111: We don't add `increaseAllowance` and `decreaseAllowance` methods that can mitigate the problem described in the comment. Is there any particular reason for not adding them here?

solidity/contracts/bank/Bank.sol
```diff
@@ -0,0 +1,299 @@
...
+        cachedChainId = block.chainid;
+        cachedDomainSeparator = buildDomainSeparator();
+    }
+
+    /// @notice Allows the Governance to upgrade the Bridge address.
+    /// @dev The function does not implement any governance delay and does not
+    ///      check the status of the Bridge. The Governance implementation needs
+    ///      to ensure all requirements for the upgrade are satisfied before
+    ///      executing this function.
+    function updateBridge(address _bridge) external onlyOwner {
```
lukasz-zimnoch, line 86: Regarding your comment: https://github.com/keep-network/tbtc-v2/pull/84/files?diff=unified&w=0#r775931935, don't we want to prepare some "template" of the update process at this point? I mean something like that:
1. We can introduce `IBridge` interface which exposes `disable` and `isDisabled` methods (or any other appropriate names). Implementations could emit special alerts to wallets upon `disable` call to force the sweep. The `isDisabled` will return `true` only if `disable` has been called and deposits have been swept.
2. We divide the update process into two steps. First, we can have `beginBridgeUpdate` which triggers the `disable` method on the old bridge.
3. The second part `finalizeBridgeUpdate` checks whether the old bridge `isDisabled` and makes the actual update. 

That's just a loose idea but I think we should make the update process explicit. Currently, it looks like the bank can update the bridge whenever it wants.

solidity/contracts/bank/Bank.sol
```diff
@@ -0,0 +1,299 @@
...
+        balanceOf[recipient] += amount;
+        emit BalanceTransferred(spender, recipient, amount);
+    }
+
+    function _approveBalance(
+        address owner,
+        address spender,
+        uint256 amount
+    ) private {
+        require(spender != address(0), "Can not approve to the zero address");
```
lukasz-zimnoch, line 280: I see we prevent against transferring to the bank address and increasing their balance. Shouldn't we protect against allowances having the bank address as spender for consistency?

solidity/contracts/bank/Bank.sol
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    /// @notice Increases balances of the provided `recipients` by the provided
+    ///         `amounts`. Can only be called by the Bridge.
+    /// @dev This function fails if the lengths of the arrays are not the same.
+    function increaseBalances(
+        address[] calldata recipients,
+        uint256[] calldata amounts
+    ) external onlyBridge {
+        for (uint256 i = 0; i < recipients.length; i++) {
```
lukasz-zimnoch, line 203: Maybe it's worth requiring that `recipients` and `amounts` have the same length to fail fast here? Worth noting that if `recipients` is shorter than `amounts` that function won't revert so the `@dev` comment is not accurate here.

lukasz-zimnoch (commented):
The code looks good! Left a couple of comments. Reviewing tests now.

solidity/contracts/bank/Bank.sol
```diff
@@ -0,0 +1,299 @@
...
+            "Can not transfer to the zero address"
+        );
+        require(
+            recipient != address(this),
+            "Can not transfer to the Bank address"
+        );
+
+        uint256 spenderBalance = balanceOf[spender];
+        require(spenderBalance >= amount, "Transfer amount exceeds balance");
+        balanceOf[spender] = spenderBalance - amount;
```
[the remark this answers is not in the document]
pdyraga: Not a bad idea. On average, it gives savings of ~182 gas units for `transfrBalance`.

solidity/contracts/bank/Bank.sol
pdyraga: Yes, let's do it. It's 162 gas units saving on average for `transferBalanceFrom`.

solidity/contracts/bank/Bank.sol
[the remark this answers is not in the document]
pdyraga: d362e4ea7a506f2cfb0e7e3fe90468eff9275487

solidity/contracts/bank/Bank.sol
pdyraga: d362e4ea7a506f2cfb0e7e3fe90468eff9275487

solidity/contracts/bank/Bank.sol
pdyraga: Good catch. No particular reason, added in https://github.com/keep-network/tbtc-v2/pull/84/commits/0e15708c0dfa651fac9af15a64652c3211b123da.

solidity/contracts/bank/Bank.sol
```diff
@@ -0,0 +1,335 @@
...
+
+    /// @notice Atomically increases the balance allowance granted to `spender`
+    ///         by the caller by the given `addedValue`.
+    function increaseBalanceAllowance(address spender, uint256 addedValue)
+        external
+    {
+        _approveBalance(
+            msg.sender,
+            spender,
+            allowance[msg.sender][spender] + addedValue
```
lukasz-zimnoch, line 122: Since allowance can be set to `type(uint256).max`, can we capture that case in unit tests and assert we revert due to an overflow?

solidity/contracts/bank/Bank.sol
pdyraga: We can but I think it does not hurt that much as transferring to the Bank. If one transfers their balance to the Bank, it is unrecoverable. If they approve Bank to use their balances, they will only lose gas. Bank can not be the `msg.sender` for `Bank.transferBalanceFrom` so I *think* there is no security risk.

lukasz-zimnoch: Ok sure, let's leave as is.

solidity/contracts/bank/Bank.sol
pdyraga: 👌🏻 https://github.com/keep-network/tbtc-v2/pull/84/commits/1cd2af67e7cd92a5e35096f21c40ed4d2dfe013d

solidity/contracts/bank/Bank.sol
pdyraga: I have #39 in the queue and I was planning to tackle it there. We also need to implement upgradeability for TBTC vault by at least transferring TBTC token ownership.

lukasz-zimnoch: Sure, got it!

solidity/contracts/bank/Bank.sol
pdyraga: Agreed. I was trying to optimize gas as much as possible, but this validation feels crucial and is not that gas expensive. Please see https://github.com/keep-network/tbtc-v2/pull/84/commits/8bd182cc69cb964710a120daf550d3113506a672.

PR #86, opened 2021-12-29
Assemble sweep proof transaction

tomaszslabon:
Refs: #79.

This PR contains functionalities in typescript for assembling transaction proofs.


tbtc-ts/src/index.ts
```diff
@@ -143,6 +145,23 @@ export interface TBTC {
...
+   * in the Bitcoin blockchain and has the required number of confirmations as
+   * well as information on the transaction that is needed for processing
+   * on-chain.
+   * @param txId - ID of the sweep transaction.
+   * @param confirmations - Required number of confirmations for the transaction.
+   * @param bitcoinClient - Bitcoin client used to interact with the network.
+   * @returns Data on a sweep transaction with the inclusion proof and
+   *          information needed for processing on-chain.
+   */
+  constructSweepProof(
```
lukasz-zimnoch, line 160: I think we should not tie getting transaction proof with sweeping. Constructing proofs of BTC transactions is generic and can be used for redemption or moving funds proofs as well. 

**My proposition**

1. Remove `constructSweepProof` from `TBTC` and expose `proveSweep` function with the following signature:
```
proveSweep(transactionHash: string, bitcoinClient: BitcoinClient): Promise<void>
```
Note we don't need `confirmations` (more details below) and I think we should use `transactionHash` not  `txId` to be consistent with the way we define a `Transaction` in `bitcoin.ts`.

2. Rename `sweepProof.ts` to just `proof.ts`. Make all the code generic by removing all `sweep` prefixes and suffixes. The `proof.ts` file should expose just one function:
```
createTransactionProof(
    transactionHash: string, 
    confirmations: number, 
    bitcoinClient: BitcoinClient
): Promise<Transaction & Proof>
```
The `Proof` interface should be the same as `TxProof` in `bitcoin.ts`. I'd cut the `Tx` prefix since we are in Bitcoin context so `Proof` should be clear enough. Regarding `TransactionMerkleBranch` I think it should remain as is because it seems to be at the right place.

3. Introduce `bridge.ts` file where `Bridge` interface will live. For now, the `Bridge` interface should expose one method `submitSweepProof` with same signature as the function in `Bridge.sol` contract and probably a getter of `txProofDifficultyFactor` constant (since it's immutable we can cache it somehow). We don't need real implementation for now (apart from a test mock). Having the contract binding hidden behind an interface will give us more flexibility in the future and will prevent leaking contract details into `tbtc-ts` library.

4. Place implementation of `proveSweep` right into `sweep.ts` file. The actual implementation should call `txProofDifficultyFactor` to obtain the required confirmations factor, call `createTransactionProof` from `proof.ts` to build the proof and submit the proof using `submitSweepProof` from `bridge.ts`

**Notice about bitcoin.ts interfaces**

Regarding `bitcoin.ts` interfaces, I think we should be very careful with adding additional fields to Bitcoin types. They should represent actual Bitcoin types (https://developer.bitcoin.org/reference/transactions.html). That said, I think we should just focus on `Transaction` interface, i.e. add `version` and `locktime` fields as numbers but not hex strings. So, at the moment, we have two representations of the Bitcoin transaction:
- `Transaction` is a deserialized form and can be used by CLIs, dApps, etc. easily
- `RawTransaction` is a serialized form and is used to directly interact with the Bitcoin chain. Bear in mind there are two formats of serialization (https://hongchao.me/anatomy-of-raw-bitcoin-transaction) and that has important consequences.

We still need a representation to interact with Ethereum which contains the four fields (version, inputs, outputs, locktime) serialized and distinguishable. We can't just throw them to `RawTransaction` because of two possible serialization formats where the four fields can represent only the basic one but not the extended format. However, we can always obtain the four fields by decomposing the `RawTransaction`. Having that said, I think we should design the following chain:
```
Transaction <--CONVERT--> RawTransaction --DECOMPOSE--> {version, inputs, outputs, locktime}
```
In the future, we can write `TransactionConverter` class that will handle conversions between `Transaction` and `RawTransaction`. For now, we need a `RawTransactionDecomposer` class or `decomposeRawTransaction` function that will handle the decomposition and transform a `RawTransaction` into the four fields needed on the Ethereum side.

What do you think?

lukasz-zimnoch (changes requested):

tbtc-ts/src/bitcoin.ts
```diff
@@ -17,6 +17,23 @@ export interface Transaction {
...
+  /**
+   * The number of confirmations the transaction has received, including the
+   * containing block hash.
+   */
+  confirmations: number
+
+  /**
+   * The full transaction payload as an un-prefixed hex string.
+   */
+  hex: string
```
lukasz-zimnoch, line 35: If we decide to follow the approach proposed in https://github.com/keep-network/tbtc-v2/pull/86#discussion_r819531383 we won't need those fields.

tbtc-ts/src/bitcoin.ts
```diff
@@ -83,6 +100,90 @@ export type UnspentTransactionOutput = TransactionOutpoint & {
...
+   * Single byte-string of 80-byte block headers, lowest height first.
+   */
+  bitcoinHeaders: string
+}
+
+/**
+ * Sweep transaction data for verifying the inclusion of a transaction in the
+ * blockchain and processing inputs and outputs.
+ */
+export interface SweepData {
```
lukasz-zimnoch, line 153: If we decide to follow the approach proposed in https://github.com/keep-network/tbtc-v2/pull/86#discussion_r819531383 we won't need it.

tbtc-ts/src/bitcoin.ts
```diff
@@ -83,6 +90,74 @@ export type UnspentTransactionOutput = TransactionOutpoint & {
...
+  /**
+   * All transaction's outputs prepended by the number of transaction outputs.
+   */
+  txOutputVector: string
+
+  /**
+   * Transaction's locktime.
+   */
+  txLocktime: string
+}
```
lukasz-zimnoch, line 117: What do you think about naming it `DecomposedRawTransaction` and use just `version, inputs, outputs, locktime` as field names?

tbtc-ts/src/bitcoin.ts
```diff
@@ -110,13 +185,128 @@ export interface Client {
...
+
+/**
+ * Gets vector of outputs as hex string from a transaction
+ * @param tx - Transaction
+ * @returns Hex string representation of transaction output vector
+ */
+function getTxOutputVector(tx: TX): string {
+  return vectorToRaw(tx.outputs)
+}
```
lukasz-zimnoch, line 292: I understand the purpose but those helpers are strictly related to transaction decomposition and can live under `decomposeRawTransaction` function to make it clear they are not part of the `bitcoin.ts` module interface. My proposition:
```
export function decomposeRawTransaction(rawTransaction: string): Info {
  const getTxVersion = (tx: TX) => {
     const buffer = bufio.write()
     buffer.writeU32(tx.version)
     return toHex(buffer)
  }

  (...other functions...)

  const tx = TX.fromRaw(Buffer.from(rawTransaction, "hex"), null)

  return {
    txVersion: getTxVersion(tx),
    txInputVector: getTxInputVector(tx),
    txOutputVector: getTxOutputVector(tx),
    txLocktime: getTxLocktime(tx),
  }
}
```

tbtc-ts/src/bitcoin.ts
```diff
@@ -110,13 +185,128 @@ export interface Client {
...
+
+/**
+ * Decomposes a transaction in the raw representation into version, vector of
+ * inputs, vector of outputs and locktime.
+ * @param rawTransaction - Raw (hex string) representation of a transaction
+ * @returns Transaction data with fields represented as strings
+ */
+export function decomposeRawTransaction(rawTransaction: string): Info {
```
lukasz-zimnoch, line 300: Should take `RawTransaction` type as parameter.

tbtc-ts/src/index.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
```diff
@@ -83,6 +100,90 @@ export type UnspentTransactionOutput = TransactionOutpoint & {
   value: number
 }
 
+/**
+ * Information about important parts of a transaction, including inputs and
+ * outputs.
+ */
+export interface TxInfo {
```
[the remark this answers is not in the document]
tomaszslabon: It was renamed to `DecomposedRawTransaction` and it represents data after decomposition from raw transaction.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Good idea. Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
```diff
@@ -83,6 +90,73 @@ export type UnspentTransactionOutput = TransactionOutpoint & {
   value: number
 }
 
+/**
+ * Represents data of a decomposed transaction.
```
lukasz-zimnoch, line 94: `...decomposed raw transaction.`?

tbtc-ts/src/bitcoin.ts
```diff
@@ -83,6 +90,73 @@ export type UnspentTransactionOutput = TransactionOutpoint & {
...
+ */
+export interface DecomposedRawTransaction {
+  /**
+   * Transaction's version.
```
lukasz-zimnoch, line 98: `Transaction version as an un-prefixed hex string.`?

tbtc-ts/src/bitcoin.ts
```diff
@@ -83,6 +90,73 @@ export type UnspentTransactionOutput = TransactionOutpoint & {
...
+   */
+  version: string
+
+  /**
+   * All transaction's inputs, prepended by the number of transaction inputs.
```
lukasz-zimnoch, line 103: `All transaction's inputs prepended by the number of transaction inputs, as an un-prefixed hex string.`

tbtc-ts/src/bitcoin.ts
```diff
@@ -83,6 +90,73 @@ export type UnspentTransactionOutput = TransactionOutpoint & {
...
+   */
+  inputs: string
+
+  /**
+   * All transaction's outputs prepended by the number of transaction outputs.
```
lukasz-zimnoch, line 108: `All transaction's outputs prepended by the number of transaction outputs, as an un-prefixed hex string.`

tbtc-ts/src/bitcoin.ts
```diff
@@ -83,6 +90,73 @@ export type UnspentTransactionOutput = TransactionOutpoint & {
...
+   */
+  outputs: string
+
+  /**
+   * Transaction's locktime.
```
lukasz-zimnoch, line 113: `Transaction locktime as an un-prefixed hex string.`

tbtc-ts/src/bitcoin.ts
```diff
@@ -83,6 +90,73 @@ export type UnspentTransactionOutput = TransactionOutpoint & {
...
+   */
+  locktime: string
+}
+
+/**
+ * Proof that a given transaction is included in the Bitcoin blockchain.
```
lukasz-zimnoch, line 119: `Data required to perform a proof that a given transaction was included in the Bitcoin blockchain.`?

tbtc-ts/src/bitcoin.ts
```diff
@@ -83,6 +90,73 @@ export type UnspentTransactionOutput = TransactionOutpoint & {
...
+ */
+export interface Proof {
+  /**
+   * The merkle proof of transaction inclusion in a block.
```
lukasz-zimnoch, line 123: `... as an un-prefixed hex string.`?

tbtc-ts/src/bitcoin.ts
```diff
@@ -83,6 +90,73 @@ export type UnspentTransactionOutput = TransactionOutpoint & {
...
+   */
+  merkleProof: string
+
+  /**
+   * Transaction index in the block (0-indexed).
+   */
+  txIndexInBlock: number
+
+  /**
+   * Single byte-string of 80-byte block headers, lowest height first.
```
lukasz-zimnoch, line 133: `... as an un-prefixed hex string.`?

tbtc-ts/src/bitcoin.ts
```diff
@@ -110,13 +184,99 @@ export interface Client {
...
+  /**
+   * Gets height of the latest mined block.
+   * @return Height of the last mined block.
+   */
+  latestBlockHeight(): Promise<number>
+
+  /**
+   * Gets concatenated chunk of block headers built on a starting block.
+   * @param blockHeight - Starting block height.
+   * @param confirmations -  Number of confirmations (subsequent blocks) built
```
lukasz-zimnoch, line 204: Maybe `chainLength`? I think `confirmations` are not valid in this context.

tbtc-ts/src/bitcoin.ts
```diff
@@ -110,13 +184,99 @@ export interface Client {
...
    * @param transaction - Transaction to broadcast.
    */
   broadcast(transaction: RawTransaction): Promise<void>
 }
 
+/**
+ * Decomposes a transaction in the raw representation into version, vector of
+ * inputs, vector of outputs and locktime.
+ * @param rawTransaction - Transaction in the raw format.
+ * @returns Transaction data with fields represented as strings.
```
lukasz-zimnoch, line 232: `as un-prefixed hex strings`?

tbtc-ts/src/bitcoin.ts
```diff
@@ -110,13 +184,99 @@ export interface Client {
...
+  getHeadersChain(blockHeight: number, confirmations: number): Promise<string>
+
+  /**
+   * Get proof of transaction inclusion in the block.
+   * @param txHash - Hash of a transaction.
+   * @param blockHeight - Height of the block where transaction was confirmed.
+   * @return Transaction inclusion proof in hexadecimal form.
+   */
+  getTransactionMerkle(
+    txHash: string,
```
lukasz-zimnoch, line 217: `transactionHash`?

tbtc-ts/src/bitcoin.ts
```diff
@@ -110,13 +184,99 @@ export interface Client {
...
+   * @return Concatenation of block headers in a hexadecimal format.
+   */
+  getHeadersChain(blockHeight: number, confirmations: number): Promise<string>
+
+  /**
+   * Get proof of transaction inclusion in the block.
+   * @param txHash - Hash of a transaction.
+   * @param blockHeight - Height of the block where transaction was confirmed.
+   * @return Transaction inclusion proof in hexadecimal form.
+   */
```
lukasz-zimnoch, line 215: Is this correct to talk about "proof" here? I think this is just about a Merkle branch allowing to perform inclusion proof for given transaction.

tbtc-ts/src/sweep.ts
```diff
@@ -371,3 +374,29 @@ async function prepareInputSignData(
...
+  bitcoinClient: BitcoinClient
+): Promise<void> {
+  const confirmations = await bridge.txProofDifficultyFactor()
+  const proof = await createTransactionProof(
+    transactionHash,
+    confirmations,
+    bitcoinClient
+  )
+  const rawTransaction = await bitcoinClient.getRawTransaction(transactionHash)
+  const info = decomposeRawTransaction(rawTransaction)
```
lukasz-zimnoch, line 400: `info` -> `decomposedRawTransaction`?

tbtc-ts/src/sweep.ts
```diff
@@ -371,3 +374,29 @@ async function prepareInputSignData(
...
+ * @returns Empty promise.
+ */
+export async function proveSweep(
+  transactionHash: string,
+  mainUtxo: UnspentTransactionOutput,
+  bridge: Bridge,
+  bitcoinClient: BitcoinClient
+): Promise<void> {
+  const confirmations = await bridge.txProofDifficultyFactor()
+  const proof = await createTransactionProof(
```
lukasz-zimnoch, line 394: `proof` contains `Transaction`. Worth dropping a TODO this can be used instead of calling `bitcoinClient.getRawTransaction` once we have a `Transaction -> RawTransaction` converter implemented.

tbtc-ts/test/data/sweep.ts
```diff
@@ -207,3 +214,186 @@ export const sweepWithMainUtxo: SweepTestData = {
     },
   },
 }
+
+/**
+ * Represents data for tests of assembling sweep transactions.
```
lukasz-zimnoch, line 219: `...assembling sweep proofs.`?

tbtc-ts/test/data/sweep.ts
```diff
@@ -207,3 +214,186 @@ export const sweepWithMainUtxo: SweepTestData = {
...
+ */
+export interface SweepProofTestData {
+  clientData: {
```
lukasz-zimnoch, line 222: What's the reasoning behind naming it `clientData`? What is the "client" here?

tbtc-ts/test/data/proof.ts
```diff
@@ -0,0 +1,347 @@
...
+  Transaction,
+  TransactionMerkleBranch,
+} from "../../src/bitcoin"
+
+/**
+ * Represents a set of data used for given proof scenario.
+ */
+export interface ProofTestData {
+  requiredConfirmations: number
+  clientData: {
```
lukasz-zimnoch, line 13: What's the reasoning behind naming it `clientData`? What is the "client" here?

tomaszslabon: It's data that that needs to be returned from the bitcoin client mock, as opposed to the expected data, or data that drives the test (arguments).

tbtc-ts/test/data/sweep.ts
tomaszslabon: It's data that that needs to be returned from the bitcoin client mock, as opposed to the expected data, or data that drives the test (arguments).

lukasz-zimnoch: Ok got it! So maybe we can avoid ambiguity by naming it, for example, `bitcoinChainData`?

tbtc-ts/test/data/proof.ts
lukasz-zimnoch: Same as in https://github.com/keep-network/tbtc-v2/pull/86#discussion_r822568649

tbtc-ts/test/data/sweep.ts
tomaszslabon: Done.

tbtc-ts/test/data/proof.ts
tomaszslabon: Done.

tbtc-ts/test/data/sweep.ts
tomaszslabon: Done.

tbtc-ts/src/sweep.ts
tomaszslabon: Done.

tbtc-ts/src/sweep.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

tbtc-ts/src/bitcoin.ts
tomaszslabon: Done.

lukasz-zimnoch (approved):

PR #87, opened 2021-12-29
TBTC Vault for Bitcoin Bank

pdyraga:
Closes https://github.com/keep-network/tbtc-v2/issues/40
~~Depends on #84~~
~~Keeping as a draft until ^ is merged.~~

`TBTCVault` is a contract minting and redeeming TBTC based on Bitcoin balances in the Bank. Every other contract working as a vault implemented on top of the Bank will have its own balance. This way, problems in a single vault are not propagated to other vaults or `Bank` depositors. In other words, one vault can not mess with the Bitcoin balances of other vaults or EOA depositors who have not approved the vault for their Bitcoin balances.

`TBTCVault` is the owner of TBTC token contract and is the only contract minting the token.

User experience for the deposit/mint flow is as follows:
- _Depositor is revealing their P2SH Bitcoin deposit to the Bridge._
- _Bridge is sweeping and updating Depositor's balance in the Bank._
- Depositor is approving `TBTCVault` to use some amount of their Bitcoin Bank balance.
- Depositor is minting TBTC token.

User experience for the redemption flow is as follows:
- Depositor is calling `TBTC.approveAndCall` to `TBTCVault`.
- Vault is burning Depositor's TBTC and increasing their balance in Bitcoin Bank.
- _Depositor can now request Bitcoin redemption from the Bank._

Separately from this PR, we will implement a simplified flow when the user in their P2SH Bitcoin deposit reveal transaction to the Bridge will be able point to which vault it wants to allocate swept deposit (the [code](https://github.com/keep-network/tbtc-v2/blob/cd434b847d1fa33a36db2bf535bafa0dd706a781/solidity/contracts/bridge/Bridge.sol#L90) is already there), and the vault will mint the token - or do something else - as a part of the sweep transaction.

Gas usage (min / max / avg):

```
···················|··············|··············|·············|·············|
|  TBTCVault       ·  mint        ·       72367  ·     123679  ·      94354  │
···················|··············|··············|·············|·············|
|  TBTCVault       ·  redeem      ·       63401  ·      68213  ·      66403  |
···················|··············|··············|·············|·············|
```


solidity/contracts/tbtc/TBTCVault.sol
```diff
@@ -0,0 +1,91 @@
...
+    ///       TBTC Vault.
+    /// @param amount Amount of TBTC to redeem
+    function redeem(uint256 amount) external {
+        _redeem(msg.sender, amount);
+    }
+
+    /// @notice Unlocks the given `amount` and transfers it back to the caller's
+    ///         balance in the Bank. Burns `amount` of TBTC from the caller's
+    ///         account.
+    /// @dev This function is doing the same as `reedeem` but it allows to
```
tomaszslabon, line 69: reedeem -> redeem

solidity/test/tbtc/TBTCVault.test.ts
```diff
@@ -0,0 +1,472 @@
...
+      it("should revert", async () => {
+        await expect(
+          vault.connect(account1).redeem(to1e18(1))
+        ).to.be.revertedWith("Burn amount exceeds balance")
+      })
+    })
+
+    context("when the redeemer has not enough TBTC", () => {
+      const mintedAmount = to1e18(1)
+      const redemeedAmount = mintedAmount.add(1)
```
tomaszslabon, line 247: redemeedAmount -> redeemedAmount

solidity/contracts/bank/Vault.sol
```diff
@@ -0,0 +1,67 @@
...
+        bank = _bank;
+    }
+
+    /// @notice Transfers the given `amount` of the Bank balance from `spender`
+    ///         to the Vault and locks this balance under `spender`'s account.
+    /// @dev Vault must have an allowance for `spender`'s balance for at least
+    ///      `amount`.
+    function lockBalance(address spender, uint256 amount) internal {
+        require(
+            bank.balanceOf(spender) > amount,
```
tomaszslabon, line 49: Shouldn't it be:
```
bank.balanceOf(spender) >= amount,
```
?

pdyraga: Ooops 🙈 
2976a842552795dec2f56f34d621eb65d873ebf0

solidity/test/tbtc/TBTCVault.test.ts
pdyraga: https://github.com/keep-network/tbtc-v2/pull/87/commits/2976a842552795dec2f56f34d621eb65d873ebf0

solidity/contracts/tbtc/TBTCVault.sol
pdyraga: https://github.com/keep-network/tbtc-v2/pull/87/commits/2976a842552795dec2f56f34d621eb65d873ebf0

solidity/contracts/tbtc/TBTCVault.sol
```diff
@@ -0,0 +1,91 @@
...
+    ///      execute redemption without an additional approval transaction.
+    ///      The function can be called only via `approveAndCall` of TBTC token.
+    /// @param from TBTC token holder executing redemption
+    /// @param amount Amount of TBTC to redeem
+    /// @param token TBTC token address
+    function receiveApproval(
+        address from,
+        uint256 amount,
+        address token,
+        bytes calldata
```
dimpar, line 79: Why do we need this param?

pdyraga: Because it's a standard signature for `approveAndCall` and it's how it's implemented in an already deployed token: https://etherscan.io/address/0x18084fbA666a33d37592fA2633fD49a74DD93a88#code

pdyraga:
I realized we probably do not need `Vault` at all, and we do not need `mapping(address => uint256) public lockedBalance`. The fact someone has TBTC token is enough to say they have their balance locked in the vault. Not keeping `lockedBalance` should also help with v1>v2 migration. I will simplify the code.

pdyraga:
Simplified the logic in f6a5feb. We do not need `mapping(address => uint256) public lockedBalance` to track the balances. The fact someone has TBTC token is enough to say they have their balance locked in the application. Even more, TBTC could be transferred to someone else and the new owner should be able to redeem even though they were not the ones who minted. Given we no longer track locked balances, there is no need for `Vault` contract. I removed that contract and renamed `TBTCVault` to just `Application`.

solidity/contracts/tbtc/Vault.sol
```diff
@@ -0,0 +1,104 @@
...
+
+/// @title TBTC application vault
+/// @notice TBTC is a fully Bitcoin-backed ERC-20 token pegged to the price of
+///         Bitcoin. It facilitates Bitcoin holders to act on the Ethereum
+///         blockchain and access the decentralized finance (DeFi) ecosystem.
+///         Vault mints and redeems TBTC based on Bitcoin balances in the
+///         Bank.
+/// @dev Vault is the owner of TBTC token contract and is the only contract
+///      minting the token.
+contract Vault {
```
lukasz-zimnoch, line 29: What do you think about naming it `TBTCVault` and placing that contract in `contracts/vault`. This can make sense since we'll introduce `IVault`. Also, in case we ever want to add a new token & vault pair, that structure seems to be more clear because we would have the token placed in `contracts/token` and the vault in `contracts/vault`. A bare `Vault` looks very generic.

pdyraga:
#84 got merged so I am undrafting this PR.

solidity/test/tbtc/Vault.test.ts
```diff
@@ -0,0 +1,442 @@
...
+        expect(await vault.bank()).to.equal(bank.address)
+      })
+
+      it("should set the TBTC token field", async () => {
+        expect(await vault.tbtcToken()).to.equal(tbtc.address)
+      })
+    })
+  })
+
+  describe("mint", () => {
```
lukasz-zimnoch, line 96: Would it make sense to add a case when one minter uses multiple vaults to mint different assets and assert the bank correctly updates its balance?

solidity/test/tbtc/Vault.test.ts
```diff
@@ -0,0 +1,442 @@
...
+          .to.emit(vault, "Minted")
+          .withArgs(account1.address, to1e18(9))
+        await expect(transactions[4])
+          .to.emit(vault, "Minted")
+          .withArgs(account2.address, to1e18(2))
+      })
+    })
+  })
+
+  describe("redeem", () => {
```
lukasz-zimnoch, line 216: Would it make sense to add a case when one redeemer uses multiple vaults to redeem different assets and assert the bank correctly updates its balance?

solidity/contracts/tbtc/Vault.sol
```diff
@@ -0,0 +1,104 @@
...
+//               ▐████▌    ▐████▌
+
+pragma solidity 0.8.4;
+
+import "../bank/Bank.sol";
+import "../token/TBTC.sol";
+
+/// @title TBTC application vault
+/// @notice TBTC is a fully Bitcoin-backed ERC-20 token pegged to the price of
+///         Bitcoin. It facilitates Bitcoin holders to act on the Ethereum
```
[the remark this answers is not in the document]
pdyraga: There is 1:1 ratio between Bitcoin balance in the Bank and TBTC minted but there is no 1:1 peg between Bitcoin price and TBTC price. Currently, TBTCv1 trades at a premium because of the minting cost and the same could happen to TBTCv2. I think we should not commit to 1:1 ratio.

solidity/contracts/tbtc/Vault.sol
pdyraga: Not a bad idea and I do agree bare `Vault` looks too generic. Done in https://github.com/keep-network/tbtc-v2/pull/87/commits/acdf8fe36f98f058a298c45d8761546bf8a57df6 and PR description is now updated as well.

solidity/test/tbtc/Vault.test.ts
pdyraga: Isn't it something already covered by `Bank.test.ts`? From the Bank's perspective, it does not matter if the balance owner is EOA or smart contract and it is just transferring balances the owner gave an allowance for.

solidity/test/tbtc/Vault.test.ts
pdyraga: IMO the same situation as with `mint`.

solidity/contracts/tbtc/Vault.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Ok, got it! Thanks for the clarification.

solidity/test/tbtc/Vault.test.ts
lukasz-zimnoch: Yep right, `Bank.test.ts` covers that. I was just thinking about making it explicit here but right, that feels redundant.

PR #88, opened 2022-01-04
Unify CI workflows

michalinacienciala:
We have CI workflows scattered across various repositories. We try to keep their implementation as similar as possible. In this PR and the ones listed below we introduce a couple of changes which unify used solutions and fix small mistakes, such as:
* making tagging of published packages consistent
* passing `NPM_TOKEN` secret as variable instead of passing it by writing to file
* handling caching of npm/yarn dependencies via `setup-node` action
* adding `workflow_dispatch` job trigger where it's missing
* using Node.js in version 14.x instead of 12.x where possible

Related PRs in other repositories:
* https://github.com/keep-network/keep-core/pull/2776
* https://github.com/keep-network/keep-ecdsa/pull/941
* https://github.com/keep-network/tbtc/pull/842
* https://github.com/keep-network/tbtc.js/pull/145
* https://github.com/keep-network/sortition-pools/pull/145
* https://github.com/keep-network/local-setup/pull/124
* https://github.com/keep-network/coverage-pools/pull/204
* https://github.com/threshold-network/solidity-contracts/pull/59

### DIFF
--- a/.github/workflows/yearn.yml
+++ b/.github/workflows/yearn.yml
@@ -44,7 +44,7 @@ jobs:
         with:
           node-version: "14.x"
           cache: "yarn"
-          cache-dependency-path: yarn/yarn.lock
+          cache-dependency-path: yearn/yarn.lock
 
       - name: Install dependencies
         run: yarn install
@@ -73,7 +73,7 @@ jobs:
         with:
           node-version: "14.x"
           cache: "yarn"
-          cache-dependency-path: yarn/yarn.lock
+          cache-dependency-path: yearn/yarn.lock
 
       # Below step is a workaround. Eslint executed in `yearn` directory
       # finds `.prettierrc.js` config in the root directory and fails if
--- a/.github/workflows/yearn.yml
+++ b/.github/workflows/yearn.yml
@@ -27,6 +27,7 @@ jobs:
           filters: |
             path-filter:
               - './yearn/**'
+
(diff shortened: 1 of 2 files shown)


.github/workflows/yearn.yml
```diff
@@ -42,6 +43,8 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: "14.x"
+          cache: "yarn"
+          cache-dependency-path: yarn/yarn.lock
```
nkuba, line 47:
```suggestion
          cache-dependency-path: yearn/yarn.lock
```

.github/workflows/yearn.yml
```diff
@@ -69,6 +72,8 @@ jobs:
...
```
nkuba, line 76:
```suggestion
          cache-dependency-path: yearn/yarn.lock
```

nkuba (changes requested):

.github/workflows/yearn.yml
michalinacienciala: Good catch, fixed in 108ceff.

.github/workflows/yearn.yml
michalinacienciala: Good catch, fixed in 108ceff.

nkuba (approved):

PR #93, opened 2022-01-05
RFC 1: Skip the Sweeping Line

beaushinkle:
This PR specs out the rough draft for a mechanism to allow depositors to pay to have their deposits immediately swept.

tagging @pdyraga for review

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -380,6 +380,33 @@ In the "Bad Example", there are 3 worst case scenarios: a user deposits at
 By intentionally offsetting the sweep schedule, we're able to reduce the worst
 case for a deposit significantly!
 
+===== Skipping the Sweeping Line
+
+Governable Parameters
+
+- `skip_sweep_timeout`: The amount of time the depositor has to reimburse the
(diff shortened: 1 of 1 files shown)


pdyraga (approved):

docs/rfc/rfc-1.adoc
```diff
@@ -545,6 +572,8 @@ Alphabetized list of Governable Parameters with additional notes.
   <<heartbeat,heartbeat>> to be successful.
 - `heartbeat_block_length`: The number of ethereum blocks until the next heartbeat.
   If set to 40, then the signers sign every 40th block.
+- `skip_sweep_timeout`: The amount of time the depositor has to reimburse the
+  operator for the gas of the sweep and collect their account balance.
```
pdyraga, line 576: My only worry is:
- The time starts counting after the deposit is swept and that the time it takes to sweep is unknown - hard UX; this timeout needs to be long.
- We probably need some minimum BTC deposit amount (for standard sweep as well!) so that the cost of sweep can not be higher than the fee from BTC amount deposited. This is a bigger issue to think about so not blocking here.

PR #94, opened 2022-01-07
Shortcut for minting TBTC straight from Bridge.sweep transaction

pdyraga:
~~Depends on #87~~ 
~~Keeping as a draft until ^ is merged.~~

`IVault` is an interface for a smart contract consuming Bank balances allowing
the smart contract to receive Bank balances right after sweeping the deposit by
the Bridge. This method allows the depositor to route their deposit revealed to
the Bridge to the particular smart contract in the same transaction the deposit
is revealed. This way, the depositor does not have to execute additional
transaction after the deposit gets swept by the Bridge.

The base cost of increasing TBTC Vault balance in the Bank, without
notifying the application with `onBalanceIncreased` call, and with one element
in `depositors` and `depositedAmounts` calldata is 33,128 gas, assuming TBTC
application balance in the Bank is already non-zero.

Assuming some TBTC was minted before, the full cost of `onBalanceIncreased` with
minting and 1 depositor is 80,202 and 110,532 with 2 depositors. The cost
increases linearly from this point, about 30k per each new depositor:
139,680 for 3 depositors, 171,192 for 4 depositors, 201,534 for 5 depositors,
and 353,197 for 10 depositors.

**Disclaimer**

To avoid merge conflicts with #90, I decided to postpone changes in the `Bridge`
to protect this mechanism against griefing by depositors routing their deposits
to non-existing vaults or vaults that exceed block gas limit when notified.
This work was captured in #94.


lukasz-zimnoch:
@pdyraga Can you solve conflicts and undraft?

solidity/contracts/vault/TBTCVault.sol
```diff
@@ -55,7 +61,29 @@ contract TBTCVault {
...
+        _mint(minter, amount);
+        bank.transferBalanceFrom(minter, address(this), amount);
+    }
+
+    /// @notice Mints the same amount of TBTC as the deposited amount for each
+    ///         depositor in the array. Can only be called by the Bank after the
+    ///         Bridge swept deposits and Bank increased balance for the
+    ///         vault.
+    /// @dev Fails if the lenght of `depositors` and `depositedAmounts`
+    ///      arrays is not the same. Fails if `depositors` array is empty.
```
lukasz-zimnoch, line 78: WIll not fail if `depositedAmounts` is longer than `depositors`. However, if we decide to validate only in `increaseBalanceAndCall`, this is no longer relevant. Also, `require(depositors.length != 0, "No depositors specified");` could be deleted in that case.

solidity/contracts/bank/Bank.sol
```diff
@@ -254,6 +257,31 @@ contract Bank is Ownable {
...
+    /// @dev The `vault` must implement `IVault` intrface.
+    /// @param vault Address of `IVault` recipient contract
+    /// @param depositors Addresses of depositors whose deposits have been swept
+    /// @param depositedAmounts Amounts deposited by individual depositors and
+    ///        swept. The `vault`'s balance in the Bank will be increased by the
+    ///        sum of all elements in this array.
+    function increaseBalanceAndCall(
+        address vault,
+        address[] calldata depositors,
+        uint256[] calldata depositedAmounts
```
lukasz-zimnoch, line 275: Shouldn't we require both arrays to have an equal length and `depositors` length is >0? Also, if we do it here we could make it clear in the contract between `Bank` and `IVault` interface that `onBalanceIncreased` is always called with valid parameters. This way we only pay once for the validation and implementations don't need to bother with that. If we decide to do that, it would be good to update the `@dev` doc.

solidity/contracts/bank/Bank.sol
```diff
@@ -254,6 +257,31 @@ contract Bank is Ownable {
...
+    ///         swept by the Bridge. This way, the depositor does not have to
+    ///         issue a separate  transaction to the `vault` contract.
+    ///         Can be called only by the Bridge.
+    /// @dev The `vault` must implement `IVault` intrface.
+    /// @param vault Address of `IVault` recipient contract
+    /// @param depositors Addresses of depositors whose deposits have been swept
+    /// @param depositedAmounts Amounts deposited by individual depositors and
+    ///        swept. The `vault`'s balance in the Bank will be increased by the
+    ///        sum of all elements in this array.
+    function increaseBalanceAndCall(
```
lukasz-zimnoch, line 272: The `vault` can be theoretically untrusted since the depositor can pass an arbitrary value. A malicious vault can implement `onBalanceIncreased` which can reentry `sweep` again or make the entire transaction out of gas. Shouldn't we protect against that situation by adding vaults whitelisting in the Bridge and/or reentrancy guards where applicable?

lukasz-zimnoch, line 272: Non-blocking but worth noting such a signature of `increaseBalanceAndCall` will require grouping deposits by `vault` during the `sweep` transaction which can have a significant gas cost overhead. I'm wondering whether we should add an auxiliary mapping in the Bridge which will address that at `revealDeposit` stage.

solidity/test/vault/TBTCVault.test.ts
```diff
@@ -439,4 +439,129 @@ describe("TBTCVault", () => {
...
+    context("when called with no depositors", () => {
+      it("should revert", async () => {
+        await expect(
+          bank.connect(bridge).increaseBalanceAndCall(vault.address, [], [])
+        ).to.be.revertedWith("No depositors specified")
+      })
+    })
+
+    context(
+      "when depositors array has different length than amounts array",
```
lukasz-zimnoch, line 481: Would be good to have the opposite case (amounts.length > depositors.length) covered.

solidity/contracts/bank/Bank.sol
pdyraga: Correct. This is covered in https://github.com/keep-network/tbtc-v2/issues/95. I think only the approved vaults should be able to use the shortcut. Everyone is welcome to implement their own vault but they need to use the standard path: increase bank balance, approve and call vault (or do the last two in one TX, see https://github.com/keep-network/tbtc-v2/issues/99). I was initially planning to implement it in this PR but did now want to cause conflicts with changes in the Bridge from #90.

An alternative approach could be to let `IVault.onBalanceIncreased` fail and execute it with a gas limit. But I think this path is way riskier and may lead to bank balances being locked forever under some vault without proper logic executed (e.g. balance transferred but tokens not minted).

pdyraga: Right. I am experimenting with the code from #90 to see if we can store the hash of deposit info instead of the entire deposit info. This should let us reduce the gas expenditure by ~25% for most cases. Then, we'd need to pass deposit info as calldata to `sweep` and this is where we can group them together by vault.

solidity/contracts/bank/Bank.sol
pdyraga: Yes, let's do it. We have length validation in `increaseBalances`, we should also have it here.

solidity/test/vault/TBTCVault.test.ts
pdyraga: https://github.com/keep-network/tbtc-v2/pull/94/commits/a4370e66f09b8da92eb7f14eb941b4c329a6baba

solidity/contracts/vault/TBTCVault.sol
pdyraga: https://github.com/keep-network/tbtc-v2/pull/94/commits/a4370e66f09b8da92eb7f14eb941b4c329a6baba

solidity/contracts/bank/Bank.sol
```diff
@@ -254,6 +257,31 @@ contract Bank is Ownable {
         _increaseBalance(recipient, amount);
     }
 
+    /// @notice Increases the given smart contract `vault`'s balance and
+    ///         notifies the `vault` contract. Called by the Bridge after
+    ///         the deposits routed by depositors to that `vault` have been
+    ///         swept by the Bridge. This way, the depositor does not have to
+    ///         issue a separate  transaction to the `vault` contract.
+    ///         Can be called only by the Bridge.
+    /// @dev The `vault` must implement `IVault` intrface.
```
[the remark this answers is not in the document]
pdyraga: https://github.com/keep-network/tbtc-v2/pull/94/commits/a4370e66f09b8da92eb7f14eb941b4c329a6baba

solidity/contracts/bank/Bank.sol
pdyraga: Done in https://github.com/keep-network/tbtc-v2/pull/94/commits/a4370e66f09b8da92eb7f14eb941b4c329a6baba.

solidity/contracts/vault/TBTCVault.sol
pdyraga: There could be a scenario where some vault allows to pass empty arrays because it does not care about
who deposited, e.g. some fundraising contract. That's why we do not validate if depositors array is non-empty in `Bank` but we validate it in `TBTCVault` - for TBTC it matters.

solidity/contracts/bank/Bank.sol
lukasz-zimnoch: Ok, wasn't aware of #95. Thanks for explanation!

lukasz-zimnoch (approved):

PR #92, opened 2022-01-05
Implement sweep function

tomaszslabon:
This PR adds implementation of the `sweep` function according to the option 1 from https://github.com/keep-network/tbtc-v2/issues/79.

PR #106, opened 2022-01-14
Implement sweep function (option 2)

lukasz-zimnoch:
Refs: #79 

### Introduction

This PR adds an implementation of the `sweep` function according to option 2 described in https://github.com/keep-network/tbtc-v2/issues/79#issuecomment-1004909533. It is an alternative approach to option 1 implemented as part of https://github.com/keep-network/tbtc-v2/pull/92 and closed eventually for the reasons described in https://github.com/keep-network/tbtc-v2/issues/79#issuecomment-1022100331.

### Implementation of `sweep` function in the `Bridge` contract

The most important change introduced by this pull request is the `sweep` function in the `Bridge` contract. It is compatible with the recent refactoring of the `revealDeposit` function, performed within https://github.com/keep-network/tbtc-v2/pull/90 and is able to sweep deposits revealed using that code.

Once an off-chain wallet performs a Bitcoin sweep transaction, it must call the `sweep` function with several parameters containing, among others, the transaction and proof data. The `sweep` function uses those parameters to perform an SPV proof of the received sweep transaction. This is done in a very similar way to deposit/redemption proofs in tBTC v1 and Summa's `bitcoin-spv-sol` library is leveraged on the low level as well. Once the proof is deemed valid, the Bitcoin transaction data are also considered valid and used for further processing.

As the next step, the Bitcoin sweep transaction output vector is processed. The code asserts it contains only one output. The single output is then used to extract the sweep transaction output value and the wallet public key hash. That hash must be 20-byte length. This fact limits the sweep transaction output to be P2PKH or P2WPKH as described in the RFC.

The crucial part of the `sweep` code is input processing. The Bitcoin sweep transaction can contain multiple inputs but each input must point to either a P2(W)SH revealed unswept deposit UTXO or an expected P2(W)PKH previous sweep UTXO. Inputs that don't fall under that rule cause a revert of the entire `sweep` transaction. If given input refers to a revealed deposit, the input's outpoint is used to build the deposit key and fetch deposit data like depositor address and amount from contract storage. If given input doesn't refer to a revealed deposit and a prior sweep exists for the given wallet, the input's outpoint is compared against the previous sweep data information. If they match, such input is considered a valid previous sweep pointer, and its value is used for further processing. Worth noting that previous sweep information is actually stored as `keccak256` hash to make gas costs lower. Because of that, the `sweep` caller must pass plain-text previous sweep information as a parameter which is validated against the stored hash and used for processing. Input processing is also the point where processed deposits are marked as swept. Marking as swept and previous sweep validation form protection against replaying a `sweep` with the same transaction data. 

Once the Bitcoin sweep transaction is processed, the total fee is calculated and each deposit incurs it evenly. Finally, the given sweep information is stored as the wallet's previous sweep and the `Bank` is called to make necessary balances increases.

### Gas costs

Current gas costs of `revealDeposit` and `sweep` functions are presented below: 

```
·-----------------------------------------------------------------|----------------------------|-------------|-----------------------------·
|                       Solc version: 0.8.4                       ·  Optimizer enabled: false  ·  Runs: 200  ·  Block limit: 30000000 gas  │
··································································|····························|·············|······························
|  Methods                                                                                                                                 │
···················|··············································|··············|·············|·············|···············|··············
|  Contract        ·  Method                                      ·  Min         ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
···················|··············································|··············|·············|·············|···············|··············
|  Bridge          ·  revealDeposit                               ·       75494  ·      99998  ·      78936  ·           62  ·          -  │
···················|··············································|··············|·············|·············|···············|··············
|  Bridge          ·  sweep                                       ·      124232  ·     421798  ·     272741  ·           17  ·          -  │
···················|··············································|··············|·············|·············|···············|··············
```
By the way of working on the `sweep` functionality, multiple gas optimizations were introduced by @eth-r in `bitcoin-spv-sol` library code, specifically within https://github.com/keep-network/bitcoin-spv/pull/3. They all were captured under [`solidity/v3.2.0-solc-0.8`](https://github.com/keep-network/bitcoin-spv/releases/tag/solidity%2Fv3.2.0-solc-0.8) tag.

### Next steps

This pull request doesn't address proving single swept balance in case the entire sweep can not be proved for whatever reasons. It also doesn't implement support for shortcut minting using the `vault` address passed during the deposit reveal. Those two topics are captured on #79 TODO list and should be addressed in follow-up PRs.


solidity/contracts/bridge/Bridge.sol
```diff
@@ -74,15 +92,37 @@ contract Bridge {
...
+        // Deposit amount in satoshi.
+        uint64 amount;
         // UNIX timestamp the deposit was revealed at.
         uint32 revealedAt;
         // Address of the tBTC vault.
         address vault;
+        // UNIX timestamp the deposit was swept at. Note this is not the
+        // time when the deposit was swept on Bitcoin chain but actually
+        // the time when the sweep proof was delivered on Ethereum chain.
+        uint32 sweptAt;
```
tomaszslabon, line 104: Not sure if it's worth it, but we could use a bool here. Unless we will be using the `timestamp` for something later on.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -74,15 +92,37 @@ contract Bridge {
...
+        // Amount of the single transaction output.
+        uint256 amount;
     }
 
+    /// @notice Confirmations on the Bitcoin chain required to successfully
+    ///         evaluate an SPV proof.
+    uint256 public immutable txProofDifficultyFactor;
+
+    /// @notice Address of the Bank this Bridge belongs to.
+    Bank public immutable bank;
```
tomaszslabon, line 120: Bank will probably be updatable. I would left a TODO that we should make it updatable for know.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -264,34 +323,309 @@ contract Bridge {
...
+        (, uint256 outputsCount) = sweepTx.outputVector.parseVarInt();
+        require(
+            outputsCount == 1,
+            "Sweep transaction must have a single output"
+        );
+
+        bytes memory output = sweepTx.outputVector.extractOutputAtIndex(0);
+        value = output.extractValue();
+        bytes memory walletPubKeyHashBytes = output.extractHash();
+        // The sweep transaction output should always be P2PKH or P2WPKH.
```
tomaszslabon, line 486: Worth adding this information in the description of `sweep`.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -264,34 +323,309 @@ contract Bridge {
...
+            uint256 inputsValue,
+            address[] memory depositors,
+            uint256[] memory depositedAmounts
+        )
+    {
+        SweepInfo storage previousSweep = sweeps[walletPubKeyHash];
+        // Flag indicating whether the previous sweep transaction output
+        // is included as an input. If current sweep is the first sweep
+        // of given wallet, we use a shortcut and initialize it with `true`.
+        bool previousSweepFlag = previousSweep.txHash == bytes32(0);
```
tomaszslabon, line 516: Could we give it a more clear name, so that we know that if this flag is set to true there was no previous sweeps?
Perhaps `isFirstSweep` or `noPreviousSweeps`.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -264,34 +323,309 @@ contract Bridge {
...
+    }
+
+    // TODO: Documentation. Mention that `txInfo` data should be validated outside.
+    function extractTxInput(
+        BitcoinTx.Info calldata txInfo,
+        uint256 inputStartingIndex
+    ) internal pure returns (bytes memory input, uint256 inputLength) {
+        // First, determine the remaining vector using current input
+        // starting index.
+        bytes memory remainingVector =
```
tomaszslabon, line 613: Every time we want to extract another input we copy the whole remaining part of the input vector.
It would be great if we could optimise it. Perhaps we could change the `determineInputLength` function, so that it can work on the whole input vector, but we could pass a parameter there that indicates the start of the input which length we want to know?

solidity/contracts/bridge/Bridge.sol
```diff
@@ -264,34 +323,309 @@ contract Bridge {
...
+
+        // Inputs processing loop.
+        for (uint256 i = 0; i < inputsCount; i++) {
+            // Check if we are at the end of the input vector.
+            if (inputStartingIndex >= sweepTx.inputVector.length) {
+                break;
+            }
+
+            (bytes memory input, uint256 inputLength) =
+                extractTxInput(sweepTx, inputStartingIndex);
```
eth-r, line 550: It should be possible to change this so that we don't need to allocate the `bytes memory input` at all. I'll make sure that `bitcoin-spv` has the functions we'll need.

solidity/contracts/bridge/Bridge.sol
eth-r: I've added this capability to `bitcoin-spv`: https://github.com/keep-network/bitcoin-spv/pull/3/commits/45dfb963d3777d3bb9ec42c92e6827113895c889#diff-8e482a4e5e5d4d6a42a321153e706c491f81a1e9bf0881b4607ef4f916132370R294

solidity/contracts/bridge/Bridge.sol
eth-r: With https://github.com/keep-network/bitcoin-spv/pull/3/commits/d9248be6f752d636bd3ea12f1fe05aadee6c8cd5 we can use `extractInputTxIdLeAt`, `extractTxIndexLeAt` and `determineInputLengthAt` to replace the `extractTxInput` call altogether.

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: I'm not sure since if I'm not wrong the Bank should be the non-upgradeable element of the system. Can you confirm @pdyraga ?

pdyraga: Can we see what is the gas difference if we make it non-immutable and if it is acceptable, start with a non-immutable version and leave a TODO to revisit it later? I am still playing with upgradeability and I see the pros and cons of both options.

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: Not sure at this point. I'd revisit that later.

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: Making it non-immutable costs `2k`. I'd just leave a TODO for now: https://github.com/keep-network/tbtc-v2/pull/106/commits/4c19bc2367eae0b24a509223dde6e981f70a1bae

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: Yep, done in https://github.com/keep-network/tbtc-v2/pull/106/commits/21ccdb5838e6f0b4f633e292dd1239737c425597

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: I had a hard time trying to get a proper name here. This flag doesn't only indicate whether we should expect the previous sweep but also indicates whether the previous sweep reference was found in case it is expected (there is a prior sweep for a given wallet). Normally, we should have two separate variables but it's Solidity and I stumbled upon `stack too deep` errors which forced me to reduce the number of local variables. This is why I gave it such a generic name which should be interpreted as: "is the previous sweep condition fulfilled?" which means either the previous sweep is not required OR it is required and found.

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/152766279380c868726c81283ee1f7e8f33751fc though I need to keep it within `parseTxInputAt` to reduce local variables count to avoid `stack too deep` errors.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -306,34 +367,398 @@ contract Bridge is Ownable {
...
+                previousSweepFlag = true;
+            } else {
+                revert("Unknown input type");
+            }
+
+            // Make the `inputStartingIndex` pointing to the next input by
+            // increasing it by current input's length.
+            inputStartingIndex += inputLength;
+        }
+
```
tomaszslabon, line 718: Should we allow the wallet to make a sweep transaction that doesn't have any deposit inputs?
Such transaction would only burn some amount of BTC form the main UTXO (needed to pay for tx fee).
Perhaps we could add a check that `processedDepositsCount` is greater than `0`?

solidity/contracts/bridge/Bridge.sol
```diff
@@ -96,19 +138,24 @@ contract Bridge is Ownable {
...
     ///         wallet is responsible for validating them before attempting to
     ///         execute a sweep.
     ///
-    /// TODO: Explore the possibility of storing just a hash of DepositInfo.
-    mapping(uint256 => DepositInfo) public unswept;
+    mapping(uint256 => DepositInfo) public deposits;
+
+    /// @notice Maps the wallet public key hash (computed using HASH160 opcode)
+    ///         to the latest sweep computed as keccak256(txHash | txOutputValue).
+    mapping(bytes20 => bytes32) public sweeps;
```
tomaszslabon, line 152: If I remember correctly we planned to rename it to `mainUtxo`.

lukasz-zimnoch: Indeed. See: https://github.com/keep-network/tbtc-v2/pull/106/commits/3556c9b054412b8fee028248574bd48e800e4cbd and also https://github.com/keep-network/tbtc-v2/pull/106/commits/ff3b59fa2e48eec63ea0f71fdce85d2812a45bd3

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: Very good catch! That behavior is actually a fraud. Added a check: https://github.com/keep-network/tbtc-v2/pull/106/commits/b1c06a9499177b81c124704d51fd9ebcc5f39f73

solidity/test/data/sweep.ts
```diff
@@ -0,0 +1,729 @@
...
+          "0x010ab93f22d004afe73c75e24a9b790854be61f8a262ff7c41022b8cce9c82" +
+          "d8730100000000ffffffff",
+        outputVector:
+          "0x02102700000000000017a914e56ff1445b513721fb0e3981817d1b17d83f95" +
+          "b087acb43a00000000001600147ac2d9378a1c47e589dfb8095ca95ed2140d27" +
+          "26",
+        locktime: "0x00000000",
+      },
+      reveal: {
+        fundingOutputIndex: 0,
```
tomaszslabon, line 368: Would it make sense to write a testcase where `fundingOutputIndex` is grater than `0`?
In theory somebody can make a deposit without using our `makeDeposit` functionality (which always causes `fundingOutputIndex` to be `0`) and deposit UTXO could be at  an index different than `0`.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+            // increasing it by current input's length.
+            inputStartingIndex += inputLength;
+        }
+
+        // Construction of the input processing loop guarantees that:
+        // `processedDepositsCount == depositors.length == depositedAmounts.length`
+        // is always true at this point. If so, we just use the first variable
+        // to assert the total count of swept deposit is bigger than zero.
+        require(
+            processedDepositsCount > 0,
```
tomaszslabon, line 722: At this moment the wallet is allowed to include any number of deposits in a sweep.
If the number of deposits is very large, we will not be able to prove such a sweep (we will run out of gas).
Perhaps we could add a limit for the number of deposits in one sweep? 
Let's say we are sure we will never run out of gas if there are 10 or less deposits.
I would leave a TODO to add such a limit (after testing).
Also we should assume all the deposits will have `vault` set, which will increase gas consumption.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+        ) = processSweepTxInputs(sweepTx.inputVector, resolvedMainUtxo);
+
+        // Compute the sweep transaction fee which is a difference between
+        // inputs amounts sum and the output amount.
+        // TODO: Check fee against max fee.
+        uint256 fee = sweepTxInputsValue - sweepTxOutputValue;
+        // Calculate fee share by dividing the total fee by deposits count.
+        // TODO: Deal with precision loss.
+        uint256 feeShare =
+            depositedAmounts.length > 0 ? fee / depositedAmounts.length : 0;
```
tomaszslabon, line 434: We added a check that there is at least one deposit in `processSweepTxInputs`. I think we can simply:
```
uint256 feeShare = fee / depositedAmounts.length;
 ```

File: solidity/contracts/bridge/Bridge.sol [hunk 12]
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+            address[] memory depositors,
+            uint256[] memory depositedAmounts
+        ) = processSweepTxInputs(sweepTx.inputVector, resolvedMainUtxo);
+
+        // Compute the sweep transaction fee which is a difference between
+        // inputs amounts sum and the output amount.
+        // TODO: Check fee against max fee.
+        uint256 fee = sweepTxInputsValue - sweepTxOutputValue;
+        // Calculate fee share by dividing the total fee by deposits count.
+        // TODO: Deal with precision loss.
```
pdyraga, line 432: Let's deal with it by having the last depositor pay the higher fee than others if there is a change, just like [it's been proposed for redemption flow](https://github.com/keep-network/tbtc-v2/pull/128#discussion_r800555359).

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+        uint256 feeShare =
+            depositedAmounts.length > 0 ? fee / depositedAmounts.length : 0;
+        // Reduce each deposit amount by fee share value.
+        for (uint256 i = 0; i < depositedAmounts.length; i++) {
+            // TODO: The feeShare can be bigger than the amount.
```
pdyraga, line 437: I think it is not possible. We have [a dust threshold](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#234-sweeping) and deposits below the dust should not be considered for sweeping at all. dust threshold must be considerably higher than max fee possible for a single deposit.

pdyraga (commented):
I was looking only at `*.sol`, leaving everything else to @tomaszslabon

solidity/contracts/bridge/Bridge.sol
pdyraga: Related: https://github.com/keep-network/tbtc-v2/pull/106#discussion_r799540214 and https://github.com/keep-network/tbtc-v2/pull/106#discussion_r800655662.

solidity/contracts/bridge/Bridge.sol
pdyraga: > At this moment the wallet is allowed to include any number of deposits in a sweep.
If the number of deposits is very large, we will not be able to prove such a sweep (we will run out of gas).
Perhaps we could add a limit for the number of deposits in one sweep?
Let's say we are sure we will never run out of gas if there are 10 or fewer deposits.

I think it should be off-chain rule in the client. Even if we have a limit of 10 deposits in the contract, it does not stop the off-chain wallet to sweep more. The sweeping will happen on the Bitcoin side anyway.

> Also we should assume all the deposits will have vault set, which will increase gas consumption.

I think deposits directed to vaults will be handled in a separate PR.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +368,398 @@ contract Bridge is Ownable {
...
+
+            DepositInfo storage deposit =
+                deposits[
+                    uint256(
+                        keccak256(abi.encodePacked(inputTxHash, inputTxIndex))
+                    )
+                ];
+
+            if (deposit.revealedAt != 0) {
+                require(deposit.sweptAt == 0, "Deposit already swept");
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/b05e2a6d3a359f4973c981bcc7c605f77ef820f4

solidity/test/data/sweep.ts
tomaszslabon: Sure, let's add a TODO.

lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/b05e2a6d3a359f4973c981bcc7c605f77ef820f4

solidity/contracts/bridge/Bridge.sol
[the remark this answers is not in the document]
lukasz-zimnoch: As @tomaszslabon mentioned in https://github.com/keep-network/tbtc-v2/pull/106#discussion_r799561162, this is no longer relevant since we added this check: https://github.com/keep-network/tbtc-v2/pull/106/commits/b1c06a9499177b81c124704d51fd9ebcc5f39f73. Removed it: https://github.com/keep-network/tbtc-v2/pull/106/commits/4e7e0d1634d5f913df07c53e2eb05d75ec05593d

lukasz-zimnoch: Good point. See https://github.com/keep-network/tbtc-v2/pull/106/commits/4e7e0d1634d5f913df07c53e2eb05d75ec05593d

solidity/contracts/bridge/Bridge.sol
```diff
@@ -77,16 +96,31 @@ contract Bridge is Ownable {
...
         // Address of the Bank vault the deposit is routed to.
         // Optional, can be 0x0.
         address vault;
+        // UNIX timestamp the deposit was swept at. Note this is not the
+        // time when the deposit was swept on Bitcoin chain but actually
+        // the time when the sweep proof was delivered on Ethereum chain.
+        uint32 sweptAt;
     }
 
+    /// @notice Confirmations on the Bitcoin chain required to successfully
```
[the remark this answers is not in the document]
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/106/commits/77613267d7b197ad56424bec80143bda4026aa32

solidity/contracts/bridge/Bridge.sol
```diff
@@ -77,16 +96,31 @@ contract Bridge is Ownable {
...
+        // Deposit amount in satoshi.
+        uint64 amount;
         // UNIX timestamp the deposit was revealed at.
         uint32 revealedAt;
         // Address of the Bank vault the deposit is routed to.
         // Optional, can be 0x0.
         address vault;
+        // UNIX timestamp the deposit was swept at. Note this is not the
+        // time when the deposit was swept on Bitcoin chain but actually
+        // the time when the sweep proof was delivered on Ethereum chain.
```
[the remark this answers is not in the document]
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/106/commits/77613267d7b197ad56424bec80143bda4026aa32

solidity/contracts/bridge/Bridge.sol
```diff
@@ -43,13 +59,16 @@ import "./BitcoinTx.sol";
...
     using BytesLib for bytes;
+    using ValidateSPV for bytes;
+    using ValidateSPV for bytes32;
 
     /// @notice Represents data which must be revealed by the depositor during
     ///         deposit reveal.
     struct RevealInfo {
         // Index of the funding output belonging to the funding transaction.
-        uint8 fundingOutputIndex;
+        uint32 fundingOutputIndex;
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yep, it does make sense. Even if `BTCUtils.extractOutputAtIndex` would not support extracting the output with an index bigger than `252`, the validation of the expected script hash will prevent the reveal transaction to happen. In that case, funds can be recovered using the refund mechanism encoded in the script.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -117,7 +157,21 @@ contract Bridge is Ownable {
         address vault
     );
 
-    event VaultStatusUpdated(address indexed vault, bool isTrusted);
+    event SweepPerformed(bytes20 walletPubKeyHash, bytes32 sweepTxHash);
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/c353b7ea0fb149254102e43cbd43247bfab03a30

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+    {
+        inputTxHash = inputVector.extractInputTxIdLeAt(inputStartingIndex);
+
+        inputTxIndex = BTCUtils.reverseUint32(
+            uint32(inputVector.extractTxIndexLeAt(inputStartingIndex))
+        );
+
+        inputLength = inputVector.determineInputLengthAt(inputStartingIndex);
+
+        return (inputTxHash, inputTxIndex, inputLength);
```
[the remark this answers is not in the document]
lukasz-zimnoch: If you use just `return` you'll receive `TypeError: Return arguments required.`.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
     ///
     ///         It is possible to prove the given sweep only one time.
     /// @param sweepTx Bitcoin sweep transaction data.
-    /// @param merkleProof The merkle proof of transaction inclusion in a block.
-    /// @param txIndexInBlock Transaction index in the block (0-indexed).
-    /// @param bitcoinHeaders Single bytestring of 80-byte bitcoin headers,
-    ///                       lowest height first.
-    function sweep(
+    /// @param sweepProof Bitcoin sweep proof data.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/ec800f4567de3a823cbb753a5b3d0b4b9a2e077f

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+    /// @param mainUtxo Data of the wallet's main UTXO. If no main UTXO
+    ///        exists for given wallet, this parameter is ignored.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/ec800f4567de3a823cbb753a5b3d0b4b9a2e077f

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
     ///
     ///         It is possible to prove the given sweep only one time.
     /// @param sweepTx Bitcoin sweep transaction data.
-    /// @param merkleProof The merkle proof of transaction inclusion in a block.
-    /// @param txIndexInBlock Transaction index in the block (0-indexed).
-    /// @param bitcoinHeaders Single bytestring of 80-byte bitcoin headers,
-    ///                       lowest height first.
-    function sweep(
+    /// @param sweepProof Bitcoin sweep proof data.
+    /// @param mainUtxo Data of the wallet's main UTXO. If no main UTXO
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/ec800f4567de3a823cbb753a5b3d0b4b9a2e077f

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+    ///        exists for given wallet, this parameter is ignored.
+    /// @dev Requirements:
+    ///      - `sweepTx` components must match the expected structure. See
+    ///        `BitcoinTx.Info` docs for reference. Their values must exactly
+    ///        correspond to appropriate Bitcoin transaction fields to produce
+    ///        a provable transaction hash.
+    ///      - The `sweepTx` should represent a Bitcoin transaction with 1..n
+    ///        inputs referring to 0..n P2(W)SH revealed unswept deposits UTXOs
+    ///        and 0..1 main P2(W)PKH UTXO. That transaction must have only
```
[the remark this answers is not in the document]
lukasz-zimnoch: Simplified it. See https://github.com/keep-network/tbtc-v2/pull/106/commits/ec800f4567de3a823cbb753a5b3d0b4b9a2e077f. It should be easier to grasp now.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+    ///        one P2(W)PKH output containing a 20-byte wallet public key hash.
+    ///      - `sweepProof` components must match the expected structure. See
+    ///        `BitcoinTx.Proof` docs for reference. The `bitcoinHeaders`
+    ///        field must contain a valid number of block headers, not less
+    ///        than the `txProofDifficultyFactor` contract constant.
+    ///      - `mainUtxo` components must point to the recent main UTXO
+    ///        of the given wallet. If there is no main UTXO, this parameter
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/ec800f4567de3a823cbb753a5b3d0b4b9a2e077f

solidity/contracts/bridge/Bridge.sol
```diff
@@ -96,19 +130,25 @@ contract Bridge is Ownable {
...
     ///         wallet is responsible for validating them before attempting to
     ///         execute a sweep.
     ///
-    /// TODO: Explore the possibility of storing just a hash of DepositInfo.
-    mapping(uint256 => DepositInfo) public unswept;
+    mapping(uint256 => DepositInfo) public deposits;
+
+    /// @notice Maps the wallet public key hash (computed using HASH160 opcode)
+    ///         to the latest wallet's main UTXO computed as
+    ///         keccak256(txHash | txOutputIndex | txOutputValue).
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/ec800f4567de3a823cbb753a5b3d0b4b9a2e077f

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+
+        return (walletPubKeyHash, value);
+    }
+
+    /// @notice Processes the Bitcoin sweep transaction input vector. It
+    ///         extracts each input and try to obtain associated deposit or
+    ///         main UTXO data, depending on the input type. Reverts
+    ///         if one of the input cannot be recognized as pointer to a
+    ///         revealed unswept deposit or expected main UTXO.
+    ///         This function also marks each processed deposit as swept.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/ec800f4567de3a823cbb753a5b3d0b4b9a2e077f

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+        )
+    {
+        // If passed `mainUtxo` parameter's value are zeroed, the main UTXO
+        // for given wallet doesn't exist and it is not expected to be included
+        // in sweep transaction input vector.
+        bool mainUtxoExpected = mainUtxo.txHash != bytes32(0);
+        bool mainUtxoFound = false;
+
+        // Determining the total number of sweep transaction inputs in the same
+        // way as for number of outputs.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/ec800f4567de3a823cbb753a5b3d0b4b9a2e077f

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+        (uint256 inputsCompactSizeUintLength, uint256 inputsCount) =
+            sweepTxInputVector.parseVarInt();
+
+        // To determine the first input starting index, we must jump over
+        // the compactSize uint which prepends the input vector. One byte must
+        // be added because of how `parseVarInt` returns the length of the
+        // compactSize uint. Refer `BTCUtils` library for more details.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/ec800f4567de3a823cbb753a5b3d0b4b9a2e077f

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+        // it was actually expected.
+        require(
+            mainUtxoExpected == mainUtxoFound,
+            "Expected main UTXO not present in sweep transaction inputs"
+        );
+
+        return (inputsValue, depositors, depositedAmounts);
+    }
+
+    /// @notice Parses a Bitcoin transaction input starting at given index.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/ec800f4567de3a823cbb753a5b3d0b4b9a2e077f

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+    /// @param inputVector Bitcoin transaction input vector. This function
+    ///        assumes vector's structure is valid so it must be validated
+    ///        using e.g. `BTCUtils.validateVin` function before it is passed
+    ///        here.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/ec800f4567de3a823cbb753a5b3d0b4b9a2e077f

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
-        //
-        // TODO Delete deposit from unswept mapping or mark it as swept
-        //      depending on the gas costs. Alternatively, do not allow to
-        //      use the same TX input vector twice. Sweep should be provable
-        //      only one time.
+        // The actual transaction proof is performed here. After that point, we
+        // can assume the transaction happened on Bitcoin chain and has
+        // a sufficient number of confirmations as determined by
+        // `txProofDifficultyFactor` constant.
+        bytes32 sweepTxHash = validateSweepTxProof(sweepTx, sweepProof);
```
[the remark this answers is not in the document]
lukasz-zimnoch: I think it's not that simple. To validate the main UTXO, you need the `walletPubKeyHash`. To get `walletPubKeyHash` you need to `processSweepTxOutput`. To `processSweepTxOutput` you need to have the transaction data validated and proven.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -96,19 +130,25 @@ contract Bridge is Ownable {
...
 
-    /// @notice Collection of all unswept deposits indexed by
+    /// @notice Collection of all revealed deposits indexed by
     ///         keccak256(fundingTxHash | fundingOutputIndex).
-    ///         The fundingTxHash is LE bytes32 and fundingOutputIndex an uint8.
+    ///         The fundingTxHash is LE bytes32 and fundingOutputIndex an uint32.
     ///         This mapping may contain valid and invalid deposits and the
     ///         wallet is responsible for validating them before attempting to
     ///         execute a sweep.
     ///
```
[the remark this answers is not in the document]
lukasz-zimnoch: Done by the way of https://github.com/keep-network/tbtc-v2/pull/106/commits/1102a2fa6deae84e4af7b23000976637ed85e588

solidity/contracts/bridge/Bridge.sol
```diff
@@ -117,7 +157,21 @@ contract Bridge is Ownable {
         address vault
     );
```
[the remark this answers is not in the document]
lukasz-zimnoch: Done in https://github.com/keep-network/tbtc-v2/pull/106/commits/1102a2fa6deae84e4af7b23000976637ed85e588. Also, due to optimizations made by Promethea in Summa's lib, I was able to simplify the value extraction by the way.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+        // That compactSize uint encodes the number of vector elements using the
+        // format presented in:
+        // https://developer.bitcoin.org/reference/transactions.html#compactsize-unsigned-integers
+        // We don't need asserting the compactSize uint is parseable since it
+        // was already checked during `validateVout` validation.
+        (, uint256 outputsCount) = sweepTxOutputVector.parseVarInt();
+        require(
+            outputsCount == 1,
+            "Sweep transaction must have a single output"
+        );
```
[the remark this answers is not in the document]
lukasz-zimnoch: Looks that's covered here: https://github.com/keep-network/bitcoin-spv/blob/b6af83da200522fbbd7a28b4dfe30566f60bc723/testVectors.json#L773. The compactSize claims there are 2 outputs but actually there is only one.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+
+        bytes memory output = sweepTxOutputVector.extractOutputAtIndex(0);
+        value = output.extractValue();
+        bytes memory walletPubKeyHashBytes = output.extractHash();
+        // The sweep transaction output should always be P2PKH or P2WPKH.
+        // In both cases, the wallet public key hash should be 20 bytes length.
+        require(
+            walletPubKeyHashBytes.length == 20,
+            "Wallet public key hash should have 20 bytes"
+        );
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yep, exactly. If this is something other than 20-byte length, that's probably an arbitrary 32-byte script hash that stoles the funds.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+            walletPubKeyHash := mload(add(walletPubKeyHashBytes, 32))
+        }
+
+        return (walletPubKeyHash, value);
+    }
+
+    /// @notice Processes the Bitcoin sweep transaction input vector. It
+    ///         extracts each input and try to obtain associated deposit or
+    ///         main UTXO data, depending on the input type. Reverts
+    ///         if one of the input cannot be recognized as pointer to a
```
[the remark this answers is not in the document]
lukasz-zimnoch: Duplicate of https://github.com/keep-network/tbtc-v2/pull/106#discussion_r798621783

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+    function processSweepTxOutput(bytes memory sweepTxOutputVector)
+        internal
+        pure
+        returns (bytes20 walletPubKeyHash, uint64 value)
+    {
+        // To determine the total number of sweep transaction outputs, we need to
+        // parse the compactSize uint (VarInt) the output vector is prepended by.
+        // That compactSize uint encodes the number of vector elements using the
+        // format presented in:
+        // https://developer.bitcoin.org/reference/transactions.html#compactsize-unsigned-integers
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/f8b97f769d0087309018cea78fa1383775fdd283

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+    ///         main UTXO data, depending on the input type. Reverts
+    ///         if one of the input cannot be recognized as pointer to a
+    ///         revealed unswept deposit or expected main UTXO.
+    ///         This function also marks each processed deposit as swept.
+    /// @param sweepTxInputVector Bitcoin sweep transaction input vector.
+    ///        This function assumes vector's structure is valid so it must be
+    ///        validated using e.g. `BTCUtils.validateVin` function before
+    ///        it is passed here.
+    /// @param mainUtxo Data of the wallet's main UTXO. If no main UTXO
+    ///        exists for given wallet, this parameter's fields should be
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/f8b97f769d0087309018cea78fa1383775fdd283

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+    ///         input vector. Size of this array is either equal to the
+    ///         number of inputs (main UTXO doesn't exist) or less by one
+    ///         (main UTXO exists and is pointed by one of the inputs).
+    function processSweepTxInputs(
+        bytes memory sweepTxInputVector,
+        BitcoinTx.UTXO memory mainUtxo
+    )
+        internal
+        returns (
+            uint256 inputsValue,
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/f8b97f769d0087309018cea78fa1383775fdd283

solidity/contracts/bridge/Bridge.sol
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
+                        keccak256(abi.encodePacked(inputTxHash, inputTxIndex))
+                    )
+                ];
+
+            if (deposit.revealedAt != 0) {
+                require(deposit.sweptAt == 0, "Deposit already swept");
+
+                if (processedDepositsCount == depositors.length) {
+                    // If this condition is true, that means a deposit input
+                    // took place of an expected main UTXO input.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/f8b97f769d0087309018cea78fa1383775fdd283

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: Can I address it separately in a follow-up PR? This will require some additional logic and additional tests cases but I don't want to make this PR bigger.

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: Changed that TODO as follows https://github.com/keep-network/tbtc-v2/pull/106/commits/da281872ff9bd594176c1678de5e1c12019ee0af and added a TODO to check the deposit amount against dust threshold in `depositReveal`.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -306,35 +356,438 @@ contract Bridge is Ownable {
...
+    ///        correspond to appropriate Bitcoin transaction fields to produce
+    ///        a provable transaction hash.
+    ///      - The `sweepTx` should represent a Bitcoin transaction with 1..n
+    ///        inputs. If the wallet has no main UTXO, all inputs should
+    ///        correspond to P2(W)SH revealed deposits UTXOs. If the wallet has
+    ///        an existing main UTXO, one of the inputs must point to that
+    ///        main UTXO and remaining inputs should correspond to P2(W)SH
+    ///        revealed deposits UTXOs. That transaction must have only
+    ///        one P2(W)PKH output locking funds on the 20-byte wallet public
+    ///        key hash.
```
pdyraga, line 376: I love the simplification from https://github.com/keep-network/tbtc-v2/commit/ec800f4567de3a823cbb753a5b3d0b4b9a2e077f but maybe we should refer to 1..n inputs mentioned earlier, especially what is that n we used later? Something like:

```suggestion
    ///      - The `sweepTx` should represent a Bitcoin transaction with 1..n
    ///        inputs. If the wallet has no main UTXO, all n inputs should
    ///        correspond to P2(W)SH revealed deposits UTXOs. If the wallet has
    ///        an existing main UTXO, one of the n inputs must point to that
    ///        main UTXO and remaining n-1 inputs should correspond to P2(W)SH
    ///        revealed deposits UTXOs. That transaction must have only
    ///        one P2(W)PKH output locking funds on the 20-byte wallet public
    ///        key hash.
```

solidity/contracts/bridge/Bridge.sol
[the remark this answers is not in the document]
pdyraga: It sucks 😞

solidity/contracts/bridge/Bridge.sol
```diff
@@ -306,35 +356,438 @@ contract Bridge is Ownable {
...
+    ///         main UTXO data, depending on the input type. Reverts
+    ///         if one of the inputs cannot be recognized as a pointer to a
+    ///         revealed deposit or expected main UTXO.
+    ///         This function also marks each processed deposit as swept.
+    /// @param sweepTxInputVector Bitcoin sweep transaction input vector.
+    ///        This function assumes vector's structure is valid so it must be
+    ///        validated using e.g. `BTCUtils.validateVin` function before
+    ///        it is passed here
+    /// @param mainUtxo Data of the wallet's main UTXO. If no main UTXO
+    ///        exists for given the wallet, this parameter's fields should be
```
pdyraga, line 617:
```suggestion
    ///        exists for the given wallet, this parameter's fields should be
```

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/b86633cecaf6187423136bd96359eeb69d63e206

solidity/contracts/bridge/Bridge.sol
pdyraga: Sure, can we just improve `// TODO: Deal with precision loss.` to be more precise? Something like:

```
// TODO: Deal with precision loss by having the last depositor pay
// the higher fee than others if there is a change, just like it has
// been proposed for the redemption flow. See:
// https://github.com/keep-network/tbtc-v2/pull/128#discussion_r800555359.

Comment (lukasz-zimnoch) [reply solidity/contracts/bridge/Bridge.sol]: See https://github.com/keep-network/tbtc-v2/pull/106/commits/e7d12a1c73fc5ad376b565b76b2017291cb9bbaa

File: solidity/contracts/bridge/Bridge.sol (hunk 42 above)
Comment (lukasz-zimnoch) [reply solidity/contracts/bridge/Bridge.sol]: Addressed within https://github.com/keep-network/tbtc-v2/pull/106/commits/e7d12a1c73fc5ad376b565b76b2017291cb9bbaa

File: solidity/contracts/bridge/Bridge.sol (hunk 35 above)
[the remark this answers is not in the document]
Comment (pdyraga) [reply solidity/contracts/bridge/Bridge.sol:600]: Added [a comment](https://github.com/keep-network/tbtc-v2/issues/120#issuecomment-1032692590) to #120 as a reminder.

File: solidity/contracts/bridge/Bridge.sol [hunk 43]
```diff
@@ -307,34 +361,410 @@ contract Bridge is Ownable {
...
-        //      only one time.
+        // The actual transaction proof is performed here. After that point, we
+        // can assume the transaction happened on Bitcoin chain and has
+        // a sufficient number of confirmations as determined by
+        // `txProofDifficultyFactor` constant.
+        bytes32 sweepTxHash = validateSweepTxProof(sweepTx, sweepProof);
+
+        // Process sweep transaction output and extract its target wallet
+        // public key hash and value.
+        (bytes20 walletPubKeyHash, uint64 sweepTxOutputValue) =
```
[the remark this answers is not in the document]
pdyraga: I have also pointed it out in https://github.com/keep-network/keep-core/pull/2792. I'll leave this conversation open for now so that we can easily go back to it but I do not consider it blocking the merge given we have the TODO left.

solidity/contracts/bridge/Bridge.sol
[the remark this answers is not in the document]
pdyraga: My understanding is that `processSweepTxOutput` is cheap to execute. If so can we do:

```
(bytes20 walletPubKeyHash, uint64 sweepTxOutputValue) = processSweepTxOutput(sweepTx.outputVector);

BitcoinTx.UTXO memory resolvedMainUtxo = BitcoinTx.UTXO(bytes32(0), 0, 0);
bytes32 mainUtxoHash = mainUtxos[walletPubKeyHash];
if (mainUtxoHash != bytes32(0)) {
  // (...)
}

bytes32 sweepTxHash = validateSweepTxProof(sweepTx, sweepProof);
```
?

In other words, we'd execute `validateSweepTxProof` after the `require` from line 410, comparing hashed `mainUtxo` parameter data against the one resolved from `mainUtxos` mapping.

Then, if the maintainer got frontrunned the TX will revert *before* validating the SPV proof, so this will be cheaper.

If this change is not a simple one because tests will start failing, let's just leave a TODO at the beginning of `submitSweepProof` to fail early if the function call gets frontrunned. Another option is that we can provide `walletPubKeyHash` as a parameter, resolve main UTXO from the mapping, execute the require, and validate the provided `walletPubKeyHash` later.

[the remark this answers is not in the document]
pdyraga: We synced with @lukasz-zimnoch out-of-band and `validateSweepTxProof` is doing a static validation of inputs as well. Just changing the order is not going to work. We will address it by just adding a TODO about frontrunning and revisit in the future.

solidity/contracts/bridge/Bridge.sol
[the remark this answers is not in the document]
pdyraga: We synced with @lukasz-zimnoch out-of-band and we *think* that the referenced test vector is probably having more corrupted data than just the compactSize unit at the beginning. This is not blocking this PR but we'll work on improving test vector in bitcoin-spv separately.

solidity/contracts/bridge/Bridge.sol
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/106/commits/19025d19f7301719469ba6f6ef49e3065c9a008c

solidity/contracts/bridge/Bridge.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Looks we have another one https://github.com/keep-network/bitcoin-spv/blob/b6af83da200522fbbd7a28b4dfe30566f60bc723/testVectors.json#L757 where the compactSize claims there is 0 outputs but actually there are two correctly formatted outputs. See the first proper vector on top for comparison.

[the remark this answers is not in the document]
pdyraga: Also, opened an issue to clarify in a comment what is the purpose of that test vector and to add one more for non-zero compactSize uint: https://github.com/keep-network/tbtc-v2/issues/131/

solidity/contracts/bridge/Bridge.sol
```diff
@@ -306,35 +356,445 @@ contract Bridge is Ownable {
...
+                if (processedDepositsCount == depositors.length) {
+                    // If this condition is true, that means a deposit input
+                    // took place of an expected main UTXO input.
+                    // In other words, there is no expected main UTXO
+                    // input and all inputs come from valid, revealed deposits.
+                    // In that case, we should ignore that input and let the
+                    // transaction fail at `require` checking whether expected
+                    // main UTXO was found.
+                    inputStartingIndex += inputLength;
+                    continue;
```
pdyraga, line 719: Actually, why do we let the loop continue instead of reverting here?

lukasz-zimnoch: Good point! There was a difference at the beginning but since we introduced some other checks during development we can revert as you noticed. See: https://github.com/keep-network/tbtc-v2/pull/106/commits/dde3398b5d85298c42e99bd04e5e45041709e91f

pdyraga:
I was looking only on `*.sol` files and there are no more comments on my end. Leaving it in your hands now @tomaszslabon.

PR #107, opened 2022-01-14
Moar Bitcoin details: documentation improvements

pdyraga:
Improved documentation by adding tables explaining raw Bitcoin transaction format, transaction inputs, and outputs. Referenced these types from `BitcoinTx.Info` struct documentation.

I decided to extract `BitcoinTx.Info` to a separate file because the tables - although helpful to understand the format - were polluting `Bridge` contract code too much.

### DIFF
--- a/solidity/contracts/bridge/BitcoinTx.sol
+++ b/solidity/contracts/bridge/BitcoinTx.sol
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: MIT
+
+// ██████████████     ▐████▌     ██████████████
+// ██████████████     ▐████▌     ██████████████
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+// ██████████████     ▐████▌     ██████████████
+// ██████████████     ▐████▌     ██████████████
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+
+pragma solidity 0.8.4;
+
+/// @title Bitcoin transaction
+/// @notice Allows to reference Bitcoin raw transaction in Solidity.
+/// @dev See https://developer.bitcoin.org/reference/transactions.html#raw-transaction-format
+///
+///      Raw Bitcon transaction data:
+///
+///      | Bytes  |     Name     |        BTC type        |        Description        |
+///      |--------|--------------|------------------------|---------------------------|
+///      | 4      | version      | int32_t (LE)           | TX version number         |
+///      | varies | tx_in_count  | compactSize uint (LE)  | Number of TX inputs       |
+///      | varies | tx_in        | txIn[]                 | TX inputs                 |
+///      | varies | tx_out count | compactSize uint (LE)  | Number of TX outputs      |
+///      | varies | tx_out       | txOut[]                | TX outputs                |
+///      | 4      | lock_time    | uint32_t (LE)          | Unix time or block number |
+///
+//
+///      Non-coinbase transaction input (txIn):
+///
+///      | Bytes  |       Name       |        BTC type        |                 Description                 |
+///      |--------|------------------|------------------------|---------------------------------------------|
+///      | 36     | previous_output  | outpoint               | The previous outpoint being spent           |
+///      | varies | script bytes     | compactSize uint (LE)  | The number of bytes in the signature script |
+///      | varies | signature script | char[]                 | The signature script, not present for P2WSH |
+///      | 4      | sequence         | uint32_t (LE)          | Sequence number                             |
+///
+///
+///      The reference to transaction being spent (outpoint):
+///
+///      | Bytes | Name  |   BTC type    |               Description                |
+///      |-------|-------|---------------|------------------------------------------|
+///      |    32 | hash  | char[32]      | Hash of the transaction to spend         |
+///      |    4  | index | uint32_t (LE) | Index of the specific output from the TX |
+///
+///
+///      Transaction output (txOut):
+///
+///      | Bytes  |      Name       |     BTC type          |             Description              |
+///      |--------|-----------------|-----------------------|--------------------------------------|
+///      | 8      | value           | int64_t (LE)          | Number of satoshis to spend          |
+///      | 1+     | pk_script_bytes | compactSize uint (LE) | Number of bytes in the pubkey script |
+///      | varies | pk_script       | char[]                | Pubkey script                        |
+///
+library BitcoinTx {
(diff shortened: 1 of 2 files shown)


solidity/contracts/bridge/Bridge.sol
```diff
@@ -15,49 +15,33 @@
...
-///         confirmed on the Bitcoin network, the wallet informs the Bridge
-///         about the sweep increasing appropriate balances in the Bank.
+///         bridge using pay-to-script-hash (P2SH) or pay-to-witness-script-hash
+///         (P2WSH) containing hashed information about the depositor’s Ethereum
+///         address. Then, the depositor reveals their desired Ethereum address
+///         along with their deposit blinding factor to the Bridge on Ethereum
+///         chain. The off-chain ECDSA wallet listens for these sorts of
+///         messages and when it gets one, it checks the Bitcoin network to make
+///         sure the deposit lines up. If it does, the off-chain ECDSA wallet
+///         may decide to pick the deposit transaction for sweeping, and when
```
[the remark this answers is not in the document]
pdyraga: The validation in the reveal function might not be sufficient. I think we should leave some freedom to the wallet given that performing full validation in the reveal function might be expensive and sometimes not even possible.

This documentation is just for the contract, the system may decide to sweep but the contract can not guarantee it.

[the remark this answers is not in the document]
michalinacienciala: OK, makes sense, thanks for explaining!

solidity/contracts/bridge/Bridge.sol
```diff
@@ -15,49 +15,33 @@
...
 
-/// @title BTC Bridge
-/// @notice Bridge manages BTC deposit and redemption and is increasing and
+/// @title Bitcoin Bridge
+/// @notice Bridge manages BTC deposit and redemption flow and is increasing and
 ///         decreasing balances in the Bank as a result of BTC deposit and
-///         redemption operations.
+///         redemption operations performed by depositors and redeemers.
 ///
 ///         Depositors send BTC funds to the most-recently-created-wallet of the
```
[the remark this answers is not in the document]
pdyraga: Makes sense, done in https://github.com/keep-network/tbtc-v2/pull/107/commits/0ea7a5fd5ec44c2227a972977e725cb55beab5b1

solidity/contracts/bridge/BitcoinTx.sol
```diff
@@ -0,0 +1,93 @@
...
+/// @title Bitcoin transaction
+/// @notice Allows to reference Bitcoin raw transaction in Solidity.
+/// @dev See https://developer.bitcoin.org/reference/transactions.html#raw-transaction-format
+///
+///      Raw Bitcon transaction data:
+///
+///      | Bytes  |     Name     |        BTC type        |        Description        |
+///      |--------|--------------|------------------------|---------------------------|
+///      | 4      | version      | int32_t (LE)           | TX version number         |
+///      | varies | tx_in_count  | compactSize uint (LE)  | Number of TX inputs       |
```
lukasz-zimnoch, line 27: Just a remark regarding all `compactSize uint (LE)` mentions. The entire `compactSize` uint is not always LE (source https://developer.bitcoin.org/reference/transactions.html#compactsize-unsigned-integers). 
- If the `compactSize` uint is only one-byte length, that byte is interpreted as the number
- If the `compactSize` uint is longer than one byte, the first byte is the flag saying how many subsequent bytes represent the number. Only those subsequent bytes are little-endian.

I think it's worth clarifying that.

solidity/contracts/bridge/BitcoinTx.sol
```diff
@@ -0,0 +1,93 @@
...
+///      | 4      | lock_time    | uint32_t (LE)          | Unix time or block number |
+///
+//
+///      Non-coinbase transaction input (txIn):
+///
+///      | Bytes  |       Name       |        BTC type        |                 Description                 |
+///      |--------|------------------|------------------------|---------------------------------------------|
+///      | 36     | previous_output  | outpoint               | The previous outpoint being spent           |
+///      | varies | script bytes     | compactSize uint (LE)  | The number of bytes in the signature script |
+///      | varies | signature script | char[]                 | The signature script, not present for P2WSH |
```
lukasz-zimnoch, line 40: The signature script field is present but empty in P2WSH.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -15,49 +15,33 @@
...
-///         funds line up. If they do, the off-chain wallet may decide to pick
-///         this transaction for sweeping, and when the sweep operation is
-///         confirmed on the Bitcoin network, the wallet informs the Bridge
-///         about the sweep increasing appropriate balances in the Bank.
+///         Depositors send BTC funds to the most recently created off-chain
+///         ECDSA wallet of the bridge using pay-to-script-hash (P2SH) or
+///         pay-to-witness-script-hash (P2WSH) containing hashed information
+///         about the depositor’s Ethereum address. Then, the depositor reveals
+///         their desired Ethereum address along with their deposit blinding
+///         factor to the Bridge on Ethereum chain. The off-chain ECDSA wallet
```
lukasz-zimnoch, line 33: Actually, the depositor needs to reveal also the refund public key hash and the refund locktime. Also, the wallet public key hash used during deposit creation should be revealed but that can be inferred based on most recently created wallet

lukasz-zimnoch (commented):
Looks very good! Just three minor comments.

solidity/contracts/bridge/Bridge.sol
pdyraga: See https://github.com/keep-network/tbtc-v2/pull/107/commits/46cee25ccb48836534f2189049db3b29482d6fd2, please.

solidity/contracts/bridge/BitcoinTx.sol
pdyraga: Please see https://github.com/keep-network/tbtc-v2/pull/107/commits/7dc4cfe3248743322c6daea17db9d04b8dc604fd.

solidity/contracts/bridge/BitcoinTx.sol
pdyraga: I think we can still say it's LE 🤔 For >= 0 && <= 252 it does not matter because we have only one byte so there is no LE/BE. For anything >=253 we have a flag byte and at least two bytes that are LE. See https://github.com/keep-network/tbtc-v2/pull/107/commits/a620f7781d15d2191e1d6265477d24417dd836d8 please.

lukasz-zimnoch: Yeah, I mean mostly the trick with the flag which should not be taken while converting from bytes to uint. The additional explanation you pushed seems to be enough to make it clear.

Issue #34, opened 2021-08-30
Stake undelegations and wallet lifecycle


pdyraga:
In tBTC v1 undelegated stake is locked until all deposits backed by the given operator are redeemed. In tBTC v2, a wallet has its own lifecycle and is no longer mapped 1:1 to deposit. We need to design a mechanism for undelegations respecting the lifecycle of a wallet.

One option would be to do off-chain heartbeats often enough and count undelegating operators as inactive for some time before their stakes release. If the wallet detects that it's losing too many members, it can move its balance to another wallet.

Questions we need to answer (and there is more for sure):
- How the heartbeat works and how often is it executed? Is it retried if failed?
- For successful heartbeats, what is happening?
- For failed heartbeats, what is happening?
- What message is submitted to the chain and who is submitting it? Is the submitter reimbursed? If so, how?
- What is the critical number of members below which we attempt to move the funds?
- What is the undelegation period? Is there any stake locking?
- Where are the funds moved to and how this mechanism looks like?

beaushinkle:
I think this should be almost fully answered by #109 (and what was already there)

> How the heartbeat works and how often is it executed? Is it retried if failed?

Operators perform an off-chain heartbeat (tss message share) every `heartbeat_block_length` blocks (governable, think  ~ 720 blocks / ~3 hours). If not >= `heartbeat` (governable, think ~70) operators participate, the heartbeat passes. If not, we record the inactive operators. If this happens a number of times >= `consecutive_failed_heartbeats` (governable, think ~3) in a row then we take all of the inactive operators from all of the failed heartbeats, put them into a big list, sign it, and have an unlucky operator disable their rewards.

> What does it mean the heartbeat failed? No signature or signature with a critical mass of inactive members?

When heartbeat was supposed to happen (based on `heartbeat_block_length` (governable, think every ~3 hours)), there were < `heartbeat` operators that showed up to the party. The active operators record the ones that didn't show up for potential later punishment (disable their rewards).

> For successful heartbeats, what is happening?

When a heartbeat is supposed to happen, there were >= `heartbeat` operators that showed up. Business continues as usual and nothing actually happens.

> For failed heartbeats, what is happening?

The active operators record the ones that didn't show up for potential later punishment. If a number of heartbeats fail in a row >= `consecutive_failed_heartbeats` (governable, think ~3) then we take all of the inactive operators from all of the failed heartbeats, put them into a big list, sign it, and have an unlucky operator disable their rewards.

[...]

PR #108, opened 2022-01-14
RFC 1: Transaction Incentives

beaushinkle:
This fleshes out some of the transaction incentives that we'll commonly use across the v2 design, and lays some foundation for clarifying the design around heartbeats: #34 

tagging @pdyraga for review

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -465,7 +465,27 @@ BTC from the wallet to another random wallet and close this wallet.
 
 [[transaction-incentives]]
 === Transaction Incentives
-stub!
+We use two main types of transaction incentives: Unlucky Operators and Escalating Rewards.
+
+[[unlucky-operator]]
+==== Unlucky Operators
+The Unlucky Operator scheme involves using the a source of pseudo-randomness
+(like the next ethereum block hash) to choose who the operator to pay for the
+gas is. If that operator refuses, the remaining operators can sign a message to
+disable their rewards, where that message is incentivized via an
+<<escalating-reward,escalating reward>> to prevent recursion.
+
+[[escalating-reward]]
+==== Escalating Rewards
+The Escalating Reward scheme involves setting a governable floor price, ceiling
+price, and velocity. Whoever submits the transaction will be paid out according
+to how far the reward has escalated from the floor to the ceiling.
+
+*Example*: say that the reward started on block 110 at a floor of 100 T
+tokens, has a velocity of 2 T per block, and has a ceiling of 1000 T tokens. Someone
+submits the transaction on block 150. Thus, they would receive `min(2 T*(block
+150 - block 110) + 100 T, 1000 T) = 180 T`
+
 
 [[bitcoin-sweeping-fee]]
 === Bitcoin Sweeping Fee


docs/rfc/rfc-1.adoc
```diff
@@ -465,7 +465,27 @@ BTC from the wallet to another random wallet and close this wallet.
 
 [[transaction-incentives]]
 === Transaction Incentives
-stub!
+We use two main types of transaction incentives: Unlucky Operators and Escalating Rewards.
+
+[[unlucky-operator]]
+==== Unlucky Operators
+The Unlucky Operator scheme involves using the a source of pseudo-randomness
```
michalinacienciala, line 472:
```suggestion
The Unlucky Operator scheme involves using the a source of pseudo-randomness
```

michalinacienciala, line 472:
```suggestion
The Unlucky Operator scheme involves using the source of pseudo-randomness
```

docs/rfc/rfc-1.adoc
```diff
@@ -465,7 +465,27 @@ BTC from the wallet to another random wallet and close this wallet.
...
+(like the next ethereum block hash) to choose who the operator to pay for the
+gas is. If that operator refuses, the remaining operators can sign a message to
+disable their rewards, where that message is incentivized via an
+<<escalating-reward,escalating reward>> to prevent recursion.
+
+[[escalating-reward]]
+==== Escalating Rewards
+The Escalating Reward scheme involves setting a governable floor price, ceiling
+price, and velocity. Whoever submits the transaction will be paid out according
```
michalinacienciala, line 481: Could you add those params to the list in 'Governable Parameters' section?

docs/rfc/rfc-1.adoc
```diff
@@ -465,7 +465,27 @@ BTC from the wallet to another random wallet and close this wallet.
...
 [[transaction-incentives]]
 === Transaction Incentives
-stub!
+We use two main types of transaction incentives: Unlucky Operators and Escalating Rewards.
+
+[[unlucky-operator]]
+==== Unlucky Operators
+The Unlucky Operator scheme involves using the a source of pseudo-randomness
+(like the next ethereum block hash) to choose who the operator to pay for the
+gas is. If that operator refuses, the remaining operators can sign a message to
```
michalinacienciala, line 474: > If that operator refuses
What do we constitute as a refusal? Do we set some time (or no. of blocks) threshold after we treat the operator as refusing to pay? If so, will it be a common governable value, or will it be determined on a case-by-case basis (different value for different usages of the unlucky operator scheme).

pdyraga:
Although both strategies are what we originally considered, I am not sure if we can implement them for tBTC v2. We do not really know when the sweeping started off-chain, so we do not know where to start increasing rewards (if we'd decide for increase reward option) or when to start marking operators as eligible to submit the result (if we decided to enforce the result submission order).

What absolutely needs to be done by operators:
1. Submit DKG result
2. Signal that the funds need to be moved between wallets if the heartbeat failed
3. (Optional) Signal the staker is not performing their duty and they should be marked as ineligible for rewards; see https://github.com/keep-network/keep-core/issues/2725 and https://github.com/keep-network/keep-core/pull/2736.

What literally anyone from the internet should be able to do:

4. Kick-off DKG
5. Provide sweep proof
6. Provide redemption proof
7. Provide proof of moving funds between wallets

I think 1-3 are pretty straightforward: we could have a pool of rewards in T and three governable parameters that specify the reward amount in T. Once the transaction is submitted, the operator receives a reward from that pool. DAO needs to fill that pool from time to time.

4-7 are more problematic because they are executed more often and we could delegate this responsibility to someone having access to the capital instead of requiring ECDSA client operators to have a lot of ETH on their operator accounts.
One option would be to delegate the responsibility to 3rd party maintainers. Anyone from the internet, not even having T but having enough ETH. Maintainers would auction every 48h or so for how much T they will do "all necessary operations" for the next 48 hours. The lowest bid wins. If someone stops doing their duty, then any other maintainer can take their place and win the reward at the end. Bitcoin blocks are not that reliable when it comes to timestamps (https://bitcoin.stackexchange.com/questions/20479/how-accurate-is-bitcoin-network-time, https://arxiv.org/pdf/1803.09028.pdf) but we could have a timeout of e.g. 6 hours. Say: if someone is submitting the sweep proof to Ethereum 6 hours later than the timestamp in Bitcoin block headers says the TX was included, it means the maintainer stopped doing their work.

@mhluongo suggested using https://github.com/gelatodigital

What I think Gelato would solve for us:
- synchronization: clients are not front-running each other with notifications
- devops: we do not need to maintain a last-resort bot

What problems I see:
- It looks that they support right now only the very basic resolver type: smart contract. Resolver tells when to call the function and with what parameters. We need more because we need to observe Ethereum chain and extract parameters from Bitcoin using Electrum. There is a beta Polywrap resolver we could use, but I see it is not supported when adding new tasks.
- We need to pay Gelato in ETH. It means the DAO needs to borrow ETH, sell T for ETH or sell TBTC for ETH once enough TBTC accumulates.

pdyraga (approved):
LGTM

docs/rfc/rfc-1.adoc
```diff
@@ -465,7 +466,36 @@ BTC from the wallet to another random wallet and close this wallet.
...
+  contract will pay out per gas for a refund transaction.
+
+There are 7 main transactions that require incentives:
+
+==== Transactions That Must Be Submitted By Operators
+- Submit distributed key generation results for new wallet creation (weekly).
+- Signal that funds need to be moved between wallets if a <<heartbeat,heartbeat>> fails.
+- Signal that a staker is not performing their duty and should be marked as
+ineligible for rewards (as in the case that they were inactive during
+heartbeats).
```
pdyraga, line 481: We will have to see how expensive it is. Sometimes the cost of marking someone as ineligible could be higher than what the DAO saves on rewards.

Not opposed to the idea, I think it makes sense, we'll just need to be very careful during the implementation.

michalinacienciala (approved):

PR #109, opened 2022-01-14
RFC 1: More Heartbeat Design

beaushinkle:
refs #34 

This PR further expands on the heartbeat design, especially around the interaction between unstaking and heartbeats.

tagging @pdyraga for review

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -475,6 +475,13 @@ unstaking might cause a wallet to fail a heartbeat, but can still be around to
 help move the funds to another wallet, and would not have their rewards turned
 off.
 
+To represent this, clients should monitor the ethereum chain for unstaking
+events to keep track of which operators are active/unstaking. Unstaking
+operators should continue to send heartbeats. Operators ignore heartbeats from
+unstaking operators for the purposes of determining wallet liveness. Operators
+*don't* ignore heartbeats from unstaking operators for determining reward
+eligibility.
+
 Moving the funds costs a bitcoin mining fee, as well as a transaction on the
 ethereum side to prove this happened. To maintain the peg, we need to reduce
 the equivalent amount of fee from the treasury's account balance (set aside for
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -474,6 +474,12 @@ unstaking might cause a wallet to fail a heartbeat, but can still be around to
 help move the funds to another wallet, and would not have their rewards turned
 off.
 
+Moving the funds costs a bitcoin mining fee, as well as a transaction on the
+ethereum side to prove this happened. To maintain the peg, we need to reduce
+the equivalent amount of fee from the treasury's account balance (set aside for
+such fees). This will need to be proven and updated ethereum-side, and
+<<transaction-incentives,properly incentivized>>.
+
 *Example:*: Say that `heartbeat = 70, consecutive_failed_heartbeats = 1`.
 Operator-1 through Operator-72 are all active while Operator-73 through
 Operator-100 are inactive. Currently, there are 72 active operators, so the
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -447,6 +447,8 @@ it.
 
 Governable Parameters:
 
+- `consecutive_failed_heartbeats`: The number of times a heartbeat can fail in
+  a row that triggers a move to close the wallet.
 - `failed_heartbeat_reward_removal_period`: The amount of time an operator is
   removed from reward eligibility after failing a heartbeat.
 - `heartbeat`: The number of group members required for a heartbeat to successful.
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-1.adoc
```diff
@@ -458,11 +460,31 @@ To make sure that older wallets are still accessible for redemption, we need to
...
+for `failed_heartbeat_reward_removal_period` amount of time. The active
+operators move the remaining BTC to another random wallet and close this
+wallet.
+
+For the purposes of heartbeats, an operator that is currently unstaking (they
+started their two-week undelegation period) does not count as a live heartbeat,
+but also does not count as an inactive operator. Thus, operators who are
+unstaking might cause a wallet to fail a heartbeat, but can still be around to
+help move the funds to another wallet, and would not have their rewards turned
+off.
```
pdyraga, line 476: Unstaking operators are earning rewards as long as they do not fail a heartbeat in one of their wallets. Does it not incentivize them to modify clients to participate in heartbeats? 🤔

docs/rfc/rfc-1.adoc
```diff
@@ -458,11 +460,31 @@ To make sure that older wallets are still accessible for redemption, we need to
...
+*Example:*: Say that `heartbeat = 70, consecutive_failed_heartbeats = 1`.
+Operator-1 through Operator-72 are all active while Operator-73 through
+Operator-100 are inactive. Currently, there are 72 active operators, so the
+heartbeat check is passing. Then, Operator-25 through Operator-30 decide to
+unstake. Since unstaking operators do not count as a live heartbeat, there
+would only be 67 heartbeats, and the wallet would begin to transfer its funds.
+
+An operator from Operator-1 through Operator-72 (including 25-30) chooses to
+publish that Operator-73 through Operator-100 should have their rewards
+disabled (after the active operators sign this).
```
pdyraga, line 487: Moving funds between wallets breaks the balance between BTC in the bridge and that represented in the bank because the wallet had to pay a fee to BTC miners. Do we need to address it somehow?

docs/rfc/rfc-1.adoc
```diff
@@ -458,11 +460,31 @@ To make sure that older wallets are still accessible for redemption, we need to
...
-ones that did can create and sign a new transaction that lists the _inactive_
-operators. Once this transaction is posted to the ethereum chain, we can
-iterate through the inactive operators, remove them from reward eligibility
-for `failed_heartbeat_reward_removal_period` amount of time, transfer the remaining
-BTC from the wallet to another random wallet and close this wallet.
+active operators record the inactive operators. If this happens a number of
+times >= `consecutive_failed_heartbeats`, the operators take a union the
+recorded _inactive_ operators, and an operator from among the active operators
+posts a transaction to disable those inactive operators from receiving rewards
+for `failed_heartbeat_reward_removal_period` amount of time. The active
```
pdyraga, line 467: This operator will be expensive, I expect 1M+ here. Marking operators as ineligible for rewards is expensive and we will have to check threshold number of ECDSA signatures on this request. Do we want to reimburse the caller from the pool of fees?

docs/rfc/rfc-1.adoc
beaushinkle: Since the heartbeat happens off-chain, and whether or not you have unstaked is public information (i think), each operator can verify that the message sent from each other operator is truthful off-chain.

For example, say that I unstake, but then modify my client to still send heartbeats. You should be able to check my staking status and see that my heartbeat was a fake heartbeat and discard it.

If those assumptions are true I'll stick that in the RFC

pdyraga: We will have to cache unstaking operators in-memory, observe events on the token staking contract and add operators to the cache every time we see an unstake event. Othwerise, every heartbeat verification would get really expensive for operator running with Alchemy/Infura.

But yeah, I think it's a good direction and what I am saying here is just an implementation detail.

beaushinkle: Are you okay with that added complexity? I'm drawing a blank for other ways to make it work:

* We want the folks who are unstaking to stick around to help transfer funds
* We don't want them to modify their clients
* We want unstaking to create heartbeat pressure (otherwise when they're gone we suddenly can't move funds)

docs/rfc/rfc-1.adoc
beaushinkle: Yeah - I'll point to the `transaction-incentives` section

beaushinkle: addressed in e28babc

docs/rfc/rfc-1.adoc
beaushinkle: addressed in 26f9ad6

docs/rfc/rfc-1.adoc
pdyraga: I think it makes sense. I am wondering if we should expect unstaking operators to participate in the heartbeat but filter them off by other clients when determining if we are still above the threshold. Then, even if the wallet fails the threshold, it does not mark unstaking but active operators as non-eligible for rewards.

This way:
- unstaking operators who are still around to sign redemption requests receive rewards,
- unstaking operators who turned off their nodes are treated just like non-unstaking operators who turned off their nodes and do not receive rewards.

What do you think?

beaushinkle: Love it!

beaushinkle: addressed in 0a0363c

Issue #95, opened 2022-01-07
Specify which vaults are allowed to use `Bank.increaseBalanceAndCall`


pdyraga:
#94 implements a shortcut for depositors to route their deposits revealed to the Bridge to a specific vault. This way, depositors can avoid additional transactions to e.g. mint TBTC. The implementation in #94 lacks the validation of which vaults are allowed to use this shortcut. Without the validation, it is possible to grief sweep proof by routing the deposit to a vault that will exceed Ethereum's block gas limit when the wallet tries to call `Bridge.sweep`. Another option to grief is to route the deposits to non-existing vaults or vaults that fail on the notification. The easiest way could be to let the governance approve vaults in the Bridge contract. This way, only trusted vaults can use the shortcut but everyone else is still welcome to use the bridge - just call the vault in a separate transaction.

PR #110, opened 2022-01-19
Bank.approveBalanceAndCall

pdyraga:
Closes #99 

```
function approveBalanceAndCall(address vault, uint256 amount) external 
```

The function sets `amount` as the allowance of a smart contract `vault` over the caller's balance and calls the vault via `receiveBalanceApproval`. TBTC Vault implements this function by minting TBTC.

This way, the Bank balance owner can mint TBTC in one transaction instead of having to execute two separate transactions: `Bank.approve` and `TBTC.mint`.

I have also renamed `IVault.onBalanceIncreased` to `receiveBalanceIncrease`. Now the naming is aligned with `IVault.receiveBalanceApproval`.

PR #111, opened 2022-01-20
Deposits can be routed only to trusted vaults

pdyraga:
Closes: #95 

Depositors can route their revealed deposits only to trusted vaults and have
trusted vaults notified about new deposits as soon as these deposits get swept.
Vaults not trusted by the Bridge can still be used by Bank balance owners on
their own responsibility - anyone can approve their Bank balance to any
address.

Governance can mark the given vault address as trusted or no longer trusted.
Vaults are not trusted by default.

Trusted vault must meet the following criteria:
- `IVault.onBalanceIncreased` must have a known, low gas cost.
- `IVault.onBalanceIncreased` must never revert.

Without restricting reveal only to trusted vaults, malicious vaults not meeting
the criteria would be able to nuke sweep proof transactions executed by ECDSA
wallet with deposits routed to them.

Additionally, `DepositRevealed` even holds now vault address.

Knowing the vault address, off-chain wallet can group deposits
in sweep batches by vault to minimize the cost of the sweep proof
transaction.

All changes in this PR increased the gas cost of reveal function by approximately
3500 gas units.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -100,7 +100,8 @@ contract Bridge {
         bytes8 blindingFactor,
         bytes20 walletPubKeyHash,
         bytes20 refundPubKeyHash,
-        bytes4 refundLocktime
+        bytes4 refundLocktime,
+        address vault
     );
 
     /// @notice Used by the depositor to reveal information about their P2(W)SH
@@ -248,7 +249,8 @@ contract Bridge {
             reveal.blindingFactor,
             reveal.walletPubKeyHash,
             reveal.refundPubKeyHash,
-            reveal.refundLocktime
+            reveal.refundLocktime,
+            reveal.vault
         );
     }
 
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -117,6 +117,8 @@ contract Bridge is Ownable {
         address vault
     );
 
+    event VaultStatusUpdated(address indexed vault, bool isTrusted);
+
     /// @notice Allows the Governance to mark the given vault address as trusted
     ///         or no longer trusted. Vaults are not trusted by default.
     ///         Trusted vault must meet the following criteria:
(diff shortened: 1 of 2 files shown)


solidity/contracts/bridge/Bridge.sol
```diff
@@ -100,9 +113,26 @@ contract Bridge {
...
-        bytes4 refundLocktime
+        bytes4 refundLocktime,
+        address vault
     );
 
+    /// @notice Allows the Governance to mark the given vault address as trusted
+    ///         or no longer trusted. Vaults are not trusted by default.
+    ///         Trusted vault must meet the following criteria:
+    ///         - `IVault.onBalanceIncreased` must have a known, low gas cost.
+    ///         - `IVault.onBalanceIncreased` must never revert.
```
tomaszslabon, line 124: `onBalanceIncreased` will be renamed to `receiveBalanceIncrease` in #110. Let's not forget to change this description?

solidity/contracts/bridge/Bridge.sol
```diff
@@ -100,9 +113,26 @@ contract Bridge {
...
+    /// @dev Without restricting reveal only to trusted vaults, malicious
+    ///      vaults not meeting the criteria would be able to nuke sweep proof
+    ///      transactions executed by ECDSA wallet with  deposits routed to
+    ///      them.
+    /// @param vault The address of the vault
+    /// @param isTrusted flag indicating whether the vault is trusted or not
+    /// @dev Can only be called by the Governance.
+    function setVaultStatus(address vault, bool isTrusted) external onlyOwner {
+        isVaultTrusted[vault] = isTrusted;
```
tomaszslabon, line 133: Maybe emit an event?

pdyraga: Good point. Added in https://github.com/keep-network/tbtc-v2/pull/111/commits/f2bff21e9109b3f8c77838548c3092ee1b9e0880.

lukasz-zimnoch (approved):

solidity/contracts/bridge/Bridge.sol
pdyraga: Done in https://github.com/keep-network/tbtc-v2/pull/110/commits/32892c8ebd63dc74b9c54c48da7894c0826f184c.

PR #112, opened 2022-01-26
Use SHA256 when dealing with P2WSH script

lukasz-zimnoch:
Refs: #79 

According to [BIP141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#p2wsh) all operations around the `P2WSH` script should use `SHA256`. In our context, that means we should use this hashing when building deposit transactions and during deposit reveal, specifically while performing expected script hash validation. Initially, we were using `HASH256` (double `SHA256`) by mistake and were producing `P2WSH` transactions which could not be unlocked by the sweeping wallet. This changeset fixes this mistake.

### DIFF
--- a/tbtc-ts/test/deposit.test.ts
+++ b/tbtc-ts/test/deposit.test.ts
@@ -29,32 +29,65 @@ describe("Deposit", () => {
     createdAt: 1640181600, // 22-12-2021 14:00:00 UTC
   }
 
-  // HEX of the expected P2WSH deposit transaction made using the code from
-  // deposit.ts and given depositData. It can be decoded with:
-  // https://live.blockcypher.com/btc-testnet/decodetx.
-  const expectedP2WSHDepositTransaction: RawTransaction = {
-    transactionHex:
-      "010000000001018348cdeb551134fe1f19d378a8adec9b146671cb67b945b71bf56" +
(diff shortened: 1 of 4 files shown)


tbtc-ts/test/deposit.test.ts
```diff
@@ -276,7 +302,10 @@ describe("Deposit", () => {
     })
 
     it("should return script with proper structure", async () => {
-      expect(script.length).to.be.equal(184) // 92 bytes
+      // Returned script should ne the same as expectedDepositScript but
```
tomaszslabon, line 305: `ne` -> `be`

lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/112/commits/a82c03ab0e972bfee187f15d51a883bb48f127ba

tbtc-ts/test/deposit.test.ts
```diff
@@ -276,7 +302,10 @@ describe("Deposit", () => {
...
+      // here we make a breakdown an assert specific parts are as expected.
```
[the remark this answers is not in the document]
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/112/commits/a82c03ab0e972bfee187f15d51a883bb48f127ba

tomaszslabon (approved):

PR #113, opened 2022-01-26
RFC 1: Max BTC Fee Revamped

beaushinkle:
This PR overhauls the design for how we determine and communicate the bitcoin mining fee.

Major changes:

* We use a 3-block target instead of 25-block. 
* We do it totally off-chain using operators in sortition-order with no punishments/incentives

I think I wrote 25-block originally because I had 25 from the automated liquidation recovery work, but 250 minutes for btc confirmation is *bad*. 3 might still be too long.

tagging @pdyraga for review

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -511,7 +511,7 @@ blockchain, the miners need to be paid a fee for their work. This fee
 fluctuates with market demand and is decently volatile.
 
 The operators need to all agree on a fee before they can construct the sweeping
-transacion, but communicating that fee is tricky. When the sortition pool
+transaction, but communicating that fee is tricky. When the sortition pool
 selects the operators, it selects them in an ordered list. We can leverage this
 on-chain order to be the fee-proposal-order. Say that the sortition pool chose
 Operators: [#71, #109, #34..., #2]. When it comes time to sweep deposits,
@@ -528,8 +528,8 @@ is, they can 51-of-100 agree to skip to the next operator.
 
 *Note*: We purposefully leave out on-chain incentives/punishments here. The
 attack vector is small and the overhead is high. We might need to revisit this
-in the future, but if someone maintianed a client fork in order to take
-advanage of manipulating btc fee consensus I would be very surprised.
+in the future, but if someone maintained a client fork in order to take
+advantage of manipulating btc fee consensus I would be very surprised.
 
 == Governable Parameters
 Alphabetized list of Governable Parameters with additional notes.
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -500,55 +500,43 @@ in order to prevent problems like operators front-running each other.
 
 [[bitcoin-sweeping-fee]]
 === Bitcoin Sweeping Fee
-
 Governable Parameters
-- `bridge_deposit_percentage_fee`: The percentage fee for depositing bitcoin.
-- `bridge_deposit_flat_fee`: The flat fee for depositing bitcoin.
+
+- `btc_fee_broadcast_timeout`: The amount of an operator has to provide a suggested BTC fee
+  before the other operators give up and try the next operator.
 
 Any time a bitcoin transaction needs to be posted and then mined on the bitcoin
 blockchain, the miners need to be paid a fee for their work. This fee
 fluctuates with market demand and is decently volatile.
 
-When the depositor posts their address reveal on ethereum, they include a
-bitcoin `max_fee` denominated in total BTC that they're willing to pay for
-three different fees:
+The operators need to all agree on a fee before they can construct the sweeping
+transacion, but communicating that fee is tricky. When the sortition pool
+selects the operators, it selects them in an ordered list. We can leverage this
+on-chain order to be the fee-proposal-order. Say that the sortition pool chose
+Operators: [#71, #109, #34..., #2]. When it comes time to sweep deposits,
+Operator#71 would be expected to query
+https://blockstream.info/api/fee-estimates for the 3-block fee and broadcast
+this fee to the rest of the operators. If the operators don't receive the fee
+from operator #71 before `btc_fee_broadcast_timeout`, has elapsed, the duty
+moves to #109, and then #34, and so on.
 
-- `bridge_deposit_percentage_fee` (say 0.2% or whatever governance decides)
-- `bridge_deposit_flat_fee` (say 0.001 mBTC or whatever governance decides) to
-  deter folks from depositing dust
-- the mining fee for getting the batch sweeping transaction to post
+The operators then validate the fee they received against
+https://blockstream.info/api/fee-estimates to make sure that nothing fishy is
+going on - that the fee isn't too high or too low (TBD what that means). If it
+is, they can 51-of-100 agree to skip to the next operator.
 
-Next, we hash the most recent ethereum block to generate the seed for a random
-number generator to generate a modulus that selects a random unlucky operator
-that needs to check the blockstream api for the 25-block target fee:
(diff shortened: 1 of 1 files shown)


michalinacienciala:
If the fee proposed by an operator (and agreed upon by other members of the group) exceeds `base_sweeping_fee_max`, then we don't perform a sweep, right? When will we try again? During next sweeping period?

docs/rfc/rfc-1.adoc
```diff
@@ -500,55 +500,43 @@ in order to prevent problems like operators front-running each other.
...
+advanage of manipulating btc fee consensus I would be very surprised.
 
 == Governable Parameters
 Alphabetized list of Governable Parameters with additional notes.
 
 - `base_sweeping_fee_max`: The highest amount of BTC that operators can
   initially propose as a fee for miners in a sweeping transaction.
-- `bridge_deposit_flat_fee`: The flat fee for depositing bitcoin.
-- `bridge_deposit_percentage_fee`: The percentage fee for depositing bitcoin.
+- `btc_fee_broadcast_timeout`: The amount of an operator has to provide a suggested BTC fee
```
michalinacienciala, line 538:
```suggestion
- `btc_fee_broadcast_timeout`: The amount of time an operator has to provide a suggested BTC fee
```

docs/rfc/rfc-1.adoc
```diff
@@ -500,55 +500,43 @@ in order to prevent problems like operators front-running each other.
...
-deposits are chosen, check out <<sweeping,sweeping>>
-
-After a particular fee has been chosen and all of the deposits have been selected,
-that deposit might not mine within the timeout period. If this happens, the
-unlucky operator should submit a transaction on-chain to multiply the fee by a
-fixed amount (like 1.1x, 1.2x, etc up to some cap like 1.4x). With this in mind,
-we should only include deposits in the initial transaction whose `max_fee` is set
-high enough to withstand a fee increase up to whatever the maximum multiplier is.
+The operators need to all agree on a fee before they can construct the sweeping
+transacion, but communicating that fee is tricky. When the sortition pool
```
michalinacienciala, line 513:
```suggestion
transaction, but communicating that fee is tricky. When the sortition pool
```

docs/rfc/rfc-1.adoc
```diff
@@ -500,55 +500,43 @@ in order to prevent problems like operators front-running each other.
...
+
+The operators then validate the fee they received against
+https://blockstream.info/api/fee-estimates to make sure that nothing fishy is
+going on - that the fee isn't too high or too low (TBD what that means). If it
+is, they can 51-of-100 agree to skip to the next operator.
+
+*Note*: We purposefully leave out on-chain incentives/punishments here. The
+attack vector is small and the overhead is high. We might need to revisit this
+in the future, but if someone maintianed a client fork in order to take
+advanage of manipulating btc fee consensus I would be very surprised.
```
michalinacienciala, line 531:
```suggestion
in the future, but if someone maintained a client fork in order to take
advantage of manipulating btc fee consensus I would be very surprised.
```

michalinacienciala (commented):
Some small typos noticed

docs/rfc/rfc-1.adoc
pdyraga, line 531: All clients validate the fee independently. We use the mechanism of one client proposing the fee so that we do not have to "medianize" multiple fee propositions. 🤷🏻

docs/rfc/rfc-1.adoc
```diff
@@ -500,55 +500,43 @@ in order to prevent problems like operators front-running each other.
...
+Operator#71 would be expected to query
+https://blockstream.info/api/fee-estimates for the 3-block fee and broadcast
+this fee to the rest of the operators. If the operators don't receive the fee
+from operator #71 before `btc_fee_broadcast_timeout`, has elapsed, the duty
+moves to #109, and then #34, and so on.
+
+The operators then validate the fee they received against
+https://blockstream.info/api/fee-estimates to make sure that nothing fishy is
+going on - that the fee isn't too high or too low (TBD what that means). If it
+is, they can 51-of-100 agree to skip to the next operator.
```
pdyraga, line 526: What does it mean they 51-of-100 agree? I thought that skipping the operator after a certain time passes is just a part of the algorithm they are following and they do not need to additionally agree (communicate) to skip the operator. 🤔

docs/rfc/rfc-1.adoc
```diff
@@ -500,55 +500,43 @@ in order to prevent problems like operators front-running each other.
...
+The operators need to all agree on a fee before they can construct the sweeping
+transacion, but communicating that fee is tricky. When the sortition pool
+selects the operators, it selects them in an ordered list. We can leverage this
+on-chain order to be the fee-proposal-order. Say that the sortition pool chose
+Operators: [#71, #109, #34..., #2]. When it comes time to sweep deposits,
+Operator#71 would be expected to query
+https://blockstream.info/api/fee-estimates for the 3-block fee and broadcast
+this fee to the rest of the operators. If the operators don't receive the fee
+from operator #71 before `btc_fee_broadcast_timeout`, has elapsed, the duty
+moves to #109, and then #34, and so on.
```
pdyraga, line 521: I think we should specify what is happening with `#71` and `#109`. Can they still propose a fee?

I would say no because then we could have a problem with 3 operators proposing different fees and which one to choose. 🤔

pdyraga (commented):
Just an idea: We could also use this mechanism to signal the wallet that it's time to perform the sweep. This could theoretically help with upgrades. For example, we could upgrade the client to wait for some additional condition before the sweep happens although it would be fine to perform the sweep right now. Because the majority of the network upgraded (and hopefully the first nth clients on the list), they wait with kicking off the sweep procedure for additional few blocks.

This could also help with time synchronization problems.

docs/rfc/rfc-1.adoc
beaushinkle: Okay - talked a little bit about this offline, producing extra signatures for off-chain stuff runs into nonce-reuse problems: https://billatnapier.medium.com/ecdsa-weakness-where-nonces-are-reused-2be63856a01a . The pocket veto strategy (simply wait to skip a bad fee) works with less complication and less message exchange

docs/rfc/rfc-1.adoc
beaushinkle: I was under the assumption that it looks something like "okay it's block 25. operator #71, that means you have to give us a fee. you have until block 45 to do so, and then it'll be operator 109's turn"

docs/rfc/rfc-1.adoc
beaushinkle: addressed in 6a3ff94

docs/rfc/rfc-1.adoc
beaushinkle: addressed in 6a3ff94

docs/rfc/rfc-1.adoc
beaushinkle: addressed in 95285dd

docs/rfc/rfc-1.adoc
beaushinkle: Okay - revised in 6a65271 

can you check that out and see if you like the phrasing there?

beaushinkle:
okay - good for another round of review

beaushinkle:
anything else ya'll want to see here?

michalinacienciala:
> anything else ya'll want to see here?

There is still the matter of what doeas it mean that a proposed fee is too small or too high. But I'm fine with merging what we have and clarifying that missing part at later stage.

michalinacienciala (approved):

pdyraga (approved):

PR #115, opened 2022-01-28
RFC 6: Transaction incentives

pdyraga:
Note: we may decide to integrate it with RFC 1 or not. Making it a separate RFC for now.

# Goal

The goal of this proposal is to reduce on-chain cost and complexity related to transaction submitter responsibilities, reduce the cost of operating a client, and provide tools to punish free-riders not operating their nodes properly.

# Overview

The scheme assumes that the majority of the group is honest and is interested in helping DAO to eliminate free-riders from the network while getting nothing in return. The group will vote to mark someone as ineligible for rewards if they consistently do not participate in off-chain heartbeats or if they skip their responsibilities for submitting on-chain transactions. 

A real-world analogy would be team members discussing one of them is not contributing to the work at all so the entire team must do the work of that free-rider and have the team report this fact to the manager.

All the responsibilities that do not have to be done by a group will be delegated to opt-in maintainers to reduce the cost of operating a node.

# Unknowns

Should the DAO/council approve the wallet's decisions to mark some members as ineligible for rewards? Are fine with the majority? Or maybe there should be an option for the council to withdraw the wallet's decision?

### DIFF
--- a/docs/rfc/rfc-6.adoc
+++ b/docs/rfc/rfc-6.adoc
@@ -120,6 +120,13 @@ inactivity, the group must decide if the cost of the transaction is worth it.
 That is, if some operator was marked as inactive a minute ago, it does not make
 sense to mark it as inactive and ineligible for rewards one more time.
 
+IMPORTANT: There should be no token incentives for calling this function by
+the group. The only incentive should be their goodwill to save some rewards by
+DAO. Moreover, in the case of off-chain heartbeat failure, clients are trying to
+avoid slashing in the future. Marking operators as ineligible for rewards should
+not increase the rewards of other group members! Rewards saved on inactive
+operators marked as ineligible should return back to the DAO. 
+
 ==== Maintainer incentives
 
 For the first release, the role of maintainers will belong to the willing
--- a/docs/rfc/rfc-6.adoc
+++ b/docs/rfc/rfc-6.adoc
@@ -34,17 +34,18 @@ On-chain transactions can be divided into three types:
 
 - Transactions that can only be submitted by group members, given that only the
   group members have the knowledge about input parameters. For example,
-  submitting DKG result transaction for ECDSA wallet or submitting relay entry
-  transaction for the random beacon.
+  <<submit-dkg-result,submitting DKG result transaction>> for ECDSA wallet or
+  submitting relay entry transaction for the random beacon.
   
 - Transactions that can be submitted by anyone, usually submitted often and
   mostly belonging to a happy path. This type of transaction should never lead
-  to slashing. For example, submit sweep proof or submit redemption proof
-  transaction for tBTC v2. 
+  to slashing. For example, <<maintainer-incentives,submit sweep proof or submit
+  redemption proof transaction>> for tBTC v2. 
 
 - Transactions that can be submitted by anyone, submitted rarely or never, and
-  usually leading to slashing. For example, challenge DKG result for ECDSA
-  wallet or report fraud proof transaction for tBTC v2.
+  usually leading to slashing. For example, <<challenge-dkg-result, challenge
+  DKG result>> for ECDSA wallet or <<report-fraud, report fraud>> transaction for
+  tBTC v2.
 
 For the first group of transactions, we will not enforce the order on-chain and
 we will expect off-chain clients to respect the submission order. If some client
@@ -74,6 +75,7 @@ and free from misbehavior than to optimize costs.
 
 ==== Group member incentives
 
+[[submit-dkg-result]]
 ===== Submit + approve DKG result
 
 No submission order will be enforced on-chain and the transaction submitter will
@@ -131,6 +133,7 @@ operators marked as ineligible should return back to the DAO. The DAO should do
 something else with those extra tokens other than increase rewards. We do not
 want to incentivize voting for your own salary.
 
+[[maintainer-incentives]]
 ==== Maintainer incentives
 
 For the first release, the role of maintainers will belong to the willing
@@ -187,6 +190,7 @@ are not concerned about front-running. Front-running may happen and it is even
 desirable to some extent because these transactions protect the health of the
 network.
 
+[[challenge-dkg-result]]
 ===== Challenge DKG result
 
 No ETH cost reimbursed. Called exceptionally, ideally never. Incentivized by T
@@ -198,6 +202,7 @@ from slashed stakes.
 No ETH cost reimbursed. Called exceptionally, ideally never. Incentivized by T
 from slashed stakes.
 
+[[report-fraud]]
 ===== Report fraud
 
 No ETH cost reimbursed. Called exceptionally, ideally never. Incentivized by T
--- a/docs/rfc/rfc-6.adoc
+++ b/docs/rfc/rfc-6.adoc
@@ -1,0 +1,212 @@
+:toc: macro
+
+= RFC 6: Transaction incentives
+
+:icons: font
+:numbered:
+toc::[]
+
+== Goal
+The goal of this proposal is to reduce on-chain cost and complexity related to
+transaction submitter responsibilities, reduce the cost of operating a client,
+and provide tools to punish free-riders not operating their nodes properly.
+
+== Overview
+The scheme assumes that the majority of the group is honest and is interested in
+helping DAO to eliminate free-riders from the network while getting nothing in
+return. The group will vote to mark someone as ineligible for rewards if they
+consistently do not participate in off-chain heartbeats or if they skip their
+responsibilities for submitting on-chain transactions. 
+
+A real-world analogy would be team members discussing one of them is not
+contributing to the work at all so the entire team must do the work of that
+free-rider and have the team report this fact to the manager.
+
+All the responsibilities that do not have to be done by a group will be
+delegated to opt-in maintainers and third parties to reduce the cost of
+operating a node.
+
+== In depth
+
+=== Transaction types
+
+On-chain transactions can be divided into three types:
+
+- Transactions that can only be submitted by group members, given that only the
+  group members have the knowledge about input parameters. For example, submit
+  DKG result transaction for ECDSA wallet or submit relay entry transaction for
+  the random beacon.
+  
+- Transactions that can be submitted by anyone, usually submitted often and
+  mostly belonging to a happy path. Never leading to slashing For example,
+  submit sweep proof or submit redemption proof transaction for tBTC v2. 
+
+- Transactions that can be submitted by anyone, submitted rarely or never, and
+  usually leading to slashing. For example, challenge DKG result for ECDSA
+  wallet or report fraud proof transaction for tBTC v2.
+
+For the first group of transactions, we will not enforce the order on-chain and
+we will expect off-chain clients to respect the submission order. If some client
+skipped their round, other clients can vote to mark the inactive client as
+ineligible for rewards. To disincentivize frontrunning, we will only reimburse
+ETH costs, and no additional reward will be allocated.
+
+The responsibility of executing the second group of transactions should be
+delegated to willing maintainers. Given that the preparation of the input data
+in some cases is complicated and may require access to the Bitcoin chain, the
+maintainer functionality should be built-in into the client and activated in the
+configuration. Eventually, we may decide to use networks like Gelato but given
+that the ECSDSA wallet could be slashed on the redemption proof timeout, such
+functionality should also be implemented in the client so that the client can
+protect the stake if required. Last but not least, Gelato Polywrap resolver
+allowing assembling more complicated transaction inputs is not yet available and
+we do not want to block the release on this functionality. Maintainers will be
+reimbursed in ETH and may receive an additional reward for performing their
+duties.
+
+Transactions belonging to the third type can be executed by anyone and all
+off-chain clients will be responsible for submitting them. No ETH will be
+reimbursed and T coming from slashed stakes will be the only optional reward.
+No submission order will be maintained and all clients will try to report
+misbehavior at the same time. It is more important to keep the network healthy
+and free from misbehavior than to optimize costs.
+
+==== Group member incentives
+
+===== Submit + approve DKG result
+
+No submission order will be enforced on-chain and the transaction submitter will
+have ETH cost reimbursed. To disincentivize front-running, no additional reward
+will be released by this transaction. Off-chain clients should have a handshake
+agreement to follow the submission order based on
+`hash(new_group_pubkey) % group_size`.
+
+The first member responsible for submitting the DKG result is a member with
+index `hash(new_group_pubkey) % group_size`. If that member did not submit the
+result,  `(hash(new_group_pubkey) % group_size) + 1` becomes responsible next.
+Group members who were inactive and skipped their responsibility might be
+reported by the wallet as ones who failed the heartbeat and marked as ineligible
+for rewards.
+
+For example, if `hash(new_group_pubkey) % group_size = 62`, `group_size = 64`,
+and member `9` submitted DKG result, members with indexes
+`{62, 63, 64, 1, 2, 3, 4, 5, 6, 7, 8}` might be marked as ineligible for
+rewards. Inactive group members are determined once the result is submitted and
+accepted by the chain.
+
+Given that the DKG result submission actually consists of two transactions:
+submit the result and approve the result after the challenge period, the DKG
+result submitter to be considered as active must submit both transactions. That
+is, if the operator submitted the first transaction but did not submit the
+second one, they are considered inactive.
+
+===== Report heartbeat failed
+
+A wallet should periodically execute off-chain heartbeats and report inactive
+operators responsible for potential heartbeat failures. The majority of wallet
+members need to come together and sign a message with their private ECDSA
+operator keys. The message should contain the list of inactive operators and
+nonce ensuring uniqueness. The same mechanism can be used for reporting
+operators who failed their duty of submitting DKG result. 
+
+No submission order will be enforced on-chain for the heartbeat failure
+transaction and the transaction submitter will have ETH cost reimbursed. To
+disincentivize front-running, no additional reward will be released by this
+transaction. Off-chain clients should have a handshake agreement to follow the
+submission order: the operator submitting this TX is the first active one.
+
+When reporting heartbeat failure as a result of DKG result submission
+inactivity, the group must decide if the cost of the transaction is worth it.
+That is, if some operator was marked as inactive a minute ago, it does not make
+sense to mark it as inactive and ineligible for rewards one more time.
+
+==== Maintainer incentives
+
+For the first release, the role of maintainers will belong to the willing
+off-chain clients approved by DAO or council. Every client will have maintainer
+functionality built-in and enabled in the configuration file.
+
+We will implement a simple reimbursement contract for maintainers proxying calls
+to Bridge functions. Only approved maintainers should be able to use the proxy
+contract. Maintainers are expected to follow round-robin distribution of work
+but the order is not going to be enforced on-chain. DAO/council will have the
+functionality of removing maintainers not performing their duties.
+DAO/council may decide to distribute additional T rewards for maintainers. This
+functionality should be supported by the Maintainer contract.
+
+Eventually, the role of maintainers can be delegated to another network such as
+Gelato.
+
+
+```
+┌────────────────────────────────────────┐   calls  ┌─────────────────────────┐
+|               Maintainer               |─────────▶|          Bridge         |
+└────────────────────────────────────────┘          └─────────────────────────┘  
+| submitSweeProof() onlyMaintainer       |          | submitSweepProof()      | 
+| submitRedemptionProof() onlyMaintainer |          | submitRedemptionProof() | 
+| startDkg() onlyMaintainer              |          | startDkg()              | 
+└────────────────────────────────────────┘          └─────────────────────────┘
+```
+
+===== Submit sweep proof
+
+ETH reimbursed, no additional reward to disincentivize front-running between
+maintainers.
+
+===== Submit redemption proof
+
+ETH reimbursed, no additional reward to disincentivize front-running between
+maintainers.
+
+===== Start DKG
+
+ETH reimbursed, no additional reward to disincentivize front-running between
+maintainers. 
+
+===== Report DKG timeout
+
+ETH reimbursed, no additional reward to disincentivize front-running between
+maintainers.
+
+==== Third party incentives
+
+
+All transactions in this section must be supported by the off-chain client. We
+are not concerned about front-running. Front-running may happen and it is even
+desirable to some extent because these transactions protect the health of the
+network.
+
+===== Challenge DKG result
+
+No ETH cost reimbursed. Called exceptionally, ideally never. Incentivized by T
+from slashed stakes.
+
+===== Notify redemption timeout
+
+
+No ETH cost reimbursed. Called exceptionally, ideally never. Incentivized by T
+from slashed stakes.
+
+===== Report fraud
+
+No ETH cost reimbursed. Called exceptionally, ideally never. Incentivized by T
+from slashed stakes.
+
+=== ETH pool
+
+DAO needs to fund ETH pool that will be used for reimbursements. The pool should
+probably be a separate contract. The pool needs to protect against malicious
+miner-operators by placing a governable gas price ceiling. It should be possible
+to withdraw unspent ETH in case we decide to replace Maintainer incentives with
+something else (for example Gelato) or decide to add more functions there
+(for example, the SPV relay updates). 
+
+== Open questions
+
+- Should the DAO/council approve the wallet's decisions to mark some members as
+  ineligible for rewards? Are fine with the majority? Or maybe there should be
+  an option for the council to withdraw the wallet's decision?
+
+- Should the calls to DKG and heartbeat functions be proxied by a separate
+  contract with reimbursement logic, similar to `Maintainer` contract?
+
--- a/docs/rfc/rfc-6.adoc
+++ b/docs/rfc/rfc-6.adoc
@@ -32,22 +32,27 @@ operating a node.
 
 On-chain transactions can be divided into three types:
 
-- Transactions that can only be submitted by group members, given that only the
-  group members have the knowledge about input parameters. For example,
-  <<submit-dkg-result,submitting DKG result transaction>> for ECDSA wallet or
-  submitting relay entry transaction for the random beacon.
-  
-- Transactions that can be submitted by anyone, usually submitted often and
-  mostly belonging to a happy path. This type of transaction should never lead
-  to slashing. For example, <<maintainer-incentives,submit sweep proof or submit
-  redemption proof transaction>> for tBTC v2. 
+==== Group member transactions
 
-- Transactions that can be submitted by anyone, submitted rarely or never, and
-  usually leading to slashing. For example, <<challenge-dkg-result, challenge
-  DKG result>> for ECDSA wallet or <<report-fraud, report fraud>> transaction for
-  tBTC v2.
+Transactions that can only be submitted by group members, given that only the
+group members have the knowledge about input parameters. For example,
+<<submit-dkg-result,submitting DKG result transaction>> for ECDSA wallet or
+submitting relay entry transaction for the random beacon.
 
-[[submission-order]]
+==== Maintainer transactions
+
+Transactions that can be submitted by anyone, usually submitted often and
+mostly belonging to a happy path. This type of transaction should never lead
+to slashing. For example, <<maintainer-incentives,submit sweep proof or submit
+redemption proof transaction>> for tBTC v2. 
+
+==== Misbehavior notifier transactions
+
+Transactions that can be submitted by anyone, submitted rarely or never, and
+usually leading to slashing. For example, <<challenge-dkg-result, challenge
+DKG result>> for ECDSA wallet or <<report-fraud, report fraud>> transaction for
+tBTC v2.
+
 ==== Submission Order
 
 For the first group of transactions, we will not enforce the order on-chain and
@@ -66,8 +71,9 @@ functionality should also be implemented in the client so that the client can
 protect the stake if required. Last but not least, Gelato Polywrap resolver
 allowing assembling more complicated transaction inputs is not yet available and
 we do not want to block the release on this functionality. Maintainers will be
-reimbursed in ETH and may receive an additional reward for performing their
-duties.
+reimbursed in ETH. Maintainers may receive an additional reward for performing
+their work but this needs to happen outside of the transactions they are
+executing.
 
 Transactions belonging to the third type can be executed by anyone and all
 off-chain clients will be responsible for submitting them. No ETH will be
@@ -76,25 +82,25 @@ No submission order will be maintained and all clients will try to report
 misbehavior at the same time. It is more important to keep the network healthy
 and free from misbehavior than to optimize costs.
 
-==== Group member incentives
+==== Group member transaction incentives
 
 [[submit-dkg-result]]
 ===== Submit + approve DKG result
 
 No submission order will be enforced on-chain and the transaction submitter will
 have ETH cost reimbursed. To disincentivize front-running, no additional reward
-will be released by this transaction. Off-chain clients should have a
-gentleman's agreement to follow the submission order based on
+will be released by this transaction. Off-chain clients should have an
+informal agreement to follow the submission order based on
 `hash(new_group_pubkey) % group_size`.
-The submission order will be unenforceable but everyone will defaultly
-configured to follow the order anyway.
+The submission order will be unenforceable on-chain but every off-chain client
+will be defaultly configured to follow the order anyway.
 
 The first member responsible for submitting the DKG result is a member with
 index `hash(new_group_pubkey) % group_size`. If that member did not submit the
 result,  `(hash(new_group_pubkey) % group_size) + 1` becomes responsible next.
 Group members who were inactive and skipped their responsibility might be
-reported by the wallet as ones who failed the heartbeat and marked as ineligible
-for rewards.
+reported by the wallet as ones who <<report-heartbeat-failed,failed the heartbeat>>
+and marked as ineligible for rewards.
 
 For example, if `hash(new_group_pubkey) % group_size = 62`, `group_size = 64`,
 and member `9` submitted DKG result, members with indexes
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,219 @@
...
+- Transactions that can be submitted by anyone, usually submitted often and
+  mostly belonging to a happy path. Never leading to slashing For example,
+  submit sweep proof or submit redemption proof transaction for tBTC v2. 
+
+- Transactions that can be submitted by anyone, submitted rarely or never, and
+  usually leading to slashing. For example, challenge DKG result for ECDSA
+  wallet or report fraud proof transaction for tBTC v2.
+
+For the first group of transactions, we will not enforce the order on-chain and
+we will expect off-chain clients to respect the submission order. If some client
```
beaushinkle, line 49: I would hyperlink "submission order" to its definition here to avoid confusion and give it its own section.

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,219 @@
...
+== In depth
+
+=== Transaction types
+
+On-chain transactions can be divided into three types:
+
+- Transactions that can only be submitted by group members, given that only the
+  group members have the knowledge about input parameters. For example, submit
+  DKG result transaction for ECDSA wallet or submit relay entry transaction for
+  the random beacon.
```
beaushinkle, line 38: In order to make the subject and verb agree here, we want "For example, submitting a DKG result transaction for an ECDSA wallet or submitting a relay entry transaction for the random beacon".

Bonus points for hyperlinking to the relevant sections :D

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,219 @@
...
+  
+- Transactions that can be submitted by anyone, usually submitted often and
+  mostly belonging to a happy path. Never leading to slashing For example,
+  submit sweep proof or submit redemption proof transaction for tBTC v2.
```
beaushinkle, line 42: "Never leading to slashing" should be replaced with a complete sentence like "This type of transaction should never lead to slashing."

Then, we can do the same sort of subject/verb agreement as above: "submitting a sweep proof..."

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,219 @@
...
+and free from misbehavior than to optimize costs.
+
+==== Group member incentives
+
+===== Submit + approve DKG result
+
+No submission order will be enforced on-chain and the transaction submitter will
+have ETH cost reimbursed. To disincentivize front-running, no additional reward
+will be released by this transaction. Off-chain clients should have a handshake
+agreement to follow the submission order based on
```
beaushinkle, line 81: Is a handshake agreement like a verbal agreement or a gentleman's agreement? Or does it mean something else?

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+For example, if `hash(new_group_pubkey) % group_size = 62`, `group_size = 64`,
+and member `9` submitted DKG result, members with indexes
+`{62, 63, 64, 1, 2, 3, 4, 5, 6, 7, 8}` might be marked as ineligible for
+rewards. Inactive group members are determined once the result is submitted and
+accepted by the chain.
+
+Given that the DKG result submission actually consists of two transactions:
+submit the result and approve the result after the challenge period, the DKG
+result submitter to be considered as active must submit both transactions. That
```
beaushinkle, line 99: "Given that the DKG result submission process consists of two transactions: submitting the result and approving the result after the challenge period ends, in order for the operator to be considered active, they must submit both transactions."

beaushinkle:
Okay - something that you mentioned in the RFC but I thought was worth repeating: if the group votes to disable an operator's reward, they *don't* get that operator's share - the DAO does. That also means the DAO should probably *do something else* with those extra tokens other than increase rewards. We don't want to incentivize voting for your own salary.

I made a bunch of grammar nitpicks, but overall I'm liking this document a bunch. I'm down to edit RFC 1 to point to this when it gets merged in, rather than to have it's own relatively pitiful section

docs/rfc/rfc-6.adoc
pdyraga: What is the difference between a verbal agreement and a gentleman's agreement? If they are the same then yes, that's what I meant here. I'll clarify.

docs/rfc/rfc-6.adoc
pdyraga: Not sure if I follow, should we make this paragraph a definition of a submission order and link from _Submit + approve DKG result_ to it?

docs/rfc/rfc-6.adoc
pdyraga: Done in https://github.com/keep-network/tbtc-v2/pull/115/commits/35c11545a30a2bb2a9a4eef2df8cab7ccd15a124 and https://github.com/keep-network/tbtc-v2/pull/115/commits/4edbb20c4f606999b727e6aeb7d78a0cc662de36.

docs/rfc/rfc-6.adoc
pdyraga: Done in https://github.com/keep-network/tbtc-v2/pull/115/commits/35c11545a30a2bb2a9a4eef2df8cab7ccd15a124

docs/rfc/rfc-6.adoc
pdyraga: Please see https://github.com/keep-network/tbtc-v2/pull/115/commits/35c11545a30a2bb2a9a4eef2df8cab7ccd15a124

docs/rfc/rfc-6.adoc
pdyraga: Please see https://github.com/keep-network/tbtc-v2/pull/115/commits/35c11545a30a2bb2a9a4eef2df8cab7ccd15a124 and https://github.com/keep-network/tbtc-v2/pull/115/commits/98c2f6890d384e3f14833fb30635d843af73f666 for my bonus points!

pdyraga:
> Okay - something that you mentioned in the RFC but I thought was worth repeating: if the group votes to disable an operator's reward, they don't get that operator's share - the DAO does. That also means the DAO should probably do something else with those extra tokens other than increase rewards. We don't want to incentivize voting for your own salary.

I tried to clarify it in https://github.com/keep-network/tbtc-v2/pull/115/commits/a24bbee9752ea31b0ac817bcd45ba8526ae1441b, please let me know if it makes sense.

pdyraga:
Side note, I realized we already have majority voting: in DKG, operators selected to the group vote on who was inactive or disqualified during the off-chain protocol and we remove IA/DQ operators from the group based on the signatures on the DKG result.

docs/rfc/rfc-6.adoc
beaushinkle: Here, were saying that we expect the off-chain clients to respect the submission order, but at this point in the document, the submission order isn't defined yet. My suggestion was to create a new header like

```
[[submission-order]]
===== Submission Order
Whenever it is prudent to not have members of the signing group race to submit a transaction and waste gas, they agree off-chain to submit in an order, defined as follows...
```

and then link to that section whenever you talk about the submission order anywhere else

beaushinkle:
> Side note, I realized we already have majority voting: in DKG, operators selected to the group vote on who was inactive or disqualified during the off-chain protocol and we remove IA/DQ operators from the group based on the signatures on the DKG result.

yeah - my understanding was that this creates a transaction that gets published rather than a totally off-chain vote, and so we don't run into the nonce problem that you mentioned when we were talking about voting to reject a fee

beaushinkle:
Okay - I have a bonus-point suggestion about pulling out the submission order into its own section, but I don't want to block on that. :+1: from me.

beaushinkle (dismissed):

michalinacienciala:
I will have a couple of comments, I'm in the middle of review, so please wait with merging a bit :)

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+network.
+
+[[challenge-dkg-result]]
+===== Challenge DKG result
+
+No ETH cost reimbursed. Called exceptionally, ideally never. Incentivized by T
+from slashed stakes.
+
+===== Notify redemption timeout
+
```
michalinacienciala, line 200:
```suggestion
```

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+maintainers.
+
+==== Third party incentives
+
+
+All transactions in this section must be supported by the off-chain client. We
+are not concerned about front-running. Front-running may happen and it is even
+desirable to some extent because these transactions protect the health of the
+network.
+
```
michalinacienciala, line 192: Could we also list 'Notify relay timeout' here?

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+===== Report fraud
+
+No ETH cost reimbursed. Called exceptionally, ideally never. Incentivized by T
+from slashed stakes.
+
+=== ETH pool
+
+DAO needs to fund ETH pool that will be used for reimbursements. The pool should
+probably be a separate contract. The pool needs to protect against malicious
+miner-operators by placing a governable gas price ceiling. It should be possible
```
michalinacienciala, line 215: Beau proposed in [RFC 1](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#transaction-incentives) following governable param for this: `max_gas_refund_price`, we could mention it here.

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+```
+┌────────────────────────────────────────┐   calls  ┌─────────────────────────┐
+|               Maintainer               |─────────▶|          Bridge         |
+└────────────────────────────────────────┘          └─────────────────────────┘  
+| submitSweeProof() onlyMaintainer       |          | submitSweepProof()      | 
+| submitRedemptionProof() onlyMaintainer |          | submitRedemptionProof() | 
+| startDkg() onlyMaintainer              |          | startDkg()              | 
+└────────────────────────────────────────┘          └─────────────────────────┘
+```
+
```
michalinacienciala, line 164: Could we also add 'Notify time to sweep' and 'Notify time to create wallet' here?

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+contract. Maintainers are expected to follow round-robin distribution of work
+but the order is not going to be enforced on-chain. DAO/council will have the
+functionality of removing maintainers not performing their duties.
+DAO/council may decide to distribute additional T rewards for maintainers. This
+functionality should be supported by the Maintainer contract.
+
+Eventually, the role of maintainers can be delegated to another network such as
+Gelato.
+
+
```
michalinacienciala, line 154:
```suggestion
```

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+ETH reimbursed, no additional reward to disincentivize front-running between
+maintainers. 
+
+===== Report DKG timeout
+
+ETH reimbursed, no additional reward to disincentivize front-running between
+maintainers.
+
+==== Third party incentives
+
```
michalinacienciala, line 186:
```suggestion
```

michalinacienciala (commented):
Left some small comments. Also, do we leave the [Transaction incentives section in RFC 1](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#transaction-incentives) as it is for now?

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+A wallet should periodically execute off-chain heartbeats and report inactive
+operators responsible for potential heartbeat failures. The majority of wallet
+members need to come together and sign a message with their private ECDSA
+operator keys. The message should contain the list of inactive operators and
+nonce ensuring uniqueness. The same mechanism can be used for reporting
+operators who failed their duty of submitting DKG result. 
+
+No submission order will be enforced on-chain for the heartbeat failure
+transaction and the transaction submitter will have ETH cost reimbursed. To
```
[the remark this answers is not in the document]
michalinacienciala: See also my comment in https://github.com/keep-network/tbtc-v2/pull/127#discussion_r798406929. That PR suggests that a missed heartbeat notification is a Type 3 transaction.

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+from slashed stakes.
+
+=== ETH pool
+
+DAO needs to fund ETH pool that will be used for reimbursements. The pool should
+probably be a separate contract. The pool needs to protect against malicious
+miner-operators by placing a governable gas price ceiling. It should be possible
+to withdraw unspent ETH in case we decide to replace Maintainer incentives with
+something else (for example Gelato) or decide to add more functions there
+(for example, the SPV relay updates).
```
nkuba, line 218: Maybe it would be worth mentioning that ETH Pool should be upgradeable (it could be connected with the possibility to withdraw unspent ETH, e.g. by transferring it to a new pool)

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+DAO needs to fund ETH pool that will be used for reimbursements. The pool should
+probably be a separate contract. The pool needs to protect against malicious
+miner-operators by placing a governable gas price ceiling. It should be possible
+to withdraw unspent ETH in case we decide to replace Maintainer incentives with
+something else (for example Gelato) or decide to add more functions there
+(for example, the SPV relay updates). 
+
+== Open questions
+
+- Should the DAO/council approve the wallet's decisions to mark some members as
```
nkuba, line 222: Ineligibility for rewards will result in the rewards coming for this particular wallet or all rewards in a specific period of time?

The majority sounded good to me at first. If we think that having 51 votes is not enough we could raise the limit to e.g. 64 votes.

But on the other hand what if there is an upgrade of the client that will result in the majority of members staying on the old version accusing members that updated the version. Or when some part of the members is using Ethereum API provider that is down. Having a safety would be wise. Maybe instead of approving the majority decision the DAO could reject the decision and bring a member back to eligible for rewards? Accepting all the decisions may be a PITA.

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+
+- Transactions that can be submitted by anyone, submitted rarely or never, and
+  usually leading to slashing. For example, <<challenge-dkg-result, challenge
+  DKG result>> for ECDSA wallet or <<report-fraud, report fraud>> transaction for
+  tBTC v2.
+
+For the first group of transactions, we will not enforce the order on-chain and
+we will expect off-chain clients to respect the submission order. If some client
+skipped their round, other clients can vote to mark the inactive client as
+ineligible for rewards. To disincentivize frontrunning, we will only reimburse
```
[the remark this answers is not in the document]
dimpar: I'm not 100% sure, but my understanding is that we need to implement and ETH pool described [here](https://github.com/keep-network/tbtc-v2/blob/98c2f6890d384e3f14833fb30635d843af73f666/docs/rfc/rfc-6.adoc#32-eth-pool) and reimbursement will happen upon calling a function on a contract that notifies about failed hearbeats. Something like [here](https://github.com/keep-network/keep-core/blob/main/solidity/random-beacon/contracts/RandomBeacon.sol#L944) but instead of transferring T tokens, we will need to reimburse ETH. Is it way off, Piotr?

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+
+For example, if `hash(new_group_pubkey) % group_size = 62`, `group_size = 64`,
+and member `9` submitted DKG result, members with indexes
+`{62, 63, 64, 1, 2, 3, 4, 5, 6, 7, 8}` might be marked as ineligible for
+rewards. Inactive group members are determined once the result is submitted and
+accepted by the chain.
+
+Given that the DKG result submission process consists of two transactions:
+submitting the result and approving the result after the challenge period ends,
+in order for the operator to be considered active, they must submit **both**
```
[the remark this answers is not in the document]
dimpar: Maybe we can store a `msg.sender` upon the first step - result submission, and in the second step of approval we should verify if `msg.sender == savedMsgSenderFromSubmissionStep`. However, what if someone frontruns the second tx all the time? Should there be some sort of slashing applied or no gas cost reimbursment?

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+transactions.
+
+===== Report heartbeat failed
+
+A wallet should periodically execute off-chain heartbeats and report inactive
```
dimpar, line 109: Are we talking about ECDSA wallet here?

dimpar, line 109: > ..should periodically execute off-chain heartbeats..

Do we know what is "periodically" or what action should trigger off-chain heartbeats? Can you please give an example?

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+operators responsible for potential heartbeat failures. The majority of wallet
```
dimpar, line 110: "..operators responsible for potential heartbeat failures.."

Is the idea to have 2 separate functions for reporting a heartbeat failures?
- first: when dkg is happening and one of the members is not responsive
- second: before the dkg, just to verify if any of the members are not responsive

docs/rfc/rfc-6.adoc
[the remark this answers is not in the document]
nkuba: In the Random Beacon V2, we give a priority to the result submitter to approve the result during a certain period of time. Maybe we could mention the same approach here?

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+off-chain clients will be responsible for submitting them. No ETH will be
+reimbursed and T coming from slashed stakes will be the only optional reward.
+No submission order will be maintained and all clients will try to report
+misbehavior at the same time. It is more important to keep the network healthy
+and free from misbehavior than to optimize costs.
+
+==== Group member incentives
+
+[[submit-dkg-result]]
+===== Submit + approve DKG result
```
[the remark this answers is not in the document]
beaushinkle: I like this idea

docs/rfc/rfc-6.adoc
pdyraga: If some operators selected to a new group were inactive during the off-chain DKG, they are marked as such in the submitted DKG result and marked as ineligible for rewards when the result is approved. This is happening in `approveDkgResult`.

When an active wallet failed a heartbeat, inactive operators are marked as ineligible for rewards, see https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#heartbeat and https://github.com/keep-network/tbtc-v2/pull/127, and the wallet needs to move funds. I imagine all of it will happen in one function to save gas on validating the signatures.

If some operator did not submit DKG result in their round other group members may decide to mark them as inactive and ineligible for rewards even though that operator was active during the off-chain DKG. This is happening in a separate function.

To sum up, I think there are three functions marking someone as ineligible for rewards.
-

docs/rfc/rfc-6.adoc
pdyraga: https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#heartbeat

pdyraga: Correct.

docs/rfc/rfc-6.adoc
pdyraga: > Ineligibility for rewards will result in the rewards coming for this particular wallet or all rewards in a specific period of time?

A particular operator is marked as ineligible for future rewards, for example, for the next 2 weeks.

> The majority sounded good to me at first. If we think that having 51 votes is not enough we could raise the limit to e.g. 64 votes.

We also need to balance the cost of validating signatures. Something we can decide during the implementation.

> But on the other hand what if there is an upgrade of the client that will result in the majority of members staying on the old version accusing members that updated the version. Or when some part of the members is using Ethereum API provider that is down. Having a safety would be wise. Maybe instead of approving the majority decision the DAO could reject the decision and bring a member back to eligible for rewards? Accepting all the decisions may be a PITA.

I think that the DAO should be able to turn rewards back on for someone. But it should not approve decisions of the group every single time. We could think about it for slashing but not for rewards ineligibility IMO.

docs/rfc/rfc-6.adoc
pdyraga: > Maybe it would be worth mentioning that ETH Pool should be upgradeable (it could be connected with the possibility to withdraw unspent ETH, e.g. by transferring it to a new pool)

It does not have to be upgradeable but yes, it needs to have a function allowing the DAO to withdraw unspent ETH. I was thinking this is obvious but if not, I can make it clear in the RFC.

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+not increase the rewards of other group members! Rewards saved on inactive
+operators marked as ineligible should return back to the DAO. The DAO should do
+something else with those extra tokens other than increase rewards. We do not
+want to incentivize voting for your own salary.
+
+[[maintainer-incentives]]
+==== Maintainer incentives
+
+For the first release, the role of maintainers will belong to the willing
+off-chain clients approved by DAO or council. Every client will have maintainer
```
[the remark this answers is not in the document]
pdyraga: I think that the `Bridge` contract should not contain any logic related to reimbursements. This should all happen in the `Maintainer` contract proxying calls. If we decide to move to Gelato, we will just remove `Maintainer` and have Gelato network calling `Bridge` directly.

If someone wants to bypass `Maintainer`, fine. They will pay for the transaction and don't get reimbursed. An important part is to make sure that in this case, all `Bridge` functions revert early to do not burn too much gas from frontrunned honest maintainer.

Yes, maintainers are approved on the `Maintainer` contract. They are not trusted anymore than 3rd party anons calling `Bridge` functions directly. The only reason to approve them is to make sure they follow the off-chain agreement and do not frontrun each other.

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+functionality built-in and enabled in the configuration file.
+
+We will implement a simple reimbursement contract for maintainers proxying calls
+to Bridge functions. Only approved maintainers should be able to use the proxy
+contract. Maintainers are expected to follow round-robin distribution of work
+but the order is not going to be enforced on-chain. DAO/council will have the
+functionality of removing maintainers not performing their duties.
+DAO/council may decide to distribute additional T rewards for maintainers. This
+functionality should be supported by the Maintainer contract.
```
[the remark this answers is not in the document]
pdyraga: I was thinking about the proxy but this could be confusing with the proxy pattern on Solidity... Any other ideas?

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+Gelato.
+
+
+```
+┌────────────────────────────────────────┐   calls  ┌─────────────────────────┐
+|               Maintainer               |─────────▶|          Bridge         |
+└────────────────────────────────────────┘          └─────────────────────────┘  
+| submitSweeProof() onlyMaintainer       |          | submitSweepProof()      | 
+| submitRedemptionProof() onlyMaintainer |          | submitRedemptionProof() | 
+| startDkg() onlyMaintainer              |          | startDkg()              |
```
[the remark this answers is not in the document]
pdyraga: `Bridge` does not care who is calling it. I'll clarify it. Also please look at [this](https://github.com/keep-network/tbtc-v2/pull/115#discussion_r799383518) comment.

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+operators marked as ineligible should return back to the DAO. The DAO should do
+something else with those extra tokens other than increase rewards. We do not
+want to incentivize voting for your own salary.
+
+[[maintainer-incentives]]
+==== Maintainer incentives
+
+For the first release, the role of maintainers will belong to the willing
+off-chain clients approved by DAO or council. Every client will have maintainer
+functionality built-in and enabled in the configuration file.
```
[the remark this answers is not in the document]
pdyraga: Could be the same, I do not see reasons for separating them.

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+operators responsible for potential heartbeat failures. The majority of wallet
+members need to come together and sign a message with their private ECDSA
+operator keys. The message should contain the list of inactive operators and
+nonce ensuring uniqueness. The same mechanism can be used for reporting
+operators who failed their duty of submitting DKG result. 
+
+No submission order will be enforced on-chain for the heartbeat failure
+transaction and the transaction submitter will have ETH cost reimbursed. To
+disincentivize front-running, no additional reward will be released by this
+transaction. Off-chain clients should have a handshake agreement to follow the
```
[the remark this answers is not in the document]
pdyraga: @beaushinkle which one?

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+submission order: the operator submitting this TX is the first active one.
```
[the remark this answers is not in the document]
pdyraga: > Or do we want to have a similar mechanism as we had for the result submission (hash(new_group_pubkey) % group_size)?

No, because this is about heartbeat failure, we are not in the context of DKG here. See https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#heartbeat.

Say we have a group of 5 operators: A,B,C,D,E and A,B are inactive and heartbeat failed because of them. The first to submit would be C, then D, then E but I intentionally left it a bit unspecified - we need to figure it out during the implementation, there are some more questions to answer here. For example, what happens if we have one operator inactive in the middle? One option would be to have C always responsible for submitting and if they don't, group signs another heartbeat failure with C included as inactive.

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+result,  `(hash(new_group_pubkey) % group_size) + 1` becomes responsible next.
+Group members who were inactive and skipped their responsibility might be
+reported by the wallet as ones who failed the heartbeat and marked as ineligible
+for rewards.
+
+For example, if `hash(new_group_pubkey) % group_size = 62`, `group_size = 64`,
+and member `9` submitted DKG result, members with indexes
+`{62, 63, 64, 1, 2, 3, 4, 5, 6, 7, 8}` might be marked as ineligible for
+rewards. Inactive group members are determined once the result is submitted and
+accepted by the chain.
```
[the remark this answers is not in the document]
pdyraga: Correct, just like we have now in beacon v2 and ECDSA wallet. This is also how it worked for beacon v1. There is no easy way to validate it on-chain.

docs/rfc/rfc-6.adoc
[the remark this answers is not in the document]
pdyraga: Implementation detail.

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+misbehavior at the same time. It is more important to keep the network healthy
+and free from misbehavior than to optimize costs.
+
+==== Group member incentives
+
+[[submit-dkg-result]]
+===== Submit + approve DKG result
+
+No submission order will be enforced on-chain and the transaction submitter will
+have ETH cost reimbursed. To disincentivize front-running, no additional reward
```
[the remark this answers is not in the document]
pdyraga: The submitter of the first TX will have a priority to call the second TX and they will get reimbursed for both when calling the second one.

docs/rfc/rfc-6.adoc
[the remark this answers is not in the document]
michalinacienciala: [Here](https://github.com/keep-network/tbtc-v2/pull/127#discussion_r798924792) Beau settles on 'informal agreement'.

docs/rfc/rfc-6.adoc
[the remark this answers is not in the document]
michalinacienciala: > The only reason to approve them is to make sure they follow the off-chain agreement and do not frontrun each other.

What kind of validations of the client are supposed to be done by the `Maintainer` contract or DAO/council members when approving a client? I'm not familiar with how it works - will DAO somehow have access to the code of the client and will be able to verify that client was not modified to skip the order of transaction submission?
Or maybe this ensuring that off-chain order is followed will be done retroactively - DAO first approves a client and later if that client misbehaves (e.g. skips the line), DAO can step in and remove that client from the list of approved maintainers?

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+`hash(new_group_pubkey) % group_size`.
+The submission order will be unenforceable but everyone will defaultly
+configured to follow the order anyway.
+
+The first member responsible for submitting the DKG result is a member with
+index `hash(new_group_pubkey) % group_size`. If that member did not submit the
+result,  `(hash(new_group_pubkey) % group_size) + 1` becomes responsible next.
+Group members who were inactive and skipped their responsibility might be
+reported by the wallet as ones who failed the heartbeat and marked as ineligible
+for rewards.
```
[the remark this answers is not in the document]
nkuba: In previous implementations, we were using a submission eligibility order that was also verified on-chain, so there was a way to determine who should submit but didn't. For clarity, we could add a sentence that we're not doing anything like that on-chain but leave it all to the majority.

docs/rfc/rfc-6.adoc
pdyraga: This RFC is for tBTC v2 and relay entry timeout is not a part of it. I'd rather not mix them there.

docs/rfc/rfc-6.adoc
pdyraga: `startDkg` is what I assumed is called when it's a time to create a wallet but now I see the name could be misleading, I'll adjust the RFC.

pdyraga: For notify time to sweep, I think sweeping is not something that will have to be triggered on-chain. The clients can decide when to start the sweep.

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+
+===== Start DKG
+
+ETH reimbursed, no additional reward to disincentivize front-running between
+maintainers. 
+
+===== Report DKG timeout
+
+ETH reimbursed, no additional reward to disincentivize front-running between
+maintainers.
```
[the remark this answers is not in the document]
pdyraga: See [31b26ab](https://github.com/keep-network/tbtc-v2/pull/115/commits/31b26aba802ff56c557d51867645928d00322af1), I did not end up with the table but I named the types, hope it's clearer now.

docs/rfc/rfc-6.adoc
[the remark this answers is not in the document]
pdyraga: See [31b26ab](https://github.com/keep-network/tbtc-v2/pull/115/commits/31b26aba802ff56c557d51867645928d00322af1)

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+DAO/council may decide to distribute additional T rewards for maintainers. This
+functionality should be supported by the Maintainer contract.
+
+Eventually, the role of maintainers can be delegated to another network such as
+Gelato.
+
+
+```
+┌────────────────────────────────────────┐   calls  ┌─────────────────────────┐
+|               Maintainer               |─────────▶|          Bridge         |
```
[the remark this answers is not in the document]
pdyraga: See [31b26ab](https://github.com/keep-network/tbtc-v2/pull/115/commits/31b26aba802ff56c557d51867645928d00322af1)

docs/rfc/rfc-6.adoc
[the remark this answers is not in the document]
pdyraga: See [31b26ab](https://github.com/keep-network/tbtc-v2/pull/115/commits/31b26aba802ff56c557d51867645928d00322af1)

docs/rfc/rfc-6.adoc
```diff
@@ -0,0 +1,228 @@
...
+[[submit-dkg-result]]
+===== Submit + approve DKG result
+
+No submission order will be enforced on-chain and the transaction submitter will
+have ETH cost reimbursed. To disincentivize front-running, no additional reward
+will be released by this transaction. Off-chain clients should have a
+gentleman's agreement to follow the submission order based on
+`hash(new_group_pubkey) % group_size`.
+The submission order will be unenforceable but everyone will defaultly
+configured to follow the order anyway.
```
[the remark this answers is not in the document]
pdyraga: See [31b26ab](https://github.com/keep-network/tbtc-v2/pull/115/commits/31b26aba802ff56c557d51867645928d00322af1)

docs/rfc/rfc-6.adoc
pdyraga: [31b26ab](https://github.com/keep-network/tbtc-v2/pull/115/commits/31b26aba802ff56c557d51867645928d00322af1)

docs/rfc/rfc-6.adoc
pdyraga: [31b26ab](https://github.com/keep-network/tbtc-v2/pull/115/commits/31b26aba802ff56c557d51867645928d00322af1)

docs/rfc/rfc-6.adoc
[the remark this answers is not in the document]
pdyraga: [31b26ab](https://github.com/keep-network/tbtc-v2/pull/115/commits/31b26aba802ff56c557d51867645928d00322af1)

docs/rfc/rfc-6.adoc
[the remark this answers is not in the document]
pdyraga: The entire idea of the approval is just to have a smaller group of people submitting maintainer transactions so that we are sure that this group is not modifying their clients to race with each other. There is no formal process. One approach could be e.g. having the bootstrap nodes to be the maintainers plus some group of other clients.

pdyraga:
OK, I think everything is addressed now. If there are no blocking comments, I'd prefer to merge this PR, and address small, stylistic, or formatting issues later.

beaushinkle (approved):

PR #125, opened 2022-02-02
RFC 1: Align RFC 1 Transaction Incentives With RFC 6

beaushinkle:
This PR rewrites the transaction incentives section to summarize and point to RFC 6 so that internal to RFC 1, the concepts can still be referenced but can be followed for more information.

marking as a draft until #115 gets merged

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -515,10 +515,9 @@ Operator-Only transactions are where only the operators have access to the
 information required to assemble the transaction with the right input
 parameters.
 
-In order to avoid all operators racing to submit the
-transaction at the same time, we have an off-chain gentleman's agreement to
-submit based on the operator's position in the group (can use the hash of the
-group's pubkey).
+In order to avoid all operators racing to submit the transaction at the same
+time, we have an off-chain informal agreement to submit based on the operator's
+position in the group (can use the hash of the group's pubkey).
 
 If the designated operator does not submit their transaction before a timeout
 expires, the duty moves to the next operator and the group can sign a
@@ -535,7 +534,7 @@ Public-Knowledge transactions are where anyone has access to the information
 required to assemble the transaction.
 
 In order to prevent wasting gas on racing to submit, we can either use an
-off-chain gentleman's agreement for the operators like in
+off-chain informal agreement for the operators like in
 <<operator-only,Operator-Only>> transactions, or we can delegate the
 transactions to a network like https://www.gelato.network/[Gelato]. We can't
 cost-effectively stop members of the public from trying to race to submit.
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -504,7 +504,7 @@ this transaction.
 [[transaction-incentives]]
 === Transaction Incentives
 
-Governable Parameters
+Governable Parameters:
 - `max_gas_refund_price`: The highest amount of gwei that the gas refund
   contract will pay out per gas for a refund transaction.
 
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -503,6 +503,11 @@ this transaction.
 
 [[transaction-incentives]]
 === Transaction Incentives
+
+Governable Parameters
+- `max_gas_refund_price`: The highest amount of gwei that the gas refund
+  contract will pay out per gas for a refund transaction.
+
 Transaction incentives are more deeply explored in link:rfc-6.adoc[RFC 6:
 Transaction Incentives]. Summarized:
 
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-1.adoc
```diff
@@ -503,37 +503,58 @@ this transaction.
...
+<<public-knowledge,Public-Knowledge>>, and <<punishment,Punishment>>.
+
+[[operator-only]]
+==== Operator-Only
+Operator-Only transactions are where only the operators have access to the
+information required to assemble the transaction with the right input
+parameters.
+
+In order to avoid all operators racing to submit the
+transaction at the same time, we have an off-chain gentleman's agreement to
```
michalinacienciala, line 519: Aren't we switching 'gentleman's' to 'nonformal' (or maybe even better 'non-formal')?

docs/rfc/rfc-1.adoc
```diff
@@ -503,37 +503,58 @@ this transaction.
...
+submit based on the operator's position in the group (can use the hash of the
+group's pubkey).
+
+If the designated operator does not submit their transaction before a timeout
+expires, the duty moves to the next operator and the group can sign a
+transaction to mark that operator as inactive. Such a transaction would be a
+<<punishment,Punishment>>.
```
michalinacienciala, line 526: I just want to make sure I understand correctly (plus have one question):
Let's say we're talking about transaction submitting DKG result. So in the case where the first operator does not submit the DKG result before the timeout, second operator would send that transaction. This transaction with the DKG result sent by operator_2 would not report that operator_1 was inactive during result submission. It would be a transaction of Operator-Only type. It would be reimbursed (maybe during approval and not during submission, but still).
To report the inactivity of operator_1 and it's ineligibility for rewards a separate transaction would need to be sent by a willing party. That would be transaction of the Punishment type. This transaction would not be reimbursable.
Is that right so far?
If so, what would be the incentive here for posting of this second transaction? We're not slashing here, and funds saved by not paying the rewards to operator_1 will not increase the rewards for other operators, so why would anybody care to post that?

docs/rfc/rfc-1.adoc
```diff
@@ -503,37 +503,58 @@ this transaction.
...
+
+In order to prevent wasting gas on racing to submit, we can either use an
+off-chain gentleman's agreement for the operators like in
+<<operator-only,Operator-Only>> transactions, or we can delegate the
+transactions to a network like https://www.gelato.network/[Gelato]. We can't
+cost-effectively stop members of the public from trying to race to submit.
+
+To compensate these transactions, whoever posts them will have the gas spent
+reimbursed by a DAO-funded eth pool in the same transaction, with governable
+limits on how much eth can be spent on gas.
```
michalinacienciala, line 545: We could mention here the `max_gas_refund_price` param, which we have listed at the bottom of the doc.

docs/rfc/rfc-1.adoc
beaushinkle: Great catch!

Your understanding is totally correct - operator_1 failed, so the duty falls to operator_2 and the wallet can sign a transaction to disable rewards for operator_1.

As you say, there's no incentive to post that transaction, and only the operators have the required info, so even though it results in a punishment we need to incentivize it like it's a Operator-Only transaction (refund the eth). I'll amend the RFC 😇

docs/rfc/rfc-1.adoc
beaushinkle: addressed in [9aedaa3](https://github.com/keep-network/tbtc-v2/pull/125/commits/9aedaa3c728976584a5f7cdcf2f9ebce41743cd2)

docs/rfc/rfc-1.adoc
beaushinkle: addressed in [cf33272](https://github.com/keep-network/tbtc-v2/pull/125/commits/cf33272c245caa20cdbe6e38d22fe80085d790c3)

docs/rfc/rfc-1.adoc
beaushinkle: addressed in [47995fd](https://github.com/keep-network/tbtc-v2/pull/125/commits/47995fd09c29a509a6661a7895588d2b7f8e22b9)

michalinacienciala (approved):

docs/rfc/rfc-1.adoc
```diff
@@ -504,36 +504,62 @@ this transaction.
 [[transaction-incentives]]
 === Transaction Incentives
 
-Governable Parameters:
-
+Governable Parameters
```
michalinacienciala, line 507:
```suggestion
Governable Parameters:

```
Looks that this got in accidentally...

beaushinkle: [753109d](https://github.com/keep-network/tbtc-v2/pull/125/commits/753109dd65daab9333599f01d8d747950547eeb6)

beaushinkle:
alrighty - good for another pass @pdyraga

pdyraga (approved):

Issue #119, opened 2022-02-02
Redemption fraud proof


pdyraga:
From [RFC 1](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#241-fraud-proof):

> When a redemption is requested, the redeeming bitcoin public key and amount are known on the ethereum chain. Any bitcoin transactions with the main wallet wallet UTXO as the input must have outputs that match those known redemption requests, otherwise the transaction was fraudulent.

A fraudulent redemption transaction is a transaction that uses the main UTXO as input and has at least one output sending funds to an address not being part of any known `RedemptionRequest`. If such a fraud occurs, the wallet is not able to do any provable subsequent redemption/sweep since the main UTXO stored in `mainUtxos` points to the UTXO used to commit fraud. We can use that fact to expose a function allowing to report that fraud which will perform SPV proof of the fraudulent transaction and mark the wallet as fraudulent*.

The situation is more complex if a fraudulent transaction contains more outputs than just the fraudulent one and those outputs are actually requested redemptions. In that case, we should process the correct redemptions but mark the wallet as fraudulent as well. We can also revert the transaction and allow for single redemption proofs (similarly to the single sweep mechanism). At this point, it's not clear which approach will be better and if both are feasible. This needs to be determined during the implementation.

_*Marking wallet as fraudulent must prevent the wallet to make sweeps and redemptions after the fraud was proven. We probably want the governance to vote about such a wallet fate to eliminate cases where the fraud is not actually fraud but, for example, a bug._

pdyraga:
After looking at all possible options and considering problems with the maximum Bitcoin transaction size, we decided to switch to a challenge-response style of fraud proofs as described in https://github.com/keep-network/tbtc-v2/pull/135. There will be a single mechanism for reporting frauds, both for deposit and redemption flow.

PR #127, opened 2022-02-02
RFC 1: Moving Funds Between Wallets

beaushinkle:
This PR fleshes out the mechanism for moving BTC between wallets, which happens after a failed heartbeat, or after a deposit puts the wallet under the dust threshold.

Out of Scope:

+ How exactly operators are punished in the case of a punishment
+ Edge cases for what happens if a wallet transfers to a wallet that was active when the transfer starts, but then immediately starts to close itself, beginning a race

depends on #125

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -814,8 +814,8 @@ advantage of manipulating btc fee consensus I would be very surprised.
 == Governable Parameters
 Alphabetized list of Governable Parameters with additional notes.
 
-- `base_btc_fee_max`: The highest amount of BTC that operators can
-  initially propose as a fee for miners of Bitcoin transaction.
+- `base_btc_fee_max`: The highest amount of BTC that operators can initially
+  propose as a fee for miners of Bitcoin transaction.
 - `btc_fee_broadcast_timeout`: The amount of time an operator has to provide a
   suggested BTC fee before the other operators give up and try the next
   operator.
@@ -827,17 +827,35 @@ Alphabetized list of Governable Parameters with additional notes.
   be considered for a sweep.
 - `failed_heartbeat_reward_removal_period`: The amount of time an operator is
   removed from reward eligibility after failing a heartbeat.
+- `fraud_challenge_defend_timeout`: The amount of time the wallet has to defend
+  against a fraud challenge.
+- `fraud_challenge_deposit_amount`: The amount of ETH the party challenging the
+  wallet for fraud needs to deposit.
+- `fraud_notifier_reward_multiplier`: The percentage of the notifier reward
+  from the staking contract the notifier of a fraud receives.
+- `fraud_slashing_amount`: The amount of stake slashed from each member of a
+  wallet for a fraud.
+- `heartbeat_block_length`: The number of ethereum blocks until the next
+  heartbeat. If set to 40, then the signers sign every 40th block.
 - `heartbeat`: The number of group members required for a
   <<heartbeat,heartbeat>> to be successful.
-- `heartbeat_block_length`: The number of ethereum blocks until the next heartbeat.
-  If set to 40, then the signers sign every 40th block.
 - `max_gas_refund_price`: The highest amount of gwei that the gas refund
   contract will pay out per gas for a refund transaction.
+- `redemption_request_timeout_redemeer_bonus_multiplier`: The percentage of the
+  notifier reward from the staking contract the redeemer receives in case of a
+  redemption timeout.
+- `redemption_request_timeout_slashing_amount`: The amount of stake slashed
+  from each member of a wallet for a timed-out redemption request.
+- `redemption_request_timeout`: The amount of time the wallet has to provide
+  redemption proof.
+- `redemption_treasury_fee`: The percentage of redeemed amount put aside as a
+  treasury fee.
 - `skip_sweep_timeout`: The amount of time the depositor has to reimburse the
   operator for the gas of the sweep and collect their account balance.
 - `sweep_max_btc`: The amount of summed non-dust unswept bitcoin deposits that
   will trigger an early sweep on a wallet.
-- `sweep_period`: The amount of time we wait between scheduled sweeps on a wallet.
+- `sweep_period`: The amount of time we wait between scheduled sweeps on a
+  wallet.
 - `sweeping_fee_bump_period`: The amount of time we wait to see if a sweeping
   tranaction is mined before increasing the fee.
 - `sweeping_fee_max_multiplier`: The highest we will try to increment the fee
@@ -850,26 +868,14 @@ Alphabetized list of Governable Parameters with additional notes.
 - `sweeping_refund_safety_time`: The amount of time prior to when a UTXO
   becomes eligible for a refund where we will not include it in a sweeping
   transaction.
-- `redemption_request_timeout`: The amount of time the wallet has to provide
-  redemption proof.
-- `redemption_request_timeout_slashing_amount`: The amount of stake slashed
-  from each member of a wallet for a timed-out redemption request.
-- `redemption_request_timeout_redemeer_bonus_multiplier`: The percentage of the
-  notifier reward from the staking contract the redeemer receives in case of a
-  redemption timeout.
-- `redemption_treasury_fee`: The percentage of redeemed amount put aside as a
-  treasury fee.
+- `wallet_closure_timeout`: The amount of time that a wallet has to
+  successfully move its funds to another wallet and inform the ethereum chain
+  before it is at risk of punishment.
 - `wallet_creation_period`: How frequently we attempt to create new wallets.
-- `wallet_max_age`: The oldest we allow a wallet to become before we transfer the funds
-  to a randomly selected wallet.
-- `wallet_min_btc`: The smallest amount of btc a wallet can hold before we
+- `wallet_max_age`: The oldest we allow a wallet to become before we transfer
+  the funds to a randomly selected wallet.
+- `wallet_min_btc`: The smallest amount of BTC a wallet can hold before we
   attempt to close the wallet and transfer the funds to a randomly selected
   wallet.
-- `fraud_slashing_amount`: The amount of stake slashed from each member of a
-  wallet for a fraud.
-- `fraud_notifier_reward_multiplier`: The percentage of the notifier reward from
-  the staking contract the notifier of a fraud receives.
-- `fraud_challenge_defend_timeout`: The amount of time the wallet has to defend
-  against a fraud challenge.
-- `fraud_challenge_deposit_amount`: The amount of ETH the party challenging the
-  wallet for fraud needs to deposit.
+- `wallet_transfer_max`: The most amount of BTC a wallet can transfer to a
+  single other wallet during a wallet closure.
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -699,7 +699,8 @@ disabled (after the active operators sign this). See
 <<transaction-incentives,transaction incentives>> for more on how to encourage
 this transaction.
 
-=== Bridge donate
+[[donate]]
+=== Bridge Donation
 
 If a wallet committed fraud, there has to be a function allowing to "donate" BTC
 to the Bridge without increasing anyone's balances. This function needs to
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -368,13 +368,13 @@ to buy back T tokens, and then uses those to reimburse the operator.
 
 Governable Parameters:
 
-- `redemption_request_timeout`: The amount of time the wallet has to provide
-  redemption proof.
-- `redemption_request_timeout_slashing_amount`: The amount of stake slashed
-  from each member of a wallet for a timed-out redemption request.
 - `redemption_request_timeout_redemeer_bonus_multiplier`: The percentage of the
   notifier reward from the staking contract the redeemer receives in case of a
   redemption timeout.
+- `redemption_request_timeout_slashing_amount`: The amount of stake slashed
+  from each member of a wallet for a timed-out redemption request.
+- `redemption_request_timeout`: The amount of time the wallet has to provide
+  redemption proof.
 - `redemption_treasury_fee`: The percentage of redeemed amount put aside as a
   treasury fee.
 - `wallet_min_btc`: The smallest amount of BTC a wallet can hold before we
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -205,7 +205,7 @@ reason), all they need to do is not submit their reveal and wait 30 days.
 [[sweeping]]
 ==== Sweeping
 
-Governable Parameters
+Governable Parameters:
 
 - `sweeping_refund_safety_time`: The amount of time prior to when a UTXO
   becomes eligible for a refund where we will not include it in a sweeping
@@ -310,7 +310,7 @@ address in 3 hours with no further interaction.
 
 ===== How Frequently To Sweep
 
-Governable Parameters
+Governable Parameters:
 
 - `sweep_period`: The amount of time we wait between scheduled sweeps on a wallet.
 - `sweep_max_btc`: The amount of summed non-dust unswept bitcoin deposits that
@@ -338,7 +338,7 @@ batches) and increasing reliability from a user experience standpoint.
 
 ===== Skipping the Sweeping Line
 
-Governable Parameters
+Governable Parameters:
 
 - `skip_sweep_timeout`: The amount of time the depositor has to reimburse the
   operator for the gas of the sweep and collect their account balance.
@@ -715,6 +715,7 @@ manually liquidate tokens to acquire BTC and donate it to the Bridge.
 === Transaction Incentives
 
 Governable Parameters:
+
 - `max_gas_refund_price`: The highest amount of gwei that the gas refund
   contract will pay out per gas for a refund transaction.
 
@@ -773,7 +774,8 @@ compensate gas.
 
 [[bitcoin-sweeping-fee]]
 === Bitcoin Sweeping Fee
-Governable Parameters
+
+Governable Parameters:
 
 - `btc_fee_broadcast_timeout`: The amount of time an operator has to provide a
   suggested BTC fee before the other operators give up and try the next
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -395,7 +395,7 @@ wallet that contains enough bitcoin to fulfil the redemption. If more BTC
 needs to be redeemed than there is in the largest wallet, then the user needs
 to submit multiple redemptions. After a redemption, if the wallet has under
 `wallet_min_btc` remaining, it transfers that BTC to a randomly selected wallet
-and closes.
+and <<closing-a-wallet,closes>>.
 
 Each redemption request is identified by a concatenation of the wallet's pubkey
 hash and redeemer's output hash (redeemer's BTC address). Such an identifier
@@ -633,13 +633,13 @@ As time passes and operators drop out of the system, a wallet becomes at risk
 of being able to meet the 51-of-100 threshold to produce signatures.
 Additionally, we want to avoid situations where operators are the custodians of
 a wallet for extended periods. To avoid these issues, once a wallet is older
-than the `wallet_max_age`, or if it drops below the liveness threshold (say, below 70 on
-a <<heartbeat,heartbeat>>), we motion to transfer the funds to another randomly
-selected wallet.
+than the `wallet_max_age`, or if it drops below the liveness threshold (say,
+below 70 on a <<heartbeat,heartbeat>>), we motion to
+<<closing-a-wallet,transfer the funds>> to another randomly selected wallet.
 
 Once a wallet no longer has funds and is not the primary wallet for new
-deposits, it can be closed and operators are no longer required to maintain
-it.
+deposits, it can be <<closing-a-wallet,closed>> and operators are no longer
+required to maintain it.
 
 [[closing-a-wallet]]
 ==== Closing A Wallet
@@ -798,8 +798,8 @@ times >= `consecutive_failed_heartbeats`, the operators take a union the
 recorded _inactive_ operators, and an operator from among the active operators
 posts a transaction to disable those inactive operators from receiving rewards
 for `failed_heartbeat_reward_removal_period` amount of time. The active
-operators move the remaining BTC to another random wallet and close this
-wallet.
+operators move the remaining BTC to another random wallet and
+<<closing-a-wallet,close>> this wallet.
 
 For the purposes of heartbeats, an operator that is currently unstaking (they
 started their two-week undelegation period) does not count as a live heartbeat,
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -641,6 +641,141 @@ Once a wallet no longer has funds and is not the primary wallet for new
 deposits, it can be closed and operators are no longer required to maintain
 it.
 
+[[closing-a-wallet]]
+==== Closing A Wallet
+
+Governable Parameters:
+
+- `heartbeat`: The number of group members required for a
+  <<heartbeat,heartbeat>> to be successful.
+- `wallet_closure_timeout`: The amount of time that a wallet has to
+  successfully move its funds to another wallet and inform the ethereum chain
+  before it is at risk of punishment.
+- `wallet_max_age`: The oldest we allow a wallet to become before we transfer the funds
+  to a randomly selected wallet.
+- `wallet_min_btc`: The smallest amount of btc a wallet can hold before we
+  attempt to close the wallet and transfer the funds to a randomly selected
+  wallet.
+
+When a wallet fails a <<heartbeat,heartbeat>> `consecutive_failed_heartbeats`
+times, ends a <<redemption,redemption>> with less than `wallet_min_btc`
+remaining, or exceeds the `wallet_max_age`, the operators begin the process of
+closing it. If there are still funds remaining, the <<operator-only,first
+operator>> posts a <<operator-only,reimbursable>> transaction to the ethereum
+chain declaring an intention to close the wallet. This transaction includes a
+timestamp hereby referred to as `move_funds_start_timestamp`.
+
+That operator proposes a BTC fee (as in <<sweeping,sweeping>>) and the
+operators use the wallet's public key hash to choose viable (it must also have
+a passing heartbeat and be above `wallet_min_btc`, and not exceeded the
+`wallet_max_age) wallet(s) to transfer the remaining funds to. They construct a
+P2PKH transaction moving the wallet's main UTXO to that wallet. For more
+details on exactly how a wallet is chosen and how the funds are split up see
+<<wallet-closure-transaction,the dedicated section>>.
+
+After this transaction is complete, <<public-knowledge,the public>> is able to
+submit a SPV proof to let ethereum know that the funds were transferred
+nonfraudulently. Later, a <<donate,donation>> can be made by governance to
+handle any difference in outstanding v2 token supply and locked BTC due to
+mining transfer fees.
+
+Later, <<public-knowledge,the public>> at any time can challenge the operator's
+wallet choice. They do so by showing that there exists a live, valid wallet
+which was created before `move_funds_start_timestamp` that _wasn't_ chosen
+_and_ should have been. These challenges are incentivized to happen sooner
+rather than later, as eventually all of the relevant operators may be fully
+unstaked and there will be no stake to slash, and thus no reward.
+
+*Notes*: Transferring the BTC to any address other than the P2PKH of one of the
+other wallet addresses is fraud, and is <<Punishment,punishable>>. Moving funds
+without sending a follow-up proof is <<Punishment,punishable>>. Failure to
+close a wallet that failed a <<heartbeat,heartbeat>> or fell below
+`wallet_min_btc`, or exceeded `wallet_max_age` is <<Punishment,punishable>>.
+Failure to complete the process before `wallet_closure_timeout` has elapsed is
+<<Punishment,punishable>>.
+
+Any unswept funds can be returned via the 30-day return script, though since
+we're redeeming from the oldest wallet and only sweeping/depositing to the
+newest wallets this hopefully won't come up often 🤞.
+
+[[wallet-closure-transaction]]
+===== Wallet Closure Transactions
+
+Governable Parameters:
+
+- `wallet_transfer_max`: The most amount of BTC a wallet can transfer to a
+  single other wallet during a wallet closure.
+
+When a wallet <<closing-a-wallet,closes>>, if there is any amount of bitcoin
+remaining, it needs to be transfered to another _live_ wallet. Our input to the
+transaction is the closing wallet's main UTXO, and we create an equal sized
+output for the `N` valid wallets, where `N = min(valid_wallets.size,
+ceil(funds_to_transfer / wallet_transfer_max))`.
+
+The destination wallets are chosen for their public key hash's modulus distance
+to the closing wallet's public key hash (both represented as a number). Ties are
+broken by the destination with the greater public key hash, and if we run out of
+viable wallets, it is okay to reuse them (though we prefer not to).
+
+For all the below examples, say that the public key hashes can range from 0 to
+99 (mod 100). Wallet#43 Means that the PKH of the Wallet is 43. Real public key
+hashes and their max mod values will be much higher in practice!
+
+Exceeding Transfer Max Example:
+
+- Wallet#90 with balance `250 BTC` needs to close.
+- `wallet_transfer_max = 100 BTC`
+- Live wallet options are Wallet#30, Wallet#80, Wallet#25, Wallet#15
+- Calculate distances:
+- Wallet#30 = min(90 - 30, 130 - 90) = min(60, 40) = 40
+- Wallet#80 = min(90 - 80, 180 - 90) = min(10, 90) = 10
+- Wallet#25 = min(90 - 25, 125 - 90) = min(65, 35) = 35
+- Wallet#15 = min(90 - 15, 115 - 90) = min(75, 25) = 25
+
+`N = min([Wallet#30, Wallet#80, Wallet#25, Wallet#15].size, ceil(250 / 100)) = min(4, ceil(2.5)) = min(4, 3) = 3`
+Split the 250 BTC 3 ways. Deposit 250/3 = 83 1/3 BTC into Wallet#80, Wallet#15, and Wallet#25.
+
+Under Transfer Max Example:
+
+- Wallet#52 with balance `50 BTC` needs to close.
+- `wallet_transfer_max = 100 BTC`
+- Live wallet options are Wallet#13, Wallet#90, Wallet#59, Wallet#42
+- Calculate distances:
+- Wallet#13 = min(52 - 13, 113 - 52) = min(39, 61) = 39
+- Wallet#90 = min(90 - 52, 152 - 90) = min(38, 62) = 38
+- Wallet#59 = min(59 - 52, 152 - 59) = min(7, 93) = 7
+- Wallet#42 = min(52 - 42, 142 - 52) = min(10, 90) = 10
+
+`N = min([Wallet#13, Wallet#90, Wallet#59, Wallet#42].size, ceil(50/100)) = min(4, ceil(0.5) = min(4,1) = 1)`
+Split the 50 BTC 1 way. Deposit 50/1 = 50 BTC into Wallet#59.
+
+Tiebreak Example:
+
+- Wallet#52 with a balance `13 BTC` needs to close.
+- `wallet_transfer_max = 100 BTC`
+- Live wallet options are Wallet#13, Wallet#60, Wallet#44, Wallet#75
+- Calculate distances:
+- Wallet#13 = min(52 - 13, 113 - 52) = min(39, 61) = 39
+- Wallet#60 = min(60 - 52, 152 - 60) = min(8, 92) = 8
+- Wallet#44 = min(52 - 44, 144 - 52) = min(8, 92) = 8
+- Wallet#75 = min(75 - 52, 152 - 75) = min(23, 77) = 23
+
+`N = min([Wallet#13, Wallet#60, Wallet#44, Wallet#75].size, ceil(13/100)) = min(4, ceil(0.13) = min(4, 1) = 1)`
+Split the 13 BTC 1 way. Since Wallet#60 is tied with Wallet#44 for the closest, we favor Wallet#60
+because 60>44. Deposit 13/1 = 13 BTC into Wallet#60.
+
+Wallet Reuse Example:
+
+- Wallet#35 with a balance `413 BTC` needs to close.
+- `wallet_transfer_max = 100 BTC`
+- Live wallet options are Wallet#20, Wallet#30
+- Calculate distances:
+- Wallet#20 = min(35 - 20, 120 - 35) = min(15, 85) = 15
+- Wallet#30 = min(35 - 30, 130 - 35) = min(5, 95) = 5
+
+`N = min([Wallet#20, Wallet#30].size, ceil(413/100)) = min(2, ceil(4.13)) = min(2, 5) = 2`
+Split the 413 BTC 2 ways. Deposit 413/2 = 206.5 BTC into Wallet#30 and Wallet#20.
+
 [[heartbeat]]
 === Heartbeats
 


beaushinkle:
also, i believe that this would close https://github.com/keep-network/tbtc-v2/issues/35 so if you agree with that (I tried to answer all of the questions in there explicitly) I can add the magic phrase here

docs/rfc/rfc-1.adoc
```diff
@@ -503,37 +539,58 @@ this transaction.
...
+
+In order to avoid all operators racing to submit the
+transaction at the same time, we have an off-chain gentleman's agreement to
+submit based on the operator's position in the group (can use the hash of the
+group's pubkey).
+
+If the designated operator does not submit their transaction before a timeout
+expires, the duty moves to the next operator and the group can sign a
+transaction to mark that operator as inactive. Such a transaction would be a
+<<punishment,Punishment>>.
```
michalinacienciala, line 562: I just want to make sure I understand correctly (plus have one question):
Let's say we're talking about transaction submitting DKG result. So in the case where the first operator does not submit the DKG result before the timeout, second operator would send that transaction. This transaction with the DKG result sent by operator_2 would not report that operator_1 was inactive during result submission. It would be a transaction of Operator-Only type. It would be reimbursed (maybe during approval and not during submission, but still).
To report the inactivity of operator_1 and it's ineligibility for rewards a separate transaction would need to be sent by a willing party. That would be transaction of the Punishment type. This transaction would not be reimbursable.
Is that right so far?
If so, what would be the incentive here for posting of this second transaction? We're not slashing here, and funds saved by not paying the rewards to operator_1 will not increase the rewards for other operators, so why would anybody care to post that?

pdyraga, line 562: I am not sure if such a transaction should be _punishment_ or _operator only_, I'd rather say the latter.

First, I do not think we should reward for this transaction. ETH reimbursement is fine but the submitter should be net-0. We do not want to incentivize this call, while we do want incentivize _punishment_ transactions. Second, only operators from the given group can assemble this transaction because it requires a sufficient number of signatures to support the motion. Such a motion needs to be proposed off-chain and signers need to work off-chain to share their signatures supporting it.

michalinacienciala:
I thought I was commenting in https://github.com/keep-network/tbtc-v2/pull/125/, but now I see got them mixed up. The #125 duplicates part of #127, this transaction incentives stuff was probably not intended here, right? I'll repost my comments in #125 in case you want to remove the Incentives changes from #127.
And I will review the rest of #127.

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,49 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+to submit a SPV proof to let ethereum know that the funds were transferred
+nonfraudulently, and which account balances need to be updated (like the
+treasur which pays the bitcoin mining fee).
+
+*Notes*: We trust the honest majority to pick a new random wallet. Transferring
+the BTC to any address other than the P2PKH of one of the other wallet
+addresses is fraud, and is <<Punishment,punishable>>. Moving funds without
+sending a follow-up proof is <<Punishment,punishable>>. Failure to close a
+wallet that failed a <<heartbeat,heartbeat>> or fell below `wallet_min_btc` is
+<<Punishment,punishable>>.
```
pdyraga, line 476: We can make it a bit more strict. When someone proposes to move the funds to another wallet, we record the target wallet on-chain. Then, non-fraudulent redemption is the one following rules from https://github.com/keep-network/tbtc-v2/issues/119 or the one that moves funds to the wallet recorded in the transaction requesting the move of funds.

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,49 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+<<redemption,redemption>> with less than `wallet_min_btc` remaining, the
+operators begin the process of closing it. If there are still funds remaining,
+the <<operator-only,first operator>> posts a <<operator-only,reimbursable>>
+transaction to the ethereum chain declaring an intention to close the wallet.
+
+That operator proposes a BTC fee (as in <<sweeping,sweeping>>) and an ethereum
+block height commitment, and they use that to construct a P2PKH transaction
+moving the wallet's main UTXO to another randomishly chosen wallet where the
+source of randomness comes from the ethereum block hash at the pre-commited
+block height.
```
pdyraga, line 464: I do not understand the pre-committed block height, can you please elaborate?

What I think we could do instead of an additional commitment is that when someone (does not even have to be an operator from that wallet - this is *public-knowledge*) notifies that a wallet has funds below the dust threshold, they submit a transaction to the chain to select another wallet where the funds should go to.

We take the blockhash (I think this is sufficient) and % N where N is the number of most active wallets we are interested in. We record blockhash % N as the target wallet assuming the amount hold by that wallet is above the dust threshold or that wallet is active; if not we decrement to move to the more recent wallets.

For example:

wallet 20 [ active ] 0.01 BTC
wallet 19 [ active ] 200 BTC
wallet 18 [ inactive ] 198 BTC
wallet 17 [inactive ] 194 BTC
...
wallet 3 [ inactive ] 10 BTC
wallet 2 [ inactive ] 5 BTC
wallet 1 [ inactive ] 0.01 BTC

With N = 4, we would be selecting from wallets 20-17 and all wallets qualify; they are either above the dust threshold (17-19) or they are below the dust threshold but active (20).

pdyraga (commented):
What is yet unclear for me:
- Is there any timeout for a wallet to move funds? If the timeout is exceeded what happens?
- How do we maintain the balance of BTC and the sum of Bank balances? I know the DAO is responsible for covering BTC fees but how exactly? Where and how is the BTC injected into the system?

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,49 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+
+After this transaction is complete, <<public-knowledge,the public>> is able
+to submit a SPV proof to let ethereum know that the funds were transferred
+nonfraudulently, and which account balances need to be updated (like the
+treasur which pays the bitcoin mining fee).
+
+*Notes*: We trust the honest majority to pick a new random wallet. Transferring
+the BTC to any address other than the P2PKH of one of the other wallet
+addresses is fraud, and is <<Punishment,punishable>>. Moving funds without
+sending a follow-up proof is <<Punishment,punishable>>. Failure to close a
```
michalinacienciala, line 474: So the follow-up proof can be sent by anyone and if it's not sent (in what time?), the operators of the emptying wallet may be punished. So usually it will be operators of the wallet who send that proof (it's in their interest to do that), but theoretically this could be done by everybody, because everybody has a sufficient knowledge to do that. Do I understand correctly?

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,49 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+wallet that failed a <<heartbeat,heartbeat>> or fell below `wallet_min_btc` is
```
michalinacienciala, line 475:
```suggestion
wallet that failed a <<heartbeat,heartbeat>>, fell below `wallet_min_btc`, or exceeded `wallet_max_age` is
```

docs/rfc/rfc-1.adoc
michalinacienciala, line 476: All those punishment transactions would lead to slashing, do I get that right?

docs/rfc/rfc-1.adoc
```diff
@@ -503,37 +539,58 @@ this transaction.
...
+<<public-knowledge,Public-Knowledge>>, and <<punishment,Punishment>>.
+
+[[operator-only]]
+==== Operator-Only
+Operator-Only transactions are where only the operators have access to the
+information required to assemble the transaction with the right input
+parameters.
+
+In order to avoid all operators racing to submit the
+transaction at the same time, we have an off-chain gentleman's agreement to
```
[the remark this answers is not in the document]
beaushinkle: I think the final consensus here was a meme 

image

Jokes aside, briefly checking the internet grammar wizards out there, I think "informal" is probably the most accurate and inclusive. I'm down to standardize on that :D

beaushinkle:
oh whoops - let me point this branch against the transaction incentives one so we get a clean diff

docs/rfc/rfc-1.adoc
beaushinkle: I like this scheme - so the public just *points out* after a redemption that the funds need to be transferred, or the operators *point out* after a failed heartbeat that the funds need to be transferred. No one but the miners are able to change the hash of this "hey we need to transfer" declaration, and it carries two benefits:

a) it determines the target wallet on-chain
b) it sets the beginning of the time-out on-chain

:+1:

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,49 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+deposits, it can be <<closing-a-wallet,closed>> and operators are no longer
+required to maintain it.
+
+[[closing-a-wallet]]
+==== Closing A Wallet
+Governable Parameters:
+
+- `wallet_min_btc`: The smallest amount of BTC a wallet can hold before we
+  attempt to close the wallet and transfer the funds to a randomly selected
+  wallet.
```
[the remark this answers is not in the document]
beaushinkle: > Maybe we will only perform that check after redemptions? That would make a bit more sense, but still somebody could make a deposit on a fresh wallet and make a redemption soon afterwards.

Yeah - that's the intention! The definition is weird out of context in a list, so I'll add some clarity.

As for the second point, active wallets can't be closed, and redemptions are still supposed to be coming from the oldest wallets (generally).

-----

Say that `wallet_min_btc = 1 BTC`. If a 3-week old wallet (no longer active) has 1.5 BTC left, and someone skips the dApp to redeem against that wallet instead of the oldest (and we allow that in our validation contract), then it would fall below the threshold and should close. I think this is okay! Closing that wallet is better than making 100 people look after a 0.5 BTC wallet until all of the older wallets get around to finally clearing out.

docs/rfc/rfc-1.adoc
beaushinkle: Yeah - everything listed here is slashable

docs/rfc/rfc-1.adoc
beaushinkle: addressed in [0704bae](https://github.com/keep-network/tbtc-v2/pull/127/commits/0704baebb3cdf5e953a6e6c4f41f4de939a80a3c)

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,49 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+
+When a wallet fails a <<heartbeat,heartbeat>> or ends a
+<<redemption,redemption>> with less than `wallet_min_btc` remaining, the
```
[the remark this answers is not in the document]
beaushinkle: addressed in [0704bae](https://github.com/keep-network/tbtc-v2/pull/127/commits/0704baebb3cdf5e953a6e6c4f41f4de939a80a3c)

docs/rfc/rfc-1.adoc
beaushinkle: e4717caa5fa2862dce4f2e184f670b441c6fc43b fleshes out the timeout mechanism here. We post a transaction when we notice that a wallet needs to be closed (anyone including operators can do it), and we use the hash of that tx to decide which wallet to send funds to. The timeout starts counting from when that tx clears.

Other than that, the rest of your understanding is spot on!

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,49 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+That operator proposes a BTC fee (as in <<sweeping,sweeping>>) and an ethereum
+block height commitment, and they use that to construct a P2PKH transaction
+moving the wallet's main UTXO to another randomishly chosen wallet where the
+source of randomness comes from the ethereum block hash at the pre-commited
+block height.
+
+After this transaction is complete, <<public-knowledge,the public>> is able
+to submit a SPV proof to let ethereum know that the funds were transferred
+nonfraudulently, and which account balances need to be updated (like the
+treasur which pays the bitcoin mining fee).
```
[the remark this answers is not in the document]
beaushinkle: addressed in [8699428](https://github.com/keep-network/tbtc-v2/pull/127/commits/869942844909fd992c7370857c45db7cd9459132)

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,49 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+  wallet.
+
+When a wallet fails a <<heartbeat,heartbeat>> or ends a
+<<redemption,redemption>> with less than `wallet_min_btc` remaining, the
+operators begin the process of closing it. If there are still funds remaining,
+the <<operator-only,first operator>> posts a <<operator-only,reimbursable>>
+transaction to the ethereum chain declaring an intention to close the wallet.
+
+That operator proposes a BTC fee (as in <<sweeping,sweeping>>) and an ethereum
+block height commitment, and they use that to construct a P2PKH transaction
```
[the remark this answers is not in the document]
beaushinkle: rewrote the whole section in [e4717ca](https://github.com/keep-network/tbtc-v2/pull/127/commits/e4717caa5fa2862dce4f2e184f670b441c6fc43b)

docs/rfc/rfc-1.adoc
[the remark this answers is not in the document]
beaushinkle: clarified in [0ea9675](https://github.com/keep-network/tbtc-v2/pull/127/commits/0ea96758d96f51a03c7df587d3485964b61e2e5f)

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,49 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+
+[[closing-a-wallet]]
+==== Closing A Wallet
+Governable Parameters:
+
+- `wallet_min_btc`: The smallest amount of BTC a wallet can hold before we
+  attempt to close the wallet and transfer the funds to a randomly selected
+  wallet.
+
+When a wallet fails a <<heartbeat,heartbeat>> or ends a
```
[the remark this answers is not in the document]
beaushinkle: addressed in [bae7e6c](https://github.com/keep-network/tbtc-v2/pull/127/commits/bae7e6c61b6028eb155321728433cd5b7a320e59)

docs/rfc/rfc-1.adoc
beaushinkle: addressed in [e4717ca](https://github.com/keep-network/tbtc-v2/pull/127/commits/e4717caa5fa2862dce4f2e184f670b441c6fc43b)

docs/rfc/rfc-1.adoc
beaushinkle: addressed in [e4717ca](https://github.com/keep-network/tbtc-v2/pull/127/commits/e4717caa5fa2862dce4f2e184f670b441c6fc43b)

docs/rfc/rfc-1.adoc
beaushinkle: addressed in [6878b4a](https://github.com/keep-network/tbtc-v2/pull/127/commits/6878b4add75c025371166f7982465278ebfc626d)

beaushinkle: Your understanding here is totally correct - I changed the incentive type to Operator-Only in [6878b4a](https://github.com/keep-network/tbtc-v2/pull/127/commits/6878b4add75c025371166f7982465278ebfc626d), and specified that we should be refunding eth.

So it ends up being an operator-only transaction that causes some punishment. Blurry lines, but the end result is that no one gets slashed tokens, only operators are able to submit it, and we need to reimburse the eth for whoever does, and we want to prevent the operators from frontrunning each other.

docs/rfc/rfc-1.adoc
```diff
@@ -503,37 +539,58 @@ this transaction.
...
+
+In order to prevent wasting gas on racing to submit, we can either use an
+off-chain gentleman's agreement for the operators like in
+<<operator-only,Operator-Only>> transactions, or we can delegate the
+transactions to a network like https://www.gelato.network/[Gelato]. We can't
+cost-effectively stop members of the public from trying to race to submit.
+
+To compensate these transactions, whoever posts them will have the gas spent
+reimbursed by a DAO-funded eth pool in the same transaction, with governable
+limits on how much eth can be spent on gas.
```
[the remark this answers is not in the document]
beaushinkle: addressed in [6878b4a](https://github.com/keep-network/tbtc-v2/pull/127/commits/6878b4add75c025371166f7982465278ebfc626d) (accidentally stuck it in with another change - sorry!)

docs/rfc/rfc-1.adoc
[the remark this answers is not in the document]
beaushinkle: I also think that standardization should happen in #125 so that this PR doesn't get muddied up

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,58 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+<<heartbeat,heartbeat>> or ends a <<redemption,redemption>> with less than
+`wallet_min_btc` remaining, or exceeds the `wallet_max_age`, the operators
+begin the process of closing it. If there are still funds remaining, the
+<<operator-only,first operator>> posts a <<operator-only,reimbursable>>
+transaction to the ethereum chain declaring an intention to close the wallet.
+
+That operator proposes a BTC fee (as in <<sweeping,sweeping>>) and, and the
+operators use the hash of the previous intention-to-close transaction as a
+source of randomness to choose viable (it must also have a passing heartbeat
+and be above `wallet_min_btc`) to transfer the remaining funds to. They
```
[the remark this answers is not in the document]
beaushinkle: [aa5fad1](https://github.com/keep-network/tbtc-v2/pull/127/commits/aa5fad1759eccd49630610dddd1625faf483d1f1)

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,58 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+  randomly selected wallet.
+
+When a wallet drops below the `heartbeat` threshold during a
+<<heartbeat,heartbeat>> or ends a <<redemption,redemption>> with less than
+`wallet_min_btc` remaining, or exceeds the `wallet_max_age`, the operators
+begin the process of closing it. If there are still funds remaining, the
+<<operator-only,first operator>> posts a <<operator-only,reimbursable>>
+transaction to the ethereum chain declaring an intention to close the wallet.
+
+That operator proposes a BTC fee (as in <<sweeping,sweeping>>) and, and the
```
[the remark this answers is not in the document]
beaushinkle: [aa5fad1](https://github.com/keep-network/tbtc-v2/pull/127/commits/aa5fad1759eccd49630610dddd1625faf483d1f1)

michalinacienciala (dismissed):

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,59 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+`wallet_min_btc` remaining, or exceeds the `wallet_max_age`, the operators
+begin the process of closing it. If there are still funds remaining, the
+<<operator-only,first operator>> posts a <<operator-only,reimbursable>>
+transaction to the ethereum chain declaring an intention to close the wallet.
+
+That operator proposes a BTC fee (as in <<sweeping,sweeping>>) and the
+operators use the hash of the previous intention-to-close transaction as a
+source of randomness to choose viable (it must also have a passing heartbeat
+and be above `wallet_min_btc`) wallet to transfer the remaining funds to. They
+construct a P2PKH transaction moving the wallet's main UTXO to that wallet.
```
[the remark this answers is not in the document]
beaushinkle: okay - [728e0e5](https://github.com/keep-network/tbtc-v2/pull/127/commits/728e0e5d1f1c7443167b8cde03c756eeb78ddec8)

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,64 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+begin the process of closing it. If there are still funds remaining, the
+<<operator-only,first operator>> posts a <<operator-only,reimbursable>>
+transaction to the ethereum chain declaring an intention to close the wallet.
+
+That operator proposes a BTC fee (as in <<sweeping,sweeping>>) and the
+operators use the hash of the previous intention-to-close transaction as a
+source of randomness to choose viable (it must also have a passing heartbeat
+and be above `wallet_min_btc`, and be within the `random_transfer_depth`)
+wallet to transfer the remaining funds to. They construct a P2PKH transaction
+moving the wallet's main UTXO to that wallet.
```
pdyraga, line 478: We [can not](https://docs.soliditylang.org/en/latest/units-and-global-variables.html) have the hash of the previous intention-to-close transaction on-chain so we will not be able to validate on-chain if the funds were moved to the right wallet. If we do not have to validate it on-chain and we can trust the majority in the wallet to move the funds to any of the `random_transfer_depth` new wallets, this is fine. But if we want to validate it on-chain, then we need to choose something else as the seed.

Options that come to my mind:
- ask the random beacon
- use the most recently created wallet's pubkey hash
- use relay entry from the latest DKG

All these options have their pros and cons. Random beacon is probably the most secure one but it adds some complexity given that the beacon can be busy at the moment of failing the heartbeat and we do want to start moving funds immediately.

On the first sight, hash of the pubkey of the most recently created wallet does not sound that bad but worth thinking more about it cc @eth-r

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,49 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+<<closing-a-wallet,transfer the funds>> to another randomly selected wallet.
 
 Once a wallet no longer has funds and is not the primary wallet for new
-deposits, it can be closed and operators are no longer required to maintain
-it.
+deposits, it can be <<closing-a-wallet,closed>> and operators are no longer
+required to maintain it.
+
+[[closing-a-wallet]]
+==== Closing A Wallet
```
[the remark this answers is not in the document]
pdyraga: FWIW, we already have the "Report heartbeat failed" transaction listed in RFC 6. The only one we are missing is the transaction submitting the proof of moving the funds. I think it should land in the Maintainer transaction incentives section.

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,64 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+operators use the hash of the previous intention-to-close transaction as a
+source of randomness to choose viable (it must also have a passing heartbeat
+and be above `wallet_min_btc`, and be within the `random_transfer_depth`)
+wallet to transfer the remaining funds to. They construct a P2PKH transaction
+moving the wallet's main UTXO to that wallet.
+
+After this transaction is complete, <<public-knowledge,the public>> is able to
+submit a SPV proof to let ethereum know that the funds were transferred
+nonfraudulently, and which account balances need to be updated (like the
+treasury which pays the bitcoin mining fee).
```
pdyraga, line 483: I think we should do the bookkeeping of Bitcoin fees off-chain to minimise gas costs. I think it is enough to emit an event with Bitcoin transaction hash so that the treasury governance can do the donation (see https://github.com/keep-network/tbtc-v2/pull/132/files#diff-1044c578112615d9cb745f17e8203a444e72ddb3c5d8d5fd97d97038952c77b1R622) after some time.

docs/rfc/rfc-1.adoc
```diff
@@ -435,13 +435,64 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+
+*Notes*: Transferring the BTC to any address other than the P2PKH of one of the
+other wallet addresses is fraud, and is <<Punishment,punishable>>. Moving funds
+without sending a follow-up proof is <<Punishment,punishable>>. Failure to
+close a wallet that failed a <<heartbeat,heartbeat>> or fell below
+`wallet_min_btc`, or exceeded `wallet_max_age` is <<Punishment,punishable>>.
```
pdyraga, line 489: But how do we know that on-chain? It feels like it's just a lack of notification about the current state from maintainers.

docs/rfc/rfc-1.adoc
pdyraga: The transaction hash is not available from Solidity function, see the parameters [here](https://docs.soliditylang.org/en/latest/units-and-global-variables.html). We can have it off-chain but not on-chain.

beaushinkle: I think it's fine that we just have it off-chain, since we're just using the information as a way to short-cut consensus.

Here's how I imagine this plays out in the happy path:

+ We detect that we need to transfer funds, and that gets posted to ethereum with a particular hash
+ The wallet reads that message, uses that hash off-chain to all seed the same RNG and pick a uniform random number between [0, `random_transfer_depth`], and picks a wallet from that number, and transfers the funds to that wallet
+ No one challenges the tx because it isn't fraudulent

Unchallengable Bad Path

+ We detect that we need to transfer funds, and that gets posted to ethereum with a particular hash
+ The wallet reads that message, colludes to send it to a wallet within [0, `random_transfer_depth`], and transfers the funds to that wallet
+ No one challenges the tx because it doesn't look fraudulent

(this requires a 51-of-100 attack with modified clients)

Challengable Bad Path

+ We detect that we need to transfer funds, and that gets posted to ethereum with a particular hash
+ The wallet reads that message, colludes to send it to a wallet outside of [0, `random_transfer_depth`], and transfers the funds to that wallet (could be a totally separate wallet)
+ challenge/response sighash etc fraud-proof process

(this requires a 51-of-100 attack with modified clients)

---

So, we open ourselves up to a 51-of-100 attack where the wallet is able to move the funds to a wallet *that they already control* to another wallet of their choosing within `[0, random_transfer_depth]` and have it not seem like fraud.

---

Asking the random beacon sounds gas expensive, using the wallet pubkey hash is past, static information which means that all wallet closures would go to the same wallet until a new wallet is created unless you also had some sort of indexing, and then folks could manipulate the index to get the funds to move into the wallet they wanted.

Can you elaborate on what the relay entry is?

So my main question at the moment is: how expensive is the attack I'm describing above vs paying the random beacon to prevent that attack? There's a decent chance that we're overengineering this at the moment and we could just do the off-chain consensus based on the intention-to-close transaction and upgrade to random-beacon later after more analysis later

docs/rfc/rfc-1.adoc
beaushinkle: What I mean here is that we want to be in a state where we have maintainers that are incentivized to post transactions about the state of wallets (they're too old, too low on funds, they failed a heartbeat, etc). Once one of those transactions gets posted, it starts a timeout. If the close-the-wallet process isn't complete before the time runs out, the wallet gets punished.

Does that answer your question, or are you asking about how we get the maintainers to keep the state up to date?

docs/rfc/rfc-1.adoc
beaushinkle: addressed in 73c7f373212d240c743d6664dbcc01342116dbd9

beaushinkle:
anything else ya'll want to see here?

docs/rfc/rfc-1.adoc
```diff
@@ -633,13 +633,65 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+handle any difference in outstanding v2 token supply and locked BTC due to
+mining transfer fees.
+
+*Notes*: Transferring the BTC to any address other than the P2PKH of one of the
+other wallet addresses is fraud, and is <<Punishment,punishable>>. Moving funds
+without sending a follow-up proof is <<Punishment,punishable>>. Failure to
+close a wallet that failed a <<heartbeat,heartbeat>> or fell below
+`wallet_min_btc`, or exceeded `wallet_max_age` is <<Punishment,punishable>>.
+Failure to complete the process before `wallet_closure_timeout` has elapsed is
+<<Punishment,punishable>>.
```
michalinacienciala, line 690: To report that timeout, anyone would be able to post a transaction calling a `notifyWalletClosureTimeout` function (or something named similarly) and if that transaction would pass the validation, first notifier would receive some percentage of the slashed amount, right?
If so, wouldn't we need to have a governable param `wallet_closure_timeout_notifier_bonus_multiplier` for managing what part of slashed amount goes to notifier?
Or maybe we could have a common `notifier_bonus_multiplier` param for redemption request timeout and wallet closure timeout notifiers?

docs/rfc/rfc-1.adoc
```diff
@@ -633,13 +633,65 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+
+Any unswept funds can be returned via the 30-day return script, though since
+we're redeeming from the oldest wallet and only sweeping/depositing to the
+newest wallet this hopefully won't come up often 🤞.
```
michalinacienciala, line 694:
```suggestion
we're redeeming from the oldest wallet and only sweeping/depositing to the
newest wallets this hopefully won't come up often 🤞.
```

docs/rfc/rfc-1.adoc
```diff
@@ -825,17 +880,48 @@ Alphabetized list of Governable Parameters with additional notes.
...
+- `redemption_request_timeout_slashing_amount`: The amount of stake slashed
+  from each member of a wallet for a timed-out redemption request.
+- `redemption_request_timeout`: The amount of time the wallet has to provide
+  redemption proof.
+- `redemption_request_timeout`: The amount of time the wallet has to provide
+  redemption proof.
+- `redemption_treasury_fee`: The percentage of redeemed amount put aside as a
+  treasury fee.
+- `redemption_treasury_fee`: The percentage of redeemed amount put aside as a
+  treasury fee.
```
michalinacienciala, line 918: Some parameters got duplicated
```suggestion
- `redemption_request_timeout_redemeer_bonus_multiplier`: The percentage of the
  notifier reward from the staking contract the redeemer receives in case of a
  redemption timeout.
- `redemption_request_timeout_slashing_amount`: The amount of stake slashed
  from each member of a wallet for a timed-out redemption request.
- `redemption_request_timeout`: The amount of time the wallet has to provide
  redemption proof.
- `redemption_treasury_fee`: The percentage of redeemed amount put aside as a
  treasury fee.
```

docs/rfc/rfc-1.adoc
```diff
@@ -633,13 +633,65 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+`wallet_min_btc` remaining, or exceeds the `wallet_max_age`, the operators
+begin the process of closing it. If there are still funds remaining, the
+<<operator-only,first operator>> posts a <<operator-only,reimbursable>>
+transaction to the ethereum chain declaring an intention to close the wallet.
+
+That operator proposes a BTC fee (as in <<sweeping,sweeping>>) and the
+operators use the hash of the previous intention-to-close transaction as a
+source of randomness to choose viable (it must also have a passing heartbeat
+and be above `wallet_min_btc`, and be within the `random_transfer_depth`)
+wallet to transfer the remaining funds to. They construct a P2PKH transaction
```
michalinacienciala, line 675: Should we send the funds to only one wallet from the set of `random_transfer_depth` the newest wallets or to all wallets from that set, in equal fractions? Piotr told me there were some conversations pointing to that second solution...
Sending to bigger set of wallets seems to be safer - if there would be some possibility for an operator who controls the closing wallet to abuse the mechanism of moving funds if they control the recipient wallet as well, then having more recipient wallets minimizes the amount of funds at risk.

michalinacienciala (changes requested):
I left some comments. I also discussed the PR with Piotr, he will have some more comments here...

docs/rfc/rfc-1.adoc
beaushinkle: Yup! Though in the interest of keeping this PR well-scoped, I think it's fine to just say that it's "punishable", and leave the details of that punishment to a follow-up PR

michalinacienciala: Ok, I'm fine with that :)

docs/rfc/rfc-1.adoc
beaushinkle: From a risk standpoint, I *greatly* prefer distributing the funds across the wallets. The main question is how much extra that costs in terms of sweeping transactions and ethereum proofs

docs/rfc/rfc-1.adoc
```diff
@@ -634,13 +634,65 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+  successfully move its funds to another wallet and inform the ethereum chain
+  before it is at risk of punishment.
+- `wallet_max_age`: The oldest we allow a wallet to become before we transfer the funds
+  to a randomly selected wallet.
+- `wallet_min_btc`: The smallest amount of btc a wallet can hold before we
+  attempt to close the wallet and transfer the funds to a randomly selected
+  wallet.
+
+When a wallet drops below the `heartbeat` threshold during a
+<<heartbeat,heartbeat>> or ends a <<redemption,redemption>> with less than
```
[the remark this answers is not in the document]
beaushinkle: [2c89353](https://github.com/keep-network/tbtc-v2/pull/127/commits/2c89353ae98c69f909cf510441b23f93df31f023)

docs/rfc/rfc-1.adoc
beaushinkle: [ffcbf40](https://github.com/keep-network/tbtc-v2/pull/127/commits/ffcbf40b22feb64fc64fada6704c1232dc89ef74)

docs/rfc/rfc-1.adoc
```diff
@@ -634,13 +634,65 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+`wallet_min_btc` remaining, or exceeds the `wallet_max_age`, the operators
+begin the process of closing it. If there are still funds remaining, the
+<<operator-only,first operator>> posts a <<operator-only,reimbursable>>
+transaction to the ethereum chain declaring an intention to close the wallet.
+
+That operator proposes a BTC fee (as in <<sweeping,sweeping>>) and the
+operators use the hash of the previous intention-to-close transaction as a
+source of randomness to choose viable (it must also have a passing heartbeat
+and be above `wallet_min_btc`, and be within the `random_transfer_depth`)
```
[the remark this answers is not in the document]
beaushinkle: [70e2369](https://github.com/keep-network/tbtc-v2/pull/127/commits/70e23696c47a148808b66445025474c9102792bf)

beaushinkle:
I added some scope notes in the PR description. I know this PR has some gaps specifically around edge cases around what happens if the wallet that is chosen becomes inactive right *after* it's chosen but *before* the funds finish transferring. I think that should be fleshed out in a follow-up lest this PR grows to monstrous size/complexity!

lukasz-zimnoch:
What is the plan for merging the UTXO representing the transferred funds with the target wallet's main UTXO? Do we want to do it via sweep or using a separate path that will involve an additional BTC transaction and proof submission on Ethereum?

beaushinkle:
> What is the plan for merging the UTXO representing the transferred funds with the target wallet's main UTXO? Do we want to do it via sweep or using a separate path that will involve an additional BTC transaction and proof submission on Ethereum?

I think we're going to want something separate but similar, definitely in a follow-up PR :D

michalinacienciala (commented):
In the `Wallet lifecycle` section we have following:
```
To avoid these issues, once a wallet is older
than the `wallet_max_age`, or if it drops below the liveness threshold (say,
below 70 on a <<heartbeat,heartbeat>>), we motion to
<<closing-a-wallet,transfer the funds>> to another randomly selected wallet.
```
Could you update that? The wallet is no longer selected randomly.

docs/rfc/rfc-1.adoc
```diff
@@ -633,13 +633,138 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+99 (mod 100). Wallet#43 Means that the PKH of the Wallet is 43. Real public key
+hashes and their max mod values will be much higher in practice!
+
+Exceeding Transfer Max Example:
+
+- Wallet#90 with balance `250 BTC` needs to close.
+- `wallet_transfer_max = 100 BTC`
+- Live wallet options are Wallet#30, Wallet#80, Wallet#25, Wallet#15
+- Calculate distances:
+- Wallet#30 = min(90 - 30, 130 - 90) = min(60, 40) = 40
```
pdyraga, line 721: We have two arguments passed to `min` because we want to look from both sides, right? That is, we can imagine the list of wallets as a circle and not as a line. Is it needed though? 

Thinking of it just as just a line would be a bit easier to implement. Are there reasons for doing it another way?

docs/rfc/rfc-1.adoc
```diff
@@ -633,13 +633,138 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+
+- Wallet#35 with a balance `413 BTC` needs to close.
+- `wallet_transfer_max = 100 BTC`
+- Live wallet options are Wallet#20, Wallet#30
+- Calculate distances:
+- Wallet#20 = min(35 - 20, 120 - 35) = min(15, 85) = 15
+- Wallet#30 = min(35 - 30, 130 - 35) = min(5, 95) = 5
+
+Deposit 100 BTC into Wallet#30, 100 BTC into Wallet#20, 100 BTC into Wallet#30,
+100 BTC into Wallet#20, 13 BTC into Wallet#30 for a total of 213 BTC into Wallet#30 and 200 BTC into Wallet#20.
```
pdyraga, line 767: Do we expect the operators to commit to which wallets they are moving the funds before they actually do it?

docs/rfc/rfc-1.adoc
```diff
@@ -694,6 +690,82 @@ Any unswept funds can be returned via the 30-day return script, though since
 we're redeeming from the oldest wallet and only sweeping/depositing to the
 newest wallets this hopefully won't come up often 🤞.
 
+[[wallet-closure-transaction]]
+===== Wallet Closure Transactions
+
+Governable Parameters:
+
+- `wallet_transfer_max`: The most amount of BTC a wallet can transfer to a
+  single other wallet during a wallet closure.
```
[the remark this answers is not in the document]
beaushinkle: That's what Promethea originally proposed, but I didn't see a compelling reason to entangle them (it confuses the naming and reduces governance flexibility) and scope-creeps the PR :D

docs/rfc/rfc-1.adoc
```diff
@@ -633,13 +633,138 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+
+When a wallet <<closing-a-wallet,closes>>, if there is any amount of bitcoin
+remaining, it needs to be transfered to another _live_ wallet. Our input to the
+transaction is the closing wallet's main UTXO, and we create an output for
+every multiple of `wallet_transfer_max` we have in BTC in the closing wallet.
+
+The destination wallets are chosen for their public key hash's modulus distance
```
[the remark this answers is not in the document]
beaushinkle: If you don't do the modulus then numbers in the middle of the range are statistically more likely to be chosen, since they're closer to more numbers than numbers at the ends. Here's a simulated output:

```
const arr = [0,1,2,3,4,5,6,7]
let results = {}
for (let i = 0; i < 1000; i++) {
  const sample = getRandomSample(arr, 3)
  const option1DistanceSquared = Math.pow(sample[0] - sample[1], 2)
  const option2DistanceSquared = Math.pow(sample[0] - sample[2], 2)
  if (option1DistanceSquared < option2DistanceSquared) {
    results[sample[1]] = (results[sample[1]] || 0) + 1
  } else if (option1DistanceSquared > option2DistanceSquared) {
    results[sample[2]] = (results[sample[2]] || 0) + 1
  } else if (sample[1] > sample[2]){
    results[sample[1]] = (results[sample[1]] || 0) + 1
  } else {
    results[sample[2]] = (results[sample[2]] || 0) + 1
  }
}

console.log(results)
```
`getRandomSample` is grabbing a random sample from `[0,1,2,3,4,5,6,7]` like `[4,2,6]`. The first number is the wallet that's being closed and the second and third are the options. Here's how often each number is chosen without a modulus from 1000 runs:

```
{
  '0': 48,
  '1': 108,
  '2': 154,
  '3': 163,
  '4': 151,
  '5': 147,
  '6': 152,
  '7': 77
}
```

But with a modulus:

```
function dist(a, b, modulus) {
  let amod = a % modulus
  let bmod = b % modulus

  if (amod < bmod) {
    const temp = amod
    amod = bmod
    bmod = temp
  } 
  return Math.min(amod-bmod, bmod-amod+modulus);
}


const arr = [0,1,2,3,4,5,6,7]
let results = {}
for (let i = 0; i < 1000; i++) {
  const sample = getRandomSample(arr, 3)
  const option1Distance = dist(sample[0], sample[1], arr.length)
  const option2Distance = dist(sample[0], sample[2], arr.length)

  if (option1Distance < option2Distance) {
    results[sample[1]] = (results[sample[1]] || 0) + 1
  } else if (option1Distance > option2Distance) {
    results[sample[2]] = (results[sample[2]] || 0) + 1
  } else if (sample[1] > sample[2]){
    results[sample[1]] = (results[sample[1]] || 0) + 1
  } else {
    results[sample[2]] = (results[sample[2]] || 0) + 1
  }
}
console.log(results)
```

We get a nice random distribution
```
{
  '0': 122,
  '1': 111,
  '2': 120,
  '3': 105,
  '4': 119,
  '5': 128,
  '6': 159,
  '7': 136
}
```

docs/rfc/rfc-1.adoc
beaushinkle: Yup! Explained it in depth [above](https://github.com/keep-network/tbtc-v2/pull/127#discussion_r828731782)

docs/rfc/rfc-1.adoc
```diff
@@ -633,13 +633,148 @@ As time passes and operators drop out of the system, a wallet becomes at risk
...
+After this transaction is complete, <<public-knowledge,the public>> is able to
+submit a SPV proof to let ethereum know that the funds were transferred
+nonfraudulently. Later, a <<donate,donation>> can be made by governance to
+handle any difference in outstanding v2 token supply and locked BTC due to
+mining transfer fees.
+
+Later, <<public-knowledge,the public>> at any time can challenge the operator's
+wallet choice. They do so by showing that there existed a live, valid wallet
+which was created after `move_funds_start_timestamp` that _wasn't_ chosen _and_
+should have been. These challenges are incentivized to happen sooner rather
```
pdyraga, line 685:
```suggestion
Later, <<public-knowledge,the public>> at any time can challenge the operator's
wallet choice. They do so by showing that there exists a live, valid wallet
which was created before `move_funds_start_timestamp` that _wasn't_ chosen _and_
should have been. These challenges are incentivized to happen sooner rather
```

We can not prove on-chain there was a live wallet without additional costs of tracking timestamps of wallet status updates. But I think it is a typo given the created before/after part. What we can prove easily in the challenge: there is a live wallet, created before `move_funds_start_timestamp` and that wallet should be chosen but it wasn't.

pdyraga (approved):

docs/rfc/rfc-1.adoc
pdyraga: Answer: no. This would not be possible. We will allow anyone to challenge their choice later.

docs/rfc/rfc-1.adoc
beaushinkle: yeah - definitely meant before! [dd68fe7](https://github.com/keep-network/tbtc-v2/pull/127/commits/dd68fe7bb46201b906f6d41f3cbcca28fb9e907b)

pdyraga (dismissed):
This looks good to me but I can't merge given that @michalinacienciala requested changes. Leaving it in your hands Michalina!

michalinacienciala (dismissed):

michalinacienciala:
I've set the automerge, but GH actions are misbehaving again ([githubststus.com](https://www.githubstatus.com/incidents/y5hdmv0p49x3)) and checks are stuck in progress..
@beaushinkle, once outage is fixed, you will probably need to amend the latest commit (or push a new one) to kick off the checks again...

beaushinkle:
Awesome! Thanks a ton ya'll

beaushinkle:
woah the history got funky

beaushinkle:
I rewrote the history in a temp branch if we want to go that route:

```
❯ git log
commit 7679d620a8a58a9bd88358e7fe4a865ffb39e7a0 (HEAD -> temp)
Author: Beau Shinkle <beaushinkle@gmail.com>
Date:   Fri Mar 25 15:32:23 2022 -0400

    Hyperlink the wallet-closure details into the rest of the doc

commit 36b5be0796030aaafe93cc8fa3abd77a607582cd
Author: Beau Shinkle <beaushinkle@gmail.com>
Date:   Fri Mar 25 15:31:47 2022 -0400

    Flesh out the details for closing a wallet

commit 4b4bd7b815520b41f173e29fa685e1c2626b5007
Author: Beau Shinkle <beaushinkle@gmail.com>
Date:   Fri Mar 25 15:29:42 2022 -0400

    Clean up the bridge donation section title

commit 7518d1e460f640d4eb5f6466f7cc5e9b132b4413
Author: Beau Shinkle <beaushinkle@gmail.com>
Date:   Fri Mar 25 15:29:29 2022 -0400

    Properly sort the global governable param list

commit 50e3dbcb1bec25ca877319a1c282872dbc577013
Author: Beau Shinkle <beaushinkle@gmail.com>
Date:   Fri Mar 25 15:28:44 2022 -0400

    Properly sort redemption governable parameters

commit d0dad6d559a70331139dd58c514046feeb43486e
Author: Beau Shinkle <beaushinkle@gmail.com>
Date:   Fri Mar 25 15:27:07 2022 -0400

    Standardize Governable Parameters heading syntax

commit be73ffa655de9704ad660f461170147640f17911 (origin/main, origin/HEAD, main)
```

pdyraga:
Yeah, let's force-push a temp branch as `rfc-1-moving-funds-between-wallets`, and we'll merge with a clean history.

beaushinkle:
okay i think this should be good to merge on green!

michalinacienciala (approved):

PR #129, opened 2022-02-08
Single deposit proof

tomaszslabon:
It can be useful to prove a single deposit when a wallet made a sweep transaction on Bitcoin, but cannot prove the whole batch of deposits.

Depends on https://github.com/keep-network/tbtc-v2/pull/106.


tomaszslabon:
It was decided values will be called with gas limit and wrapped in a try-catch.

PR #128, opened 2022-02-03
Redemption request and proof

lukasz-zimnoch:
Closes: #118 
Depends on: #106 

### Introduction

This pull request adds an implementation of the redemption flow to the `Bridge` contract. That feature allows redeeming Bitcoins for anyone having a TBTC balance in the `Bank` contract. The redemption flow consists of two functions: `requestRedemption` and `submitRedemptionProof` which are both described in the following sections.

### Requesting a redemption

Redemption must be initiated with a call to the `requestRedemption` function. The caller must pass the wallet it wants to redeem Bitcoins from, the output script which should be used to lock redeemed funds, and the requested amount. Having that information, the function performs some preconditions checks to make sure such a request is possible. 

If the request can be made, the function builds the redemption key by concatenating the wallet public key hash with the redeemer output script and applying `keccak256` at the end. That redemption key is used to store the request in the contract state and identify the redemption request during the request lifetime. That means there can be only one pending redemption request asking for redemption from a specific wallet to the specific output script at the same time.

Once the request is put to the contract storage, the function interacts with the `Bank` contract to take the necessary TBTC amount from the redeemer balance and add it to the `Bridge` contract balance. The specific amount taken is always equal to the amount of BTC requested by the redeemer. 

From this point, the request is registered in the system and the wallet should take action to handle it in a proper way and on time. Each request can be terminated in several ways. 

The happy path is when the request is successfully handled by the wallet within the given time window. In that case, the redeemer should receive a BTC amount locked on the output script they passed upon making the request. Worth noting, the actual BTC amount locked on the redeemer output script is always lower than the requested amount since treasury and Bitcoin transaction fees must be incurred. However, each request has an acceptable range that prevents abuse. Once the request is closed successfully, its redemption key can be reused by further requests.

The non-happy paths are when the wallet doesn't handle the request on time and such a timeout is reported or the wallet sends an incorrect amount not satisfying the request's boundaries, which is considered a fraud. In both cases, the redeemer should probably get back their TBTC taken upon request in the appropriate amount, and the wallet should be punished for committing misbehavior. This behavior will be implemented by functions described in the **Next steps** section.

### Proving a Bitcoin redemption transaction

[...]

PR #132, opened 2022-02-10
RFC 1: Spec out details of redemption

pdyraga:
RFC 1 is updated with more details about the redemption mechanism. Specifically:
- redemption constraints related to EVM limitations,
- how the redemption proof should behave,
- what is happening in case of a timeout,
- what is happening in case of fraud,
- how to rebalance the Bridge after fraud.

This change is an outcome of a long discussion with @lukasz-zimnoch and @tomaszslabon regarding the problems that occurred when implementing Solidity code in #128. The code in #128 has been adjusted to follow the approach described here.

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -382,6 +382,17 @@ key than the same `signingGroupPubkey` is fraud.
 
 Governable Parameters:
 
+- `redemption_request_timeout`: The amount of time the wallet has to provide
+  redemption proof.
+- `redemption_request_timeout_slashing_amount`: The amount of stake slashed from
+  each member of a wallet for a timed-out redemption request.
+- `redemption_request_timeout_redemeer_bonus_multiplier`: The percentage of the
+  notifier reward from the staking contract the redeemer receives in case of a
+  redemption timeout.
+- `redemption_fraud_slashing_amount`: The amount or stake slashed from each
+  member of a wallet for a redemption fraud.
+- `redempton_fraud_notifier_reward_multiplier`: The percentage of the notifier
+  reward from the staking contract the notifier of redemption fraud receives.
 - `wallet_min_btc`: The smallest amount of BTC a wallet can hold before we
   attempt to close the wallet and transfer the funds to a randomly selected
   wallet.
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -457,6 +457,38 @@ not disincentivize it for waiting for enough redemption requests to accumulate,
 the redemption timeout should be long enough, for example, 72 hours and ideally
 even longer in the early days of the system.
 
+==== Redemption proof
+
+Based on the information provided in the submitted redemption proof, pending
+redemption requests possibly satisfied by the proof should be retrieved and each
+of them should be validated separately. 
+
+If a pending redemption request exists for every UTXO of the redemption
+transaction but the main wallet UTXO, and BTC value of each UTXO is within the
+expected range, redemption proof is accepted and redeemed Bitcoin balances in
+the Bridge are cleared out. It is important to note that timeout is not
+validated in this function. That is, if the timeout was not reported, the wallet
+can process the entire redemption proof successfully even though the timeout for
+the oldest redemption request already passed.
+
+If a pending redemption request for the given UTXO does not exist, and that
+redemption request was not earlier reported as timed-out or fraudulent because
+of the UTXO value not being within the expected range, the entire redemption
+proof transaction should revert.
+
+If the value of the given UTXO matching pending redemption request is not in the
+expected range, this is a fraud, and the entire redemption proof transactions
+should revert unless that fraud was already reported. 
+
+If a pending redemption request for the given UTXO does not exist but that
+redemption request was earlier reported as timed-out or fraudulent because of
+the UTXO value not being within the expected range, that UTXO is skipped, and
+the redemption proof transaction gets accepted, assuming there are more UTXOs in
+the redemption transaction.
+
+Submit redemption proof function updates the main wallet UTXO. Bitcoin balances
+in the bridge are cleared out for successfully processed redemption requests.
+
 ==== Fraud Proof
 
 When a redemption is requested, the redeeming bitcoin public key and amount are
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -733,12 +733,12 @@ advantage of manipulating btc fee consensus I would be very surprised.
 Alphabetized list of Governable Parameters with additional notes.
 
 - `base_btc_fee_max`: The highest amount of BTC that operators can
-  initially propose as a fee for miners for Bitcoin transaction.
+  initially propose as a fee for miners of Bitcoin transaction.
 - `btc_fee_broadcast_timeout`: The amount of time an operator has to provide a
   suggested BTC fee before the other operators give up and try the next
   operator.
 - `btc_fee_max`: The highest amount of BTC that operators can eventually
-  propose as a fee for miners for Bitcoin transaction.
+  propose as a fee for miners of Bitcoin transaction.
 - `consecutive_failed_heartbeats`: The number of times a heartbeat can fail in
   a row that triggers a move to close the wallet.
 - `dust_threshold`: The minimum bitcoin deposit amount for the transaction to
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -517,6 +517,36 @@ unblocking the wallet for future redemptions.
 The transaction must revert if the given redemption request was already reported
 as timed out.
 
+===== Redemption fraud
+
+Redemption proof containing UTXO identified as fraud is reverting the redemption
+proof submit transaction. Such a fraud must first be reported by calling a
+dedicated function.
+
+There are two possible fraudulent scenarios.
+
+First, there is no redemption requested for the given UTXO. In this scenario,
+each wallet member is slashed for `redemption_fraud_slashing_amount` and the
+notifier receives a reward as specified by
+`redempton_fraud_notifier_reward_multiplier`. The fraudulent output information
+is stored on-chain.
+
+Second, there was a redemption request for the given UTXO but the UTXO value is
+not within the expected range. In this case, the value difference is returned to
+the redemeer as their balance. Each wallet member is slashed for
+`redemption_fraud_slashing_amount` and the notifier receives a reward as
+specified by `redempton_fraud_notifier_reward_multiplier`. The fraudulent output
+information is stored on-chain.
+
+The transaction must revert if the given fraud was already reported. 
+
+If the wallet has still some BTC or if there is a redemption transaction with
+some part of outputs fraudulent and some part of outputs valid, it should be
+possible to submit redemption proof. Because information about fraudulent
+outputs is stored on-chain, these outputs can be skipped by the Bridge and the
+main wallet UTXO can be updated. This mechanism allows proving whatever can
+still be proved and unblocking the wallet for future redemptions.
+
 === Wallet Lifecycle
 
 Governable Parameters:
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -640,6 +640,18 @@ disabled (after the active operators sign this). See
 <<transaction-incentives,transaction incentives>> for more on how to encourage
 this transaction.
 
+=== Bridge donate
+
+If a wallet committed fraud, there has to be a function allowing to "donate" BTC
+to the Bridge without increasing anyone's balances. This function needs to
+validate SPV proof of the donate transaction and update the main wallet UTXO.
+This function should also be used to donate the Bridge to compensate for Bitcoin
+fees burned on moving funds between wallets that failed a heartbeat.
+
+The BTC to donate needs to come from the coverage pool funds. The DAO - or some
+delegate of the DAO - should be able to claim the coverage from the pool and
+manually liquidate tokens to acquire BTC and donate it to the Bridge.
+
 [[transaction-incentives]]
 === Transaction Incentives
 
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -737,6 +737,13 @@ Alphabetized list of Governable Parameters with additional notes.
 - `sweeping_refund_safety_time`: The amount of time prior to when a UTXO
   becomes eligible for a refund where we will not include it in a sweeping
   transaction.
+- `redemption_request_timeout`: The amount of time the wallet has to provide
+  redemption proof.
+- `redemption_request_timeout_slashing_amount`: The amount of stake slashed from
+  each member of a wallet for a timed-out redemption request.
+- `redemption_request_timeout_redemeer_bonus_multiplier`: The percentage of the
+  notifier reward from the staking contract the redeemer receives in case of a
+  redemption timeout.
 - `wallet_creation_period`: How frequently we attempt to create new wallets.
 - `wallet_max_age`: The oldest we allow a wallet to become before we transfer the funds
   to a randomly selected wallet.
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -412,6 +412,51 @@ to submit multiple redemptions. After a redemption, if the wallet has under
 `wallet_min_btc` remaining, it transfers that BTC to a randomly selected wallet
 and closes.
 
+Each redemption request is identified by a concatenation of the wallet's pubkey
+hash and redeemer's output hash (redeemer's BTC address). Such an identifier
+allows retrieving pending redemption requests in a gas-efficient way based on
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-1.adoc
```diff
@@ -401,12 +412,140 @@ to submit multiple redemptions. After a redemption, if the wallet has under
...
+can be improved by the dApp by selecting the previous-oldest wallet without a
+pending redemption request to the given BTC address in case a new redemption is
+requested before the pending one gets cleared out.
+
+At a minimum, pending redemption request needs to capture the following
+information:
+- Expected range of value for redemption UTXO: needed to validate the amount
+  released by the system when processing the redemption proof. Fees are
+  governable and the expected amount redeemed needs to be captured at the moment
+  of the redemption request.
```
michalinacienciala, line 430: > Expected range of value for redemption UTXO

I don't fully understand this, why are we talking about a range here? I'm not sure I get the intention of this paragraph right, but how I understand it is that we want to make sure that the share of the `Y` fee charged from the redeemer will be computed based on the value of `Y` at the moment of redemption request (so that if `Y` got increased by the governance before redemption request got processed, redeemer would not be charged more than he initially anticipated). But I don't see how that corresponds to need for storing an expected range for value of UTXO and how would that range be calculated.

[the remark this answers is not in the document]
pdyraga: Here I am talking about the redemption fee, `fee`. It is mentioned earlier but not listed on the list of governable parameters. I can fix it. This is what you called `Y`.

Also, I think we should have another governable parameter `max_btc_tx_fee`, that will protect against the situation when the wallet decides to spend 99% of the redeemed BTC on Bitcoin network fees.

`X` you mentioned does not exist. Bitcoin network receives its fee from the redeemed Bitcoin, it means that the redemeer receives a bit less BTC than the redeemed amount minus `fee` for the treasury.

pdyraga: Say that I want to redeem 5 TBTC from my balance.

Right now, the redemption fee for the treasury is 0.1 TBTC.

I know that I  will receive no more than 4.9 BTC because 0.1 TBTC was a redemption fee paid to the system.

"No more" than 4.9 BTC because there is Bitoin fee paid to the miners. If that fee is 3 sat, then I will receive `4.9 BTC - 3 sat`.

At the moment of requesting the redemption, we do not know what the Bitcoin miner fee will be. We also do not know if the treasury fee is not going to be altered. That's why we either need to capture expected ranges or capture max Bitcoin fee and treasury fee and calculate if UTXO value (BTC I received) is above the acceptable value or not.

docs/rfc/rfc-1.adoc
```diff
@@ -401,13 +412,151 @@ to submit multiple redemptions. After a redemption, if the wallet has under
...
+not processing individual redemption requests immediately. At the same time, in
+case of a high number of redemption requests, the wallet is incentivized to
+process them in batches to minimize the wait time for Bitcoin and Ethereum chain
+confirmations between redemptions. To minimize the risk for the wallet and to do
+not disincentivize it for waiting for enough redemption requests to accumulate,
+the redemption timeout should be long enough, for example, 72 hours and ideally
+even longer in the early days of the system.
+
+A timed-out redemption request is considered a <<heartbeat,heartbeat>> failure
+and the wallet is ordered to move the BTC to another random wallet. The wallet
```
michalinacienciala, line 462: In this case, that closing of a wallet will happen as a result of `Notify redemption timeout` transaction, not `Notify heartbeat failed` - do I get that right?

docs/rfc/rfc-1.adoc
```diff
@@ -401,13 +412,151 @@ to submit multiple redemptions. After a redemption, if the wallet has under
...
+that the redeemer can not use the same redemption BTC address when there is
+already one redemption request pending from the given wallet. User experience
+can be improved by the dApp by selecting the previous-oldest wallet without a
+pending redemption request to the given BTC address in case a new redemption is
+requested before the pending one gets cleared out.
+
+At a minimum, pending redemption request needs to capture the following
+information:
+
+- Expected range of value for redemption UTXO: needed to validate the amount
```
beaushinkle, line 428: Check for my understanding: Is this a range because the mining fee is a range? As in, say that we're trying to redeem 1 TBTC, the redemption fee is 1% (0.01 TBTC), and the mining fee ranges from 0.0001 BTC to 0.0003 BTC (paid in TBTC). 

We need to validate that the redemption address receives from `1-0.01-0.0001=0.9899` to `1-0.01-0.0001=0.9897` BTC.

michalinacienciala: Yes, exactly as you wrote, it's because we don't know the mining fee. Piotr explained this to me in this comment:
https://github.com/keep-network/tbtc-v2/pull/132#discussion_r803873543

docs/rfc/rfc-1.adoc
```diff
@@ -382,6 +382,17 @@ key than the same `signingGroupPubkey` is fraud.
...
+  redemption proof.
+- `redemption_request_timeout_slashing_amount`: The amount of stake slashed from
+  each member of a wallet for a timed-out redemption request.
+- `redemption_request_timeout_redemeer_bonus_multiplier`: The percentage of the
+  notifier reward from the staking contract the redeemer receives in case of a
+  redemption timeout.
+- `redemption_fraud_slashing_amount`: The amount or stake slashed from each
+  member of a wallet for a redemption fraud.
+- `redempton_fraud_notifier_reward_multiplier`: The percentage of the notifier
+  reward from the staking contract the notifier of redemption fraud receives.
```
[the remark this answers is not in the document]
pdyraga: https://github.com/keep-network/tbtc-v2/pull/132/commits/14bf3cf0403089b102ffaaa37ad84cfc83526fe9

docs/rfc/rfc-1.adoc
pdyraga: Correct!

docs/rfc/rfc-1.adoc
pdyraga: Correct. In practice, we store the current redemption (treasury fee) and max BTC network fee in the `RedemptionRequest` but it is easier to explain in English as an acceptable range given that in the end, we are validating a range calculated based on the captured values of the parameters:

https://github.com/keep-network/tbtc-v2/blob/73fcd4832ad1a1cea43d9826b49f634453337f24/solidity/contracts/bridge/Bridge.sol#L113-L125

https://github.com/keep-network/tbtc-v2/blob/73fcd4832ad1a1cea43d9826b49f634453337f24/solidity/contracts/bridge/Bridge.sol#L1433-L1441

`txMaxFee` should be set to an exceptionally high value, say 1000x expected fee. It is there only to protect against wallets controlled by Bitcoin miners and stealing all the redeeming value in a form of a fee.

pdyraga: One question though: should the redemption fee be a % of the redeemed value of a fixed fee in sat? Right now, the latter is implemented.

pdyraga:
This is ready for another chance!

docs/rfc/rfc-1.adoc
beaushinkle: In my head it was a percentage of redeemed value - that's how I did all of the profitability modeling (otherwise it incentivizes big redemptions and now it's based on number of redemptions rather than amount redeemed)

pdyraga: We will have to update it @lukasz-zimnoch

docs/rfc/rfc-1.adoc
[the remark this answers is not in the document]
michalinacienciala: @pdyraga, could you add that `fee` parameter to the list of governable params in this section and at the end of the doc? And include in the explanation that it's not a flat fee, but %?

docs/rfc/rfc-1.adoc
lukasz-zimnoch: Here it is https://github.com/keep-network/tbtc-v2/pull/136

docs/rfc/rfc-1.adoc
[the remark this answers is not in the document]
pdyraga: Done in https://github.com/keep-network/tbtc-v2/pull/132/commits/bd81af627ffcfea060602ac851a789d26c5e20cc.

pdyraga:
It is ready for another chance!

docs/rfc/rfc-1.adoc
```diff
@@ -382,14 +382,23 @@ key than the same `signingGroupPubkey` is fraud.
...
+  treasury fee.
 - `wallet_min_btc`: The smallest amount of BTC a wallet can hold before we
   attempt to close the wallet and transfer the funds to a randomly selected
   wallet.
 
 To initiate a redemption, a user with a swept balance > `x` supplies a bitcoin
 address. Then, the system calculates the redemption fee `fee`, and releases an
 amount of bitcoin `y` such that `x = y + fee` to the supplied bitcoin address.
-The remaining `fee` sold by the system to buy back `T` tokens (more about this
+The remaining `fee` is sold by the system to buy back `T` tokens (more about this
```
michalinacienciala, line 401: Could we replace `fee` here with `redemption_treasury_fee`?

docs/rfc/rfc-1.adoc
```diff
@@ -401,13 +412,151 @@ to submit multiple redemptions. After a redemption, if the wallet has under
...
+
+If the wallet has still some BTC or if there is a redemption transaction with
+some part of outputs fraudulent and some part of outputs valid, it should be
+possible to submit redemption proof. Because information about fraudulent
+outputs is stored on-chain, these outputs can be skipped by the Bridge and the
+main wallet UTXO can be updated. This mechanism allows proving whatever can
+still be proved and unblocking the wallet by updating its main UTXO.
+
+The wallet is ordered to move the BTC to another random wallet and is no longer
+accepting redemption requests.
```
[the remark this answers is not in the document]
pdyraga: Redesigned in: https://github.com/keep-network/tbtc-v2/pull/135

We can have fraud proofs!

docs/rfc/rfc-1.adoc
```diff
@@ -215,7 +215,7 @@ Governable Parameters
   will trigger an early sweep on a wallet.
 - `dust_threshold`: The minimum bitcoin deposit amount for the transaction to
   be considered for a sweep.
-- `base_sweeping_fee_max`: The highest amount of BTC that operators can
+- `base_btc_fee_max`: The highest amount of BTC that operators can
   initially propose as a fee for miners in a sweeping transaction.
```
michalinacienciala, line 219:
```suggestion
  initially propose as a fee for miners of Bitcoin transaction.
```

docs/rfc/rfc-1.adoc
```diff
@@ -602,11 +729,13 @@ advantage of manipulating btc fee consensus I would be very surprised.
 == Governable Parameters
 Alphabetized list of Governable Parameters with additional notes.
 
-- `base_sweeping_fee_max`: The highest amount of BTC that operators can
-  initially propose as a fee for miners in a sweeping transaction.
+- `base_btc_fee_max`: The highest amount of BTC that operators can
+  initially propose as a fee for miners for Bitcoin transaction.
```
michalinacienciala, line 733:
```suggestion
  initially propose as a fee for miners of Bitcoin transaction.
```

docs/rfc/rfc-1.adoc
```diff
@@ -602,11 +729,13 @@ advantage of manipulating btc fee consensus I would be very surprised.
...
 - `btc_fee_broadcast_timeout`: The amount of time an operator has to provide a
   suggested BTC fee before the other operators give up and try the next
   operator.
+- `btc_fee_max`: The highest amount of BTC that operators can eventually
+  propose as a fee for miners for Bitcoin transaction.
```
michalinacienciala, line 738:
```suggestion
  propose as a fee for miners of Bitcoin transaction.
```

docs/rfc/rfc-1.adoc
```diff
@@ -215,7 +215,7 @@ Governable Parameters
   will trigger an early sweep on a wallet.
 - `dust_threshold`: The minimum bitcoin deposit amount for the transaction to
   be considered for a sweep.
-- `base_sweeping_fee_max`: The highest amount of BTC that operators can
+- `base_btc_fee_max`: The highest amount of BTC that operators can
   initially propose as a fee for miners in a sweeping transaction.
 - `sweeping_fee_bump_period`: The amount of time we wait to see if a sweeping
```
michalinacienciala, line 220: Could we add `btc_fee_max` here as well?

docs/rfc/rfc-1.adoc
```diff
@@ -267,7 +267,7 @@ base_fee * fee_multiplier`. We repeat until either the transaction posts or
 *Note*: We do not allow users to specify a max btc fee. When users deposit,
 they're agreeing to be swept at whatever fee the operators decide is
 appropriate (based on https://blockstream.info/api/fee-estimates). Operators
-cannot pick a starting fee higher than `base_sweeping_fee_max`.
+cannot pick a starting fee higher than `base_btc_fee_max`.
```
michalinacienciala, line 270: Could we mention here that fee cannot be bumped above `btc_fee_max` level?

docs/rfc/rfc-1.adoc
pdyraga: We are in the Governable Parameters of Sweeping section, I intentionally left it this way so that we are more specific the transaction we are talking about is the sweeping transaction.

docs/rfc/rfc-1.adoc
pdyraga: 428af5bc8ee9c07f2a4883d66c5351fea66a3525

docs/rfc/rfc-1.adoc
pdyraga: 428af5bc8ee9c07f2a4883d66c5351fea66a3525

docs/rfc/rfc-1.adoc
michalinacienciala: I see, makes sense.

docs/rfc/rfc-1.adoc
pdyraga: 89cfc7d20db222d71cddd56fdc74c2a8f93a964d

docs/rfc/rfc-1.adoc
pdyraga: c7935fa107dc1cc665e7c18d1a2fea06c2ff81da

docs/rfc/rfc-1.adoc
pdyraga: c7935fa107dc1cc665e7c18d1a2fea06c2ff81da

docs/rfc/rfc-1.adoc
```diff
@@ -272,7 +272,7 @@ cannot pick a starting fee higher than `base_btc_fee_max` and they can never
 choose a fee higher than `btc_fee_max`.
 
 When the transaction clears, and the information has made its way
-over the relay maintainer, then another transaction needs to be created to on
+over the relay maintainer, then another transqaction needs to be created to on
```
michalinacienciala, line 275: This got introduced accidentally, could you fix that as well? :)

pdyraga: Ooops, see https://github.com/keep-network/tbtc-v2/pull/132/commits/552171477ba5f1ba08700c9246497cafb2a1882b - I squashed two last commits with typo fixes.

docs/rfc/rfc-1.adoc
```diff
@@ -382,14 +384,24 @@ key than the same `signingGroupPubkey` is fraud.
 
 Governable Parameters:
 
+- `redemption_request_timeout`: The amount of time the wallet has to provide
+  redemption proof.
+- `redemption_request_timeout_slashing_amount`: The amount of stake slashed from
```
beaushinkle, line 389: What is the unit here? Is it a percentage or raw T tokens?

docs/rfc/rfc-1.adoc
```diff
@@ -401,6 +413,112 @@ to submit multiple redemptions. After a redemption, if the wallet has under
...
+redemption requests, the wallet takes a certain risk on itself by waiting and
+not processing individual redemption requests immediately. At the same time, in
+case of a high number of redemption requests, the wallet is incentivized to
+process them in batches to minimize the wait time for Bitcoin and Ethereum chain
+confirmations between redemptions. To minimize the risk for the wallet and to do
+not disincentivize it for waiting for enough redemption requests to accumulate,
+the redemption timeout should be long enough, for example, 72 hours and ideally
+even longer in the early days of the system.
+
+A timed-out redemption request is considered a <<heartbeat,heartbeat>> failure
```
beaushinkle, line 462: How do we want this to interact with the `consecutive_failed_heartbeats` param?

```
- `consecutive_failed_heartbeats`: The number of times a heartbeat can fail in
  a row that triggers a move to close the wallet.
```

3 good options:

+ it counts for a single failure
+ it counts for a governable number of failures
+ it counts for +Inf failures (immediately moves into closing the wallet)

beaushinkle:
I'm going to get this merged and then handle follow-ups in separate PR's 🚀

beaushinkle (approved):

Issue #118, opened 2022-02-02
Redemption proof


pdyraga:
The redemption will be triggered by a redeemer submitting a `RedemptionRequest` to the `Bridge` consisting of a wallet pubkey hash, redeemed amount, and redeemer's BTC address. `RedemptionRequest` should also capture the current block's timestamp so that it can later be used for notifying about the redemption timeout.

The selected wallet will perform the redemption on Bitcoin bridge by constructing a 1 input (main wallet's UTXO) and >=1 output transaction:
- 1 output when the redemption from the wallet is the last redemption the wallet can handle, that is 0 BTC will be left in the wallet. In this case, the output should point to the redeemer's BTC address
- N > 1 outputs where N-1 is the number of redemption requests handled in the batch; e.g. 2 redemption outputs pointing to redemeers' BTC addresses and 1 output being the new main wallet UTXO and pointing to the wallet pubkey hash.

Once the transaction gains 6 confirmations on Bitcoin, anyone can `submitRedemptionProof` on Ethereum.

`submitRedemptionProof` should accept parameters required to prepare and execute SPV proof on the redemption transaction and to locate the redeeming wallet. First, the redemption TX struct containing version, inputVector, outputVector and locktime. Next params will be the redeeming wallet pubkey hash, probably (to be determined) main UTXO value, and a struct containing SPV proof data. 

`submitRedemptionProof` will then perform a validation of the input. Specifically, it will use wallet pubkey hash passed in params to access the `mainUtxos` mapping (currently called `sweeps`) and get the hash of wallet's main UTXO. The hash is built as `keccak256(txHash | txOutputValue)` (as proposed in https://github.com/keep-network/tbtc-v2/pull/106) and the main UTXO value is passed as parameter. The missing `txHash` will be taken from the transaction single input's outpoint, encoded in the `inputVector`. Having all pieces required to construct the hash, the `submitRedemptionProof` will compute it, and compare with the hash taken from mainUtxos mapping. If they match, further processing is possible. This is safe because by validating the SPV proof we validate that the wallet with the given main UTXO value submitted by the user is the one that actually performed the redemption.

PR #136, opened 2022-02-22
Compute treasury fee as request percentage

lukasz-zimnoch:
Refs: https://github.com/keep-network/tbtc-v2/issues/118

So far, the treasury fee was taken as a fixed satoshi amount from each deposit, regardless of its requested amount. That approach incentivizes big redemptions and does not fit the profitability model. This change updates that behavior and computes the treasury fee as a percentage of each redemption request. The initial value of this parameter is `0.05%` just like in tBTC v1.


tomaszslabon (dismissed):

lukasz-zimnoch:
For reviewers: I recommend looking at the commit history because https://github.com/keep-network/tbtc-v2/pull/136/commits/476f8af636e0f1a98c118db16a1348769910a879 introduces a lot of noise due to Prettier changes introduced in https://github.com/keep-network/tbtc-v2/pull/158.

solidity/contracts/bank/Bank.sol
```diff
@@ -222,22 +222,23 @@ contract Bank is Ownable {
         );
         require(v == 27 || v == 28, "Invalid signature 'v' value");
 
-        bytes32 digest = keccak256(
```
pdyraga, line 225: These changes look weird because it's exactly what was altered by linter in @nkuba' PR #158. I don't say we should block this PR but we need to figure out what is happening so that we do not start linting back and forth between PRs.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -1065,7 +1071,13 @@ contract Bridge is Ownable {
             "There is a pending redemption request from this wallet to the same address"
         );
 
-        uint64 treasuryFee = redemptionTreasuryFee;
+        // No need to check whether `amount - treasuryFee - txMaxFee > 0`
+        // since the `redemptionDustThreshold` should force that condition
+        // to be always true.
+        // TODO: Is it okay to lose the precision here?
```
pdyraga, line 1077: Solidity will truncate (round down) the fee but we subtract it from the requested `amount` to calculate `pendingRedemptionsValue` so I think we are not losing any dust. Some rounding would need to happen anyway. Am I missing something?

solidity/test/bridge/Bridge.test.ts
```diff
@@ -54,6 +54,10 @@ const fixture = async () => {
   await bank.updateBridge(bridge.address)
   await bridge.connect(deployer).transferOwnership(governance.address)
 
+  // Set the redemption dust threshold to 0.001 BTC, i.e. 10x smaller than
+  // the initial value in the Bridge in order to save test Bitcoins.
+  await bridge.setRedemptionDustThreshold(100000)
```
pdyraga, line 59: I am not following. Why can't we operate on existing numbers, BTC transactions we used for tests already happened.

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: Just an old TODO I left at the beginning and forgot to remove it at the end. Removed it https://github.com/keep-network/tbtc-v2/pull/136/commits/5657086ce3a129119844fcf54eb46017bd3fa0e5

solidity/test/bridge/Bridge.test.ts
lukasz-zimnoch: There are two reasons. First, we have a very limited amount of testnet BTC and we cannot easily recover everything so it is far easier to operate on smaller amounts. Second, some BTC transactions are used in multiple scenarios where different requests are made against. Having a smaller dust threshold makes reusing easier because the possible amount range is wider. This is especially useful in next PRs https://github.com/keep-network/tbtc-v2/pull/153 and https://github.com/keep-network/tbtc-v2/pull/155

pdyraga (approved):

PR #153, opened 2022-02-22
Remaining unit tests for `requestRedemption`

lukasz-zimnoch:
Refs: https://github.com/keep-network/tbtc-v2/issues/118
Depends on: #136 

This change covers all test scenarios for `requestRedemption` function.


solidity/test/bridge/Bridge.test.ts
```diff
@@ -1493,164 +1493,323 @@ describe("Bridge", () => {
...
+              "0x220020ef0b4d985752aa5ef6243e4c6f6bebc2a007e7d671ef27d4b1d0db8dcc93bc1c"
+            const redeemerOutputScriptP2PKH =
+              "0x1976a914f4eedc8f40d4b8e30771f792b065ebec0abaddef88ac"
+            const redeemerOutputScriptP2SH =
+              "0x17a914f4eedc8f40d4b8e30771f792b065ebec0abaddef87"
+
             context(
               "when redeemer output script does not point to the wallet public key hash",
               () => {
                 context("when amount is not below the dust threshold", () => {
```
pdyraga, line 1509: s/not below/above ?

solidity/test/bridge/Bridge.test.ts
```diff
@@ -1659,16 +1818,63 @@ describe("Bridge", () => {
...
+                    // for tests is 100000. A value lower by 1 sat should
+                    // trigger the tested condition.
+                    await expect(
+                      bridge
+                        .connect(thirdParty)
+                        .requestRedemption(
+                          walletPubKeyHash,
+                          mainUtxo,
+                          redeemerOutputScriptP2WPKH,
+                          99999
```
pdyraga, line 1875: We could export constants from the fixture, [just like we do in ecdsa tests](https://github.com/keep-network/keep-core/blob/b0acf0e54eb7d82c6ab339d0d09f0c542c2649ea/solidity/ecdsa/test/fixtures/index.ts#L19-L31). Then, we'd import constants from here and reference them.

pdyraga, line 1875: We could export constants from the fixture, [just like we do in ecdsa tests](https://github.com/keep-network/keep-core/blob/b0acf0e54eb7d82c6ab339d0d09f0c542c2649ea/solidity/ecdsa/test/fixtures/index.ts#L19-L31). Then, we'd import constants from here and reference them.

pdyraga (approved):

solidity/test/bridge/Bridge.test.ts
lukasz-zimnoch: Not quite. The "not below" is used because it should be greater or equal.

Issue #137, opened 2022-02-22
Deposit fee


pdyraga:
We should have a governable (possibly 0) deposit % fee, cut during the sweep proof. The fee should go to the same treasury as the redemption fee.

PR #160, opened 2022-03-07
Deposit treasury fee

lukasz-zimnoch:
Closes: #137 

This change introduces a deposit treasury fee which works in the same way as the existing redemption fee. This is achieved using the governable `depositTreasuryFeeDivisor` parameter that determines how much of the deposited amount should be taken as a treasury fee. Since that parameter is governable and may change over time, the actual fee value is computed and stored upon deposit reveal. During the sweep proof submission, the fee is cut from each depositor's balance and the total collected fee is used to increase the treasury balance accordingly. This changeset also involves some refactoring intended to reduce the number of declared variables in order to avoid `stack too deep` error fired by the Solidity compiler.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -101,10 +101,38 @@ contract Bridge is Ownable {
         // Address of the Bank vault the deposit is routed to.
         // Optional, can be 0x0.
         address vault;
+        // Treasury TBTC fee in satoshi at the moment of deposit reveal.
+        uint64 treasuryFee;
         // UNIX timestamp the deposit was swept at. Note this is not the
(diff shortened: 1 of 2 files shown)


solidity/contracts/bridge/Bridge.sol
```diff
@@ -101,12 +101,40 @@ contract Bridge is Ownable {
         // Address of the Bank vault the deposit is routed to.
         // Optional, can be 0x0.
         address vault;
+        // Treasury TBTC fee in satoshi at the moment of deposit reveal.
+        uint64 treasuryFee;
```
pdyraga, line 105: Wow, we had exactly 512-(160+64+32+160+32) = 64 bits left. Lucky us!

pdyraga (approved):

PR #162, opened 2022-03-09
Deposit transaction maximum fee

lukasz-zimnoch:
This change implements a mechanism allowing to limit the maximum transaction fee incurred by each deposit being part of a sweep transaction. Transactions that burn too much value for fees are not accepted as proper sweeps and can be successfully challenged as frauds. This fact should force wallets to use decent fees for sweep transactions and discourage fraudulent behavior.


tomaszslabon (approved):

pdyraga (approved):

pdyraga:
Non-blocking comments about initial value. Let's revisit it.

Issue #150, opened 2022-02-22
Deal with precision loss when validating the sweep proof


pdyraga:
We need to deal with the precision loss in `submitSweepProof` function [when calculating Bitcoin network fee share.](https://github.com/keep-network/tbtc-v2/blob/73fcd4832ad1a1cea43d9826b49f634453337f24/solidity/contracts/bridge/Bridge.sol#L609-L613)

See https://github.com/keep-network/tbtc-v2/pull/128#discussion_r800555359

PR #164, opened 2022-03-09
Handle precision loss during sweep transaction fee distribution

lukasz-zimnoch:
Closes: #150 
Depends on: #162 

So far, the sweep transaction fee was distributed over all deposits but the potential remainder was lost due to Solidity rounding. This change mitigates that problem by isolating the indivisible remainder and pushing it towards the last swept deposit determined according to the order of the inputs.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -653,20 +653,34 @@ contract Bridge is Ownable {
         // all deposits.
         uint256 totalTreasuryFee = 0;
 
-        // Calculate the transaction fee per deposit by dividing the total
-        // transaction fee by deposits count. The total fee is just the
-        // difference between inputs amounts sum and the output amount.
-        // TODO: Deal with precision loss by having the last depositor pay
-        //       the higher transaction fee than others if there is a change,
-        //       just like it has been proposed in discussion:
-        //       https://github.com/keep-network/tbtc-v2/pull/128#discussion_r800555359.
-        uint256 txFee = (inputsInfo.inputsTotalValue - sweepTxOutputValue) /
-            inputsInfo.depositedAmounts.length;
+        // Determine the transaction fee that should be incurred by each deposit
+        // and the indivisible remainder that should be additionally incurred
+        // by the last deposit.
+        (
+            uint256 depositTxFee,
+            uint256 depositTxFeeRemainder
+        ) = sweepTxFeeDistribution(
+                inputsInfo.inputsTotalValue,
+                sweepTxOutputValue,
+                inputsInfo.depositedAmounts.length
+            );
 
-        require(txFee <= depositTxMaxFee, "Transaction fee is too high");
+        // Make sure the highest value of the deposit transaction fee does not
+        // exceed the maximum value limited by the governable parameter.
+        require(
+            depositTxFee + depositTxFeeRemainder <= depositTxMaxFee,
+            "Transaction fee is too high"
+        );
 
         // Reduce each deposit amount by treasury fee and transaction fee.
         for (uint256 i = 0; i < inputsInfo.depositedAmounts.length; i++) {
+            // The last deposit should incur the deposit transaction fee
+            // remainder.
+            uint256 depositTxFeeIncurred = i ==
+                inputsInfo.depositedAmounts.length - 1
+                ? depositTxFee + depositTxFeeRemainder
+                : depositTxFee;
+
             // There is no need to check whether
             // `inputsInfo.depositedAmounts[i] - inputsInfo.treasuryFees[i] - txFee > 0`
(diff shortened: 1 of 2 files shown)


lukasz-zimnoch:
Ready for review but do not merge until #162 lands on main.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -653,28 +653,42 @@ contract Bridge is Ownable {
...
+        );
 
         // Reduce each deposit amount by treasury fee and transaction fee.
         for (uint256 i = 0; i < inputsInfo.depositedAmounts.length; i++) {
+            // The last deposit should incur the deposit transaction fee
+            // remainder.
+            uint256 depositTxFeeIncurred = i ==
+                inputsInfo.depositedAmounts.length - 1
+                ? depositTxFee + depositTxFeeRemainder
+                : depositTxFee;
```
tomaszslabon, line 682: `depositTxFeeRemainder` will be at most `depositedAmounts.length - 1`.
If we assume the number of swept deposits is no greater than `10`, then `depositTxFeeRemainder` will be at most `9`.

So if we have 10 deposits, and tx fee is 999, the cost will spread like this:
99, 99, 99, 99, 99, 99, 99, 99, 99, 108.

9 satoshis paid more is not much and we can live with this.

Alternatively we could:
1) Force the wallet to always choose a fee that is divisible by the number of deposits
2) Or spread the cost more evenly:
  99 100 100 100 100 100 100 100 100 100

 So that the remainder is split evenly among the last depositors. 
 If the remainder is 9, add 1 to the cost of the last 9 depositors.
 If the remainder is 4, add 1 to the cost of the last 4 depositors.
 If the remainder is 1, add 1 to the cost of the last depositor.

lukasz-zimnoch: Yep, that's right, there are some on-chain alternatives. However, pushing the remainder to the last deposit seems to be the simplest and most gas-effective option at the moment. Currently, 9 satoshi is ~0.00354 USD so I think the additional cost for the last deposit is negligible. We can also implement the "divisible fee" behavior in the off-chain client at any moment.

pdyraga: Although I do agree with @tomaszslabon that the two options he proposed are more fair, I also agree with @lukasz-zimnoch that we should choose the least expensive option given the gas consumption. Eventually, it is the depositor who will pay in the treasury fee for the gas consumption of the sweep proof transaction so option 2 could end up as theoretically more fair but practically, depositors would receive less value. Option 1 feels too dangerous to operators, in my opinion. The off-chain signing protocol will be complex enough even without additional constraints on the fee.

solidity/test/bridge/Bridge.test.ts
```diff
@@ -892,30 +892,52 @@ describe("Bridge", () => {
 
                       it("should update the depositors balances", async () => {
                         // The sum of sweep tx inputs is 4148000 satoshi. The output
-                        // value is 4145001 so the fee is 2999. There is 5 deposits
-                        // so 599 satoshi fee should be incurred per deposit.
-                        // Each deposit should also incur the treasury fee whose
-                        // initial value is 0.05% of the deposited amount.
+                        // value is 4145001 so the sweep transaction fee is 2999.
```
pdyraga, line 895: Clear example that `depositTxMaxFee = 1000; // 1000 satoshi` in the constructor does not make sense 😬 

Not in this PR, but can we increase it to 100k? I know there is a TODO to revisit these values but 1000 is definitely not enough.

pdyraga (approved):
Looks good to me, leaving it in @tomaszslabon's hands.

solidity/test/bridge/Bridge.test.ts
lukasz-zimnoch: Yeah this was just an initial value that will be revisited anyway. Also remember `depositTxMaxFee` is max transaction fee per deposit. Not for the entire sweep transaction.

solidity/test/bridge/Bridge.test.ts
```diff
@@ -892,30 +892,52 @@ describe("Bridge", () => {
...
+                        // There are 5 deposits so the fee per deposit is 599
+                        // and the indivisible remainder is 4 which means the
+                        // last deposit should incur 604 satoshi. Worth noting
```
tomaszslabon, line 898: Shouldn't it be 603 instead of 604?

lukasz-zimnoch: Yep, right. Fixed https://github.com/keep-network/tbtc-v2/pull/164/commits/2e8a2b2f9a3a033fe4ddd8f76ba966394d839b8e

tomaszslabon (approved):

PR #174, opened 2022-03-18
Configure git to don't use unauthenticated protocol

michalinacienciala:
We're adding a step that configures git to use `https://` protocol
instead of `git://` when downloading files.
We need this step because the `@keep-network/tbtc` which we update in next
steps has a dependency to `@summa-tx/relay-sol@2.0.2` package, which
downloads one of its sub-dependencies via unathenticated `git://`
protocol. That protocol is no longer supported. See
https://github.blog/2021-09-01-improving-git-protocol-security-github/.

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -58,6 +58,14 @@ jobs:
           cache: "yarn"
           cache-dependency-path: solidity/yarn.lock
 
+      # We need this step because the `@keep-network/tbtc` which we update in
+      # next step has a dependency to `@summa-tx/relay-sol@2.0.2` package, which
+      # downloads one of its sub-dependencies via unathenticated `git://`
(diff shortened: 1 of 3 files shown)


.github/workflows/contracts.yml
```diff
@@ -110,6 +118,14 @@ jobs:
       - name: Install dependencies
         run: yarn install --frozen-lockfile
 
+      # We need this step because the `@keep-network/tbtc` which we update in
```
lukasz-zimnoch, line 121: NB and not this PR but a general comment to that workflow. We are probably referring to the wrong package because `@keep-network/tbtc` is actually v1 which is not used from within v2. We should probably produce a new package called `@keep-network/tbtc-v2`

lukasz-zimnoch (approved):

Issue #139, opened 2022-02-22
Closing wallets


pdyraga:
When ECDSA wallet reached its maximum age or when it holds less BTC than the dust threshold, Bridge should request to move funds to some other wallet and close the wallet that is too old/below the dust threshold.

Additionally, there should be a timeout and slashing in case the wallet refuses to move funds. 

This mechanism needs to be integrated with the heartbeat failure from https://github.com/keep-network/keep-core/issues/2839

PR #172, opened 2022-03-16
Wallets creation

lukasz-zimnoch:
Closes: #138 
Depends on: #159 

The first change introduced by this pull request is the `Wallets` library that will take care of integration with ECDSA wallets. This library will take over all the wallet-related code currently living in the `Bridge` as well as the new code responsible for the wallet lifecycle. The main motivation behind using a library is the need of reducing `Bridge` byte size and achieve better code organization.

Secondly, that PR also introduces the possibility of requesting the creation of new wallets once all conditions described in the RFC 1 are met, registering wallets created by the ECDSA registry, and managing their internal state according to Bridge-related events.

Changes introduced by this PR have the following impact on `Bridge` contract size and the gas costs of their main functions:

| Parameter                    | Before     | After      |
|------------------------------|------------|------------|
| `revealDeposit` cost         | 89931 gas  | 89913 gas  |
| `submitSweepProof` cost      | 254707 gas | 254653 gas |
| `requestRedemption` cost     | 107784 gas | 107428 gas |
| `submitRedemptionProof` cost | 143325 gas | 143100 gas |
| `Bridge` size                | 20.217 KB  | 20.156 KB  |

PR #133, opened 2022-02-10
Fraud proofs

tomaszslabon:
This PR adds a capability of fraud proofs.
Any incorrect way of spending an UTXO is considered a fraud.
If the wallet commits a fraud it can be challenged.
The wallet will have a certain amount of time to defend against this challenge.

The most important part of this PR is extracting UTXO information from the preimage.
The process of constructing the preimage can be seen for example in the `bcoin` library, see:
- [signatureHashV0](https://github.com/bcoin-org/bcoin/blob/ff7e43bfd935b778e8d01017470007b9d609a6af/lib/primitives/tx.js#L467) for non-witness preimage
- [signatureHashV1](https://github.com/bcoin-org/bcoin/blob/ff7e43bfd935b778e8d01017470007b9d609a6af/lib/primitives/tx.js#L637) for witness preimage.

Seeing [BIP-143](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki) can also be helpful.

The cost of added functions:
- `submitFraudChallenge`: `~112k` 
- `defeatFraudChallenge`: `71k`-`95k` (depends on type of preimage and number of inputs, tested with max 6 inputs)
- `notifyFraudChallengeDefeatTimeout`: `~53k`
 

Since storing of the spent main UTXO's hash was added, it increased the gas usage (around `21000` per transaction):
- `submitSweepProof` (tested using the 5-deposit input test): `320681` -> `331290`
- `submitRedemptionProof` (tested with multiple output test): `155240` -> `173251`

PR #176, opened 2022-03-21
Wallets closing

lukasz-zimnoch:
Closes: #139 
Depends on: #172 
Depends on: https://github.com/keep-network/keep-core/pull/2876
Depends on: https://github.com/keep-network/keep-core/pull/2879

Here we introduce the next part of the wallet lifecycle management code, focused on wallet closing. According to RFC 1 and updates from https://github.com/keep-network/tbtc-v2/pull/127, the wallet can be closed due to the following reasons:

- Wallet caused a heartbeat failure
- Wallet became too old or their balance dropped below the minimum BTC amount threshold
- Wallet caused a redemption request timeout

This pull request aims to handle those cases by covering the scope described below.

### Wallet caused a heartbeat failure

As described in https://github.com/keep-network/keep-core/issues/2839, the ECDSA registry will have a mechanism that will allow reporting a failed heartbeat of a wallet and notifying the Bridge about that event. The implementation details of this mechanism are out of this PR's scope though here we handle the Bridge notification part. This is achieved by implementing the `__ecdsaWalletHeartbeatFailedCallback` function introduced by https://github.com/keep-network/keep-core/pull/2876 that will receive the failed heartbeat notification and act accordingly by triggering the wallet moving funds process.

###  Wallet became too old or their balance dropped below the minimum BTC amount threshold

A wallet can naturally reach the end of its life by achieving the maximum age or by dropping its balance below the minimum BTC amount threshold. This case is handled by this pull request by exposing the `notifyCloseableWallet` function in the `Bridge` that will trigger the moving funds process of a closeable wallet.

###  Wallet caused a redemption request timeout

A wallet can cause a redemption request timeout and such a situation can be reported using the `notifyRedemptionTimeout` function exposed by the Bridge (not yet implemented but specified in a TODO). To handle that case, this pull request introduces the `notifyRedemptionTimedOut` exposed by the `Wallets` library that should be called from within the future `notifyRedemptionTimeout` function. The `notifyRedemptionTimedOut` function will perform all necessary actions (e.g. slashing) and trigger the moving funds process of the wallet that caused the timeout.

### Wallet moving funds process

All three aforementioned cases trigger the wallet moving funds process. This process consists of the following steps:
- Check all case-specific preconditions (e.g. failed heartbeat notification comes from ECDSA registry or closeable wallet is actually closeable)
- Check whether the wallet is actually in `Live` state
- Depending on the wallet balance:
   - If the balance is zero, close the wallet immediately and notify the ECDSA registry about wallet closure. Worth noting that this PR only calls an exposed method of the `IWalletRegistry` and doesn't focus on any implementation details there. See https://github.com/keep-network/keep-core/issues/2864 for details
   - If the balance is above zero, initialize the moving funds process. Worth noting that PR only prepares the moving funds process to be started. Actual implementation will be done in the follow-up PRs.
- If the move funds request refers to the current active wallet, unset the active wallet to allow new wallet creation

### Wallet termination

Wallet termination is not in the scope of this pull request and will be handled separately. Currently, it is in the scope of #133 and #177 

### Gas consumption and contract size

Changes presented in this PR has the following impact on contract size and gas consumption:

| Parameter                                   | Before    | After     |
|---------------------------------------------|-----------|-----------|
| `__ecdsaWalletHeartbeatFailedCallback` cost | Undefined | 45756 gas |
| `notifyCloseableWallet` cost                | Undefined | 42391 gas |
| `Bridge` size                               | 20.156 KB | 20.779 KB |

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -493,7 +493,7 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     /// @notice A callback function that is called by the ECDSA Wallet Registry
     ///         once a wallet heartbeat failure is detected.
     /// @param ecdsaWalletID Wallet's unique identifier.
-    /// @param publicKeyY Wallet's public key's X coordinate.
+    /// @param publicKeyX Wallet's public key's X coordinate.
     /// @param publicKeyY Wallet's public key's Y coordinate.
     /// @dev Requirements:
     ///      - The only caller authorized to call this function is `registry`
@@ -507,7 +507,21 @@ contract Bridge is Ownable, EcdsaWalletOwner {
         wallets.notifyWalletHeartbeatFailed(publicKeyX, publicKeyY);
     }
 
-    // TODO: Documentation.
+    /// @notice Notifies about wallet exhaustion and triggers the
+    ///         wallet closure process.
+    /// @param walletPubKeyHash 20-byte public key hash of the wallet
+    /// @param walletMainUtxo Data of the wallet's main UTXO, as currently
+    ///        known on the Ethereum chain.
+    /// @dev Requirements:
+    ///      - Wallet must not be set as the current active wallet
+    ///      - Wallet must exceed the wallet maximum age OR the wallet BTC
+    ///        balance must be lesser than the minimum threshold. If the latter
+    ///        case is true, the `walletMainUtxo` components must point to the
+    ///        recent main UTXO of the given wallet, as currently known on the
+    ///        Ethereum chain. If the wallet has no main UTXO, this parameter
+    ///        can be empty as it is ignored since the wallet balance is
+    ///        assumed to be zero.
+    ///      - Wallet must be in Live state
     function notifyWalletExhausted(
         bytes20 walletPubKeyHash,
         BitcoinTx.UTXO calldata walletMainUtxo
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -512,8 +512,8 @@ contract Bridge is Ownable, EcdsaWalletOwner {
         wallets.notifyWalletHeartbeatFailed(publicKeyX, publicKeyY);
     }
 
-    /// @notice Notifies about wallet exhaustion and triggers the
-    ///         wallet closure process.
+    /// @notice Notifies that the wallet is either old enough or has too few
+    ///         satoshis left and qualifies to be closed.
     /// @param walletPubKeyHash 20-byte public key hash of the wallet
     /// @param walletMainUtxo Data of the wallet's main UTXO, as currently
     ///        known on the Ethereum chain.
@@ -527,11 +527,11 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     ///        can be empty as it is ignored since the wallet balance is
     ///        assumed to be zero.
     ///      - Wallet must be in Live state
-    function notifyWalletExhausted(
+    function notifyCloseableWallet(
         bytes20 walletPubKeyHash,
         BitcoinTx.UTXO calldata walletMainUtxo
     ) external {
-        wallets.notifyWalletExhausted(walletPubKeyHash, walletMainUtxo);
+        wallets.notifyCloseableWallet(walletPubKeyHash, walletMainUtxo);
     }
 
     /// @notice Gets details about a registered wallet.
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -497,9 +497,9 @@ contract Bridge is Ownable, EcdsaWalletOwner {
 
     /// @notice A callback function that is called by the ECDSA Wallet Registry
     ///         once a wallet heartbeat failure is detected.
-    /// @param ecdsaWalletID Wallet's unique identifier.
-    /// @param publicKeyX Wallet's public key's X coordinate.
-    /// @param publicKeyY Wallet's public key's Y coordinate.
+    /// @param ecdsaWalletID Wallet's unique identifier
+    /// @param publicKeyX Wallet's public key's X coordinate
+    /// @param publicKeyY Wallet's public key's Y coordinate
     /// @dev Requirements:
     ///      - The only caller authorized to call this function is `registry`
     ///      - Wallet must be in Live state
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -298,6 +298,13 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     /// @notice State related with wallets.
     Wallets.Data internal wallets;
 
+    event WalletCreationPeriodUpdated(uint32 newCreationPeriod);
+
+    event WalletBtcBalanceRangeUpdated(
+        uint64 newMinBtcBalance,
+        uint64 newMaxBtcBalance
+    );
+
     event NewWalletRequested();
 
     event NewWalletRegistered(
@@ -375,12 +382,45 @@ contract Bridge is Ownable, EcdsaWalletOwner {
         // TODO: Revisit initial values.
         wallets.init(_ecdsaWalletRegistry);
         wallets.setCreationPeriod(1 weeks);
-        wallets.setBtcBalanceRange(1 * 1e8, 10 * 1e8); // 1 BTC - 10 BTC
+        wallets.setBtcBalanceRange(1 * 1e8, 10 * 1e8); // [1 BTC, 10 BTC]
         wallets.setMaxAge(8 weeks);
     }
 
-    // TODO: Add function `onNewWalletCreated` according to discussion:
-    //       https://github.com/keep-network/tbtc-v2/pull/128#discussion_r809885230
+    /// @notice Updates parameters used by the `Wallets` library.
+    /// @param creationPeriod New value of the wallet creation period
+    /// @param minBtcBalance New value of the minimum BTC balance
+    /// @param maxBtcBalance New value of the maximum BTC balance
+    /// @dev Requirements:
+    ///      - Caller must be the contract owner.
+    ///      - Minimum BTC balance must be greater than zero
+    ///      - Maximum BTC balance must be greater than minimum BTC balance
+    function updateWalletsParameters(
+        uint32 creationPeriod,
+        uint64 minBtcBalance,
+        uint64 maxBtcBalance
+    ) external onlyOwner {
+        wallets.setCreationPeriod(creationPeriod);
+        wallets.setBtcBalanceRange(minBtcBalance, maxBtcBalance);
+    }
+
+    /// @return creationPeriod Value of the wallet creation period
+    /// @return minBtcBalance Value of the minimum BTC balance
+    /// @return maxBtcBalance Value of the maximum BTC balance
+    function getWalletsParameters()
+        external
+        view
+        returns (
+            uint32 creationPeriod,
+            uint64 minBtcBalance,
+            uint64 maxBtcBalance
+        )
+    {
+        creationPeriod = wallets.creationPeriod;
+        minBtcBalance = wallets.minBtcBalance;
+        maxBtcBalance = wallets.maxBtcBalance;
+
+        return (creationPeriod, minBtcBalance, maxBtcBalance);
+    }
 
     /// @notice Allows the Governance to mark the given vault address as trusted
     ///         or no longer trusted. Vaults are not trusted by default.
@@ -418,7 +458,7 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     ///        conditions must be true:
     ///        - The active wallet BTC balance is above the minimum threshold
     ///          and the active wallet is old enough, i.e. the creation period
-    ///           was elapsed since its creation time
+    ///          was elapsed since its creation time
     ///        - The active wallet BTC balance is above the maximum threshold
     function requestNewWallet(BitcoinTx.UTXO calldata activeWalletMainUtxo)
         external
@@ -471,7 +511,7 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     /// @param walletPubKeyHash The 20-byte wallet public key hash (computed
     ///        using Bitcoin HASH160 over the compressed ECDSA public key)
     /// @return Wallet details.
-    function getRegisteredWallet(bytes20 walletPubKeyHash)
+    function getWallet(bytes20 walletPubKeyHash)
         external
         view
         returns (Wallets.Wallet memory)
@@ -482,6 +522,7 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     /// @notice Gets the public key hash of the active wallet.
     /// @return The 20-byte public key hash (computed using Bitcoin HASH160
     ///         over the compressed ECDSA public key) of the active wallet.
+    ///         Returns bytes20(0) if there is no active wallet at the moment.
     function getActiveWalletPubKeyHash() external view returns (bytes20) {
         return wallets.activeWalletPubKeyHash;
     }
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -275,6 +275,8 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     ///           successfully
     ///         - `notifyRedemptionTimeout` in case the request was reported
     ///           to be timed out
+    ///         - `submitRedemptionFraudProof` in case the request was handled
+    ///           in an fraudulent way amount-wise.
     mapping(uint256 => RedemptionRequest) public pendingRedemptions;
 
     /// @notice Collection of all timed out redemptions requests indexed by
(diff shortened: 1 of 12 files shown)


solidity/contracts/bridge/Wallets.sol
```diff
@@ -159,6 +173,14 @@ library Wallets {
         emit WalletBtcBalanceRangeUpdated(minBtcBalance, maxBtcBalance);
     }
 
+    /// @notice Sets the wallet maximum age.
+    /// @param maxAge New value of the wallet maximum age
```
tomaszslabon, line 177: We should put a dot at the end of the @param description (we do that throughout the Bridge.sol file).

lukasz-zimnoch: AFAIK the convention is the opposite. We should rather remove the dots in places where they occur

tomaszslabon: Yes, the convention is the opposite.

solidity/contracts/bridge/Wallets.sol
```diff
@@ -305,6 +327,145 @@ library Wallets {
...
+    ///         zero, the wallet is requested to move their funds first.
+    ///         Otherwise, the wallet is closed immediately and the ECDSA
+    ///         registry is notified about this fact. If the wallet closure
+    ///         request refers to the current active wallet, such a wallet
+    ///         is no longer considered active and the active wallet slot
+    ///         is unset allowing to trigger a new wallet creation immediately.
+    /// @param walletPubKeyHash 20-byte public key hash of the wallet
+    /// @dev Requirements:
+    ///      - Wallet must be in Live state
+    function requestWalletClosure(Data storage self, bytes20 walletPubKeyHash)
```
pdyraga, line 431: I think we should rename this function to `moveFunds`. It should simplify state transition: wallet is `Live` and needs to get Closed. OK, the first step is always to `MoveFunds`. If the wallet has no funds to move, we can use a shortcut and mark it as `Closed`. Otherwise, we always wait for the confirmation.

```
[Live] -- moveFunds() --> [MovingFunds] -- notifyFundsMoved() --> [Closed]
[Live] -- moveFunds(); it is already empty!  --> [Closed]
```

solidity/contracts/bridge/Wallets.sol
```diff
@@ -305,6 +327,145 @@ library Wallets {
...
+        Data storage self,
+        bytes20 walletPubKeyHash
+    ) external {
+        // TODO: Perform slashing of wallet operators.
+
+        WalletState walletState = self
+            .registeredWallets[walletPubKeyHash]
+            .state;
+
+        require(walletState != WalletState.Unknown, "Unknown wallet");
```
pdyraga, line 375: Why don't we limit it only to `Live` and `MovingFunds` wallets?

solidity/contracts/bridge/Wallets.sol
```diff
@@ -305,6 +327,145 @@ library Wallets {
...
+    function notifyWalletHeartbeatFailed(
+        Data storage self,
+        bytes32 publicKeyX,
+        bytes32 publicKeyY
+    ) external {
+        require(
+            msg.sender == address(self.registry),
+            "Caller is not the ECDSA Wallet Registry"
+        );
+
```
pdyraga, line 346: Should we add a `require` statement that the wallet needs to be `Live`?

EDIT: I noticed it is checked in `requestWalletClosure` but please see the rest of my comments, I think we should have this `require` here.

solidity/contracts/bridge/Wallets.sol
pdyraga: Actually, this is another argument for renaming `requestWalletClosure` to `moveFunds`. See that the wallet that is `MovingFunds` can still have a redemption timeout and we should only request to `moveFunds` when wallet is in `Live` state.

solidity/contracts/bridge/Wallets.sol
```diff
@@ -305,6 +327,145 @@ library Wallets {
...
+        bytes20 walletPubKeyHash,
+        BitcoinTx.UTXO calldata walletMainUtxo
+    ) external {
+        require(
+            self.activeWalletPubKeyHash != walletPubKeyHash,
+            "Active wallet cannot be considered exhausted"
+        );
+
+        /* solhint-disable-next-line not-rely-on-time */
+        bool walletMaxAge = block.timestamp >=
```
pdyraga, line 408:
```suggestion
        bool walletOldEnough = block.timestamp >=
```

solidity/contracts/bridge/Wallets.sol
```diff
@@ -305,6 +327,145 @@ library Wallets {
...
+            self.registeredWallets[walletPubKeyHash].createdAt + self.maxAge;
+
+        require(
+            walletMaxAge ||
+                getWalletBtcBalance(self, walletPubKeyHash, walletMainUtxo) <
+                self.minBtcBalance,
+            "Wallet exhaustion conditions are not met"
```
pdyraga, line 415:
```suggestion
            "Wallet needs to be old enough or have too few satoshis"
```

solidity/contracts/bridge/Bridge.sol
```diff
@@ -478,13 +495,43 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
+    /// @notice A callback function that is called by the ECDSA Wallet Registry
+    ///         once a wallet heartbeat failure is detected.
+    /// @param ecdsaWalletID Wallet's unique identifier.
+    /// @param publicKeyX Wallet's public key's X coordinate.
+    /// @param publicKeyY Wallet's public key's Y coordinate.
+    /// @dev Requirements:
+    ///      - The only caller authorized to call this function is `registry`
+    ///      - Wallet must be in Live state
     function __ecdsaWalletHeartbeatFailedCallback(
+        /* solhint-disable-next-line no-unused-vars */
```
pdyraga, line 507: Instead of disabling solhint, I think we can do something like:
```
function __ecdsaWalletHeartbeatFailedCallback(
        bytes32,
        bytes32 publicKeyX,
        bytes32 publicKeyY
    ) external override {
```

and then, we should not even need to explain unused `ecdsaWalletID` in `@param`s.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -478,13 +495,43 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
+    ///      - Wallet must not be set as the current active wallet
+    ///      - Wallet must exceed the wallet maximum age OR the wallet BTC
+    ///        balance must be lesser than the minimum threshold. If the latter
+    ///        case is true, the `walletMainUtxo` components must point to the
+    ///        recent main UTXO of the given wallet, as currently known on the
+    ///        Ethereum chain. If the wallet has no main UTXO, this parameter
+    ///        can be empty as it is ignored since the wallet balance is
+    ///        assumed to be zero.
+    ///      - Wallet must be in Live state
+    function notifyWalletExhausted(
```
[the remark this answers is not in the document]
pdyraga: Alternatively, `notifyWalletToClose` or `notifyCloseableWallet`. I think this is easier to reason about - someone noticed there is a wallet that can be closed and they are notifying the `Bridge` about it. There is no additional concept of a wallet being *exhausted*. A wallet can be closed in this and that conditions and that's it.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -478,13 +495,43 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
         bytes32 ecdsaWalletID,
         bytes32 publicKeyX,
         bytes32 publicKeyY
     ) external override {
-        // TODO: Implementation.
+        wallets.notifyWalletHeartbeatFailed(publicKeyX, publicKeyY);
+    }
+
+    /// @notice Notifies about wallet exhaustion and triggers the
+    ///         wallet closure process.
```
pdyraga, line 516: Let's be more specific. My proposition: 

Notifies that the wallet is either old enough or has too few satoshis left and qualifies to be closed.

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: Not sure if I follow, why should we add an additional `require` here?

pdyraga: No, I think we should do the validation of the state here and do not repeat it in `requestWalletClosure` (`closeWallet`). I think this will be way easier to reason about when state the requirements are defined at the very beginning of the transition path.

solidity/contracts/bridge/Wallets.sol
pdyraga: What we discussed with @lukasz-zimnoch out of band:

image

Only LIVE and MOVING_FUNDS wallet can be slashed.
MOVING_FUNDS wallet can have redemption timeouts reported and be slashed for that.
MOVING_FUNDS can not go to CLOSED until all redemptions are completed.

The last point prevents us from slashing the wallet that ended in `CLOSED` or `TERMINATED` states. These two are truly last possible states and no on-chain activity is possible for these wallets.

pdyraga: tl;dr; we should require the state to be LIVE or MOVING_FUNDS to call this function.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -380,42 +390,49 @@ contract Bridge is Ownable, EcdsaWalletOwner {
         wallets.init(_ecdsaWalletRegistry);
         wallets.setCreationPeriod(1 weeks);
         wallets.setBtcBalanceRange(1 * 1e8, 10 * 1e8); // [1 BTC, 10 BTC]
+        wallets.setMaxAge(8 weeks);
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/176/commits/b9819c2775286907163f4fa564dd275bd8ba7567

solidity/contracts/bridge/Bridge.sol
```diff
@@ -478,13 +495,43 @@ contract Bridge is Ownable, EcdsaWalletOwner {
         wallets.registerNewWallet(ecdsaWalletID, publicKeyX, publicKeyY);
     }
 
-    // TODO: Documentation.
+    /// @notice A callback function that is called by the ECDSA Wallet Registry
+    ///         once a wallet heartbeat failure is detected.
+    /// @param ecdsaWalletID Wallet's unique identifier.
+    /// @param publicKeyX Wallet's public key's X coordinate.
+    /// @param publicKeyY Wallet's public key's Y coordinate.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/176/commits/0c37743398e40cde4dc8814de345f5bab299f261

solidity/contracts/bridge/Bridge.sol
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/176/commits/c10e1a1b8f474a801681ec1766337e624364d961

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/176/commits/c10e1a1b8f474a801681ec1766337e624364d961

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/176/commits/c10e1a1b8f474a801681ec1766337e624364d961

solidity/contracts/bridge/Wallets.sol
```diff
@@ -305,6 +327,145 @@ library Wallets {
...
+    ///         (e.g. redemption request timeout) and requests slashing
+    ///         of the wallet operators. Triggers wallet closure process
+    ///         only if the wallet is still in the Live state. That means
+    ///         multiple action timeouts can be reported for the same
+    ///         wallet but only the first report will trigger wallet closure
+    ///         process.
+    /// @param walletPubKeyHash 20-byte public key hash of the wallet
+    /// @dev Requirements:
+    ///      - Wallet must be in a state other than Unknown
+    function notifyWalletActionTimedOut(
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 5a5fab0a

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: See 52889324

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/176/commits/14dd3972d432744920b63f447bcb51361d4dd7ba

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/176/commits/14dd3972d432744920b63f447bcb51361d4dd7ba

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: See 04039625

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: See c7e3a122

solidity/contracts/bridge/Wallets.sol
```diff
@@ -305,6 +327,152 @@ library Wallets {
...
+            "ECDSA wallet must be in Live state"
+        );
+
+        /* solhint-disable-next-line not-rely-on-time */
+        bool walletOldEnough = block.timestamp >=
+            wallet.createdAt + self.maxAge;
+
+        require(
+            walletOldEnough ||
+                getWalletBtcBalance(self, walletPubKeyHash, walletMainUtxo) <
```
pdyraga, line 427: It is how it is described in the RFC so let's not block this PR on it but I think we need to think more cc @beaushinkle. Say that `wallet_min_btc` (here `minBtcBalance`) is set to 10 BTC. It means if the current active wallet is old enough and it has at least 10 BTC we allow for new wallet creation. But later, if the same wallet has just 1 satoshi less, we consider it closeable. Shouldn't we have some additional margin?

solidity/test/bridge/Bridge.Wallets.test.ts
```diff
@@ -705,4 +716,773 @@ describe("Bridge - Wallets", () => {
...
+              )
+            })
+          })
+
+          context("when wallet balance is greater than zero", () => {
+            const walletMainUtxo = {
+              txHash:
+                "0xc9e58780c6c289c25ae1fe293f85a4db4d0af4f305172f2a1868ddd917458bdf",
+              txOutputIndex: 0,
+              txOutputValue: 1,
```
pdyraga, line 1149: Is it possible to send 1 sat on Bitcoin? I wonder if we shouldn't have some sort of dust threshold for moving funds functionality below which, we just allow to close the wallet with the dust remaining.

pdyraga (approved):

solidity/contracts/bridge/Wallets.sol
```diff
@@ -305,6 +327,152 @@ library Wallets {
...
+
+    /// @notice Handles a notification about a wallet redemption timeout
+    ///         and requests slashing of the wallet operators. Triggers the
+    ///         wallet moving funds process only if the wallet is still in the
+    ///         Live state. That means multiple action timeouts can be reported
+    ///         for the same wallet but only the first report requests the
+    ///         wallet to move their funds.
+    /// @param walletPubKeyHash 20-byte public key hash of the wallet
+    /// @dev Requirements:
+    ///      - Wallet must be in Live or MovingFunds state
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yep that's clear but I'll add it once we have redemption timeouts implemented and everything is clear.

solidity/test/bridge/Bridge.Wallets.test.ts
lukasz-zimnoch: Yep, it is possible. You can construct a transaction with multiple outputs that are used to fund the mining fee and add one output with 1 satoshi (or even 0 satoshis). Regarding the dust threshold, this is worth exploring but note that we are not able to easily check the balance in all scenarios that trigger the moving funds process.

solidity/contracts/bridge/Wallets.sol
```diff
@@ -305,6 +327,152 @@ library Wallets {
...
+        bytes32 publicKeyY
+    ) external {
+        require(
+            msg.sender == address(self.registry),
+            "Caller is not the ECDSA Wallet Registry"
+        );
+
+        // Compress wallet's public key and calculate Bitcoin's hash160 of it.
+        bytes20 walletPubKeyHash = bytes20(
+            EcdsaLib.compressPublicKey(publicKeyX, publicKeyY).hash160()
```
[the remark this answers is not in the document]
lukasz-zimnoch: Will change it in the follow-up PR.

solidity/test/bridge/Bridge.Wallets.test.ts
pdyraga: Captured in https://github.com/keep-network/tbtc-v2/issues/180.

solidity/contracts/bridge/Wallets.sol
pdyraga: Captured in https://github.com/keep-network/tbtc-v2/issues/180.

PR #183, opened 2022-03-29
RFC 1: Specks of Dust

beaushinkle:
Fun (for some) philosophical aside: if you had to choose that one random person suffer the worst thing imaginable or for an arbitrarily large number of people (feel free to imagine a future population of humans that is multi-planetary or multi-dimensional) all suffer from having a speck of dust briefly irritate their eyes, which would you pick? If you picked specks of dust, are you thinking of a big enough number of people?

---

This PR first disambiguates `wallet_min_btc` into `wallet_min_closure_btc` and `wallet_min_creation_btc`, not adding any functionality but disentangling the jobs. The same number previously controlled the new wallet creation threshold and the wallet closure threshold, but there was no reason for that to be the case.

Then, we introduce a third param `wallet_dust_leftover` which specifies the smallest amount of bitcoin we're willing to transfer to another wallet. Anything less than that and we just abandon the funds. For concrete examples, think a few thousand sats or a few dollars worth. Transferring it will cost more than the gas to prove the tx, so we just abandon the funds, and reduce our reserve buffer that can be replenished via the `donate` function.

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -651,8 +651,10 @@ Governable Parameters:
 - `wallet_closure_timeout`: The amount of time that a wallet has to
   successfully move its funds to another wallet and inform the ethereum chain
   before it is at risk of punishment.
-- `wallet_max_age`: The oldest we allow a wallet to become before we transfer the funds
-  to a randomly selected wallet.
+- `wallet_dust_leftover`: The smallest amount of btc that we will transfer to
+  another wallet when a wallet closes. Any amount under this is abandoned.
+- `wallet_max_age`: The oldest we allow a wallet to become before we transfer
+  the funds to a randomly selected wallet.
 - `wallet_min_closure_btc`: The smallest amount of btc a wallet can hold before
   we attempt to close the wallet and transfer the funds to a randomly selected
   wallet.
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-1.adoc
```diff
@@ -1010,8 +1015,10 @@ Alphabetized list of Governable Parameters with additional notes.
 - `wallet_creation_period`: How frequently we attempt to create new wallets.
 - `wallet_max_age`: The oldest we allow a wallet to become before we transfer
   the funds to a randomly selected wallet.
-- `wallet_min_btc`: The smallest amount of BTC a wallet can hold before we
-  attempt to close the wallet and transfer the funds to a randomly selected
+- `wallet_min_closure_btc`: The smallest amount of BTC a wallet can hold before
+  we attempt to close the wallet and transfer the funds to a randomly selected
   wallet.
+- `wallet_min_creation_btc`: The minimum amount of BTC an active wallet needs
+  to have before we allow for the creation of a new active wallet.
```
pdyraga, line 1022: We should add `wallet_dust_leftover` to this list.

pdyraga (approved):

docs/rfc/rfc-1.adoc
beaushinkle: [7107234](https://github.com/keep-network/tbtc-v2/pull/183/commits/71072349b551700ab4b7a17a28e14ea5dd936bab)

PR #186, opened 2022-04-01
Handle TODOs in Fraud Proofs

tomaszslabon:
This PR solves TODOs introduced during implementation of Fraud Proofs:
1) Replaced allocating `bytes memory` with `bytes4` during sighash type extraction:
    - the gas usage of `defeatFraudChallenge` decreased `3-4k` 
2) Removes some TODOs doomed as unnecessary
3) Changes order of fraud related functions in the `Bridge`.


lukasz-zimnoch:
Changes look good to me! However, I have three concerns:

1. Using separate X and Y coords for the public key. I recall we were having a discussion about that in the PR that introduced fraud proofs and we haven't achieved an agreement about that: https://github.com/keep-network/tbtc-v2/pull/133#discussion_r834337870. @pdyraga, you were rather against the separation, what's your standpoint now?
2. I see two very important TODOs related to frauds are still not handled: https://github.com/keep-network/tbtc-v2/blob/6abb9d0dbcca4f4b9aaa0904b7568f9ba923b40f/solidity/contracts/bridge/Frauds.sol#L310. Do we plan to have a separate PR for them?
3. Can we take an opportunity and re-organize the fraud code within the Bridge? I see that code is still mixed with redemptions-related functions. I have already done that within https://github.com/keep-network/tbtc-v2/pull/184/commits/11e05638ee1bfeed19efca783305c62dfc59ddaa (PR #184). But, that will cause conflicts. If we handle it here I will just revert those changes in PR 184. WDYT @tomaszslabon ?

tomaszslabon:
> Changes look good to me! However, I have three concerns:
> 
> 1. Using separate X and Y coords for the public key. I recall we were having a discussion about that in the PR that introduced fraud proofs and we haven't achieved an agreement about that: [Fraud proofs #133 (comment)](https://github.com/keep-network/tbtc-v2/pull/133#discussion_r834337870). @pdyraga, you were rather against the separation, what's your standpoint now?
> 2. I see two very important TODOs related to frauds are still not handled: https://github.com/keep-network/tbtc-v2/blob/6abb9d0dbcca4f4b9aaa0904b7568f9ba923b40f/solidity/contracts/bridge/Frauds.sol#L310
>    . Do we plan to have a separate PR for them?
> 3. Can we take an opportunity and re-organize the fraud code within the Bridge? I see that code is still mixed with redemptions-related functions. I have already done that within [11e0563](https://github.com/keep-network/tbtc-v2/commit/11e05638ee1bfeed19efca783305c62dfc59ddaa) (PR [Submit moving funds proof #184](https://github.com/keep-network/tbtc-v2/pull/184)). But, that will cause conflicts. If we handle it here I will just revert those changes in PR 184. WDYT @tomaszslabon ?

1. Is up to @pdyraga.
2. I thought we could handle it in a separate PR.
3. Moved the fraud-related functions ahead of the redemption functions in 1da356d0b4bac0a802c7c1c94157a695d51cb2bd.

pdyraga:
> 1. Using separate X and Y coords for the public key. I recall we were having a discussion about that in the PR that introduced fraud proofs and we haven't achieved an agreement about that: [Fraud proofs #133 (comment)](https://github.com/keep-network/tbtc-v2/pull/133#discussion_r834337870). @pdyraga, you were rather against the separation, what's your standpoint now?

I looked into it again and I think we should leave the previous version (`bytes calldata walletPublicKey`). Fraud functions are not called often - hopefully never - and 3/4k gas units saving is not as important as consistency of API. ECDSA `WalletRegistry. getWalletPublicKey()` [returns uncompressed 64-byte key](https://github.com/keep-network/keep-core/blob/67156f5b2c647aa239125f6d704907592c583e70/solidity/ecdsa/contracts/WalletRegistry.sol#L706-L711) and we expect the same format for DKG result. When DKG result is submitted, we [emit an event](https://github.com/keep-network/keep-core/blob/67156f5b2c647aa239125f6d704907592c583e70/solidity/ecdsa/contracts/WalletRegistry.sol#L99) that contains the [public key as `bytes`](https://github.com/keep-network/keep-core/blob/67156f5b2c647aa239125f6d704907592c583e70/solidity/ecdsa/contracts/libraries/EcdsaDkg.sol#L82). The only two situations when we passed x and y separately, was when `WalletRegistry` talked to `Bridge`. But I think it is fine - for internal communication we can always use separate coordinates and for the public API always `bytes`.

pdyraga (approved):

PR #187, opened 2022-04-05
Fix typo in conditions for challenging intention-to-move tx

michalinacienciala:
There was one typo in #185.

When challenge of intention_to_move transaction is posted, we should
check if there are "better" wallets which wallet could choose at the time
when it was posting the intention_to_move transaction.


michalinacienciala:
I'm closing this PR without merging - we decided to change the approach and drop the challenging of the choice of wallets. The section will need to be updated and the sentence I was trying to update will not be there anymore.

PR #189, opened 2022-04-08
Check wallet state during submit sweep proof

tomaszslabon:
This PR adds a check of the wallet state in `submitSweepProof` (only `Live` and `MovingFunds` wallets can accept the proof).

### DIFF
--- a/solidity/test/bridge/Bridge.test.ts
+++ b/solidity/test/bridge/Bridge.test.ts
@@ -2162,7 +2162,72 @@ describe("Bridge", () => {
     })
 
     context("when the wallet state is neither Live or MovingFunds", () => {
-      // TODO: Implement
+      const data: SweepTestData = SingleP2SHDeposit
+      const { fundingTx, reveal } = data.deposits[0]
+
+      const testData = [
+        {
+          testName: "when wallet state is Unknown",
+          walletState: walletState.Unknown,
+        },
+        {
+          testName: "when wallet state is Closed",
+          walletState: walletState.Closed,
+        },
+        {
+          testName: "when wallet state is Terminated",
+          walletState: walletState.Terminated,
+        },
+        // TODO: Implement tests for state `Closed` when it's added
+      ]
+
+      testData.forEach((test) => {
+        context(test.testName, () => {
+          before(async () => {
+            await createSnapshot()
+
+            // Initially set the state to Live, so that the deposit can be revealed
+            await bridge.setWallet(reveal.walletPubKeyHash, {
+              ecdsaWalletID: ethers.constants.HashZero,
+              mainUtxoHash: ethers.constants.HashZero,
+              pendingRedemptionsValue: 0,
+              createdAt: await lastBlockTime(),
+              moveFundsRequestedAt: 0,
+              state: walletState.Live,
+            })
+
+            await relay.setCurrentEpochDifficulty(data.chainDifficulty)
+            await relay.setPrevEpochDifficulty(data.chainDifficulty)
+
+            await bridge.revealDeposit(fundingTx, reveal)
+
+            // Simulate the wallet's state has changed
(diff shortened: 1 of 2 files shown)


solidity/test/bridge/Bridge.test.ts
```diff
@@ -968,1024 +1023,1171 @@ describe("Bridge", () => {
...
+        },
+        {
+          testName: "when wallet state is Closed",
+          walletState: walletState.Closed,
+        },
+        {
+          testName: "when wallet state is Terminated",
+          walletState: walletState.Terminated,
+        },
+        // TODO: Implement tests for state `Closed` when it's added
```
lukasz-zimnoch, line 2181: This is about `Closing` state, not `Closed`. However, I think we can drop that TODO anyway. If we'll introduce the `Closing` state, we will be forced to revisit all test cases like that anyway.

solidity/test/bridge/Bridge.test.ts
```diff
@@ -968,1024 +1023,1171 @@ describe("Bridge", () => {
...
-        // deposits in same sweep batch should have the same value
-        // of that field.
-        const { walletPubKeyHash } = data.deposits[0].reveal
+      })
+    })
 
-        before(async () => {
-          await createSnapshot()
+    context("when the wallet state is MovingFunds", () => {
+      // The execution of `submitSweepProof` is the same for wallets in
```
lukasz-zimnoch, line 2063: I'd even remove the specific assertions altogether and leave just one "should succeed" that will make sure the function doesn't revert when called for a `MovingFunds` wallet.

solidity/test/bridge/Bridge.test.ts
```diff
@@ -968,1024 +1023,1171 @@ describe("Bridge", () => {
...
+        // Simulate the wallet's state has changed to MovingFunds
+        const wallet = await bridge.getWallet(reveal.walletPubKeyHash)
+        await bridge.setWallet(reveal.walletPubKeyHash, {
+          ecdsaWalletID: wallet.ecdsaWalletID,
+          mainUtxoHash: wallet.mainUtxoHash,
+          pendingRedemptionsValue: wallet.pendingRedemptionsValue,
+          createdAt: wallet.createdAt,
+          moveFundsRequestedAt: wallet.moveFundsRequestedAt,
+          state: walletState.MovingFunds,
         })
```
lukasz-zimnoch, line 2098: That can be simplified to:
```
        await bridge.setWallet(reveal.walletPubKeyHash, {
          ...wallet,
          state: walletState.MovingFunds,
        })
```

lukasz-zimnoch (commented):
Just three comments! Otherwise, looks good to me and leaving the last say for @pdyraga

solidity/test/bridge/Bridge.test.ts
tomaszslabon: Removed the TODO in b90c534b4001e7a99cce681451174b1eee74f5fc.

solidity/test/bridge/Bridge.test.ts
tomaszslabon: I added this in 15a2d96b8d7e39b8204ae9c3e818259a57782b94:
```      
it("should succeed", async () => {
        await expect(
          bridge.submitSweepProof(
            data.sweepTx,
            data.sweepProof,
            data.mainUtxo
          )
        ).not.to.be.reverted
 })
 ```

solidity/test/bridge/Bridge.test.ts
tomaszslabon: Done in cbbe1bd2097ee552daf4e4016a7e35fe7620bfe3.

lukasz-zimnoch (approved):

pdyraga (approved):

Issue #178, opened 2022-03-25
Rename `tbtc-ts` to `typescript`


pdyraga:
We want to rename `tbtc-v2/tbtc-ts` to `tbtc-v2/typescript` to keep it aligned with `/solidity`. The package should be published as `@keep-network/tbtc-v2.ts` (or `tbtc.ts` starting with v2 version: still to be decided)

lukasz-zimnoch:
Maybe we should publish it as `@keep-network/tbtc-v2.ts`? The `@keep-network/tbtc.ts` may be confusing as it looks like the TypeScript version of `tbtc.js` for v1.

pdyraga:
> Maybe we should publish it as `@keep-network/tbtc-v2.ts`? The `@keep-network/tbtc.ts` may be confusing as it looks like the TypeScript version of `tbtc.js` for v1.

Yes, good point! I'll edit the issue description.

nkuba:
I prefer not including `v2` in the package name and start versioning with `2.0.0`. We did similar for [@keep-network/random-beacon](https://github.com/keep-network/keep-core/blob/4d99ad4089c259d259ef8cb2bd061dbc1d6aa796/solidity/random-beacon/package.json#L2-L3) and [@keep-network/ecdsa](https://github.com/keep-network/keep-core/blob/4d99ad4089c259d259ef8cb2bd061dbc1d6aa796/solidity/ecdsa/package.json#L2-L3).

We should also align [`@keep-network/tbtc`](https://github.com/keep-network/tbtc-v2/blob/31003503c9e83d24a2ec34e71f158e9948fe1e9a/solidity/package.json#L2-L3) with this convention.

lukasz-zimnoch:
> I prefer not including v2 in the package name and start versioning with 2.0.0. We did similar for [@keep-network/random-beacon](https://github.com/keep-network/keep-core/blob/4d99ad4089c259d259ef8cb2bd061dbc1d6aa796/solidity/random-beacon/package.json#L2-L3) and [@keep-network/ecdsa](https://github.com/keep-network/keep-core/blob/4d99ad4089c259d259ef8cb2bd061dbc1d6aa796/solidity/ecdsa/package.json#L2-L3).
We should also align [@keep-network/tbtc](https://github.com/keep-network/tbtc-v2/blob/31003503c9e83d24a2ec34e71f158e9948fe1e9a/solidity/package.json#L2-L3) with this convention.

[...]

PR #191, opened 2022-04-11
Rename tBTC v2 bindings library to `typescript`

lukasz-zimnoch:
Closes: #178 

This PR renames `tbtc-ts` to `typescript` and set the NPM package name to `@keep-network/tbtc.ts` with initial version `2.0.0-dev`.


nkuba (approved):

tomaszslabon (approved):

typescript/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@keep-network/tbtc-v2-ts",
-  "version": "0.1.0",
+  "name": "@keep-network/tbtc.ts",
+  "version": "2.0.0-dev",
```
pdyraga, line 3: Let's not block on it but consider it as something that we still need to decide: I commented about it [in the issue](https://github.com/keep-network/tbtc-v2/issues/178#issuecomment-1096264677) and I think the package name should be `tbtc-v2` and the version should be `1.0.0-dev`.

PR #192, opened 2022-04-12
Disable slither job

michalinacienciala:
There is a bug in slither that causes the slither command to return an
error (https://github.com/crytic/slither/issues/1106). Bug is already
marked as closed, but a package with the fix hasn't been released yet.
We're temporarily disabling the slither job, until a new release is
availible.


.github/workflows/contracts.yml
```diff
@@ -262,51 +262,53 @@ jobs:
...
-      - name: Configure git to don't use unauthenticated protocol
-        run: git config --global url."https://".insteadOf git://
-
-      - name: Install dependencies
-        run: yarn install
-
-      - name: Run Slither
-        run: slither --hardhat-artifacts-directory build .
+  # The check commented out because it was bringing more problems than value.
+  # It was raising a lot of false flags, which had to be individually supressed.
```
pdyraga, line 266: I would rather say we are commenting it out until https://github.com/crytic/slither/issues/1106 is resolved and the fix is released. I think Slither brings *some* value but it's just not working at the moment. We need a fix, then we can reconsider.

michalinacienciala: Ok, I've forced-pushed a change - updated the comment and commit message.

pdyraga: Can we say the same in the commit message and in this comment (line 265/266). Right now they are conflicting with each other.

michalinacienciala: Sorry, that was a mistake with me not committing the change and only force-pushing the amended commit message 🤦 ... Should be good now.

pdyraga (approved):

PR #196, opened 2022-04-13
Refactor one: Extract deposit reveal and bridge state into libraries

pdyraga:
A Refactoring of the Bridge contract is required given the contract
size constraints. This change is the first step on this journey.
Deposit reveal functionality is extracted to a separate, externally
linked library. Bridge state is also extracted to a separate structure
to allow passing it easily between external libraries. The state - at
least for now - remains in the Bridge contract and the library does
not have to be deployed separately from the Bridge contract.

Bridge contract size comparison before (main) and after:
```
|  Bridge               ·     27.528  ·       27.528  │
|  Bridge               ·     25.187  ·        2.341  │
```

Gas costs comparison before (main) and after:
```
|  revealDeposit          ·    91842  ·    94484  ·     92493  │
|  revealDeposit          ·    94028  ·    96659  ·     94690  │

|  submitSweepProof       ·   181754  ·   336087  ·    255233  │
|  submitSweepProof       ·   181743  ·   336100  ·    255234  │

|  submitRedemptionProof  ·   117746  ·   194802  ·    163500  │
|  submitRedemptionProof  ·   117680  ·   194572  ·    163334  │
```


solidity/contracts/bridge/Bridge.sol
```diff
@@ -2288,4 +2097,52 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
+    ///         The fundingTxHash is bytes32 (ordered as in Bitcoin internally)
+    ///         and fundingOutputIndex an uint32. This mapping may contain valid
+    ///         and invalid deposits and the wallet is responsible for
+    ///         validating them before attempting to execute a sweep.
+    function deposits(uint256 depositKey)
+        external
+        view
+        returns (Deposit.Request memory)
+    {
+        // TODO: rename to getDeposit?
```
lukasz-zimnoch, line 2145: Yep, I think it's a good idea since we fetch a single deposit here.

solidity/contracts/bridge/BridgeState.sol
```diff
@@ -0,0 +1,54 @@
...
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+
+pragma solidity ^0.8.9;
+
+import "./Deposit.sol";
+
+library BridgeState {
```
lukasz-zimnoch, line 20: Can we drop a short `@notice` explaining the purpose of this library?

solidity/contracts/bridge/Deposit.sol
```diff
@@ -0,0 +1,247 @@
...
+pragma solidity ^0.8.9;
+
+import {BTCUtils} from "@keep-network/bitcoin-spv-sol/contracts/BTCUtils.sol";
+import {BytesLib} from "@keep-network/bitcoin-spv-sol/contracts/BytesLib.sol";
+
+import "./BitcoinTx.sol";
+import "./BridgeState.sol";
+import "./Wallets.sol";
+
+library Deposit {
```
lukasz-zimnoch, line 25: Can we drop a short @notice explaining the purpose of this library?

solidity/contracts/bridge/Deposit.sol
```diff
@@ -0,0 +1,247 @@
...
+    ///        the P2(W)SH BTC deposit transaction,
+    ///      - `reveal.refundLocktime` must be the refund locktime used in the
+    ///        P2(W)SH BTC deposit transaction,
+    ///      - BTC deposit for the given `fundingTxHash`, `fundingOutputIndex`
+    ///        can be revealed only one time.
+    ///
+    ///      If any of these requirements is not met, the wallet _must_ refuse
+    ///      to sweep the deposit and the depositor has to wait until the
+    ///      deposit script unlocks to receive their BTC back.
+    function revealDeposit(
```
lukasz-zimnoch, line 124: Maybe just `reveal` since we're already in `Deposit.sol` context? Non-blocking though.

lukasz-zimnoch (approved):
LGTM! Just a couple of tiny comments!

solidity/contracts/bridge/Bridge.sol
pdyraga: Yeah, though we will later (in the next PRs) add functions such as `function pendingRedemptions(uint256 redemptionKey)` and `function spentMainUTXOs(uint256 utxoKey)` so we would need to add `get*` to all of them for consistency and I am not sure which option is better. Let's decide at the very end once we'll have all functions there.

solidity/contracts/bridge/BridgeState.sol
pdyraga: Sure, let's do it in the final PR. I need to see the full picture because some functionalities are reused between libraries (see outbound transactions).

solidity/contracts/bridge/Deposit.sol
pdyraga: The problem is with `DepositRequest` and `RedemptionRequest`. Ideally, they would be used as `Deposit.Request` and `Redemption.Request`. Unfortunately, TypeScript binding generator is getting crazy when there are two `Request`s in the context of `Bridge` contract and it is generating bindings only for one of them. I suspect we are going to have similar problems with Go bindings. In this PR, `DepositRequest` was renamed to just `Request` but I think I'll rename it back in the PR I am currently working on. I realized this is a problem when working on redemption library in #202.

solidity/contracts/bridge/Deposit.sol
pdyraga: Let me do it in a final PR, please. I want to have all libraries split so that I can see the code reused as well (see the previous comment).

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: Sure! I'll keep this convo open for visibility.

solidity/contracts/bridge/Deposit.sol
lukasz-zimnoch: Thanks for the clarification! Wasn't aware of the issues with the TS generator. Let's leave it as is then.

pdyraga: See https://github.com/keep-network/tbtc-v2/pull/206 please.

Issue #122, opened 2022-02-02
Integrate Typescript library with Electrum


pdyraga:
Right now the Typescript library can assemble P2SH deposit transaction, assemble sweep transaction, and assemble sweep proof. This code is not yet integrated with Electrum and some manual wiring needs to be done when interacting with Bitcoin testnet. This issue aims at integrating the library with electrum/electrs.

PR #197, opened 2022-04-13
Electrum-based implementation of the Bitcoin client

lukasz-zimnoch:
Closes: #122 

This pull request introduces an implementation of the Bitcoin client, based on the Electrum server.


tomaszslabon:
@lukasz-zimnoch The reason why we added this `electrum`-based implementation of the `BitcoinClient` is to enable us to write system tests, right?

lukasz-zimnoch:
> @lukasz-zimnoch The reason why we added this `electrum`-based implementation of the `BitcoinClient` is to enable us to write system tests, right?

That's the first reason. It will be also needed for the dApp and CLI tools.

tomaszslabon (approved):

PR #198, opened 2022-04-13
Refactor two: Extract logic resolving sweeping wallet to separate function, make Bridge fields mutable

pdyraga:
~~Depends on #196~~

The part of the logic resolving the wallet, validating its state and
validating the main wallet UTXO has been moved to a separate function.
This is needed to avoid "stack too deep" problems when refactoring the
code moving sweeping functionality to a separate library, out of the
`Bridge` contract (will be done in the next PR).

Gas usage increased a little (~164 units worst case):
```
|  submitSweepProof       ·   181754  ·   336087  ·    255233  │
|  submitSweepProof       ·   181884  ·   336251  ·    255380  │
```

Also, made `Bridge` `immutable` fields mutable.
These fields will have to be extracted into `BridgeState` structure to fix the
problem with `Bridge` contract size (in the next PR) so we will no
longer be able to use the `immutable` keyword for them. Also, these fields 
need to become governable at some point. Doing this change in a separate PR
allows to clearly compare the gas costs of library refactoring that will be done
in a separate PR.

Gas before vs gas after:
```
|  revealDeposit           ·    94028  ·    96659  ·    94690  │
|  revealDeposit           ·    94028  ·    96659  ·    94690  │
|  submitSweepProof        ·   181884  ·   336251  ·   255380  │
|  submitSweepProof        ·   190576  ·   344943  ·   264072  │
|  submitRedemptionProof   ·   117680  ·   194572  ·   163334  │
|  submitRedemptionProof   ·   126354  ·   203246  ·   171551  │
```
Gas consumption increased by ~8600 units for `submitSweepProof` and for
`submitRedemptionProof`.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -750,6 +750,10 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     /// @param mainUtxo Data of the wallet's main UTXO, as currently known on
     ///        the Ethereum chain. If no main UTXO exists for the given wallet,
     ///        this parameter is ignored
+    /// @dev Requirements:
+    ///     - Sweeping wallet must be either in Live or MovingFunds state.
+    ///     - If the main UTXO of the sweeping wallet exists in the storage,
+    ///       the passed `mainUTXO` parameter must be equal to the stored one.
     function resolveSweepingWallet(
         bytes20 walletPubKeyHash,
         BitcoinTx.UTXO calldata mainUtxo
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -670,7 +670,7 @@ contract Bridge is Ownable, EcdsaWalletOwner {
         (
             Wallets.Wallet storage wallet,
             BitcoinTx.UTXO memory resolvedMainUtxo
-        ) = resolveWallet(walletPubKeyHash, mainUtxo);
+        ) = resolveSweepingWallet(walletPubKeyHash, mainUtxo);
 
         // Process sweep transaction inputs and extract all information needed
         // to perform deposit bookkeeping.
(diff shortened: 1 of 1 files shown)


pdyraga:
Rebased over `main` given #196 is merged. Undrafting.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -771,6 +742,53 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
 
+    /// @notice Resolves sweeping wallet based on the provided wallet public key
+    ///         hash. Validates the wallet state and current main UTXO, as
+    ///         currently known on the Ethereum chain.
+    /// @param walletPubKeyHash public key hash of the wallet proving the sweep
+    ///        Bitcoin transaction.
+    /// @param mainUtxo Data of the wallet's main UTXO, as currently known on
+    ///        the Ethereum chain. If no main UTXO exists for the given wallet,
+    ///        this parameter is ignored
+    function resolveWallet(
```
lukasz-zimnoch, line 753: That function should be used only during the sweep. What do you think about naming it `resolveSweepingWallet`?

lukasz-zimnoch, line 753: That function has some requirements, can we list them in `@dev` section?
```
/// @dev Requirements:
         - Sweeping wallet must be either in Live or MovingFunds state
         - If the main UTXO of the sweeping wallet exists in the storage, the passed `mainUTXO`
           parameter must be equal to the stored one
```

pdyraga: Good idea. 3fc6675b07f30a32c35de58dabeee9a426ecb92c

pdyraga: Yep. 60114020a0dda22a3210901e720af574312cb946

lukasz-zimnoch (approved):

Issue #193, opened 2022-04-12
Typescript code for creating redemption transaction in Bitcoin


tomaszslabon:
Create a function that assembles Bitcoin transaction for redemptions.

PR #199, opened 2022-04-13
Assemble Bitcoin redemption transaction 

tomaszslabon:
Closes https://github.com/keep-network/tbtc-v2/issues/193.
This PR enables assembling redemption transaction in the Bitcoin network.
Two functions were added:
- `redeemDeposits`
- `createRedemptionTransaction`


typescript/src/redemption.ts
```diff
@@ -0,0 +1,168 @@
...
+   * The amount of Bitcoins in satoshis that is subtracted from the amount in
+   * redemption request and used to pay the treasury fee.
+   * The value should be equal exactly the value of treasury fee in the Bridge
+   * on-chain contract at the time the redemption request was made.
+   */
+  treasuryFee: BigNumber
+}
+
+// TODO: Description
+export async function redeemDeposits(
```
lukasz-zimnoch, line 49: I think we need to modify that function and add integration with the Bridge. A redemption key is built using two things:
- Wallet public key hash
- Redeemer output script

The first one can be derived from the `walletPrivateKey` (taking the public key and then using HASH160 on it). The second one can be derived from the BTC address passed by the redeemer. All other information is held by the Bridge. That means we should:
- Remove the `RedemptionRequest` interface altogether. It will be required in future for requesting redemptions but with more fields so we can safely get rid of it now
- Pass just the `redeemerAddresses` array instead of `redemptionRequests`
- Build redemption requests keys and fetch those redemptions from the Bridge
- Compute the acceptable redemption amount for each redemption to determine the correct output value
- Build and broadcast the transaction

typescript/src/redemption.ts
```diff
@@ -0,0 +1,168 @@
...
+  )
+
+  // Note that `broadcast` may fail silently (i.e. no error will be returned,
+  // even if the transaction is rejected by other nodes and does not enter the
+  // mempool, for example due to an UTXO being already spent).
+  await bitcoinClient.broadcast(transaction)
+}
+
+// TODO: Description
+export async function createRedemptionTransaction(
```
lukasz-zimnoch, line 79: Would be good to expose that function in the main `TBTC` interface as well

tomaszslabon: This function is now [exposed in the TBTC interface](https://github.com/keep-network/tbtc-v2/blob/793b5df4b506d1016d155f4556182479230fd28f/typescript/src/index.ts#L213)

typescript/src/redemption.ts
tomaszslabon: I think I've done it in recent commits.
The `redeemDeposits` function has `redeemerAddresses`, `walletPrivateKey` and `bridge` passed as parameters
and builds the key and gets the pending redemption requests from the `Bridge`.

The `createRedemptionTransaction` function does the assembling of the transaction.

typescript/test/utils/mock-bridge.ts
```diff
@@ -38,4 +44,24 @@ export class MockBridge implements Bridge {
       resolve(this._difficultyFactor)
     })
   }
+
+  getPendingRedemptions(
```
tomaszslabon, line 48: It would be better to name this function `pendingRedemption()`, so that the interface of the `Bridge` is the same as the online contract. However it causes a conflict with the setter `set pendingRedemptions()`.

So I can either keep it as is, or rename the setter to `setPendingRedemption()`, and name the getter `pendingRedemptions()`.

@lukasz-zimnoch What do you think?

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    const redemptionKey = utils.solidityKeccak256(
+      ["bytes20", "bytes"],
+      [walletPubKeyHash, redeemerOutputScript]
+    )
+
+    const pendingRedemption = await bridge.getPendingRedemptions(redemptionKey)
+
+    // If the pending redemption was found in the Bridge, add it to the list of
+    // redemption requests. Otherwise, skip it.
```
tomaszslabon, line 281: If the address does not represent a pending redemption request, we skip it.
Alternatively, we could revert the whole transaction if at leat one redemption is not found in the Bridge.
@lukasz-zimnoch What do you think?

lukasz-zimnoch: Definitely revert. The `submitRedemptionProof` in the `Bridge` is all or nothing. We should do the same here.

typescript/test/utils/mock-bridge.ts
lukasz-zimnoch: Yeah, we should name it `pendingRedemptions` to match the Bridge API. Regarding the setter, maybe we can add `requestRedemption` instead? It's also a Bridge's function and it could add a redemption to the mock's `_pendingRedemptions` array.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+ *      on-chain contract. If the given address (along with the wallet public
+ *      key corresponding to the provided private key) does not represent a
+ *      valid pending redemption, this address will be skipped.
+ * @param bridge The interface to the Bridge on-chain contract.
+ * @param walletPrivateKey The private key of the wallet in the WIF format.
+ * @param redeemerAddresses The addresses that will be the recipients of the
+ *                          redeemed Bitcoins.
+ * @returns The list of redemption requests.
+ */
+async function prepareRedemptionRequests(
```
lukasz-zimnoch, line 258: Let's move that function between `redeemDeposits` and `createRedempionTransaction`. This will make navigation easier since it is used earlier than the functions defined above.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+    compressed: decodedPrivateKey.compressed,
+  })
+
+  return keyRing.getPublicKey().toString("hex")
+}
+
+/**
+ * Prepares a list of redemption requests based on the provided redeemer
+ * addresses and the wallet key. The information on the redemption requests
+ * is acquired by connecting to the provided Bridge contract interface.
```
lukasz-zimnoch, line 246: `The information on the redemption requests is acquired by connecting to the provided Bridge contract interface.` -> `Details of the redemption requests are fetched from the provided Bridge on-chain contract handle`.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+ * @dev It is up to the caller of this function to ensure that each of the
+ *      addresses represent a valid pending redemption request in the Bridge
+ *      on-chain contract. If the given address (along with the wallet public
+ *      key corresponding to the provided private key) does not represent a
+ *      valid pending redemption, this address will be skipped.
```
lukasz-zimnoch, line 251: Let's revert, not skip.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+ * @param bridge The interface to the Bridge on-chain contract.
```
lukasz-zimnoch, line 252: `The interface` -> `The handle`.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+function getPublicKey(privateKey: string): string {
+  const decodedPrivateKey = wif.decode(privateKey)
+
+  const keyRing = new bcoin.KeyRing({
+    privateKey: decodedPrivateKey.privateKey,
+    compressed: decodedPrivateKey.compressed,
+  })
+
+  return keyRing.getPublicKey().toString("hex")
+}
```
lukasz-zimnoch, line 241: Those are helpers of the `prepareRedemptionRequests`. Let's move them below that function.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+ */
+async function prepareRedemptionRequests(
+  bridge: Bridge,
+  walletPrivateKey: string,
+  redeemerAddresses: string[]
+): Promise<RedemptionRequest[]> {
+  const walletPublicKey = getPublicKey(walletPrivateKey)
+
+  const walletPubKeyHash =
+    "0x" + hash160.digest(Buffer.from(walletPublicKey, "hex")).toString("hex")
```
lukasz-zimnoch, line 266: Maybe 
```
`0x${hash160.digest(Buffer.from(walletPublicKey, "hex")).toString("hex")}`
```

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+  return "0x" + scriptLength.toString(16) + scriptAsStr
+}
+
+/**
+ * Returns the Bitcoin public key in the compressed format corresponding to
+ * the provided private key.
+ * @param privateKey - The private key of in the WIF format.
+ * @returns The public key of as a string.
+ */
+function getPublicKey(privateKey: string): string {
```
lukasz-zimnoch, line 232: Maybe `derivePublicKey`

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+  const scriptLength = scriptAsStr.length / 2 // get length in bytes
+
+  return "0x" + scriptLength.toString(16) + scriptAsStr
+}
+
+/**
+ * Returns the Bitcoin public key in the compressed format corresponding to
+ * the provided private key.
+ * @param privateKey - The private key of in the WIF format.
+ * @returns The public key of as a string.
```
lukasz-zimnoch, line 230: Worth noting its format, i.e. it is not prefixed by `0x` and has `02` or `03` prefix since it is compressed.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  return {
+    transactionHex: transaction.toRaw().toString("hex"),
+  }
+}
+/**
+ * Builds a length prefixed output script based on the provided address (P2PKH,
+ * P2WPKH, P2SH or P2WSH).
+ * @param address - Bitcoin address from which the output script will be build
+ * @returns The output script as a string.
```
lukasz-zimnoch, line 216: Worth noting it is `0x`-prefixed.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+ */
+function buildOutputScript(address: string): string {
```
lukasz-zimnoch, line 218: I'd not call it building. The script is derived from the address. So maybe `deriveOutputScript`?

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+  const script = bcoin.Script.fromAddress(address)
+  const scriptAsStr = script.toRaw().toString("hex")
+  const scriptLength = scriptAsStr.length / 2 // get length in bytes
+
+  return "0x" + scriptLength.toString(16) + scriptAsStr
```
lukasz-zimnoch, line 223: Let's do it in the same way as we did for tBTC v1 (see https://github.com/keep-network/tbtc.js/blob/9838c06f21082cc050fab5f1304b95d6fa3db6a7/src/Deposit.js#L866):
```
const rawOutputScript = bcoin.Script.fromAddress(address).toRaw()
return `0x${Buffer.concat([Buffer.from([rawOutputScript.length]), rawOutputScript]).toString("hex")}`
```

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+    throw new Error("There must be at least one request to redeem")
+  }
+
+  const decodedWalletPrivateKey = wif.decode(walletPrivateKey)
+
+  const walletKeyRing = new bcoin.KeyRing({
+    witness: witness,
+    privateKey: decodedWalletPrivateKey.privateKey,
+    compressed: decodedWalletPrivateKey.compressed,
+  })
```
lukasz-zimnoch, line 138: Let's use `createKeyRing` from `deposit.ts`. I think we can also move that function to `bitcoin.ts`. Let's also take an opportunity and use it in `deposit-sweep.ts`.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+    bcoin.Coin.fromTX(
+      bcoin.MTX.fromRaw(mainUtxo.transactionHex, "hex"),
+      mainUtxo.outputIndex,
+      -1
+    ),
+  ]
+
+  const transaction = new bcoin.MTX()
+
+  let txFee = 0
```
lukasz-zimnoch, line 153: `txTotalFee`?

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+  let totalOutputValue = 0
```
lukasz-zimnoch, line 154: `totalOutputsValue`?

lukasz-zimnoch (changes requested):
Here are the comments regarding the code. Will review tests in the next round, once those comments are addressed.

typescript/src/redemption.ts
tomaszslabon: Done in e0e634e06e5a2a14c4c4f574b8625b2d4c8d2664.

typescript/src/bridge.ts
```diff
@@ -4,6 +4,16 @@ import {
...
 
+import { BigNumberish, BigNumber } from "ethers"
+
+export interface PendingRedemption {
+  redeemer: string
+  requestedAmount: BigNumber
+  treasuryFee: BigNumber
+  txMaxFee: BigNumber
+  requestedAt: number
+}
```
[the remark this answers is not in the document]
tomaszslabon: We can do that, but notice these two interfaces are not exactly the same:
* `PendingRedemption` from `bridge.ts` is exactly what is returned from the `Bridge` contract (including field names).  It is unprocessed. The  field `txMaxFee` is the maximum fee share for that particular pending redemption.

* `RedemptionRequest` from `redemption.ts`  contains already processed data that `createRedemptionTransaction` needs - the `feeShare` is the actual fee share that the will be used which does not necessarily equal `txMaxfree` (it can be smaller, for now I just copy `txMaxfree` to `feeShare`, I left a TODO how we can handle it).

typescript/src/bridge.ts
```diff
@@ -27,4 +37,12 @@ export interface Bridge {
    * @returns Proof difficulty factor.
    */
   txProofDifficultyFactor(): Promise<number>
+
+  /**
+   * Gets a pending redemption from the on-chain contract.
+   * @param redemptionKey The redemption key that identifies the pending
+   *        redemption built as keccak256(walletPubKeyHash | redeemerOutputScript).
```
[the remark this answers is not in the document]
tomaszslabon: Done in 61db9faba39eb0acac1108fce641a777653e99ce.

typescript/test/utils/mock-bridge.ts
tomaszslabon: Done in d98dc0c42863be578c7311faa4904cbb46da6eca.

typescript/src/bridge.ts
```diff
@@ -27,4 +37,12 @@ export interface Bridge {
...
+   * @returns Promise with pending redemption.
+   */
+  getPendingRedemptions(redemptionKey: BigNumberish): Promise<PendingRedemption>
```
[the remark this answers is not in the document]
tomaszslabon: Done in d98dc0c42863be578c7311faa4904cbb46da6eca.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  /**
+   * The amount of Bitcoins in satoshis that is subtracted from the amount in
+   * redemption request and used to pay for the transaction.
+   * The value should not be greater than the max fee in the Bridge on-chain
+   * contract at the time the redemption request was made.
+   * The sum of the fee shares of all the output values makes the total fee of
+   * the redemption transaction.
+   */
+  feeShare: BigNumber
```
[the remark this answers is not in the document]
tomaszslabon: Done in 1d1a4061ead1b5dc27709ae910931fb33f537fdd.

typescript/src/redemption.ts
tomaszslabon: Done in e0e634e06e5a2a14c4c4f574b8625b2d4c8d2664 and c42a9c806a64215abd36fde97e5e8d0ace4f6392.

typescript/src/index.ts
```diff
@@ -160,6 +165,57 @@ export interface TBTC {
...
+  /**
+   * Creates a Bitcoin redemption transaction.
+   * The transaction will have a single input (main UTXO) and an output for each
+   * redemption request provided and a change output if the redemption requests
+   * do not consume all the Bitcoins from the main UTXO.
+   * @dev The caller is responsible for ensuring the redemption request list is
+   *      correctly formed:
+   *        - there is at least one redemption
+   *        - the `requestedAmount` in each redemption request is greater than
+   *          the sum of its `feeShare` and `treasuryFee`
```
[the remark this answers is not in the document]
tomaszslabon: Done in 1d1a4061ead1b5dc27709ae910931fb33f537fdd.

typescript/src/redemption.ts
tomaszslabon: Done in 4af4c77ecae3df9614963646301b01eb2295808b.

typescript/src/redemption.ts
tomaszslabon: Done in 17157246593ce9f23251e1e09657604775b5def9.

typescript/src/redemption.ts
tomaszslabon: Done in 17157246593ce9f23251e1e09657604775b5def9.

typescript/src/redemption.ts
tomaszslabon: Done in ed8950381d4f9391cfb1e7f9200b9c876598719e.

typescript/src/redemption.ts
tomaszslabon: Done in 12832ed4b093566712a8201dff0e089379c114b3.

typescript/src/redemption.ts
tomaszslabon: Done in 0e7002cd00ddd49533ead638b3dd42c0fe81ef47.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+ * @param bitcoinClient - The Bitcoin client used to interact with the network
+ * @param bridge - The Interface used to interact with the Bridge on-chain contract
+ * @param walletPrivateKey = The private kay of the wallet in the WIF format
+ * @param mainUtxo - The main UTXO of the wallet. Must be P2(W)PKH
+ * @param redeemerAddresses - The list of redeemer addresses
+ * @param witness - The parameter used to decide the type of the change output.
+ *                  P2WPKH if `true`, P2PKH if `false`
+ * @returns Empty promise.
+ */
+export async function redeemDeposits(
```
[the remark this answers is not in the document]
tomaszslabon: Done in 3f538ae0a0a7b1c493dc15b880e248a00279fcc6.

typescript/src/index.ts
```diff
@@ -160,6 +165,57 @@ export interface TBTC {
...
+   * broadcasting it. The change UTXO resulting from the transaction becomes
+   * the new main UTXO of the wallet.
+   * @dev It is up to the caller to ensure the wallet key and each of the redeemer
+   *      addresses represent a valid pending redemption request in the Bridge.
+   * @param bitcoinClient - The Bitcoin client used to interact with the network
+   * @param bridge - The Interface used to interact with the Bridge on-chain contract
+   * @param walletPrivateKey = The private kay of the wallet in the WIF format
+   * @param mainUtxo - The main UTXO of the wallet. Must be P2(W)PKH
+   * @param redeemerAddresses - The list of redeemer addresses.
+   * @param witness - The parameter used to decide the type of the change output.
```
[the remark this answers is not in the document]
tomaszslabon: Done in 3f538ae0a0a7b1c493dc15b880e248a00279fcc6.

typescript/src/index.ts
```diff
@@ -160,6 +165,57 @@ export interface TBTC {
...
+   * Redeems deposited tBTC by creating a redemption transaction transferring
+   * Bitcoins from the wallet's main UTXO to the redeemer addresses and
+   * broadcasting it. The change UTXO resulting from the transaction becomes
+   * the new main UTXO of the wallet.
+   * @dev It is up to the caller to ensure the wallet key and each of the redeemer
+   *      addresses represent a valid pending redemption request in the Bridge.
+   * @param bitcoinClient - The Bitcoin client used to interact with the network
+   * @param bridge - The Interface used to interact with the Bridge on-chain contract
+   * @param walletPrivateKey = The private kay of the wallet in the WIF format
+   * @param mainUtxo - The main UTXO of the wallet. Must be P2(W)PKH
```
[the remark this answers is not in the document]
tomaszslabon: Done in 3f538ae0a0a7b1c493dc15b880e248a00279fcc6.

typescript/src/index.ts
```diff
@@ -160,6 +165,57 @@ export interface TBTC {
...
+
+  /**
+   * Redeems deposited tBTC by creating a redemption transaction transferring
+   * Bitcoins from the wallet's main UTXO to the redeemer addresses and
+   * broadcasting it. The change UTXO resulting from the transaction becomes
+   * the new main UTXO of the wallet.
+   * @dev It is up to the caller to ensure the wallet key and each of the redeemer
+   *      addresses represent a valid pending redemption request in the Bridge.
+   * @param bitcoinClient - The Bitcoin client used to interact with the network
+   * @param bridge - The Interface used to interact with the Bridge on-chain contract
```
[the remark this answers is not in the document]
tomaszslabon: Done in 3f538ae0a0a7b1c493dc15b880e248a00279fcc6.

typescript/src/index.ts
```diff
@@ -160,6 +165,57 @@ export interface TBTC {
...
     bitcoinClient: BitcoinClient
   ): Promise<void>
+
+  /**
+   * Redeems deposited tBTC by creating a redemption transaction transferring
+   * Bitcoins from the wallet's main UTXO to the redeemer addresses and
+   * broadcasting it. The change UTXO resulting from the transaction becomes
+   * the new main UTXO of the wallet.
+   * @dev It is up to the caller to ensure the wallet key and each of the redeemer
+   *      addresses represent a valid pending redemption request in the Bridge.
```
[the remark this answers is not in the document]
tomaszslabon: Done in 3f538ae0a0a7b1c493dc15b880e248a00279fcc6.

typescript/src/index.ts
```diff
@@ -160,6 +165,57 @@ export interface TBTC {
     bridge: Bridge,
     bitcoinClient: BitcoinClient
   ): Promise<void>
+
+  /**
+   * Redeems deposited tBTC by creating a redemption transaction transferring
```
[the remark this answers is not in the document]
tomaszslabon: Done in 3f538ae0a0a7b1c493dc15b880e248a00279fcc6.

typescript/src/index.ts
```diff
@@ -160,6 +165,57 @@ export interface TBTC {
...
+    mainUtxo: UnspentTransactionOutput,
+    redeemerAddresses: string[],
+    witness: boolean
+  ): Promise<void>
+
+  /**
+   * Creates a Bitcoin redemption transaction.
+   * The transaction will have a single input (main UTXO) and an output for each
+   * redemption request provided and a change output if the redemption requests
+   * do not consume all the Bitcoins from the main UTXO.
```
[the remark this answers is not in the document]
tomaszslabon: Done in ce10a047b696a90439691068db1bd6df86e60f2c.

typescript/src/index.ts
```diff
@@ -160,6 +165,57 @@ export interface TBTC {
...
+   * @dev The caller is responsible for ensuring the redemption request list is
+   *      correctly formed:
+   *        - there is at least one redemption
+   *        - the `requestedAmount` in each redemption request is greater than
+   *          the sum of its `feeShare` and `treasuryFee`
+   *        - the redeemer address in each redemption request is of a standard
+   *          type (P2PKH, P2WPKH, P2SH, P2WSH).
+   * @param walletPrivateKey  - The private key of the wallet in the WIF format
+   * @param mainUtxo - The main UTXO of the wallet. Must be P2(W)PKH
```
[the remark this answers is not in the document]
tomaszslabon: Done in ce10a047b696a90439691068db1bd6df86e60f2c.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+
+/**
+ * Contains information needed to fulfill a redemption request.
+ */
+export interface RedemptionRequest {
+  /**
+   * The address that is the recipient of the redeemed Bitcoins.
+   * The type of the address must be P2PKH, P2WPKH, P2SH or P2WSH.
+   */
+  redeemerAddress: string
```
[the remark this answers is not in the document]
tomaszslabon: I'll handle this one, once we decide what to do with [this comment](https://github.com/keep-network/tbtc-v2/pull/199#discussion_r872350930). If we decide to remove `PendingRedemption` structure, then `redeemerAddress` may mean both ETC and BTC address.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+  /**
+   * The amount of Bitcoins in satoshis that is requested to be redeemed.
+   * The actual value of the output in the Bitcoin transaction will be decreased
+   * by the sum of the fee share and the treasury fee for this particular output.
+   */
+  requestedAmount: BigNumber
+
+  /**
+   * The amount of Bitcoins in satoshis that is subtracted from the amount in
+   * redemption request and used to pay the treasury fee.
```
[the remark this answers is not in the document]
tomaszslabon: Done in e2af1dd67eccd10a61d4f5dca2c3b1d8c7df6413.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,299 @@
...
+  treasuryFee: BigNumber
+
+  /**
+   * The amount of Bitcoins in satoshis that is subtracted from the amount in
+   * redemption request and used to pay for the transaction.
+   * The value should not be greater than the max fee in the Bridge on-chain
+   * contract at the time the redemption request was made.
+   * The sum of the fee shares of all the output values makes the total fee of
+   * the redemption transaction.
+   */
```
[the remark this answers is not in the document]
tomaszslabon: Done in e2af1dd67eccd10a61d4f5dca2c3b1d8c7df6413.

typescript/src/redemption.ts
tomaszslabon: Done in a30b960784890495cef286d93d544fc0220a50d2.

typescript/src/redemption.ts
tomaszslabon: The function for building the wallet public key was removed in f97831fd678ea03de8ffb8e17d6400a1f6f440af.
`createKeyRing` was used instead.

typescript/src/bridge.ts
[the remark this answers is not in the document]
lukasz-zimnoch: Yes and this is wrong because we tie ourselves to the `Bridge` specific implementation, not their interface. If we do that, the current ETH Bridge implementation details will leak into our library thus making the potential cross-chain support much harder.  At this point, we do not need to care what exactly is returned from the Ethereum `Bridge` contract. We must care about building a consistent universal model of all "entities". This is why I propose to stick with `RedemptionRequest` defined in `redemption.ts`. This is our core type representing a redemption request. All implementation-specific flavors will live separately. For example, we could have an `ethereum.ts` file containing a `Bridge` implementation that would use a contract binding and use some internal type to represent a redemption request.

typescript/src/bridge.ts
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/199#discussion_r873496483. The `Bridge` interface defined in `bridge.ts` is not supposed to map 1:1 to the on-chain contract. Anyway, we will need an intermediary layer between them that will handle all the details.

typescript/src/bridge.ts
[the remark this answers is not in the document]
tomaszslabon: OK, it makes sense. Done in 5c98ec9c19f2a3d2681e738493650312002d4c12.
We now use [a single interface](https://github.com/keep-network/tbtc-v2/blob/4712bc31a42f6c172ee6657eecd607781616cfb2/typescript/src/redemption.ts#L17) to represent the redemption returned from the Bridge, as well as the input data to `createRedemptionTransaction` function.

typescript/src/bridge.ts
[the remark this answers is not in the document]
tomaszslabon: Ok, then. Done in https://github.com/keep-network/tbtc-v2/commit/61db9faba39eb0acac1108fce641a777653e99ce.

typescript/src/redemption.ts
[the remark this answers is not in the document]
tomaszslabon: I changed the description of `redeemerAddress` to:
```
  /**
   * The address of the redeemer.
   */
  redeemerAddress: string
```

I wouldn't mention whether the address is `0x` prefixed or not, as this field can represent BTC address which are not hex strings:
```
    // P2PKH address
    "mmTeMR8RKu6QzMGTG4ipA71uewm3EuJng5",
    // P2WPKH address
    "tb1qgycg0ys3c4xlgc8ysnwln2kqp89n3mn5ts7z3l",
    // P2SH address
    "2Mxy76sc1qAxiJ1fXMXDXqHvVcPLh6Lf12C",
    // P2WSH address
    "tb1qs63s8nwjut4tr5t8nudgzwp4m3dpkefjzpmumn90pruce0cye2tq2jkq0y",
```

typescript/src/redemption.ts
```diff
@@ -0,0 +1,291 @@
...
+import { Bridge } from "./bridge"
+
+/**
+ * Represents a redemption request.
+ */
+export interface RedemptionRequest {
+  /**
+   * The address of the redeemer.
+   */
+  redeemerAddress: string
```
lukasz-zimnoch, line 21: Step 1:
I think we should remove that field and introduce two other ones instead:
```
redeemer: Identifier
redeemerOutputScript: string
```
The `redeemerOutputScript` should NOT be prefixed with `0x`. Let's leave prefixing to the `Bridge` implementation.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,291 @@
...
+
+  /**
+   * The amount of Bitcoins in satoshis that is subtracted from the amount of
+   * the redemption request and used to pay the transaction network fee.
+   * The value must not exceed the value of the Bridge's txMaxFee parameter
+   * that was in force at the moment the redemption request was made. The sum
+   * of the txFee values taken from all redemption outputs is equivalent to
+   * the total network fee of the entire redemption transaction.
+   */
+  txFee: BigNumber
```
lukasz-zimnoch, line 46: Step 2:
I think it makes more sense to have just the `txMaxFee` parameter here. The actual fee used should be determined during `createRedemptionTransaction` and embedded in the Bitcoin transaction.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,291 @@
...
+ * the new main UTXO of the wallet.
+ * @dev It is up to the caller to ensure the wallet key and each of the redeemer
+ *      addresses represent a valid pending redemption request in the Bridge.
+ *      If this is not the case, an exception will be thrown.
+ * @param bitcoinClient - The Bitcoin client used to interact with the network
+ * @param bridge - The handle to the Bridge on-chain contract
+ * @param walletPrivateKey = The private kay of the wallet in the WIF format
+ * @param mainUtxo - The main UTXO of the wallet. Must match the main UTXO
+ *        held by the on-chain Bridge contract.
+ * @param redeemerAddresses - The list of redeemer addresses
```
lukasz-zimnoch, line 67: Step 3: I think the redeemer's Bitcoin address should be used only when the redeemer requests the redemption. From that point, we should use `redeemerOutputScript` as the address doesn't matter for the wallet and the `Bridge`. Hence, I propose to rename `redeemerAddresses` to `redeemerOutputScripts`. I think we can also keep the `deriveOutputScript` function as it will be useful for future `requestRedemption` function. If we do that change, let's update the `@dev` section of docs and propagate it to `index.ts`.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,291 @@
...
+ *      on-chain contract. An exception will be thrown if any of the addresses
+ *      (along with the wallet public key corresponding to the provided private
+ *      key) does not represent a valid pending redemption.
+ * @param bridge The handle to the Bridge on-chain contract
+ * @param walletPrivateKey The private key of the wallet in the WIF format
+ * @param redeemerAddresses The addresses that will be the recipients of the
+ *        redeemed Bitcoins
+ * @returns The list of redemption requests.
+ */
+async function prepareRedemptionRequests(
```
lukasz-zimnoch, line 123: Step 4: Let's refactor this function and work with `redemptionOutputScripts`. I'd also rename that function to `fetchRedemptionRequests` as the requests are not prepared here. Let's make sure all comments and docs refer to the redeemer output scripts instead of addresses (and propagate the changes `index.ts`).

typescript/src/redemption.ts
```diff
@@ -0,0 +1,291 @@
...
+        "Provided redeemer address and wallet public key do not identify a pending redemption"
+      )
+    }
+
+    // TODO: Use the value of fee that was set in the Bridge (max fee) as the
+    // `txFee` for now.
+    // In the future allow the caller to propose the value of transaction fee.
+    // If the proposed transaction fee is smaller than the sum of fee shares from
+    // all the outputs then use the proposed fee and add the difference to outputs
+    // proportionally.
```
lukasz-zimnoch, line 158: Step 5: I agree with this TODO but I think that the `createRedemptionTransaction` function should be responsible for setting the fee. Let's move that TODO there.

typescript/src/redemption.ts
```diff
@@ -0,0 +1,291 @@
...
+ * the redemption), an output for each redemption request provided, and a change
+ * output if the redemption requests do not consume the entire amount of the
+ * single input.
+ * @dev The caller is responsible for ensuring the redemption request list is
+ *      correctly formed:
+ *        - there is at least one redemption
+ *        - the `requestedAmount` in each redemption request is greater than
+ *          the sum of its `txFee` and `treasuryFee`
+ *        - the redeemer address in each redemption request is of a standard
+ *          type (P2PKH, P2WPKH, P2SH, P2WSH).
```
lukasz-zimnoch, line 200: Step 6: Let's remove this doc as it is no longer valid (remember about `index.ts`)

typescript/src/redemption.ts
```diff
@@ -0,0 +1,291 @@
...
+      address.isWitnessScripthash()
+    ) {
+      transaction.addOutput({
+        script: bcoin.Script.fromAddress(address),
+        value: outputValue.toNumber(),
+      })
+    } else {
+      throw new Error("Redemption address must be P2PKH, P2WPKH, P2SH or P2WSH")
+    }
+  }
```
lukasz-zimnoch, line 265: Step 7: This validation will be no longer needed. Addresses should be already validated at the redemption request stage. Regarding output script, we can create a `bcoin.Script` instance based on the given request's `redeemerOutputScript` field.

lukasz-zimnoch (changes requested):
Given we merged https://github.com/keep-network/tbtc-v2/pull/292 recently, let's do the following refactoring to keep things consistent across the entire library:

typescript/src/redemption.ts
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/199/commits/4b3be0a5f176ab3e5733f1876ff60fa6d3ca53a2.

typescript/src/redemption.ts
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/199/commits/4b3be0a5f176ab3e5733f1876ff60fa6d3ca53a2.

typescript/src/redemption.ts
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/199/commits/4b3be0a5f176ab3e5733f1876ff60fa6d3ca53a2.

typescript/src/redemption.ts
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/199/commits/4b3be0a5f176ab3e5733f1876ff60fa6d3ca53a2.

typescript/src/redemption.ts
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/199/commits/4b3be0a5f176ab3e5733f1876ff60fa6d3ca53a2.

typescript/src/redemption.ts
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/199/commits/4b3be0a5f176ab3e5733f1876ff60fa6d3ca53a2.

typescript/test/data/redemption.ts
```diff
@@ -0,0 +1,484 @@
...
+      "679f1a813835dc5a1b65321077cdccaf08f98cbf04ca96d09b1600000000001600148" +
+      "db50eb52063ea9d98b3eac91489a90f738986f602483045022100ed5fa06ea5e9d4a9" +
+      "f0cf0df86a2cd473f693e5bda3d808ba82b04ee26d72b73f0220648f4d7bb25be7819" +
+      "22349d382cf0f32ffcbbf89c483776472c2d15644a48d67012103989d253b17a6a0f4" +
+      "1838b84ff0d20e8898f9d7b1a98f2564da4cc29dcf8581d900000000",
+  },
+  pendingRedemptions: [
+    {
+      redemptionKey:
+        "0xe7a0e8b724c3ee7708701137cb57f0a742e284b4ad0b21cc16a3b3fbe733521a",
```
lukasz-zimnoch, line 60: Can we use un-prefixed `redemptionKey` across this dataset?

typescript/test/data/redemption.ts
```diff
@@ -0,0 +1,484 @@
+import { BigNumber, BytesLike } from "ethers"
```
lukasz-zimnoch, line 1: I think we can improve the naming of datasets:
```
singleP2PKHAddressRedemption -> singleP2PKHRedemptionWithChange
singleP2WPKHAddressRedemption -> singleP2WPKHRedemptionWithChange
singleP2SHAddressRedemption -> singleP2SHRedemptionWithChange
singleP2WSHAddressRedemption -> singleP2WSHRedemptionWithChange
multipleAddressesRedemption -> multipleRedemptionsWithChange
noChangeRedemption -> multipleRedemptionsWithoutChange
p2PKHChangeRedemption -> I think it is redundant to singleP2PKHAddressRedemption so let's remove it
```
Does it makes sense?

lukasz-zimnoch (commented):
Just two comments regarding tests and I'm good to merge!

typescript/test/data/redemption.ts
tomaszslabon: Done in d510584750db537a224f7da442247f318680551d.

typescript/test/data/redemption.ts
tomaszslabon: As discussed, we decided to keep the prefix, as the result of `solidityKeccak256` is prefixed.

lukasz-zimnoch (approved):

PR #200, opened 2022-04-14
Refactor three: Extract deposit sweeping into a separate library

pdyraga:
~~Depends on #198~~

A refactoring of the Bridge contract is required given the contract size
constraints. This change is another step on this journey. Sweep
functionality is extracted to a separate, externally linked library. No
changes to the logic in the sweeping code were applied.
There were to small changes to the code and docs in `BridgeState`
committed separately: 7b4fefd1b40f51e9f49152cbebb95242f37a8efd, 0cb7816ce2f1ff338f3fb72fdfd82920dfb7d9b6.

Bridge contract size comparison before and after:
```
 |  Bridge               ·     24.852   │
 |  Bridge               ·     20.689   │
```

Gas costs comparison before and after:
```
 |  revealDeposit           ·    94028  ·    96659  ·    94690   │
 |  revealDeposit           ·    96041  ·    98672  ·    96703   │
 |  submitSweepProof        ·   190576  ·   344943  ·   264072   │
 |  submitSweepProof        ·   196251  ·   350622  ·   269763   │
 |  submitRedemptionProof   ·   126354  ·   203246  ·   171551   │
 |  submitRedemptionProof   ·   126280  ·   203172  ·   171481   │
```

PR #201, opened 2022-04-14
Refactor four: extract parts of redemption transaction outputs processing code to separate functions

pdyraga:
~~Depends on #200~~

Extracted code processing single non-change redemption output to
a separate function and split the code processing all outputs into two
functions. This should help  with the clarity of the code but most
importantly, help with fixing "stack too deep" problems after extracting
redemption code to an external library, out of the `Bridge` contract.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -1318,7 +1318,7 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     ///         separately before calling this function.
     ///         Reverts if output is neither requested pending redemption nor
     ///         requested and reported timed-out redemption.
-    ///         This function also marks each pending request as processed by 
+    ///         This function also marks each pending request as processed by
     ///         removing it from `pendingRedemptions` mapping.
     /// @param walletPubKeyHash 20-byte public key hash (computed using Bitcoin
     //         HASH160 over the compressed ECDSA public key) of the wallet which
@@ -1332,7 +1332,7 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     ///         for a timed-out redemption request.
     /// @return treasuryFee The treasury fee from this single redemption output.
     ///         This value needs to be summed up with treasury fees of all other
-    ///         outputs to evaluate the total treasury fee for the entire 
+    ///         outputs to evaluate the total treasury fee for the entire
     ///         redemption transaction. This value is 0 for a timed-out
     ///         redemption request.
     function processNonChangeRedemptionTxOutput(
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -1267,8 +1267,6 @@ contract Bridge is Ownable, EcdsaWalletOwner {
                     // key from the mapping to make it reusable for further
                     // requests.
                     delete pendingRedemptions[redemptionKey];
-
-                    processedRedemptionsCount++;
                 } else {
                     // If we entered here, the output is not a redemption
                     // request but there is still a chance the given output is
@@ -1295,9 +1293,9 @@ contract Bridge is Ownable, EcdsaWalletOwner {
                             outputValue <= redeemableAmount,
                         "Output value is not within the acceptable range of the timed out request"
                     );
-
-                    processedRedemptionsCount++;
                 }
+
+                processedRedemptionsCount++;
             }
 
             // Make the `outputStartingIndex` pointing to the next output by
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -1231,69 +1231,16 @@ contract Bridge is Ownable, EcdsaWalletOwner {
                 info.changeIndex = uint32(i);
                 info.changeValue = outputValue;
             } else {
-                // If we entered here, that the means the given output is
-                // supposed to represent a redemption. Build the redemption key
-                // to perform that check.
-                uint256 redemptionKey = uint256(
-                    keccak256(abi.encodePacked(walletPubKeyHash, outputScript))
-                );
-
-                if (pendingRedemptions[redemptionKey].requestedAt != 0) {
-                    // If we entered here, that means the output was identified
-                    // as a pending redemption request.
-                    RedemptionRequest storage request = pendingRedemptions[
-                        redemptionKey
-                    ];
-                    // Compute the request's redeemable amount as the requested
-                    // amount reduced by the treasury fee. The request's
-                    // minimal amount is then the redeemable amount reduced by
-                    // the maximum transaction fee.
-                    uint64 redeemableAmount = request.requestedAmount -
-                        request.treasuryFee;
-                    // Output value must fit between the request's redeemable
-                    // and minimal amounts to be deemed valid.
-                    require(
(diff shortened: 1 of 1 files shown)


pdyraga:
Changes rebased over `main` given #200 is merged. Undrafting!

solidity/contracts/bridge/Bridge.sol
```diff
@@ -1195,21 +1211,54 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
+    ///        vector. This function assumes vector's structure is valid so it
+    ///        must be validated using e.g. `BTCUtils.validateVout` function
+    ///        before it is passed here
+    /// @param walletPubKeyHash 20-byte public key hash (computed using Bitcoin
+    //         HASH160 over the compressed ECDSA public key) of the wallet which
+    ///        performed the redemption transaction.
+    /// @param processInfo RedemptionTxOutputsProcessingInfo identifying output
+    ///        starting index, the number of outputs and possible wallet change
+    ///        P2PKH and P2WPKH scripts.
+    function processRedemptionTxOutputs(
```
lukasz-zimnoch, line 1242: Would it work if we don't extract this function and just stick with `processNonChangeRedemptionTxOutput`? I see some cons of having it here:
- It overloads the `processRedemptionTxOutputs` name so we have two functions named that way and the difference between them is very vague. On the other hand, it is quite difficult to name it differently.  This is the first sign indicating that this function should not be split.
- Usage of `RedemptionTxOutputsProcessingInfo` looks kinda weird as we could compute those parameters entirely within the body of this function. The point where we end with `processRedemptionTxOutputs` no.1 and start `processRedemptionTxOutputs` no.2 seems to be unintuitive. This is the second argument against splitting.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -1312,8 +1310,87 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
+    ///         Reverts if outut is neither requested pending redemption nor
+    ///         requested and reported timed-out redemption.
+    ///         This function also marks each request as processed by removing
+    ///         them from `pendingRedemptions` mapping.
+    /// @param walletPubKeyHash 20-byte public key hash (computed using Bitcoin
+    //         HASH160 over the compressed ECDSA public key) of the wallet which
+    ///        performed the redemption transaction.
+    /// @param outputScript Non-change output script to be processed
+    /// @param outputValue Value of the output being processed
+    function processNonChangeRedemptionTxOutput(
```
lukasz-zimnoch, line 1329: I think we should document return parameters and make it super-explicit that they are both `0` for timed out redemptions and they are not counted into the `totalBurnableValue` and `totalTreasuryFee`. At a first sight, I thought we changed the original behavior here.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -1312,8 +1310,87 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
             "Redemption transaction must process at least one redemption"
         );
+    }
+
+    /// @notice Processes a single redemption transaction output. Tries to
+    ///         identify output as a pending redemption request or reported
+    ///         redemption timeout. Output script passed to this function must
+    ///         not be the change output. Such output needs to be identified
+    ///         separately before calling this function.
+    ///         Reverts if outut is neither requested pending redemption nor
```
lukasz-zimnoch, line 1320: Typo `outut`

solidity/contracts/bridge/Bridge.sol
```diff
@@ -1312,8 +1310,87 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
+    ///         requested and reported timed-out redemption.
+    ///         This function also marks each request as processed by removing
```
lukasz-zimnoch, line 1322: `This function also marks each request` -> `This function also marks each pending request` to make it explicit?

solidity/contracts/bridge/Bridge.sol
```diff
@@ -1221,88 +1270,37 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
                 // If we entered here, that means the change output with a
                 // proper non-zero value was found.
-                info.changeIndex = uint32(i);
-                info.changeValue = outputValue;
+                resultInfo.changeIndex = uint32(i);
+                resultInfo.changeValue = outputValue;
             } else {
                 // If we entered here, that the means the given output is
                 // supposed to represent a redemption. Build the redemption key
                 // to perform that check.
```
lukasz-zimnoch, line 1287: Redemption key is no longer built here but within `processNonChangeRedemptionTxOutput`.

solidity/contracts/bridge/Bridge.sol
pdyraga: > Would it work if we don't extract this function and just stick with processNonChangeRedemptionTxOutput? 

Unfortunately no. That was my first approach but the "stack too deep" error was still there. What's more, I think we will have to approach a refactoring of this function one more time once we add support to routing deposits to vaults. When optimizing gas costs in the PR I am working on, I tried to capture `self.bank` before using it to increase balances (the very end of the function) and got hit by "stack too deep" again.

> It overloads the processRedemptionTxOutputs name so we have two functions named that way and the difference between them is very vague. On the other hand, it is quite difficult to name it differently. This is the first sign indicating that this function should not be split.

I think it is fine if one function is preparing parameters for another. I used this pattern in the past (even outside of Solidity) and I think it can improve the clarity of the code and avoid too long functions. But sure, it's just a subjective opinion. 

> Usage of RedemptionTxOutputsProcessingInfo looks kinda weird as we could compute those parameters entirely within the body of this function. The point where we end with processRedemptionTxOutputs no.1 and start processRedemptionTxOutputs no.2 seems to be unintuitive. 

Agreed, I absolutely don't love it. I tried [various options](https://soliditydeveloper.com/stacktoodeep) and this one was working without a serious overhaul that I tried to avoid during this refactor. I completely agree it would be ideal to avoid it. I think we can try to approach it again when adding support for vaults. We may need to come up with another composition of logic because, as mentioned earlier, the "stack too deep" risk is still there.

lukasz-zimnoch: Ok, thanks for the clarification! If we can't do anything here, let's just leave it as is. Seems like an acceptable trade-off in the context of what you said.

solidity/contracts/bridge/Bridge.sol
pdyraga: 027b0bab7715569acf2430189f0b3bf1c14ec3d4

solidity/contracts/bridge/Bridge.sol
pdyraga: 027b0bab7715569acf2430189f0b3bf1c14ec3d4

solidity/contracts/bridge/Bridge.sol
pdyraga: 027b0bab7715569acf2430189f0b3bf1c14ec3d4

solidity/contracts/bridge/Bridge.sol
pdyraga: 027b0bab7715569acf2430189f0b3bf1c14ec3d4

lukasz-zimnoch (approved):

lukasz-zimnoch (approved):

Issue #146, opened 2022-02-22
Typescript code for revealing deposits


pdyraga:
We need to add support for revealing deposits in the typescript tBTC v2 library. Needed for smart contract system tests and dApp.

PR #205, opened 2022-04-15
Get rid of `getActiveWalletPublicKey` function

lukasz-zimnoch:
Refs: #146 

That function was marked as deprecated as it does not fit the current design. This pull request removes it.

### DIFF
--- a/typescript/src/sweep.ts
+++ b/typescript/src/sweep.ts
@@ -10,11 +10,7 @@ import {
   decomposeRawTransaction,
   isCompressedPublicKey,
 } from "./bitcoin"
-import {
-  createDepositScript,
-  getActiveWalletPublicKey,
-  DepositData,
-} from "./deposit"
+import { createDepositScript, DepositData } from "./deposit"
 import { Bridge } from "./bridge"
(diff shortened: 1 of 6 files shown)


typescript/src/sweep.ts
```diff
@@ -247,7 +243,7 @@ async function signMainUtxoInput(
 
 /**
  * Creates and sets `scriptSig` for the transaction input at the given index by
- * combining signature, signing group public key and deposit script.
```
tomaszslabon, line 250: Could we replace the description of `blindingFactor` and use `wallet public key` instead of `signing group public key`?
  
  ``` 
  /**
   * An 8 bytes number. Must be unique for given Ethereum address, signing group
   * public key and refund public key.
   */
  blindingFactor: BigNumber
  ```

lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/205/commits/ee7308ba81a9e34e9bc099addbc2f0e2f0a8d1f6

tomaszslabon (approved):

PR #207, opened 2022-04-19
Refactor nine: adjusting fraud library

pdyraga:
This PR is adjusting the library for frauds to follow the same approach as the rest of the libraries used by the `Bridge`.

First, `Fraud` does no longer keep its state but use the one from `BridgeState` passed as a `self` storage pointer. Second, the fraud challenge defeat logic that was previously split into two functions lives now under one, `defeatFraudChallange` function. Finally, adjusted the naming of functions and files.

I recommend reviewing this PR commit-by-commit. 

Bytecode size before and after:
```
 |  Bridge               ·      9.854  │      
 |  Bridge               ·      8.695  │
 |  Frauds               ·      6.829  │           
 |  Fraud                ·      6.531  │
```

Gas consumption before and after:
```
 |  submitFraudChallenge                  ·   107733  ·   107899  ·  107748  │
 |  submitFraudChallenge                  ·   107706  ·   107866  ·  107721  │
 |  defeatFraudChallenge                  ·    63988  ·    81386  ·   68444  │
 |  defeatFraudChallenge                  ·    62327  ·    79737  ·   66791  │
 |  notifyFraudChallengeDefeatTimeout     ·        -  ·        -  ·   49202  │
 |  notifyFraudChallengeDefeatTimeout     ·        -  ·        -  ·   49153  │
```


lukasz-zimnoch (approved):
LGTM! Leaving the final pass for @tomaszslabon!

Also, not for this PR but I realized that `redemptionParameters` function still returns `treasury` and `txProofDifficultyFactor` despite it no longer assigns those parameters.

pdyraga:
> Also, not for this PR but I realized that redemptionParameters function still returns treasury and txProofDifficultyFactor despite it no longer assigns those parameters.

👌🏻 I will clean this up in the final PR.

solidity/contracts/bridge/Fraud.sol
```diff
@@ -111,20 +95,22 @@ library Frauds {
...
     ///        fraud challenge deposit
     ///      - The signature (represented by r, s and v) must be generated by
-    ///        the wallet behind `walletPubKey` during signing of `sighash`
+    ///        the wallet behind `walletPublicKey` during signing of `sighash`
     ///      - Wallet can be challenged for the given signature only once
-    function submitChallenge(
-        Data storage self,
+    function submitFraudChallenge(
+        BridgeState.Storage storage self,
+        Wallets.Data storage wallets,
```
tomaszslabon, line 107: We added a new parameter 
```
Wallets.Data storage wallets
```
so we should probably add the corresponding `@param` to the function description.

solidity/contracts/bridge/Fraud.sol
```diff
@@ -230,19 +192,33 @@ library Frauds {
...
     ///      - before a defeat attempt is made the transaction that spends the
     ///        given UTXO must be proven in the Bridge
-    function defeatChallenge(
-        Data storage self,
+    function defeatFraudChallenge(
+        BridgeState.Storage storage self,
         bytes calldata walletPublicKey,
         bytes calldata preimage,
+        bool witness,
         address treasury
```
tomaszslabon, line 200: I think we don't need to pass the `treasury` parameter here, as we take it from `BridgeState` structure and we have access to this structure here.
So we can access `treasury` with `self.treasury` here.

solidity/contracts/bridge/Fraud.sol
```diff
@@ -329,58 +305,51 @@ library Frauds {
...
+    ///        produced for. See BIP-143 for reference
+    /// @param witness Flag indicating whether the preimage was produced for a
+    ///        witness input. True for witness, false for non-witness input.
+    /// @return utxoKey UTXO key that identifies spent input.
+    /// @dev Requirements:
+    ///      - `walletPublicKey` and `sighash` calculated as `hash256(preimage)`
+    ///        must identify an open fraud challenge
+    ///      - the preimage must be a valid preimage of a transaction generated
+    ///        according to the protocol rules and already proved in the Bridge
+    function unwrapChallenge(
```
tomaszslabon, line 325: I'm not sure if we really need the `unwrapChallenge` function after the refactor.
It duplicates some work that is also performed in `defeatFraudChallenge` (calculating sighs, calculating the challenge key, accessing the challenge).
I think we should just copy the contents of `unwrapChallenge` function into `Fraud.defeatFraudChallenge, remove the duplicate work and remove the `unwrapChallenge`.

solidity/contracts/bridge/Fraud.sol
pdyraga: The convention so far is that we are not adding those storage pointers to the parameter list. IMO they do not require explanation and ideally, they would be passed implicitly.

tomaszslabon: OK

solidity/contracts/bridge/Fraud.sol
pdyraga: Good idea! Dropped in 3e3e03be7be628b097d6ec5272098b265ed46805.

solidity/contracts/bridge/Fraud.sol
pdyraga: I love it! Please see e0c361f8c64db5fcd9993659105692d0bf7ab7d0.

tomaszslabon (approved):

lukasz-zimnoch (approved):

pdyraga:
> Also, not for this PR but I realized that `redemptionParameters` function still returns `treasury` and `txProofDifficultyFactor` despite it no longer assigns those parameters.

Cleaned up in #211

Issue #179, opened 2022-03-28
Moving funds


lukasz-zimnoch:
Wallets should be able to move their funds to another wallet under specific circumstances described in https://github.com/keep-network/tbtc-v2/pull/127

PR #208, opened 2022-04-19
Moving funds commitment

lukasz-zimnoch:
Refs: #179 

This pull request introduces the moving funds commitment feature, as specified in [RFC-1](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#26-wallet-lifecycle). Specifically, it adds the `submitMovingFundsCommitment` function that registers the target wallets commitment for the given source wallet if the following prerequisites are met:
- That function must be called only by one of the source wallet operators
- The source wallet must be in `MovingFunds` state
- The source waller must not have any pending redemptions
- The source wallet must not have a prior commitment submitted earlier
- The count of submitted target wallets must correspond to the expected count, computed according to the equation `N = min(valid_wallets.size, ceil(funds_to_transfer / wallet_transfer_max))` and `N > 0`
- Each target wallet must be in `Live` state
- Each target wallet must be different than the source wallet

Once the commitment is registered, the source wallet can make an actual Bitcoin transaction targeting the pre-committed wallets and submit a successful proof of that transaction in order to terminate the moving funds process with respect to the protocol rules.

The gas cost of the new `submitMovingFundsCommitment` function is `82580`. However, that cost does not include the call to the ECDSA registry (`isWalletMember`) since that dependency is mocked in the unit tests.


solidity/contracts/bridge/Wallets.sol
```diff
@@ -537,6 +560,68 @@ library Wallets {
...
+
+        uint64 walletBtcBalance = getWalletBtcBalance(
+            self,
+            walletPubKeyHash,
+            walletMainUtxo
+        );
+        uint256 expectedTargetWalletsCount = Math.min(
+            self.liveWalletsCount,
+            Math.ceilDiv(walletBtcBalance, self.maxBtcTransfer)
+        );
```
tomaszslabon, line 596: Could we add a description how we perform calculations of expected target wallets count like that?

lukasz-zimnoch: This is straight from RFC-1. What exactly do you propose to describe here?

solidity/contracts/bridge/Bridge.sol
```diff
@@ -408,6 +418,59 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
+    ///        but they can be read from appropriate `DkgResultSubmitted`
+    ///        and `DkgResultApproved` events.
+    ///      - The `walletMemberIndex` must be in range [1, walletMembersIDs.length]
+    ///      - The caller must be the member of the source wallet signing group
+    ///        at the position indicated by `walletMemberIndex` parameter
+    ///      - The `walletMainUtxo` components must point to the recent main
+    ///        UTXO of the source wallet, as currently known on the Ethereum
+    ///        chain. If the source wallet has no main UTXO, this parameter
+    ///        can be empty as it is ignored since the wallet balance is
+    ///        assumed to be zero.
```
pdyraga, line 451: Does this part make sense for this function? If there is no main UTXO, there is no moving funds, right?

solidity/test/bridge/Bridge.MovingFunds.test.ts
```diff
@@ -45,10 +49,501 @@ describe("Bridge - Moving funds", () => {
...
+
+        context("when the commitment was not submitted yet", () => {
+          // The wallet created using the `walletDraft` has no commitment
+          // submitted by default. No need to do anything here.
+
+          context("when the caller is a member of the source wallet", () => {
+            const walletMembersIDs = [1, 2, 3, 4, 5]
+            const walletMemberIndex = 2
+
+            let caller
```
tomaszslabon, line 93: I would give it a type, so that it's not of type `any`:
```
let caller: SignerWithAddress
```

solidity/contracts/bridge/Wallets.sol
```diff
@@ -456,6 +464,115 @@ library Wallets {
...
+        uint256 expectedTargetWalletsCount = Math.min(
+            self.liveWalletsCount,
+            Math.ceilDiv(walletBtcBalance, self.walletMaxBtcTransfer)
+        );
+
+        // This requirement fails only when `liveWalletsCount` is zero. In
+        // that case, the system cannot accept the commitment and must provide
+        // new wallets first.
+        require(expectedTargetWalletsCount > 0, "No target wallets available");
```
pdyraga, line 552: The active wallet can not be notified as closeable (see `notifyCloseableWallet`) but it can fail the heartbeat and transition into `MovingFunds` state. If there is no other wallet in the system, the active wallet can not move its funds.

We end up with a scenario when we have one `MovingFunds` wallet in the system, not accepting new deposits, not accepting redemption requests, so there is no way to increase or decrease its balance, and there is no way to create a new wallet in another way than by governance action lowering `walletMaxBtcBalance`. There is a way to fix it by the governance but it is not ideal. What is worse though is that it complicates timeouts for completing the moving funds operation. If there are no live wallets in the system, and the wallet can not commit to target wallets, the time for moving funds should not start counting.

A similar scenario is when the active wallet failed the heartbeat and transitioned into `MovingFunds` state and some other wallet qualified as closeable, entered the `MovingFunds` state, and has no `Live` wallets in the system it could move the funds to. 

One idea I had would be to disallow the active wallet to fail the heartbeat but it does not solve the problem given that nothing stops the active wallet to enter `MovingFunds` state by a timed-out redemption.

The only solution I see is to alter the algorithm for requesting a new wallet, and let the new wallet creation to be requested if the active wallet is in `MovingFunds` state, no matter its age or balance. What do you think? cc @beaushinkle @michalinacienciala

solidity/contracts/bridge/BridgeState.sol
```diff
@@ -174,6 +176,11 @@ library BridgeState {
         // active wallet. Can be unset to the zero value under certain
         // circumstances.
         bytes20 activeWalletPubKeyHash;
+        // Current live wallets count;
+        uint32 liveWalletsCount;
+        // The maximum BTC transfer than can be sent to a single target wallet
+        // during the moving funds process.
```
pdyraga, line 182:
```suggestion
        ///  The maximum amount that can be transferred to a single target wallet
        /// during the moving funds process
```

solidity/contracts/bridge/Bridge.sol
```diff
@@ -906,20 +980,25 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     ///         used to decide about wallet creation.
     /// @return walletMaxAge The maximum age of a wallet in seconds, after which
     ///         the wallet moving funds process can be requested.
+    /// @return walletMaxBtcTransfer The maximum BTC transfer in satoshi than
+    ///         can be sent  to a single target wallet during the moving
+    ///         funds process.
```
pdyraga, line 985:
```suggestion
    /// @return walletMaxBtcTransfer The maximum BTC amount in satoshi than
    ///         can be transferred  to a single target wallet during the moving
    ///         funds process.
```

solidity/contracts/bridge/BridgeState.sol
```diff
@@ -198,16 +206,21 @@ library BridgeState {
     /// @param _walletMaxAge New value of the wallet maximum age in seconds,
     ///        indicates the maximum age of a wallet in seconds, after which
     ///        the wallet moving funds process can be requested
+    /// @param _walletMaxBtcTransfer New value of the wallet maximum BTC
+    ///        transfer in satoshi, than can be sent  to a single target wallet
+    //         during the moving funds process.
```
pdyraga, line 211:
```suggestion
    /// @param _walletMaxBtcTransfer New value of the wallet maximum BTC
    ///        transfer in satoshi, determines the maximum amount that can be
    ///        transferred to a single target wallet during the moving funds process
```

solidity/contracts/bridge/Wallets.sol
```diff
@@ -456,6 +464,115 @@ library Wallets {
...
+    ///        chain. If the source wallet has no main UTXO, this parameter
+    ///        can be empty as it is ignored since the wallet balance is
+    ///        assumed to be zero.
+    ///      - At least one Live wallet must exist in the system
+    ///      - Submitted target wallets count must match the expected count
+    ///        `N = min(liveWalletsCount, ceil(walletBtcBalance / walletMaxBtcTransfer))`
+    ///        where `N > 0`
+    ///      - Each target wallet must be not equal to the source wallet
+    ///      - Each target wallet must be in Live state
+    function notifyWalletMovingFundsCommitmentSubmitted(
```
pdyraga, line 504: Why in `Wallets.sol` and not in `MovingFunds.sol`? 🤔

pdyraga (commented):
Was looking just at `*sol`.

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: Technically, this is possible because a wallet can have 0 BTC and be moved to the `MovingFunds` state. Of course, we'll handle that case with the "low balance close" function and a rational wallet should call it instead of submitting a commitment.

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: > (...) there is no way to create a new wallet in another way than by governance action lowering walletMaxBtcBalance.

This is not true. If an active wallet goes `MovingFunds`, the active wallet slot gets cleared out and a new wallet creation can be requested immediately (see `Wallets.requestNewWallet`). That basically solves the problem you described.

michalinacienciala: > We end up with a scenario when we have one MovingFunds wallet in the system, not accepting new deposits, not accepting redemption requests, so there is no way to increase or decrease its balance, and there is no way to create a new wallet in another way than by governance action lowering walletMaxBtcBalance

> The only solution I see is to alter the algorithm for requesting a new wallet, and let the new wallet creation to be requested if the active wallet is in MovingFunds state, no matter its age or balance

My understanding is that when the wallet marked as 'active' goes to `MovingFunds ` state, the 'active' mark gets removed (`delete self.activeWalletPubKeyHash` in `moveFunds` function). And the function for creating a new wallet allows for creation of the wallet when there's no 'active' wallet set.
Am I missing something?

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: The responsibility boundaries between those two are very fluid 😀 But, this is how I see it:
- `Wallets` should be responsible for wallet state updates (ideally)
- `MovingFunds` library is here because we needed to reduce the size of the `Bridge`. It groups everything related to moving funds that would normally live in the `Bridge`. Giving the above point about `Wallets`, the `MovingFunds` library should delegate wallet management to `Wallets`. A good example is moving funds proof where BTC tx processing is done in `MovingFunds` and wallet state check and update is made in `Wallets`.

Moving funds commitment is strictly related to wallet state updates. This is why `Wallets` library does the heavy lifting and `MovingFunds` just delegates the call. We can do it directly from `Bridge` but if we want to keep things consistent, that should pass through `MovingFunds` imho.

solidity/contracts/bridge/Wallets.sol
michalinacienciala: Oh, I see Łukasz wrote basically the same :)

solidity/test/bridge/Bridge.MovingFunds.test.ts
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/208/commits/26678877b05f029adf65a6c168f407e25b78bd3b

solidity/contracts/bridge/Bridge.sol
```diff
@@ -167,6 +169,12 @@ contract Bridge is Ownable, EcdsaWalletOwner {
         bytes32 sighash
     );
 
+    event MovingFundsCommitmentSubmitted(
+        bytes20 walletPubKeyHash,
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/208/commits/bda99ca8116b209828e66f3f5cb6e68c308b1bc8

solidity/contracts/bridge/BridgeState.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/208/commits/9cd2868d8576ccf65009f4ebbf1e3e1e2c6baeb4

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/208/commits/9cd2868d8576ccf65009f4ebbf1e3e1e2c6baeb4

solidity/contracts/bridge/BridgeState.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/208/commits/9cd2868d8576ccf65009f4ebbf1e3e1e2c6baeb4

solidity/contracts/bridge/Bridge.sol
```diff
@@ -408,6 +418,59 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
+    ///         proof submission.
+    /// @param walletPubKeyHash 20-byte public key hash of the source wallet
+    /// @param walletMainUtxo Data of the source wallet's main UTXO, as
+    ///        currently known on the Ethereum chain.
+    /// @param walletMembersIDs Identifiers of the source wallet signing group
+    ///        members
+    /// @param walletMemberIndex Position of the caller in the source wallet
+    ///        signing group members list
+    /// @param targetWallets List of 20-byte public key hashes of the target
+    ///        wallets that the source wallet commits to
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/208/commits/9cd2868d8576ccf65009f4ebbf1e3e1e2c6baeb4

solidity/contracts/bridge/Bridge.sol
```diff
@@ -408,6 +418,59 @@ contract Bridge is Ownable, EcdsaWalletOwner {
         self.notifyRedemptionTimeout(walletPubKeyHash, redeemerOutputScript);
     }
 
+    /// @notice Submits the moving funds target wallets commitment.
+    ///         Once all requirements are met, that function registers the
+    ///         target wallets commitment and opens the way for moving funds
+    ///         proof submission.
+    /// @param walletPubKeyHash 20-byte public key hash of the source wallet
+    /// @param walletMainUtxo Data of the source wallet's main UTXO, as
+    ///        currently known on the Ethereum chain.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/208/commits/9cd2868d8576ccf65009f4ebbf1e3e1e2c6baeb4

solidity/contracts/bridge/Bridge.sol
```diff
@@ -870,20 +937,25 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     ///         used to decide about wallet creation.
     /// @return walletMaxAge The maximum age of a wallet in seconds, after which
     ///         the wallet moving funds process can be requested.
+    /// @return walletMaxBtcTransfer The maximum BTC transfer in satoshi than
+    ///         can be sent  to a single target wallet during the moving
+    ///         funds process.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/208/commits/9cd2868d8576ccf65009f4ebbf1e3e1e2c6baeb4

solidity/contracts/bridge/Bridge.sol
```diff
@@ -897,21 +969,27 @@ contract Bridge is Ownable, EcdsaWalletOwner {
     /// @param walletMaxAge New value of the wallet maximum age in seconds,
     ///        indicates the maximum age of a wallet in seconds, after which
     ///        the wallet moving funds process can be requested
+    /// @param walletMaxBtcTransfer New value of the wallet maximum BTC
+    ///        transfer in satoshi, than can be sent  to a single target wallet
+    //         during the moving funds process.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/208/commits/9cd2868d8576ccf65009f4ebbf1e3e1e2c6baeb4

solidity/contracts/bridge/Bridge.sol
```diff
@@ -332,6 +341,11 @@ contract Bridge is Ownable, EcdsaWalletOwner {
         return self.activeWalletPubKeyHash;
     }
 
+    /// @return Current count of wallets being in the Live state.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/208/commits/9cd2868d8576ccf65009f4ebbf1e3e1e2c6baeb4

solidity/contracts/bridge/Wallets.sol
beaushinkle: > The active wallet can not be notified as closeable (see notifyCloseableWallet) but it can fail the heartbeat and transition into MovingFunds state. If there is no other wallet in the system, the active wallet can not move its funds.

> We end up with a scenario when we have one MovingFunds wallet in the system, not accepting new deposits, not accepting redemption requests, so there is no way to increase or decrease its balance, and there is no way to create a new wallet in another way than by governance action lowering walletMaxBtcBalance. There is a way to fix it by the governance but it is not ideal. What is worse though is that it complicates timeouts for completing the moving funds operation. If there are no live wallets in the system, and the wallet can not commit to target wallets, the time for moving funds should not start counting.

I think the simple solution is that a wallet should only be allowed to go into the `MovingFunds` state if there's a wallet for it to move its funds to, regardless of circumstance. To accomplish this, we could make all of the `intention-to-close` transactions (including the one we send for failed heartbeats) include a live wallet that we validate as being live at the time.

beaushinkle: `requestNewWallet` seems like it would take too long (iirc, it take 4-5 days), and by then the timeout for wallet transfer would have probably elapsed unless we make that timeout *very* long.

If we make it so that a wallet can't enter the `MovingFunds` state without another live wallet in the system, as in all `intention-to-close` transactions require a live wallet, then that solves heartbeats and whatnot, but doesn't solve timed out redemptions.

For timed out redemptions, what's the ideal behavior? There's a live wallet on the system, it'll be days before we can get a new one, it just failed to redeem, and it has some amount of bitcoin. What do we want it to do?

1) continue to accept deposits or stop accepting deposits
2) continue trying to serve redemptions or stop serving redemptions

If we want it to to continue, we slash for the failed redemption but don't go through with the closing process. If we want it to stop, we put it in the `MovingFunds` state with it's own, longer timeout. Maybe this looks like a new state called `MovingFundsToNewWallet` or something with a 7.

solidity/contracts/bridge/Bridge.sol
pdyraga: Shouldn't we instead revert if there is no main UTXO? As you say, in this case, calling this function does not make sense and the system is entering into a weird state where one wallet member can declare to which live wallets it is moving the zero balance. I would rather not allow a wallet to enter the path it should not be on.

solidity/contracts/bridge/Wallets.sol
pdyraga: > * `Wallets` should be responsible for wallet state updates (ideally)

Exactly! 💯 

But see that `notifyWalletMovingFundsCommitmentSubmitted` is not changing the state of the wallet. The wallet was in MovingFunds state and the wallet is still in MovingFunds state. It's only accepting information that will be later used by `MovingFunds.submitMovingFundsProof`. I think we should have this code in `MovingFunds` library. `MovingFunds.submitMovingFundsProof` is delegating to `Wallets.notifyWalletFundsMoved` because it's changing the state of the wallet and it makes perfect sense. `submitMovingFundsCommitment` is not changing it.

solidity/contracts/bridge/Wallets.sol
pdyraga: We discussed this problem on chat and agreed to expose a separate function allowing to reset the timeout for moving funds in case there is no active wallet in the system.

Example:
- Wallet transitioned to MovingFunds state.
- The wallet wants to commit to target wallets but there is no single Live wallet in the system. There are 4 days before the timeout for moving funds is hit.
- After 2 days, there are still no Live wallets. The wallet is calling a function to reset moving funds timeout. The wallet has again 4 days to move funds.

@lukasz-zimnoch Can we leave a TODO in `MovingFunds` library?

solidity/test/bridge/Bridge.MovingFunds.test.ts
```diff
@@ -45,10 +49,501 @@ describe("Bridge - Moving funds", () => {
...
+
+                    for (let i = 0; i < liveWallets.length; i++) {
+                      // eslint-disable-next-line no-await-in-loop
+                      await bridge.setWallet(liveWallets[i], {
+                        ...walletDraft,
+                        state: walletState.Live,
+                      })
+                    }
+
+                    await bridge.setLiveWalletsCount(liveWallets.length)
```
pdyraga, line 163: Non-blocking but we could consider increasing the number of live wallets as a part of the stub logic in `setWallet`. We are setting it separately everywhere and I think it's easy to forget for someone who is not that familiar with this code, so potentially error-prone.

solidity/test/bridge/Bridge.MovingFunds.test.ts
```diff
@@ -45,10 +49,501 @@ describe("Bridge - Moving funds", () => {
...
+                                    walletMemberIndex,
+                                    targetWallets
+                                  )
+                              })
+
+                              after(async () => {
+                                await restoreSnapshot()
+                              })
+
+                              it("should store the target wallets commitment for given wallet", async () => {
```
pdyraga, line 213: *the given

solidity/test/bridge/Bridge.MovingFunds.test.ts
```diff
@@ -45,10 +49,501 @@ describe("Bridge - Moving funds", () => {
...
+                              "Submitted target wallet cannot be equal to the source wallet"
+                            )
+                          })
+                        }
+                      )
+                    }
+                  )
+
+                  context(
+                    "when the submitted target wallets count is other as the expected",
```
pdyraga, line 303: *other than

solidity/test/bridge/Bridge.MovingFunds.test.ts
```diff
@@ -45,10 +49,501 @@ describe("Bridge - Moving funds", () => {
...
+        context("when the commitment was already submitted", () => {
+          before(async () => {
+            await createSnapshot()
+
+            await bridge.setWallet(ecdsaWalletTestData.pubKeyHash160, {
+              ...walletDraft,
+              state: walletState.MovingFunds,
+              // Set a non-zero commitment to make this scenario work.
+              movingFundsTargetWalletsCommitmentHash:
+                "0x959e95e0bd83d34878f77ead61cb4e955bf5e3bdc9e16cdfbd51c4c20ab7e6b4",
```
pdyraga, line 436: We could instead just call `submitMovingFundsCommitment` in the `before`. I think calling it is *cleaner* than manipulating stored data.

solidity/contracts/bridge/Wallets.sol
pdyraga: Also, looking at it again, I think this part:
https://github.com/keep-network/tbtc-v2/blob/5c95ed5c36e440c9aca25ec4f2694f9d8966f84f/solidity/contracts/bridge/Wallets.sol#L490-L503

should be in `MovingFunds.submitMovingFundsProof`. All before-state validations, such as `wallet.state == WalletState.MovingFunds`, should stay in `Wallets.sol` IMO in `notify*` functions.

pdyraga:
I went through all files and I am done with comments. Once they are all addressed I am good to merge.

solidity/contracts/bridge/Bridge.sol
pdyraga: Also, please see this discussion: https://github.com/keep-network/tbtc-v2/pull/213#discussion_r856944271.

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: > But see that notifyWalletMovingFundsCommitmentSubmitted is not changing the state of the wallet.

I think we have ambiguity when talking about "state". When I said:

> Wallets should be responsible for wallet state updates (ideally)

I meant that `Wallets` should take care of updates touching all fields in the `Wallet` struct, not just the `state` field. This is why I think it should NOT be done in `MovingFunds` because there is a more specific `Wallets` library that can manage everything about wallets. If we move those parts to `MovingFunds` we will scatter the wallet management logic and we will not have everything about wallets in one place as we aimed for.

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: I did not go with reverting here because before we agreed that `movingFunds` function should no longer transition to `Closed/Closing` state (https://github.com/keep-network/tbtc-v2/pull/213#discussion_r856944271) and we will have a function to close wallets with a balance below the dust threshold as a shortcut. The "no main UTXO set" case is just a specific case of a balance below the dust threshold. If we want to keep `movingFunds` function closing the wallets below the dust immediately and revert upon commitment in the same case without having a separate function for low balance close, we can run into troubles. Suppose such a case:
- Wallet has 1001 satoshi and the dust threshold is 1000 sat
- Wallet is requested to move its funds and the `balance < dust` condition is false so the wallet should push a commitment and cannot be immediately closed
- Meanwhile, the governance change the dust threshold to be 1500 sat 
- Wallet cannot submit the commitment because `balance < dust` is true now and gets slashed eventually.

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: Done.

solidity/contracts/bridge/Wallets.sol
pdyraga: We are already having cases where fields from `Wallets` struct are updated outside of `Wallets` library, one example is `Redemption` and `Sweep` changing `mainUtxoHash` and I think this is fine. 

Updating `mainUtxoHash` from `Redemption` and `Sweep` library is natural. In the same way as updating `movingFundsTargetWalletsCommitmentHash` from `MovingFunds` library is natural for me.

Also, see that the fact we route this call through `MovingFunds` is a sign that this is `MovingFunds`-related activity. 

I think `MovingFunds` should be only for state transitions and wallet lifecycle between transitions. It needs to validate before & after state when doing the transition and `Redemption`, `Sweep`, `MovingFunds` should be for particular activities validating the state required to perform such an activity.

lukasz-zimnoch: > We are already having cases where fields from Wallets struct are updated outside of Wallets library, one example is Redemption and Sweep changing mainUtxoHash and I think this is fine.

Yeah, but we talked about refactoring it by exposing appropriate functions from `Wallets` library and avoiding direct manipulation of the wallet state. 

> Also, see that the fact we route this call through MovingFunds is a sign that this is MovingFunds-related activity.

Indeed it is. However, the commitment is attached strictly to the wallet state. This is something other than, for example, a deposit which is a separate being and lies in a part of the contract state managed strictly by the `Deposit` library. Also, `MovingFunds` is here just because moving funds consist of some code that process BTC transaction. Otherwise, all this stuff could live directly in `Wallets` library.

> I think MovingFunds should be only for state transitions and wallet lifecycle between transitions. It needs to validate before & after state when doing the transition and Redemption, Sweep, MovingFunds should be for particular activities validating the state required to perform such an activity.

Agreed but see, the commitment is just a substep of the transition from `MovingFunds` to `Closing/Closed`. The wallet cannot change the state if the commitment is wrong or not submitted at all.

solidity/contracts/bridge/Bridge.sol
pdyraga: What I am trying to say in https://github.com/keep-network/tbtc-v2/pull/213#discussion_r856944271 is that if the wallet is closeable and has just the dust (or nothing) the `notifyCloseable` function should not put it into MovingFunds state but put it into Closing state immediately.

If we do it this way, we are not asking a wallet that has no funds (or just a dust of funds) to move funds and hope that wallet will not commit to a target wallets list. 

If the wallet has no funds (or just dust) we should not ask it to move funds at all and do not even give it a chance to commit to target wallets because MovingFunds state for no funds or dust of funds does not make sense.

Regarding the scenario you mentioned, we do not need to revert from this function if the balance is below the dust but I *think* we should revert if there is no main UTXO because first, the balance is zero so there is nothing to move, and second, proving the move of funds with no main UTXO is not possible. In practice, for no mainUTXO case, there should be no scenario when this function is callable because closeable notification should turn the wallet into Closing state immediately (?)

solidity/test/bridge/Bridge.MovingFunds.test.ts
lukasz-zimnoch: It's done that way because `submitMovingFundsCommitment` requires a specific state first and that will make the setup more complicated. Those are unit tests that should work in a specific initial state, and I think we should achieve that state in the most straightforward way since it is not the thing that matters here.

solidity/contracts/bridge/Wallets.sol
pdyraga: > Yeah, but we talked about refactoring it by exposing appropriate functions from Wallets library and avoiding direct manipulation of the wallet state.

How would such a function look like for Redemption/Sweeping though? Is it just about having an internal setter function in `Wallets.sol` for updating mainUTXO? If so, we can do the same for setting target wallets commitment 🤷🏻

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: So let's summarize. You propose:
- `moveFunds` should allow putting the wallet into `Closing` state. We just change that this occurs in the case when the balance is below dust, not just when main UTXO is not set
- `submitMovingFundsCommitment` reverts if main UTXO is not set (balance is zero) but doesn't care about the dust threshold
- There is a separate function for low balance close because there can be a case where `moveFunds` does not close the wallet but after handling all pending redemptions, the wallet can have a balance below dust and may not be able to transfer such a low balance to another wallet

The previous version (which I tried to implement here and in #139) was:
- `moveFunds` always put the wallet into `MovingFunds` without taking care of the balance
- `submitMovingFundsCommitment` allows putting a commitment as soon as there are no pending redemptions and doesn't care about the balance
-  There is a separate function for low balance close that can be called if the wallet is in `MovingFunds` and the balance is below dust. It can be called even if the commitment was submitted

Your proposition makes sense but makes the code and wallet lifecycle diagram a little bit more complex. Are we sure that this is justified well?

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: Not quite, that can be more complex than just simple getters/setters. Let's look at requesting redemption. We can expose `canAcceptRedemptionRequest` where we will check the wallet state is `Live` and the wallet has enough of free balance to handle that redemption. Regarding setting the main UTXO for sweeping/redemption, for the sweep, that's indeed a simple setter but for redemption, there is an additional logic that checks whether we should set or unset the main UTXO altogether.

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: Ok, let's do it. Will push changes around the commitment in this PR and those about `moveFunds` in #139. The additional function for low balance close will be a new PR anyway.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -408,6 +418,59 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
+    /// @param walletMainUtxo Data of the source wallet's main UTXO, as
+    ///        currently known on the Ethereum chain
+    /// @param walletMembersIDs Identifiers of the source wallet signing group
+    ///        members
+    /// @param walletMemberIndex Position of the caller in the source wallet
+    ///        signing group members list
+    /// @param targetWallets List of 20-byte public key hashes of the target
+    ///        wallets that the source wallet commits to move the funds to
+    /// @dev Requirements:
+    ///      - The source wallet must be in the MovingFunds state
```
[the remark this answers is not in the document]
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/208/commits/62a583d79c9a7d1024292462b9164835e715b2fa

solidity/contracts/bridge/Wallets.sol
```diff
@@ -456,6 +464,115 @@ library Wallets {
...
+
+        // This requirement fails only when `liveWalletsCount` is zero. In
+        // that case, the system cannot accept the commitment and must provide
+        // new wallets first.
+        require(expectedTargetWalletsCount > 0, "No target wallets available");
+
+        require(
+            targetWallets.length == expectedTargetWalletsCount,
+            "Submitted target wallets count is other than expected"
+        );
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, will fix it.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -408,6 +418,59 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
+    ///      - The source wallet must not have pending redemption requests
+    ///      - The source wallet must not have prior commitments submitted earlier
```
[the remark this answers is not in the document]
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/208/commits/70aae21de4c0036948930e312a8ab23e77d3eed1

solidity/contracts/bridge/Bridge.sol
```diff
@@ -792,6 +861,11 @@ contract Bridge is Ownable, EcdsaWalletOwner {
         return self.activeWalletPubKeyHash;
     }
 
+    /// @return The current count of wallets being in the Live state.
```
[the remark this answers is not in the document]
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/208/commits/70aae21de4c0036948930e312a8ab23e77d3eed1

solidity/contracts/bridge/BridgeState.sol
```diff
@@ -174,6 +176,11 @@ library BridgeState {
         // active wallet. Can be unset to the zero value under certain
         // circumstances.
         bytes20 activeWalletPubKeyHash;
+        // Current live wallets count;
```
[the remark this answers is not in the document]
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/208/commits/70aae21de4c0036948930e312a8ab23e77d3eed1

solidity/test/bridge/Bridge.MovingFunds.test.ts
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/208/commits/207daf09222830e9ff4805510006a59cc74f36ef

solidity/test/bridge/Bridge.MovingFunds.test.ts
lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/208/commits/207daf09222830e9ff4805510006a59cc74f36ef

tomaszslabon:
I don't have any further comments about this PR.

solidity/test/bridge/Bridge.MovingFunds.test.ts
lukasz-zimnoch: Good idea! See https://github.com/keep-network/tbtc-v2/pull/208/commits/1c90387f65f4ed8553231487f7c0efdc92bbcd3d

solidity/contracts/bridge/Wallets.sol
pdyraga: What if we decide about it at the very end, in a single PR, once we have all elements in place and for now stick to the current convention?

I am not saying I do not like this idea but to see if it is actually good or not, I would need to see the code. It may turn out that the code is easier to read but it may also turn out that there are too many calls to follow.

So far, the convention was that `Wallets.sol` is about requesting/registering new wallet and state transitions - checking the in and out state and performing the transition.

All validations specific to the given action were done in respective libraries. For example, the mentioned `Redemption.requestRedemption` checks the state of the wallet and its balance. In the end, it is modifying the wallet's main UTXO as something that is redemption/sweep-specific.

"Move funds" is a special case of redemption. Using this analogy, commitment for target wallets is like requesting redemption, and proving the movement of funds is like redemption proof. Using this analogy, I think the entire code of `notifyWalletMovingFundsCommitmentSubmitted` should be in `MovingFunds` and this part should be in `MovingFunds.submitMovingFundsProof`:
https://github.com/keep-network/tbtc-v2/blob/d8aabc13bf4bcc701ece6ff36e5c5eb7f1c441ef/solidity/contracts/bridge/Wallets.sol#L610-L623

Again - not saying I do not like it - but I am just worried that the final result may not be as clean as we think it's going to be and I would rather do it at the end once we have all the building blocks instead of having two competing conventions in the code now. 

I do not think the current convention is that bad, to be honest. When I am looking for a logic specific to the given action, I can see all the validations (requirements) in one place. And when I want to see if the wallet can transition from state x to state y, I can also see it all in one place with no cluttering code in between.

lukasz-zimnoch: Ok, there is no point in blocking too long here so let's move the commitment to `MovingFunds`. Just for the sake of potential future refactoring, I'll dump some concerns I have in my brain to have them captured somewhere:

The problem I see with the current convention is that it is not precise enough to be called a convention. This is the reason why we have this discussion at all.  The responsibility of `Wallets` library after the Bridge size-oriented refactoring is not clear and makes the navigation non-intuitive for someone not familiar with the code. We both have different points of view regarding that library and I bet another person will have a yet different opinion 😄 

In my opinion, the `Wallets` library must have one of the two possible shapes:
- We follow the rules about validating and updating the wallet's high-level state (everything living in `Wallet` struct) outside and the `Wallets` library only registers the new wallets and exposes methods to make state transitions: `moveFunds`, `terminateWallet`, `closeWallet`. The Bridge is responsible to decide how and when the state transitions should be made. This option is basically the extension of what we have now. That also means `Wallets` should not have any `notify*` functions exposed. 
- We put everything wallet-related into `Wallets` library, just as I proposed earlier. The Bridge just calls appropriate methods from `Wallets` library to validate and update wallet's high-level state.

What we have now is something between those two shapes and this is the main problem in my opinion.

pdyraga: Makes sense!

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: See c94ffd57

solidity/contracts/bridge/Wallets.sol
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/208/commits/524766aa17a0f681893114991fea28956f4a50f8

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/208/commits/a740fb56349974c13cb3931510d5e238bd69e570

pdyraga (approved):
Nice work @lukasz-zimnoch! 🐗

PR #212, opened 2022-04-21
Moving funds timeout

lukasz-zimnoch:
Closes: #179 
Depends on: #208

PR #213, opened 2022-04-22
Introduce wallet `Closing` state

lukasz-zimnoch:
Depends on: #212

### DIFF
--- a/solidity/contracts/bridge/Wallets.sol
+++ b/solidity/contracts/bridge/Wallets.sol
@@ -40,8 +40,12 @@ library Wallets {
         ///      fulfill their pending redemption requests although new
         ///      redemption requests and new deposit reveals are not accepted.
         MovingFunds,
-        /// @dev The wallet moved or redeemed all their funds and cannot
-        ///      perform any action.
+        /// @dev The wallet moved or redeemed all their funds and is in the
+        ///      closing period where they can be subject of fraud challenges
+        ///      and must defend against them.
+        Closing,
+        /// @dev The wallet finalized the closing period successfully and
+        ///      cannot perform any action in the Bridge.
         Closed,
         /// @dev The wallet committed a fraud that was reported. The wallet is
         ///      blocked and can not perform any actions in the Bridge.
@@ -68,6 +72,9 @@ library Wallets {
         // UNIX timestamp indicating the moment the wallet was requested to
         // move their funds.
         uint32 movingFundsRequestedAt;
+        // UNIX timestamp indicating the moment the wallet's closing period
+        // started.
+        uint32 closingStartedAt;
         // Current state of the wallet.
         WalletState state;
         // Moving funds target wallet commitment submitted by the wallet. It
@@ -84,6 +91,11 @@ library Wallets {
     );
 
     event WalletMovingFunds(
+        bytes32 indexed ecdsaWalletID,
+        bytes20 indexed walletPubKeyHash
+    );
+
+    event WalletClosing(
         bytes32 indexed ecdsaWalletID,
         bytes20 indexed walletPubKeyHash
     );
@@ -348,9 +360,7 @@ library Wallets {
         moveFunds(self, walletPubKeyHash);
     }
 
-    /// @notice Requests a wallet to move their funds. If the wallet balance
-    ///         is zero, the wallet is closed immediately and the ECDSA
-    ///         registry is notified about this fact. If the move funds
+    /// @notice Requests a wallet to move their funds. If the move funds
     ///         request refers to the current active wallet, such a wallet
     ///         is no longer considered active and the active wallet slot
     ///         is unset allowing to trigger a new wallet creation immediately.
@@ -362,19 +372,12 @@ library Wallets {
         bytes20 walletPubKeyHash
     ) internal {
         Wallet storage wallet = self.registeredWallets[walletPubKeyHash];
+        // Initialize the moving funds process.
+        wallet.state = WalletState.MovingFunds;
+        /* solhint-disable-next-line not-rely-on-time */
+        wallet.movingFundsRequestedAt = uint32(block.timestamp);
 
-        if (wallet.mainUtxoHash == bytes32(0)) {
-            // If the wallet has no main UTXO, that means its BTC balance
-            // is zero and it should be closed immediately.
-            closeWallet(self, walletPubKeyHash);
-        } else {
-            // Otherwise, initialize the moving funds process.
-            wallet.state = WalletState.MovingFunds;
-            /* solhint-disable-next-line not-rely-on-time */
-            wallet.movingFundsRequestedAt = uint32(block.timestamp);
-
-            emit WalletMovingFunds(wallet.ecdsaWalletID, walletPubKeyHash);
-        }
+        emit WalletMovingFunds(wallet.ecdsaWalletID, walletPubKeyHash);
 
         if (self.activeWalletPubKeyHash == walletPubKeyHash) {
             // If the move funds request refers to the current active wallet,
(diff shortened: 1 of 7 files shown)


solidity/contracts/bridge/Wallets.sol
```diff
@@ -40,8 +40,12 @@ library Wallets {
         ///      fulfill their pending redemption requests although new
         ///      redemption requests and new deposit reveals are not accepted.
         MovingFunds,
-        /// @dev The wallet moved or redeemed all their funds and cannot
-        ///      perform any action.
+        /// @dev The wallet moved or redeemed all their funds and is in the
+        ///      closing period where they can be subject of fraud challenges
+        ///      and must defend against them.
```
pdyraga, line 45: We could mention for future-us and auditors that this state is needed to protect against deposit frauds on deposits revealed but not swept and that the period in which the wallet is in a closing state must be greater that the deposit refund time plus some time margin.

solidity/contracts/bridge/Wallets.sol
```diff
@@ -362,19 +372,12 @@ library Wallets {
...
         Wallet storage wallet = self.registeredWallets[walletPubKeyHash];
+        // Initialize the moving funds process.
+        wallet.state = WalletState.MovingFunds;
+        /* solhint-disable-next-line not-rely-on-time */
+        wallet.movingFundsRequestedAt = uint32(block.timestamp);
 
-        if (wallet.mainUtxoHash == bytes32(0)) {
-            // If the wallet has no main UTXO, that means its BTC balance
-            // is zero and it should be closed immediately.
-            closeWallet(self, walletPubKeyHash);
```
pdyraga, line 369: I know we talked about it but I am no longer sure if it is a good move. I would rather transition that wallet to Closing state. See what is happening: we know that the wallet has no funds, yet we ask the wallet to move the funds. The wallet should respond - in a separate transaction - *but I have no funds to move!* and we should close it, but it is also possible, that the wallet [will send the commitment](https://github.com/keep-network/tbtc-v2/pull/208#discussion_r854952434) and then will try to submit a proof of 0-funds moved with no input UTXO and submitting this proof with no main UTXO is not possible:

https://github.com/keep-network/tbtc-v2/blob/5c95ed5c36e440c9aca25ec4f2694f9d8966f84f/solidity/contracts/bridge/MovingFunds.sol#L80-L82

tomaszslabon, line 369: If we decide never close the wallet in the `moveFunds` function then we will have to update the description of `notifyRedemptionTimeout` as it states the wallet can be closed if it no longer has the main UTXO.

solidity/test/bridge/Bridge.test.ts
```diff
@@ -5397,404 +5317,191 @@ describe("Bridge", () => {
...
+            let redemptionRequest: {
+              redeemer: string
+              requestedAmount: BigNumber
+              treasuryFee: BigNumber
+              txMaxFee: BigNumber
+              requestedAt: number
+            }
 
-                // Simulate the wallet's main UTXO removed
-                await bridge.unsetWalletMainUtxo(data.wallet.pubKeyHash)
```
tomaszslabon, line 5456: We can remove `unsetWalletMainUtxo` function as it is no longer used.

solidity/test/bridge/Bridge.Frauds.test.ts
```diff
@@ -413,6 +421,10 @@ describe("Bridge - Fraud", () => {
           testName: "when wallet state is Unknown",
           walletState: walletState.Unknown,
         },
+        {
+          testName: "when wallet state is Closing",
```
tomaszslabon, line 425: We should allow to call `submitFraudChallenge` when the wallet is in `Closing` state.
So we should change the `require` in `submitFraudChallenge` to allow to call the function when the state is `Closing` and add unit tests (probably just require that `submitFraudChallenge` does not revert).

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: According to our discussion https://github.com/keep-network/tbtc-v2/pull/208#discussion_r857396683, I'm keeping this condition in `moveFunds`. In the mentioned discussion, we agreed that `moveFunds` should close all wallets whose balance is below the dust threshold, not just those who have main UTXO unset. Unfortunately, this is not feasible since that needs to pass the wallet main UTXO to resolve the balance and at least one "intention-to-close" transaction (the wallet heartbeat failure) cannot do that easily. However, I think this is not a problem if we rely just on the presence of the main UTXO. That makes the behavior of the system quite consistent:
- `submitMovingFundsCommitment` function reverts if wallet's main UTXO is unset so a zero-balance wallet cannot submit the commitment
- `moveFunds` begins wallet closing in case its main UTXO is unset so a zero-balance wallet will not enter to `MoviungFunds`
- There will be a separate function that will allow closing a wallet with a non-zero but below-dust balance

lukasz-zimnoch: No longer relevant because of: https://github.com/keep-network/tbtc-v2/commit/a704bf38a45f7ba6f64ed597a13603b40e6e1c08

solidity/test/bridge/Bridge.test.ts
lukasz-zimnoch: No longer relevant because of: https://github.com/keep-network/tbtc-v2/commit/a704bf38a45f7ba6f64ed597a13603b40e6e1c08

solidity/test/bridge/Bridge.Frauds.test.ts
lukasz-zimnoch: Indeed, this is covered in https://github.com/keep-network/tbtc-v2/issues/220. I plan to do it in a follow-up PR to avoid complicating this one that focuses only on introducing the new state.

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/213/commits/2c05425bca213ea7220f7f8efe4cb4dd96b8ce22

tomaszslabon (dismissed):

solidity/contracts/bridge/Wallets.sol
pdyraga: Yeah, I also think it's a consistent approach. The wallet may be non-zero but below the dust threshold at the moment of transitioning to MoveFunds but it could also become non-zero but below the dust threshold after transitioning to MoveFunds, as a result of finalizing all outstanding redemptions. That function we are going to add will serve both cases.

Just one comment though:
> There will be a separate function that will allow closing a wallet with a non-zero but below-dust balance

There is also a possibility that the wallet will end up with zero balance after finalizing all pending redemptions. That new function needs to take this case into account.

solidity/contracts/bridge/Wallets.sol
```diff
@@ -386,13 +400,35 @@ library Wallets {
...
+        wallet.closingStartedAt = uint32(block.timestamp);
+
+        emit WalletClosing(wallet.ecdsaWalletID, walletPubKeyHash);
+    }
+
+    /// @notice Finalizes the closing period of the given wallet and notifies
+    ///         the ECDSA registry about this fact.
+    /// @param walletPubKeyHash 20-byte public key hash of the wallet
+    /// @dev Requirements:
+    ///      - The caller must make sure that the wallet is in the Closing state
```
pdyraga, line 425: If we are doing a transition between states, then I think `Wallets.sol` needs to check the initial state. But let's see how this ends up in practice at the very end when we have all the building pieces in place.

pdyraga (approved):

pdyraga:
@lukasz-zimnoch I think the PR description needs to be updated, we still have a shortcut between `Live` and `Closing`. Let's update it before the merge.

solidity/contracts/bridge/Wallets.sol
lukasz-zimnoch: The initial state checked by the caller before calling the transition function is the in-force convention in the `Wallets` lib at the moment. I agree we should revisit it since a lot of things changed since that convention was introduced.

lukasz-zimnoch:
> @lukasz-zimnoch I think the PR description needs to be updated, we still have a shortcut between `Live` and `Closing`. Let's update it before the merge.

Done 🚀

PR #215, opened 2022-04-25
Remove unnecessary TODO

tomaszslabon:
We decided to remove the TODOs because only a selected set of maintainers will be reimbursed for calling this function, and those maintainers will be working on round-robin duty. There is a low risk that someone will try to frontrun the call. At the same time, doing this change would complicate the code so we think it is not worth it.

### DIFF
--- a/solidity/contracts/bridge/Redemption.sol
+++ b/solidity/contracts/bridge/Redemption.sol
@@ -430,10 +430,6 @@ library Redemption {
         BitcoinTx.UTXO calldata mainUtxo,
         bytes20 walletPubKeyHash
     ) external {
-        // TODO: Just as for `submitSweepProof`, fail early if the function
-        //       call gets frontrunned. See discussion:
-        //       https://github.com/keep-network/tbtc-v2/pull/106#discussion_r801745204
-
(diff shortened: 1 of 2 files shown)


pdyraga:
We need to describe in the PR description and in the commit description *why* it's been decided this way.

And the reason is that only a selected set of maintainers will be reimbursed for calling this function, and those maintainers will be working on round-robin duty. There is a low risk that someone will try to frontrun the call. At the same time, doing this change would complicate the code so we think it is not worth it.

tomaszslabon:
> We need to describe in the PR description and in the commit description _why_ it's been decided this way.
> 
> And the reason is that only a selected set of maintainers will be reimbursed for calling this function, and those maintainers will be working on round-robin duty. There is a low risk that someone will try to frontrun the call. At the same time, doing this change would complicate the code so we think it is not worth it.

I changed the description.

pdyraga:
Let's also amend the commit to explain it in the details section.

tomaszslabon:
> Let's also amend the commit message to explain it in the details section.

Done.

pdyraga (approved):

PR #219, opened 2022-04-26
Use the latest Slither 0.8.3 in CI

michalinacienciala:
Slither 0.8.3 fixes issues with false validation failures from
crytic/slither#1067.

TODO:

- [x] remove TODO from the workflow once slither errors get fixed

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -284,11 +284,9 @@ jobs:
 
   contracts-slither:
     needs: contracts-detect-changes
-    # TODO: remove 'pull_request' condition before merging PR to main
     if: |
       github.event_name == 'push'
         || needs.contracts-detect-changes.outputs.path-filter == 'true'
-        || github.event_name == 'pull_request'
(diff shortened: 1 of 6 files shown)


.github/workflows/contracts.yml
```diff
@@ -282,53 +282,53 @@ jobs:
...
-  #       run: git config --global url."https://".insteadOf git://
-
-  #     - name: Install dependencies
-  #       run: yarn install
-
-  #     - name: Run Slither
-  #       run: slither --hardhat-artifacts-directory build .
+  contracts-slither:
+    needs: contracts-detect-changes
+    # TODO: remove 'pull_request' condition before merging PR to main
```
pdyraga, line 287: We should be good to drop this now @michalinacienciala. 

Can we undraft this PR?

michalinacienciala: Undrafted, we should be ok to merge

pdyraga (approved):

Issue #152, opened 2022-02-22
Address remaining TODOs in the Bridge contract


pdyraga:
Most of the TODOs have been captured separately: #42, #43, #44, #45, #46. Some small TODOs such as [potential gas cost optimizations](https://github.com/keep-network/tbtc-v2/blob/73fcd4832ad1a1cea43d9826b49f634453337f24/solidity/contracts/bridge/Bridge.sol#L1383-L1385) were not captured separately and should be handled during the work on this issue. We should work on this issue as the last one after everything from #42-#46 gets addressed.

PR #223, opened 2022-04-26
Updating deposit related parameters

tomaszslabon:
Refs #https://github.com/keep-network/tbtc-v2/issues/152
This PR adds functionalities for updating parameters related to deposits in Bridge.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -158,6 +158,12 @@ contract Bridge is Ownable, EcdsaWalletOwner {
 
     event VaultStatusUpdated(address indexed vault, bool isTrusted);
 
+    event DepositParametersUpdated(
+        uint64 depositDustThreshold,
+        uint64 depositTreasuryFeeDivisor,
+        uint64 depositTxMaxFee
+    );
+
     event WalletParametersUpdated(
         uint32 walletCreationPeriod,
         uint64 walletMinBtcBalance,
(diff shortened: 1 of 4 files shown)


pdyraga:
👀

solidity/contracts/bridge/Bridge.sol
```diff
@@ -762,7 +768,40 @@ contract Bridge is Ownable, EcdsaWalletOwner {
...
+    ///        For example, if the treasury fee needs to be 2% of each deposit,
+    ///        the `depositTreasuryFeeDivisor` should be set to `50`
+    ///        because `1/50 = 0.02 = 2%`
+    /// @param depositTxMaxFee New value of the deposit tx max fee in satoshis.
+    ///        It is the maximum amount of BTC transaction fee that can
+    ///        be incurred by each swept deposit being part of the given sweep
+    ///        transaction. If the maximum BTC transaction fee is exceeded,
+    ///        such transaction is considered a fraud
+    /// @dev Requirements:
+    ///      - Deposit treasury fee divisor must be greater than zero
```
pdyraga, line 792: Shouldn't we require `depositDustThreshold > 0`? 0-deposit does not make sense 🤔 Unless this is already taken care of in Bitcoin cc @lukasz-zimnoch

pdyraga, line 792: We should probably require `depositTxMaxFee > 0`.

solidity/test/bridge/Bridge.test.ts
```diff
@@ -188,6 +188,87 @@ describe("Bridge", () => {
     redemptionTimeout = (await bridge.redemptionParameters()).redemptionTimeout
   })
 
+  describe("updateDepositParameters", () => {
```
pdyraga, line 191: `Bridge.Parameters.test.ts` please.

pdyraga (approved):

solidity/contracts/bridge/Bridge.sol
tomaszslabon: Done in the follow-up PR in https://github.com/keep-network/tbtc-v2/pull/229/commits/c0a2ec4316c35a476d000a24e6e5c4d479c4c800

Deposit dust threshold is used for checking the value of deposits:
```
         require(
            fundingOutputAmount >= self.depositDustThreshold,
            "Deposit amount too small"
        );
```
Since there can not be deposits with value `0`, we require it to be at least `1`.

tomaszslabon: Yes, by allowing it to be `0`, we would force the transaction not to have fee at all.
So it should be positive.
Done in https://github.com/keep-network/tbtc-v2/pull/229/commits/f8a3ddcc8bec832d03183e0696751abb4051e261.

solidity/test/bridge/Bridge.test.ts
tomaszslabon: Done in 49f3a5d7a6cabc18a383e4b72c7f4077cdf79499.

PR #231, opened 2022-04-27
Governable

nkuba:
Instead of using Open Zeppelin's Ownable implementation, we switch to
Governable. Which is a simpler and smaller contract that is upgrade-safe.
Also, this approach is used in other projects
@threshold-network/solidity-contract, @keep-network/random-beacon, and
@keep-network/ecdsa.

Refs https://github.com/keep-network/keep-core/pull/2921


solidity/deploy/08_transfer_governance.ts
```diff
@@ -0,0 +1,20 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types"
+import { DeployFunction } from "hardhat-deploy/types"
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
```
pdyraga, line 4: Should we use this function in deployment scripts now?

pdyraga (commented):
Just one question @nkuba, everything else looks good to me.

solidity/deploy/08_transfer_governance.ts
nkuba: I'm sorry, I don't get your question. 

We used to transfer the ownership, now we transfer the governance. It's just a function name change, the result is the same.

pdyraga: My fault, never mind. 🤦🏻

pdyraga (approved):

Issue #220, opened 2022-04-26
Wallet lifecycle updates


lukasz-zimnoch:
The wallet lifecycle diagram reached its final shape. Here is the list of changes that must be made in the code to adjust it to the final diagram:
- [x] Introduce `Closing` state that begins the closing period instead of closing the wallet immediately (`MovingFunds` → `Closing` → `Closed` instead of `MovingFunds`-> `Closed`) (https://github.com/keep-network/tbtc-v2/pull/213)
- [x] Implement finalization of the wallet closing period (`Closing` → `Closed`) (https://github.com/keep-network/tbtc-v2/pull/227)
- [x] Implement immediate closing for moving funds wallets that are below the dust threshold (`MovingFunds` → `Closing`). No need to make sure that there is no pending redemptions here if the moving funds dust threshold is low enough (below the redemption dust threshold). (https://github.com/keep-network/tbtc-v2/pull/232)
- [x] Fraud challenges should be possible to submit for `Closing` wallets (https://github.com/keep-network/tbtc-v2/pull/222)
- [x] Fraud challenge timeout should be possible to report for `Closing` wallets. (https://github.com/keep-network/tbtc-v2/pull/222)
- [x] Fraud challenge timeout should be possible to report for `Terminated` wallets since there can be multiple challenges for the same wallet and the first challenge timeout will make the wallet `Terminated` . In that case, we should just return the challenger deposit. Notifier reward cannot be paid since the wallet cannot be slashed. (https://github.com/keep-network/tbtc-v2/pull/222)
- [x] Fraud challenge timeout should cause wallet termination (https://github.com/keep-network/tbtc-v2/pull/222)
- [x] Break the `walletMinBtcBalance` parameter into `walletMinCreationBtcBalance` and `walletMinClosureBtcBalance` parameters as per https://github.com/keep-network/tbtc-v2/issues/180 (https://github.com/keep-network/tbtc-v2/pull/233)

Issue #180, opened 2022-03-28
Potential improvements to wallet lifecycle


pdyraga:
This issue is a followup of #176 review/discussion. 

1. Say that `wallet_min_btc` (`minBtcBalance` in Solidity code) is set to 10 BTC. It means that if the current active wallet is old enough and it has at least 10 BTC we allow for wallet creation to kick off. But later, if the same wallet has just 1 satoshi less (10 BTC - 1 sat), we consider it closeable. Shouldn't we have some additional margin?

2. We attempt to move funds if the wallet balance is below `wallet_min_btc`. What if it is 1 sat? Can we guarantee sending 1 sat is doable on Bitcoin? This feels like a potential attack vector when someone redeems such an amount from the given wallet so that the amount left is not transferable and the wallet gets slashed for not moving funds to another wallet in time. Shouldn't we have some additional dust threshold so that closing the wallet is possible without moving funds?

beaushinkle:
After hashing this out internally, here's the tentative solution:

We disambiguate `wallet_min_btc` into three separate governable params: 
+ The minimum amount of btc the active wallet needs to have before we make a new active wallet (`wallet_min_creation_btc`)
+ The minimum amount of btc a non-active wallet needs to have before we motion to close+transfer (`wallet_min_closure_btc`)
+ The amount of btc at which if a non-active wallet has less than, we motion to close and *don't* transfer (`wallet_dust_leftover`).

So if `wallet_dust_leftover = 1000 sats` and `wallet_min_closure_btc = 1 btc`, and the current balance of a wallet is 5 btc then when:

+ That wallet gets a 4.5 btc redemption (leaving it at 0.5 btc), we close+transfer that 0.5 btc to another wallet
+ That wallet gets a (5-500 sat) redemption (leaving it 500 sats), we close that wallet, abandoning those 500 sats. We can maintain the peg via a governance donation.

beaushinkle:
@pdyraga good to close this?

PR #232, opened 2022-04-27
Handle dust balances during moving funds process

lukasz-zimnoch:
Refs: #220
Refs: #180 
Depends on: #227

Moving funds wallets can have a low BTC balance that would make the BTC transfer infeasible due to BTC network fees. To overcome that problem, we allow closing moving funds wallets whose balances are below the moving funds dust threshold without performing the BTC transaction.


tomaszslabon (dismissed):

solidity/contracts/bridge/Bridge.sol
```diff
@@ -226,6 +229,7 @@ contract Bridge is Governable, EcdsaWalletOwner {
         self.redemptionTimeout = 172800; // 48 hours
         self.movingFundsTxMaxTotalFee = 10000; // 10000 satoshi
         self.movingFundsTimeout = 7 days;
+        self.movingFundsDustThreshold = 1000; // 1000 satoshi
```
tomaszslabon, line 232: This value should probably be higher than `movingFundsTxMaxTotalFee` since that parameter is the highest fee we think may be necessary to pay for the moving funds transaction.

lukasz-zimnoch: Makes sense! See https://github.com/keep-network/tbtc-v2/pull/232/commits/7836dbd33df102464250d4d4ff0eb1addd506609

solidity/contracts/bridge/Bridge.sol
```diff
@@ -119,6 +119,8 @@ contract Bridge is Governable, EcdsaWalletOwner {
 
     event MovingFundsTimedOut(bytes20 walletPubKeyHash);
 
+    event MovingFundsBelowDustReported(bytes20 walletPubKeyHash);
```
pdyraga, line 122: I would call it just `MovingFundsBelowDust` or `MovingFundsWentBelowDust` (the first one preferred). `*Reported* feels to me like we should expect a second part of the action.

pdyraga (approved):

PR #235, opened 2022-04-27
Use hardhat deployments as test fixtures

nkuba:
We have deployment scripts for contracts that should be used in tests, instead of duplicating the deployments of contracts. This PR updates tests to use the contracts deployed with the hardhat scripts.


solidity/test/bridge/Bridge.MovingFunds.test.ts
```diff
@@ -1567,6 +1572,8 @@ describe("Bridge - Moving funds", () => {
         before(async () => {
           await createSnapshot()
 
+          walletRegistry.closeWallet.reset()
```
pdyraga, line 1575: Self-note: check if it is really needed once tests are green.

pdyraga (commented):
Tests are failing but the conception looks good to me. Would love @lukasz-zimnoch to also take a look given it's quite a wide change. On my end, once tests are green I am good to merge (assuming the test fix will be easy change).

nkuba:
@pdyraga looks like it's green

solidity/test/fixtures/bridge.ts
```diff
@@ -0,0 +1,66 @@
...
+    {
+      libraries: {
+        Deposit: (await ethers.getContract("Deposit")).address,
+        Sweep: (await ethers.getContract("Sweep")).address,
+        Redemption: (await ethers.getContract("Redemption")).address,
+        Wallets: (await ethers.getContract("Wallets")).address,
+        Fraud: (await ethers.getContract("Fraud")).address,
+        MovingFunds: (await ethers.getContract("MovingFunds")).address,
+      },
+    }
```
pdyraga, line 53: It is used only in one place I think so we could also define it in a separate function to do not pollute what's returned from `bridgeFixture()`.

solidity/test/bridge/Bridge.MovingFunds.test.ts
pdyraga: Yes, it is needed.

pdyraga (approved):

PR #236, opened 2022-04-28
Slashing of wallet members on moving funds timeout

tomaszslabon:
This PR adds slashing on wallet operators on fraud challenge defeat timeout notification.
Two new parameters was added: 
 - `movingFundsTimeoutSlashingAmount`,
 - `movingFundsTimeoutNotifierRewardMultiplier`.
 
 Executed call to `ecdsaWalletRegistry`'s `seize` function.

PR #234, opened 2022-04-27
Slashing of wallet members on fraud challenge defeat timeout

tomaszslabon:
This PR adds slashing of wallet operators on fraud challenge defeat timeout notification.

PR #238, opened 2022-04-28
Slashing of wallet members on redemption timeout notifications

tomaszslabon:
This PR adds slashing on wallet operators on redemption timeout notification.
Two new parameters were added:
 - `redemptionTimeoutSlashingAmount`,
 - `redemptionTimeoutNotifierRewardMultiplier`.

Executed call to `ecdsaWalletRegistry`'s `seize` function.


lukasz-zimnoch (dismissed):
LGTM! Leaving the final pass for @pdyraga

pdyraga:
👀

solidity/contracts/bridge/Bridge.sol
```diff
@@ -434,31 +438,47 @@ contract Bridge is Governable, EcdsaWalletOwner {
...
-    ///         be closed or marked as `MovingFunds` (depending on the presence
-    ///         or absence of the wallet's main UTXO) and the wallet will no
-    ///         longer be marked as the active wallet (if it was marked as such).
+    ///         The results of calling this function:
+    ///         - the pending redemptions value for the wallet will be decreased
+    ///           by the requested amount (minus treasury fee),
+    ///         - the tokens taken from the redeemer on redemption request will
+    ///           be returned to the redeemer,
+    ///         - the request will be moved from pending redemptions to
+    ///           timed-out redemptions.
```
pdyraga, line 449: Non-blocking but let's align punctuation marks:
- `,` everywhere but the last one and `.` on the last one OR
- Treat each one as a sentence, start with a capital letter and end with a dot `.`

solidity/contracts/bridge/Bridge.sol
```diff
@@ -434,31 +438,47 @@ contract Bridge is Governable, EcdsaWalletOwner {
...
     /// @dev Requirements:
+    ///      - The wallet must be in the Live or MovingFunds or Terminated state
     ///      - The redemption request identified by `walletPubKeyHash` and
     ///        `redeemerOutputScript` must exist
+    ///      - The expression `keccak256(abi.encode(walletMembersIDs))` must
+    ///        be exactly the same as the hash stored under `membersIdsHash`
+    ///        for the given `walletID`. Those IDs are not directly stored
+    ///        in the contract for gas efficiency purposes but they can be
+    ///        read from appropriate `DkgResultSubmitted` and `DkgResultApproved`
+    ///        events.
```
pdyraga, line 470:
```suggestion
    ///        events of the `WalletRegistry` contract.
```

pdyraga: Can we also update `Bridge.notifyFraudChallengeDefeatTimeout` and `Fraud.notifyFraudChallengeDefeatTimeout` the same way? I noted it during the review of #234.

solidity/contracts/bridge/Redemption.sol
```diff
@@ -765,29 +765,41 @@ library Redemption {
...
     /// @dev Requirements:
+    ///      - The wallet must be in the Live or MovingFunds or Terminated state
     ///      - The redemption request identified by `walletPubKeyHash` and
     ///        `redeemerOutputScript` must exist
+    ///      - The expression `keccak256(abi.encode(walletMembersIDs))` must
+    ///        be exactly the same as the hash stored under `membersIdsHash`
+    ///        for the given `walletID`. Those IDs are not directly stored
+    ///        in the contract for gas efficiency purposes but they can be
+    ///        read from appropriate `DkgResultSubmitted` and `DkgResultApproved`
+    ///        events.
```
pdyraga, line 795:
```suggestion
    ///        events of the `WalletRegistry` contract.
```

solidity/test/bridge/Bridge.test.ts
```diff
@@ -5398,8 +5433,14 @@ describe("Bridge", () => {
...
-              it("should slash the wallet's operators", async () => {
-                // TODO: Add test when operator slashing is implemented
+              it("should call the ECDSA wallet registry's seize function", async () => {
+                expect(walletRegistry.seize).to.have.been.calledOnceWith(
+                  redemptionTimeoutSlashingAmount,
+                  redemptionTimeoutNotifierRewardMultiplier,
+                  await thirdParty.getAddress(),
+                  data.wallet.ecdsaWalletID,
+                  walletMembersIDs
+                )
```
pdyraga, line 5443:
```
describe("notifyRedemptionTimeout", () => {
    context("when redemption request exists", () => {
      context("when the redemption request has timed out", () => {
        context("when the wallet is in Live state", () => {
          context("when the wallet is the active wallet", () => {
            context("when the wallet does not have the main UTXO set", () => {
```

I am confused about it. If the wallet is Live and has no main UTXO set, how could it be requested for redemption? cc @lukasz-zimnoch

pdyraga (approved):
Left some comments. Let's address them in the last PR from the series of slashing PRs, #236.

solidity/contracts/bridge/Bridge.sol
tomaszslabon: Done in ad7f9b2c328b3b7bbbe1da8523a66816368ab9df.

solidity/contracts/bridge/Bridge.sol
tomaszslabon: This comment is used throughout the bridge and its libraries multiple time.
I modified every occurrence of it in 4f71462c58380aa3997fb91645eb80c3af0f86a1.

solidity/test/bridge/Bridge.test.ts
tomaszslabon: @pdyraga 
It seems you are right (if the wallet is in the `Live` state and has a pending redemption request, it must have a `mainUTXO` as well).
So the situation in which the `Live` wallet has no `mainUTXO` cannot happen in reality.

However, in unit tests we could do that by modifying the state of the `BridgeStub` (we have `unsetWalletMainUtxo` function).
I added this "when the wallet does not have the main UTXO set" `context`, because In the flow of `notifyRedemptionTimeout` we reuse the `moveFunds` and there is an if statement that checks for presence or absence of `mainUTXO` and I wanted to test it.  
I didn't realise the wallet lifecycle does not allow for this.

So to fix this I suggest we remove `context("when the wallet does not have the main UTXO set"...)`  from the Live state.

Issue #237, opened 2022-04-28
Donations


lukasz-zimnoch:
The tBTC v2 system requires a donation mechanism that will allow to re-balance the system externally in case of a peg breach. The simplest way to achieve this is using the existing deposit mechanism and combining it with vault functionality (developed as part of #123). We can write a vault that will not increase depositors' balances nor its own balance. This is equivalent to a donation since new BTCs are added to the system but no new tBTCs are produced. The scope of this issue is writing such a vault.

PR #240, opened 2022-04-29
Reveal donation

lukasz-zimnoch:
Refs: #237 

This change exposes the `revealDonation` function that allows revealing an external BTC donation. The donation logic lives in the `Donation.sol` library. Once a donation is revealed to the `Bridge`, a merging funds request is created in order to notify the recipient wallet about the need of merging the obtained UTXO with the current main UTXO and updating the BTC balance. The merging funds mechanism is not exclusively connected with donations and will be also used to merge UTXOs received through the moving funds process, hence the logic lives in a separate `MergingFunds` library. Regarding wallet lifecycle diagram, the `revealDonation` function allows revealing donations only to wallets being in the `Live` state.


lukasz-zimnoch:
This approach turned out to have serious problems. There is a possibility to reveal regular sweep and moving funds transactions as donations. This opens a wide attack surface which is hard to mitigate since there are no simple and gas-effective solutions to distinguish those transactions. Because of that, we decided to close this PR and pivot to donations based on a dedicated vault.

PR #241, opened 2022-04-29
Prevent redemption requests with values lower than moving funds dust threshold

tomaszslabon:
This PR removes an unnecessary TODO and adds checks that ensure `redemptionDustThreshold` is always greater than `movingFundsDustThreshold`.
The check is needed to prevent redemption requests with values smaller or equal to `movingFundsDustThreshold`.

Refs #152


pdyraga:
> Wallets in the Closing state should not allow notifyRedemptionTimeout to be called.

The question though is: why?

I think we have three paths for a wallet to enter closing state:
1. Moving funds was requested, the wallet has no main UTXO set in the Bridge, and it moved to "Closing state". No UTXO  means the wallet's balance is 0 so no redemptions could be requested.
2. Wallet moved its funds. For a wallet to move the funds it first needs to commit to the list of target wallets. To commit to the list of target wallets, it needs to have all its requested redemptions completed.
3. The last possibility - covered by @lukasz-zimnoch in #232 - is when the wallet entered MovingFunds state but has just the dust. As a result, we allow ignoring the dust and transition to the Closing state. 

1 and 2 are easy. But for 3, to make it all consistent, I think we should ensure that the redemption `redemptionDustThreshold > movingFundsDustThreshold`. Then, we can be sure there are no redemptions requested and the wallet can safely transition to Closing state in `notifyMovingFundsBelowDust`.

If it makes sense, can we make it explicit in the documentation and add proper validation to functions updating redemption dust threshold and moving funds threshold?

tomaszslabon:
> > Wallets in the Closing state should not allow notifyRedemptionTimeout to be called.
> 
> The question though is: why?
> 
> I think we have three paths for a wallet to enter closing state:
> 
> 1. Moving funds was requested, the wallet has no main UTXO set in the Bridge, and it moved to "Closing state". No UTXO  means the wallet's balance is 0 so no redemptions could be requested.
> 2. Wallet moved its funds. For a wallet to move the funds it first needs to commit to the list of target wallets. To commit to the list of target wallets, it needs to have all its requested redemptions completed.
> 3. The last possibility - covered by @lukasz-zimnoch in [Handle dust balances during moving funds process #232](https://github.com/keep-network/tbtc-v2/pull/232) - is when the wallet entered MovingFunds state but has just the dust. As a result, we allow ignoring the dust and transition to the Closing state.
> 
> 1 and 2 are easy. But for 3, to make it all consistent, I think we should ensure that the redemption `redemptionDustThreshold > movingFundsDustThreshold`. Then, we can be sure there are no redemptions requested and the wallet can safely transition to Closing state in `notifyMovingFundsBelowDust`.
> 
> If it makes sense, can we make it explicit in the documentation and add proper validation to functions updating redemption dust threshold and moving funds threshold?

@pdyraga (CC @lukasz-zimnoch) Done in 46bc178317c57d08f20b50d3fd95311006ab3c11.

lukasz-zimnoch (approved):
LGTM! Leaving the final approval for @pdyraga

pdyraga:
👀

solidity/contracts/bridge/Bridge.sol
```diff
@@ -1159,7 +1160,8 @@ contract Bridge is Governable, EcdsaWalletOwner {
     ///        timeout receives. The value must be in the range [0, 100]
     /// @dev Requirements:
     ///      - Moving funds transaction max total fee must be greater than zero
-    ///      - Moving funds dust threshold must be greater than zero
+    ///      - Moving funds dust threshold must be greater than zero and smaller
+    ///        than redemption dust threshold
```
pdyraga, line 1164:
```suggestion
    ///      - Moving funds dust threshold must be greater than zero and lower
    ///        than the redemption dust threshold
```

[Here](https://english.stackexchange.com/a/26473) is a good explanation:

> Definitely use greater/less when writing in a maths context, as you say. And computing, where it's common to write and even say LT/GT (even some computer languages themselves support these forms). But outside these specialist areas it's not just 'colloquial' to use smaller - it's overwhelmingly more 'standard'.

solidity/contracts/bridge/BridgeState.sol
```diff
@@ -88,7 +88,9 @@ library BridgeState {
         // the moving funds process. Moving funds wallets having their BTC
         // balance below that value can begin closing immediately as
         // transferring such a low value may not be possible due to
-        // BTC network fees.
+        // BTC network fees. The value of this parameter must always be smaller
+        // than `redemptionDustThreshold` in order to prevent redemption requests
+        // with values smaller or equal to `movingFundsDustThreshold`.
         uint64 movingFundsDustThreshold;
```
pdyraga, line 94:
```suggestion
        // BTC network fees. The value of this parameter must always be lower
        // than `redemptionDustThreshold` in order to prevent redemption requests
        // with values lower or equal to `movingFundsDustThreshold`.
        uint64 movingFundsDustThreshold;
```

solidity/contracts/bridge/BridgeState.sol
```diff
@@ -133,6 +135,9 @@ library BridgeState {
         // `redemptionTreasuryFeeDivisor` and `redemptionTxMaxFee`
         // parameters in order to make requests that can incur the
         // treasury and transaction fee and still satisfy the redeemer.
+        // Additionally, the value of this parameter must always be greater
+        // than `movingFundsDustThreshold` in order to prevent redemption
+        // requests with values smaller or equal to `movingFundsDustThreshold`.
```
pdyraga, line 140:
```suggestion
        // requests with values lower or equal to `movingFundsDustThreshold`.
```

solidity/contracts/bridge/BridgeState.sol
```diff
@@ -495,7 +501,8 @@ library BridgeState {
     ///        timeout receives. The value must be in the range [0, 100]
     /// @dev Requirements:
     ///      - Moving funds transaction max total fee must be greater than zero
-    ///      - Moving funds dust threshold must be greater than zero
+    ///      - Moving funds dust threshold must be greater than zero and smaller
```
pdyraga, line 504:
```suggestion
    ///      - Moving funds dust threshold must be greater than zero and lower
```

solidity/contracts/bridge/BridgeState.sol
```diff
@@ -521,8 +528,9 @@ library BridgeState {
         );
 
         require(
-            _movingFundsDustThreshold > 0,
-            "Moving funds dust threshold must be greater than zero"
+            _movingFundsDustThreshold > 0 &&
+                _movingFundsDustThreshold < self.redemptionDustThreshold,
+            "Moving funds dust threshold must be greater than zero and smaller than redemption dust threshold"
```
pdyraga, line 533:
```suggestion
            "Moving funds dust threshold must be greater than zero and lower than redemption dust threshold"
```

solidity/test/bridge/Bridge.Parameters.test.ts
```diff
@@ -444,11 +451,40 @@ describe("Bridge - Parameters", () => {
...
               )
           ).to.be.revertedWith(
-            "Moving funds dust threshold must be greater than zero"
+            "Moving funds dust threshold must be greater than zero and smaller than redemption dust threshold"
           )
         })
       })
 
+      context(
+        "when new moving funds dust threshold is not smaller than redemption dust threshold",
```
pdyraga, line 460:
```suggestion
        "when new moving funds dust threshold is not lower than redemption dust threshold",
```

solidity/contracts/bridge/Bridge.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/241/commits/f2a5d262f26294ddcbdaf4295aa33b3cc18a5418.

solidity/contracts/bridge/BridgeState.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/241/commits/f2a5d262f26294ddcbdaf4295aa33b3cc18a5418.

solidity/contracts/bridge/BridgeState.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/241/commits/f2a5d262f26294ddcbdaf4295aa33b3cc18a5418.

solidity/contracts/bridge/BridgeState.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/241/commits/f2a5d262f26294ddcbdaf4295aa33b3cc18a5418.

solidity/contracts/bridge/BridgeState.sol
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/241/commits/f2a5d262f26294ddcbdaf4295aa33b3cc18a5418.

solidity/test/bridge/Bridge.Parameters.test.ts
tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/241/commits/f2a5d262f26294ddcbdaf4295aa33b3cc18a5418.

pdyraga (approved):

PR #244, opened 2022-05-02
Submitting the moved funds sweep proof

lukasz-zimnoch:
Refs: #243 

This pull request aims to introduce the ability to submit a moved funds sweep proof. That action is supposed to be done by the target wallets which received some BTC during a moving funds process performed by a closing wallet. This is achieved by introducing the `submitMovedFundsSweepProof` function.

### The `submitMovedFundsSweepProof` function

This function is supposed to be called once the sweeping wallet performs a BTC transaction that sweeps the moved funds UTXO with their own main UTXO. It expects a strict transaction structure. The output vector must contain a single output locking funds on the sweeping wallet's 20-byte public key hash. The input vector must contain the swept UTXO as the first input and the wallet's main UTXO as the second one. Such a constraint allows avoiding the iteration thus makes the code more gas-effective. The second input is optional, and it is required only if the sweeping wallet has a main UTXO set in the Bridge. If all prerequisites are met, the `submitMovedFundsSweepProof` sets the single transaction output as the new sweeping wallet main UTXO thus updates its BTC balance.

### Gas costs

Here are the gas costs of the new `submitMovedFundsSweepProof` function and the updated version of `submitMovingFundsProof` function:

```
·------------------------------------------------------------|---------------------------|--------------|-----------------------------·
|                    Solc version: 0.8.9                     ·  Optimizer enabled: true  ·  Runs: 1000  ·  Block limit: 30000000 gas  │
·····························································|···························|··············|······························
|  Methods                                                                                                                            │
····················|········································|·············|·············|··············|···············|··············
|  Contract         ·  Method                                ·  Min        ·  Max        ·  Avg         ·  # calls      ·  eur (avg)  │
····················|········································|·············|·············|··············|···············|··············
|  BridgeStub       ·  submitMovedFundsSweepProof            ·     144706  ·     157802  ·      149952  ·            5  ·          -  │
····················|········································|·············|·············|··············|···············|··············
|  BridgeStub       ·  submitMovingFundsProof                ·     150656  ·     221045  ·      197157  ·            9  ·          -  │
····················|········································|·············|·············|··············|···············|··············
```

Issue #243, opened 2022-05-02
Sweep moved funds


lukasz-zimnoch:
A wallet that received some funds through the moving funds process must sweep those funds with its main UTXO. To achieve that we need to:
- [x] Expose a function allowing to submit the sweep funds proof (https://github.com/keep-network/tbtc-v2/pull/244)
- [x] A `MovingFunds` wallet must handle all sweep requests before being able to submit their moving funds commitment (https://github.com/keep-network/tbtc-v2/pull/261)
- [x] Expose a function allowing to notify the sweep funds timeout (https://github.com/keep-network/tbtc-v2/pull/262)
- [x] Integrate with the fraud mechanism in a way that will disallow reporting a sweep funds transaction as a fraud (https://github.com/keep-network/tbtc-v2/pull/264)

PR #261, opened 2022-05-04
Prevent moving funds commitment if wallet has pending moved funds sweep requests

lukasz-zimnoch:
Refs: #243 
Depends on: #244 

It may happen that a live wallet that received some funds as the effect of a moving funds process performed by a closing wallet turns into the `MovingFunds` state itself. In that case, we must force that wallet to sweep all received UTXOs with their own main UTXO, before moving their own funds. Otherwise, the received UTXOs can remain blocked on a closed wallet. This can be achieved by forcing the wallet to handle all moved funds sweep requests before being able to submit the moving funds commitment. This is similar to how we handle pending redemptions for a `MovingFunds` wallet.


lukasz-zimnoch:
Ready for review. Keeping it draft until #244 lands on main.

pdyraga:
👀

solidity/contracts/bridge/MovingFunds.sol
```diff
@@ -416,6 +422,11 @@ library MovingFunds {
                 uint32(block.timestamp),
                 0
             );
+            // We added a new moved funds sweep request for the target wallet
+            // so we must increment their request counter.
+            self
+                .registeredWallets[targetWalletPubKeyHash]
+                .pendingMovedFundsSweepRequestsCount++;
```
pdyraga, line 429: Not in this PR but I wonder if we shouldn't emit an event in this line so the target wallet could realize they have a transaction to sweep.

Right now it's possible but a bit complicated:
- Wallet needs to monitor all `MovingFundsCommitmentSubmitted` in the network.
- Once `MovingFundsCommitmentSubmitted` is emitted, the wallet needs to check if it is on the list of `targetWallets`.
- If wallet is on the list, it needs to monitor `MovingFundsCompleted` for the given source `walletPubKeyHash`

If we emitted an event here, the wallet would only care about it. I guess the decision factor should be gas costs.

solidity/test/bridge/Bridge.MovingFunds.test.ts
```diff
@@ -2249,6 +2327,20 @@ describe("Bridge - Moving funds", () => {
                                   ).to.be.equal(await lastBlockTime())
                                 })
 
+                                it("should decrease the sweeping wallet's pending requests count", async () => {
+                                  // The `setPendingMovedFundsSweepRequest` call
+                                  // made as part of `runMovedFundsSweepScenario`
+                                  // set this counter to 1. Eventually, it
```
pdyraga, line 2333: What about replacing "Eventually" with "After submitting the proof of sweeping the moved funds"?

pdyraga (approved):

PR #262, opened 2022-05-04
Moved funds sweep timeout

lukasz-zimnoch:
Refs: #243
Depends on: #261 

Here we introduce the `notifyMovedFundsSweepTimeout` function that is meant to be used to notify about a timed out moved funds sweep process. Just like in the moving funds process, the recipient wallet must sweep the obtained funds with their own main UTXO in a finite time window. In case this time is exceeded, a third party can notify about this fact thus cause the wallet termination and a severe slashing of wallet operators' stakes.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -183,7 +183,10 @@ contract Bridge is Governable, EcdsaWalletOwner {
         uint32 movingFundsTimeout,
         uint96 movingFundsTimeoutSlashingAmount,
         uint256 movingFundsTimeoutNotifierRewardMultiplier,
-        uint64 movedFundsMergeTxMaxTotalFee
+        uint64 movedFundsMergeTxMaxTotalFee,
+        uint32 movedFundsMergeTimeout,
+        uint96 movedFundsMergeTimeoutSlashingAmount,
+        uint256 movedFundsMergeTimeoutNotifierRewardMultiplier
     );
 
     event WalletParametersUpdated(
@@ -243,6 +246,9 @@ contract Bridge is Governable, EcdsaWalletOwner {
         self.movingFundsTimeoutSlashingAmount = 10000 * 1e18; // 10000 T
         self.movingFundsTimeoutNotifierRewardMultiplier = 100; //100%
         self.movedFundsMergeTxMaxTotalFee = 10000; // 10000 satoshi
+        self.movedFundsMergeTimeout = 7 days;
+        self.movedFundsMergeTimeoutSlashingAmount = 10000 * 1e18; // 10000 T
+        self.movedFundsMergeTimeoutNotifierRewardMultiplier = 100; //100%
         self.fraudChallengeDepositAmount = 2 ether;
         self.fraudChallengeDefeatTimeout = 7 days;
         self.fraudSlashingAmount = 10000 * 1e18; // 10000 T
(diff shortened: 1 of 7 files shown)


lukasz-zimnoch:
Ready for review. Keeping it draft until https://github.com/keep-network/tbtc-v2/pull/261 lands on main.

pdyraga:
👀

solidity/contracts/bridge/Bridge.sol
```diff
@@ -123,6 +123,12 @@ contract Bridge is Governable, EcdsaWalletOwner {
 
     event MovedFundsSwept(bytes20 walletPubKeyHash, bytes32 sweepTxHash);
 
+    event MovedFundsSweepRequestTimedOut(
```
pdyraga, line 126: We have `RedemptionTimedOut` so maybe `MovedFundsSweepTimedOut`?

tomaszslabon (dismissed):

solidity/contracts/bridge/Bridge.sol
pdyraga: I think `MovedFundsSweepTimedOut` makes sense especially given the parameter names:
https://github.com/keep-network/tbtc-v2/blob/b06948ece5083d659bbcd3782bda531959e750e4/solidity/contracts/bridge/Bridge.sol#L193-L195

solidity/contracts/bridge/Bridge.sol
```diff
@@ -123,6 +123,12 @@ contract Bridge is Governable, EcdsaWalletOwner {
...
+        bytes20 walletPubKeyHash,
+        bytes32 movingFundsTxHash,
+        uint32 movingFundsTxOutputIndex
```
pdyraga, line 129: Should we say *moving* or *moved*? I have not yet reached that part of the code but I think we require `submitMovingFundsProof` to be called first. In that case, funds are moved and await for sweeping.

lukasz-zimnoch: *Moving* because here we refer to the moving funds transaction submitted via `submitMovingFundsProof`. This is the transaction that produced the UTXO being subject of the moved funds sweep request.

solidity/contracts/bridge/MovingFunds.sol
```diff
@@ -58,6 +58,19 @@ library MovingFunds {
         bytes movingFundsTxOutputVector;
     }
 
+    /// @notice Represents moved funds sweep request state.
+    enum MovedFundsSweepRequestState {
+        /// @dev The request is unknown to the Bridge.
+        Unknown,
+        /// @dev Request is pending and can be either processed or timed out.
```
pdyraga, line 65:
```suggestion
        /// @dev Request is pending and can become either processed or timed out.
```

solidity/contracts/bridge/MovingFunds.sol
```diff
@@ -70,8 +83,8 @@ library MovingFunds {
         uint64 value;
         // UNIX timestamp the request was created at.
         uint32 createdAt;
-        // UNIX timestamp the funds were swept at.
-        uint32 sweptAt;
+        // Current state of the request.
```
pdyraga, line 86:
```suggestion
        // The current state of the request.
```

solidity/test/bridge/Bridge.MovingFunds.test.ts
```diff
@@ -3075,6 +3170,466 @@ describe("Bridge - Moving funds", () => {
...
+
+                before(async () => {
+                  await createSnapshot()
+
+                  await bridge.setWallet(
+                    movedFundsSweepRequest.walletPubKeyHash,
+                    {
+                      ...(await bridge.wallets(
+                        movedFundsSweepRequest.walletPubKeyHash
+                      )),
```
pdyraga, line 3298: I do not entirely follow what is happening here. Can you clarify please?

solidity/contracts/bridge/Bridge.sol
pdyraga: Makes sense!

solidity/contracts/bridge/MovingFunds.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/262/commits/9fa5f0d18fe99ddca5b5b4b6c3f312c719aca13f

solidity/contracts/bridge/MovingFunds.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/262/commits/9fa5f0d18fe99ddca5b5b4b6c3f312c719aca13f

solidity/test/bridge/Bridge.MovingFunds.test.ts
lukasz-zimnoch: We change the wallet `state` while preserving other fields. We don't have an update function in the stub so we just use the getter to get wallet fields and set them through the setter with `state` field overwritten.

pdyraga: Ah! Yeah, it makes sense. Can you please copy&paste this explanation to the comment above this line?

pdyraga (approved):

solidity/test/bridge/Bridge.MovingFunds.test.ts
pdyraga: (in the next PR, I am merging this one)

Issue #254, opened 2022-05-03
Do not allow to use heartbeat messages for fraud challenges


pdyraga:
The problematic scenario is when a group signs some message (say every nth block hash) with their threshold ECDSA group public key, and then, malicious group member uses the signed message as [a sighash](https://github.com/keep-network/tbtc-v2/blob/5583abebb8c63bbc57e523d90b6072ec4ffb5d3b/solidity/contracts/bridge/Bridge.sol#L770) for a fraud challenge. Since that sighash was never used on Bitcoin, the fraud challenge can not be defeated. 

One potential option to address this problem is to establish a format for a heartbeat message and to make sure:
- message in this format can never become a sighash
- message in this format is rejected as a challenge.

Another option would be to never use tECDSA key for heartbeats but this option is not ideal for heartbeats given the fact operator is online does not mean they still have their private key share.

pdyraga:
It's been solved for the `Bridge` contract in #263 and #271.

For Random Beacon, the situation is different because we do not use a challenge-defeat scheme there. Instead, the notifier needs to sign `msg.sender` [with the group leaked key](https://github.com/keep-network/keep-core/blob/222be833fd41dac7255974aa885693cb278ecdd2/solidity/random-beacon/contracts/RandomBeacon.sol#L1202). Also, the documentation says:

> One example of a heartbeat protocol is signing some piece of information every n-th block and making sure this piece of information cannot be used for `RandomBeacon.reportUnauthorizedSigning()`. Specifically, the signed piece of information can not become `msg.sender` for `reportUnauthorizedSigning` call.

Such a heartbeat message could be, for example, `RandomBeacon` contract address. We do not need such strong heartbeat guarantees from the Random Beacon signing group. Even if the key material is lost, nothing terrible happens - the group is slashed, and a new group is created. For TBTC, the heartbeat must be stronger given that the key material loss is not only about slashing wallet signing group members but also about losing access to BTC stored in the wallet.

PR #263, opened 2022-05-04
Establish expected format for wallet ECDSA signing group heartbeat messages

pdyraga:
Refs #254

The problematic scenario is when a group signs some message (say every nth block
hash) with their threshold ECDSA group public key, and then, malicious group
member uses the signed message as a sighash for a fraud challenge. Since that
sighash was never used on Bitcoin, the fraud challenge can not be defeated.

We address this problem by establishing a format for a heartbeat message making
sure that  message in this format can never become a sighash of Bitcoin
transaction.

We construct the heartbeat message, as a 16-byte message. The first 8 bytes are
0xffffffffffffffff. The last 8 bytes are for an arbitrary uint64, being
a signed heartbeat nonce (for example, the last Ethereum block hash).

This PR is just establishing the format of a heartbeat message. Integration
with Fraud library will be implemented separately.

References:
https://github.com/bcoin-org/bcoin/blob/67590290858ef5eabc94178b716fad6bb6286b71/lib/primitives/tx.js#L441
https://github.com/bcoin-org/bcoin/blob/67590290858ef5eabc94178b716fad6bb6286b71/lib/primitives/tx.js#L467
https://github.com/bcoin-org/bcoin/blob/67590290858ef5eabc94178b716fad6bb6286b71/lib/primitives/tx.js#L637
https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#specification


solidity/contracts/bridge/Heartbeat.sol
```diff
@@ -0,0 +1,104 @@
...
+///      For this reason, we construct the heartbeat message, as a 40-byte
+///      message. The first 8 bytes are 0xffffffffffffffff. The last 32 bytes
+///      are for an arbitrary uint64, being a signed heartbeat nonce (for
+///      example, the last Ethereum block hash).
+library Heartbeat {
+    using BytesLib for bytes;
+
+    /// @notice Determines if the signed byte array is in fact valid wallet
+    ///         heartbeat message that is not fraudulent.
+    /// @dev Wallet heartbeat message must be exactly 40 bytes long with the first
```
tomaszslabon, line 86: Perhaps add `@param message`?

beaushinkle:
> The problematic scenario is when a group signs some message (say every nth block
hash) with their threshold ECDSA group public key, and then, malicious group
member uses the signed message as a sighash for a fraud challenge. Since that
sighash was never used on Bitcoin, the fraud challenge can not be defeated.

Ah, I see! The way that we do the fraud challenge scheme looks like this:

> For every UTXO spent by the wallet in an incorrect way, anyone should be able
to provide a fraud proof. Given that a Bitcoin transaction could be so large
that proving it on Ethereum would be impossible, fraud proofs need to be
processed with a challenge-response approach. 

> When a wallet unlocks a UTXO it needs to calculate a sighash and provide
a signature over that sighash, one for each unlocked UTXO. Fraud is reported
for a UTXO by providing a sighash along with the wallet's signature over that
sighash. From that moment, the wallet has a certain time to defend itself
against the challenge and prove that the UTXO was spent in an honest way.

So the malicious wallet provides the wallet's sighash *of the heartbeat* in order to present an unwinnable challenge to the honest operators. The solution is to make sure that the output of the heartbeat sighash is distinguishable from normal bitcoin transaction sighashes, so that malicious operators don't have this attack vector.

solidity/contracts/bridge/Heartbeat.sol
```diff
@@ -0,0 +1,104 @@
...
+///
+///      For this reason, we construct the heartbeat message, as a 40-byte
+///      message. The first 8 bytes are 0xffffffffffffffff. The last 32 bytes
+///      are for an arbitrary uint64, being a signed heartbeat nonce (for
+///      example, the last Ethereum block hash).
+library Heartbeat {
+    using BytesLib for bytes;
+
+    /// @notice Determines if the signed byte array is in fact valid wallet
+    ///         heartbeat message that is not fraudulent.
```
beaushinkle, line 85: Maybe "Determines if the signed byte array is a valid, non-fraudulent heartbeat message."

solidity/contracts/bridge/Heartbeat.sol
pdyraga: Good idea! Done in 8491ae4c68fac89ba7f23c391fa11989a4759c91.

solidity/contracts/bridge/Heartbeat.sol
```diff
@@ -0,0 +1,104 @@
...
+///        8  bytes  for tx_out.value
+///        1  byte   for tx_out.pk_script_bytes
+///        1  byte   for tx_out.pk_script
+///        4  bytes  for lock_time
+///
+///
+///      A SIGHASH flag is used to indicate which part of the transaction is
+///      signed by the ECDSA signature. There are currently 3 flags:
+///      SIGHASH_ALL, SIGHASH_NONE, SIGHASH_SINGLE, and different combinations
+///      of these falgs.
```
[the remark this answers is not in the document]
pdyraga: Ooops. Fixed in 8491ae4c68fac89ba7f23c391fa11989a4759c91.

pdyraga:
It's ready for another chance. I am not merging master for now - there are many PRs that are being merged and do not want to merge it multiple times. Leaving it to the reviewer who will approve & merge.

lukasz-zimnoch (approved):

Issue #173, opened 2022-03-16
Break `Bridge.test.ts` into smaller parts


lukasz-zimnoch:
The `Bridge.test.ts` file became too large and should be split into smaller files.

pdyraga:
My proposition:
- `Bridge.Vaults.test.ts` (`setVaultStatus`)
- `Bridge.Deposit.test.ts` (`revealDeposit`, `submitSweepProof`)
- `Bridge.Redemption.test.ts` (`requestRedemption`, `submitRedemptionProof`, `notifyRedemptionTimeout`)

PR #264, opened 2022-05-05
Integrate moved funds sweeps with fraud mechanism

lukasz-zimnoch:
Closes: #243 
Depends on: #262

Signatures that unlock main UTXOs being subject of processed moved funds sweep request can be challenged as frauds. Wallets must be able to defeat those challenges. This change adds a path to the `defeatFraudChallenge` function which makes such defeats possible.

### DIFF
--- a/solidity/contracts/bridge/Fraud.sol
+++ b/solidity/contracts/bridge/Fraud.sol
@@ -22,6 +22,7 @@ import {CheckBitcoinSigs} from "@keep-network/bitcoin-spv-sol/contracts/CheckBit
 import "./BitcoinTx.sol";
 import "./EcdsaLib.sol";
 import "./BridgeState.sol";
+import "./MovingFunds.sol";
 import "./Wallets.sol";
 
 /// @title Bridge fraud
@@ -226,7 +227,10 @@ library Fraud {
 
         // Check that the UTXO key identifies a correctly spent UTXO.
         require(
-            self.deposits[utxoKey].sweptAt > 0 || self.spentMainUTXOs[utxoKey],
+            self.deposits[utxoKey].sweptAt > 0 ||
+                self.spentMainUTXOs[utxoKey] ||
+                self.movedFundsSweepRequests[utxoKey].state ==
+                MovingFunds.MovedFundsSweepRequestState.Processed,
             "Spent UTXO not found among correctly spent UTXOs"
         );
 
--- a/solidity/contracts/test/BridgeStub.sol
+++ b/solidity/contracts/test/BridgeStub.sol
@@ -46,6 +46,17 @@ contract BridgeStub is Bridge {
         }
     }
 
+    function setProcessedMovedFundsSweepRequests(BitcoinTx.UTXO[] calldata utxos) external {
+        for (uint256 i = 0; i < utxos.length; i++) {
+            uint256 utxoKey = uint256(
+                keccak256(
+                    abi.encodePacked(utxos[i].txHash, utxos[i].txOutputIndex)
+                )
+            );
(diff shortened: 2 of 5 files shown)


lukasz-zimnoch:
Ready for review. Keeping it draft until https://github.com/keep-network/tbtc-v2/pull/262 lands on main.

solidity/contracts/bridge/Fraud.sol
```diff
@@ -226,7 +227,10 @@ library Fraud {
 
         // Check that the UTXO key identifies a correctly spent UTXO.
         require(
-            self.deposits[utxoKey].sweptAt > 0 || self.spentMainUTXOs[utxoKey],
+            self.deposits[utxoKey].sweptAt > 0 ||
+                self.spentMainUTXOs[utxoKey] ||
+                self.movedFundsSweepRequests[utxoKey].state ==
+                MovingFunds.MovedFundsSweepRequestState.Processed,
```
tomaszslabon, line 233: This will work, but as a simpler alternative we could consider renaming `spentMainUtxos` to `provedUtxos` (or something like that) and put there both of the inputs of the "funds moved" transaction.

`provedUTXO` could hold all the inputs proved as correctly spent from the `submit*Proof` functions.

(Just a suggestion, it can stay as it is).

lukasz-zimnoch: That will cost more gas in `submitMovedFundsSweepProof`. Now we just change the `state` of an existing request but in the proposed approach we'll need to put a new entry to the `provedUtxos` mapping. Also, I think it's cleaner to have the current approach where things are separated. We need to hold `movedFundsSweepRequests` mapping anyway.

pdyraga:
👀

solidity/contracts/test/BridgeStub.sol
```diff
@@ -46,6 +46,21 @@ contract BridgeStub is Bridge {
...
+            uint256 utxoKey = uint256(
+                keccak256(
+                    abi.encodePacked(utxos[i].txHash, utxos[i].txOutputIndex)
+                )
+            );
+            self.movedFundsSweepRequests[utxoKey].state = MovingFunds
+                .MovedFundsSweepRequestState
+                .Processed;
+        }
+    }
```
pdyraga, line 62: When working on https://github.com/keep-network/tbtc-v2/issues/173, let's try splitting `BridgeStub` into smaller, per-test-file stubs. The contract size is not a problem but we have so many functions in the stub now that it's sometimes hard to figure out which one is needed to test which functionality, and looking at the test code which one is from stub and which one is from the actual contract

solidity/test/data/fraud.ts
```diff
@@ -51,15 +86,16 @@ export const nonWitnessSignSingleInputTx: FraudTestData = {
     "a90f738986f60000000001000000",
   sighash: "0x5d09cd07392c7163335b67eacc999491a3794c15b88e2b59094be5c5b157064b",
   witness: false,
-  deposits: [
+  deposits: [],
+  spentMainUtxos: [],
+  movedFundsSweepRequests: [
```
pdyraga, line 91: Non-blocking but I think we should improve unit tests. Right now, the `self.movedFundsSweepRequests[utxoKey].state == MovingFunds.MovedFundsSweepRequestState.Processed` is covered kind of "accidentally", because we register `nonWitnessSignSingleInputTx` case as `movedFundsSweepRequests`.

I think we should distinguish in unit tests swept deposits UTXOs, spent main UTXOs, and moved funds sweep requests and have those three cases tested explicitly for all types of transactions (witness, non-witness, single input, multiple inputs).

pdyraga (approved):

Issue #267, opened 2022-05-09
WalletRegistry imported deploy script doesn't work with hardhat-deploy-ethers


nkuba:
In https://github.com/keep-network/tbtc-v2/pull/242 we switched dependency to `@keep-network/ecdsa` to the one that deploys an upgradable proxy for `WalletRegistry`. With this change the deployment scripts here stopped working with an error:
```
ERROR processing /home/runner/work/tbtc-v2/tbtc-v2/solidity/node_modules/@keep-network/ecdsa/export/deploy/04_deploy_wallet_registry.js:
Error: The requested contract was not found. Make sure the source code is available for compilation
```
See: https://github.com/keep-network/tbtc-v2/runs/6320923006?check_suite_focus=true

After investigation, it turned out that it is a problem with incompatibility in the plugins we use. When I removed dependency to `hardhat-deploy-ethers` and used `@nomiclabs/hardhat-ethers` instead the deployment script worked fine.

This problem needs further investigation and fixing the deployment script to work correctly with the imported deployment of `WalletRegistry` as a proxy contract.

PR #242, opened 2022-04-30
Bridge deployed as upgradeable proxy

nkuba:
Depends on https://github.com/keep-network/tbtc-v2/pull/235
Refs https://github.com/keep-network/keep-core/pull/2926
Closes https://github.com/keep-network/tbtc-v2/issues/141

This PR implements the upgradable deployment of the _Bridge_ contract.
The work here is very much based on the PR for _WalletRegistry_ upgradable deployment (https://github.com/keep-network/keep-core/pull/2926).

For upgradable deployments, we use [@keep-network/hardhat-helpers](https://github.com/keep-network/hardhat-helpers/) library that uses [openzeppelin-upgrades](https://github.com/OpenZeppelin/openzeppelin-upgrades) plugin under the hood.

## Constructor, Initializer

We moved most of the operations from the constructor to an initializer, which is a [requirement of the upgradable contracts](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#initializers).

## Gaps

We added gaps (`__gap`) to reserve storage slots for possible future upgrades. The gaps were added to base contracts and structs that hold data for libraries. There was no need to add gaps to structs that are stored in mappings, as mapping values are [stored in different slots](https://docs.soliditylang.org/en/v0.8.13/internals/layout_in_storage.html#mappings-and-dynamic-arrays).

Read more:
- https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
- https://forum.openzeppelin.com/t/storage-gaps-in-libraries-and-structs/28171

## Initializable

We use OpenZeppelin's implementation of the `Initializable` contract. The latest contract version exposes `reinitializer` function that should be used in new versions of the upgradable contract. Unfortunately, this feature is available right now only in an RC package `v4.6.0-rc.0`. We use this package version and hope that a final version is released by the time we deploy our contracts.

## ProxyAdmin

`ProxyAdmin` is a contract that is used to manage upgrades of an upgradable proxy contract. The deployments scripts were updated to transfer the ownership of the `ProxyAdmin` to the ESDM (Emergency Security Developer Multisig). In the future, once a DAO is established we want to replace `ProxyAdmin` with a [`ProxyAdminWithDeputy`](https://github.com/threshold-network/solidity-contracts/blob/main/contracts/governance/ProxyAdminWithDeputy.sol).

## WalletRegistry deployment

[...]

PR #268, opened 2022-05-09
Integration test for slashing gas usage

tomaszslabon:
WIP


solidity/test/fixtures/bridge-full-deployment.ts
```diff
@@ -0,0 +1,212 @@
+import { deployments, ethers, helpers } from "hardhat"
```
pdyraga, line 1: @nkuba What's your position on this? Should we reference random beacon and ecdsa deployment scripts or do it in one bundle like here? Given https://github.com/keep-network/tbtc-v2/issues/267 maybe we should start with `bridge-full-deployment.ts` and remove it as a part of the work on #267?

I think we generally need to come up with a consistent approach for testing in `tbtc-v2` given we have more and more deployment scripts added. It feels to me that the current approach is:
- Unit tests DO NOT use deployment scripts given that testing with a real `WalletRegistry` would be too complex. Instead, for Bridge unit tests, we always load the fixture from `fixtures/bridge.ts`.
- The only test that uses deployment scripts is `Bridge.Deployment.test.ts` - currently in progress in #242.
- We do need integration tests. @tomaszslabon's work in #268 is the first attempt. We need one, consistent approach to load the entire system that would be reused by all integration tests (see the question to @nkuba).

Does it make sense? cc @dimpar @lukasz-zimnoch

pdyraga: I've just realized `fixtures/bridge.ts` uses `await deployments.fixture()` and uses Smock to deploy `WalletRegistry` fake at the address of already deployed `WalletRegistry` from the deployment script. (I am surprised it works at all.) This is kind of spooky if we think we want to add proxy to the mix BUT it also means `bridge-full-deployment.ts` is [probably not needed](https://github.com/keep-network/tbtc-v2/blob/87a36ea6393db95e1391736a84cb8721ce3087e7/solidity/deploy/05_deploy_bridge.ts#L12-L15) (?)

nkuba: > What's your position on this? Should we reference random beacon and ecdsa deployment scripts or do it in one bundle like here?

We should base the tests on the deployment scripts. This way we can verify the deployment script. I don't like duplicating the deployment here, as it's likely the scripts become inconsistent at some point.

> Unit tests DO NOT use deployment scripts given that testing with a real WalletRegistry would be too complex. Instead, for Bridge unit tests, we always load the fixture from fixtures/bridge.ts.

Incorrect. The tests that use `bridgeFixture` from `fixtures/bridge.ts` are using the deployment scripts.
https://github.com/keep-network/tbtc-v2/blob/87a36ea6393db95e1391736a84cb8721ce3087e7/solidity/test/fixtures/bridge.ts#L17
The `walletRegistry` is faked there.

> We do need integration tests. @tomaszslabon's work in 
Integration test for slashing gas usage #268 is the first attempt. We need one, consistent approach to load the entire system that would be reused by all integration tests (see the question to @nkuba).

We do need the integration (or system) test that will work on the full deployment of all the modules, without any stubs.

PR #269, opened 2022-05-09
Donation vault

lukasz-zimnoch:
Closes: #237 

BTC donations are needed to re-balance the system in some rare cases when the BTC:tBTC peg can be breached. This change introduces a new `DonationVault` that is meant to handle BTC donations. That vault is plugged-in into the deposit flow. An external actor willing to make a BTC donation just makes a regular deposit targeting the `DonationVault`. Once that deposit is swept by the target wallet, the depositor's tBTC balance is not increased in the Bank, and the deposit is routed to the `DonationVault` which ignores it as well. In effect, deposited BTC remains swept by the wallet but no tBTC-related side effects take place.

### DIFF
--- a/solidity/contracts/vault/DonationVault.sol
+++ b/solidity/contracts/vault/DonationVault.sol
@@ -25,9 +25,8 @@ import "../bank/Bank.sol";
 ///         allows making donations using existing Bank balances.
 ///
 ///         BEWARE: ALL BTC DEPOSITS TARGETING THIS VAULT ARE NOT REDEEMABLE
-///         AND THERE IS NO WAY TO RESTORE THE DONATED BALANCE! DO NOT DEPOSIT
-///         AGAINST THIS VAULT OR DONATE YOUR EXISTING BANK BALANCE UNLESS YOU
-///         REALLY KNOW WHAT YOU ARE DOING!
+///         AND THERE IS NO WAY TO RESTORE THE DONATED BALANCE.
+///         USE THIS VAULT ONLY WHEN YOU REALLY KNOW WHAT YOU ARE DOING!
 contract DonationVault is IVault {
     Bank public bank;
 
--- a/solidity/contracts/vault/DonationVault.sol
+++ b/solidity/contracts/vault/DonationVault.sol
@@ -20,12 +20,14 @@ import "../bank/Bank.sol";
 
 /// @title BTC donation vault
 /// @notice Vault that allows making BTC donations to the system. Upon deposit,
-///         this vault does not increase depositors' tBTC balances and always
-///         decreases its own tBTC balance in the same transaction.
+///         this vault does not increase depositors' balances and always
+///         decreases its own balance in the same transaction. The vault also
+///         allows making donations using existing Bank balances.
 ///
-///         BEWARE: ALL BTC DEPOSITS TARGETING THIS VAULT ARE NOT REDEEMABLE AND
-///         THERE IS NO WAY TO RESTORE THE TBTC BALANCE! DO NOT DEPOSIT AGAINST
-///         THIS VAULT UNLESS YOU REALLY KNOW WHAT YOU ARE DOING!
+///         BEWARE: ALL BTC DEPOSITS TARGETING THIS VAULT ARE NOT REDEEMABLE
+///         AND THERE IS NO WAY TO RESTORE THE DONATED BALANCE! DO NOT DEPOSIT
+///         AGAINST THIS VAULT OR DONATE YOUR EXISTING BANK BALANCE UNLESS YOU
+///         REALLY KNOW WHAT YOU ARE DOING!
 contract DonationVault is IVault {
     Bank public bank;
 
@@ -96,7 +98,7 @@ contract DonationVault is IVault {
     }
 
     /// @notice Ignores the deposited amounts and does not increase depositors'
-    ///         individual tBTC balances. Decreases its own tBTC balance
+    ///         individual balances. The vault decreases its own tBTC balance
     ///         in the Bank by the total deposited amount.
     /// @param depositors Addresses of depositors whose deposits have been swept
     /// @param depositedAmounts Amounts deposited by individual depositors and
--- a/solidity/contracts/vault/DonationVault.sol
+++ b/solidity/contracts/vault/DonationVault.sol
@@ -29,7 +29,7 @@ import "../bank/Bank.sol";
 contract DonationVault is IVault {
     Bank public bank;
 
-    event DonationReceived(address[] depositors, uint256[] depositedAmounts);
+    event DonationReceived(address donator, uint256 donatedAmount);
 
     modifier onlyBank() {
         require(msg.sender == address(bank), "Caller is not the Bank");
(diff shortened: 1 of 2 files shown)


pdyraga:
👀

solidity/contracts/vault/DonationVault.sol
```diff
@@ -0,0 +1,81 @@
...
+
+pragma solidity ^0.8.9;
+
+import "./IVault.sol";
+import "../bank/Bank.sol";
+
+/// @title BTC donation vault
+/// @notice Vault that allows making BTC donations to the system. Upon deposit,
+///         this vault does not increase depositors' tBTC balances and always
+///         decreases its own tBTC balance in the same transaction.
```
pdyraga, line 24:
```suggestion
/// @notice Vault that allows making BTC donations to the system. Upon deposit,
///         this vault does not increase depositor's balances and always
///         decreases its own balance in the same transaction.
```

TBTC is only relevant for `TBTCVault`. All other system components operate on a balance in the Bank.

solidity/contracts/vault/DonationVault.sol
```diff
@@ -0,0 +1,81 @@
...
+///
+///         BEWARE: ALL BTC DEPOSITS TARGETING THIS VAULT ARE NOT REDEEMABLE AND
+///         THERE IS NO WAY TO RESTORE THE TBTC BALANCE! DO NOT DEPOSIT AGAINST
```
pdyraga, line 27:
```suggestion
///         THERE IS NO WAY TO RESTORE THE BALANCE! DO NOT DEPOSIT AGAINST
```

solidity/contracts/vault/DonationVault.sol
```diff
@@ -0,0 +1,81 @@
...
+        external
+        override
+        onlyBank
+    {
+        revert("Donation vault cannot receive balance approval");
+    }
+
+    /// @notice Ignores the deposited amounts and does not increase depositors'
+    ///         individual tBTC balances. Decreases its own tBTC balance
+    ///         in the Bank by the total deposited amount.
```
pdyraga, line 59:
```suggestion
    /// @notice Ignores the deposited amounts and does not increase depositors'
    ///         individual balances. The vault decreases its own balance
    ///         in the Bank by the total deposited amount.
```

solidity/contracts/vault/DonationVault.sol
```diff
@@ -0,0 +1,81 @@
...
+        bank = _bank;
+    }
+
+    /// @notice Function not supported by the `DonationVault`. Always reverts.
+    function receiveBalanceApproval(address, uint256)
+        external
+        override
+        onlyBank
+    {
+        revert("Donation vault cannot receive balance approval");
```
pdyraga, line 54: Why not to allow someone make a donation from their Bank balance? 

I think we should allow it by `Bank.approveBalanceAndCall` as well as implement `DonationVault.donate()` for already approved balances.

lukasz-zimnoch: Yep, good idea! Here it is: https://github.com/keep-network/tbtc-v2/pull/269/commits/4f2b9d0bcc28ff55a08e12881d2c28331b3000e0

solidity/contracts/vault/DonationVault.sol
lukasz-zimnoch: See 2efa4ced

solidity/contracts/vault/DonationVault.sol
lukasz-zimnoch: See 2efa4ced

solidity/contracts/vault/DonationVault.sol
lukasz-zimnoch: See 2efa4ced

lukasz-zimnoch:
Ready for another shot @pdyraga !

solidity/contracts/vault/DonationVault.sol
```diff
@@ -0,0 +1,126 @@
...
+/// @title BTC donation vault
+/// @notice Vault that allows making BTC donations to the system. Upon deposit,
+///         this vault does not increase depositors' balances and always
+///         decreases its own balance in the same transaction. The vault also
+///         allows making donations using existing Bank balances.
+///
+///         BEWARE: ALL BTC DEPOSITS TARGETING THIS VAULT ARE NOT REDEEMABLE
+///         AND THERE IS NO WAY TO RESTORE THE DONATED BALANCE! DO NOT DEPOSIT
+///         AGAINST THIS VAULT OR DONATE YOUR EXISTING BANK BALANCE UNLESS YOU
+///         REALLY KNOW WHAT YOU ARE DOING!
```
pdyraga, line 30: Let's simplify it a bit. I was really confused with *or* and *unless* when I was reading it for the first time :)

```suggestion
///         BEWARE: ALL BTC DEPOSITS TARGETING THIS VAULT ARE NOT REDEEMABLE
///         AND THERE IS NO WAY TO RESTORE THE DONATED BALANCE.
///         USE THIS VAULT ONLY WHEN YOU REALLY KNOW WHAT YOU ARE DOING!
```

solidity/contracts/vault/DonationVault.sol
```diff
@@ -0,0 +1,126 @@
...
+contract DonationVault is IVault {
+    Bank public bank;
+
+    event DonationReceived(address donator, uint256 donatedAmount);
```
pdyraga, line 34: Can we replace `donator` with `donor`? The second option is a way most common word.

solidity/test/vault/DonationVault.test.ts
```diff
@@ -0,0 +1,283 @@
...
+          .connect(account1)
+          .approveBalanceAndCall(vault.address, 1000)
+      })
+
+      after(async () => {
+        await restoreSnapshot()
+      })
+
+      it("should decrease donator's balance", async () => {
+        expect(await bank.balanceOf(account1.address)).to.be.equal(0)
```
pdyraga, line 198: Can we `increaseBalance` for a bit more than 1000 to confirm just the amount from `approveBalanceAndCall` is donated?

solidity/test/vault/DonationVault.test.ts
```diff
@@ -0,0 +1,283 @@
...
+    context("when called with correct parameters", () => {
+      let tx: ContractTransaction
+
+      before(async () => {
+        await createSnapshot()
+
+        await bank.connect(bridge).increaseBalance(account1.address, 1000)
+        await bank
+          .connect(account1)
+          .increaseBalanceAllowance(vault.address, 1000)
```
pdyraga, line 122: Can we `increaseBalance` and `increaseBalanceAllowance` for a bit more than `1000` to confirm the amount passed as `donate` parameter is taken into account?

lukasz-zimnoch: See 52fc175

solidity/contracts/vault/DonationVault.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/269/commits/63eeb9baeb7f699785b9c3111102a0324ef54513

solidity/contracts/vault/DonationVault.sol
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/269/commits/1624c3a7b29da55b57d9701840a0cd230c2ee3b8

Issue #249, opened 2022-05-03
Implement governance contract for Bridge


pdyraga:
`Bridge` contract exposes a lot of `update*` functions allowing to update values of parameters. We should implement `BridgeGovernance` contract similar to `WalletRegistryGovernance` and `RandomBeaconGovernance`.

Such a contract should:
- implement a governable delay for updating parameters,
- allow transferring the ownership of `Bridge` to some other contract,
- allow to update a single parameter without having to pass values of other parameters (less error-prone update).

PR #273, opened 2022-05-10
Bridge governance

dimpar:
Closes https://github.com/keep-network/tbtc-v2/issues/249

Implementation of the Bridge Governance contract.

TODO:
[ ] Resize the contract size! With all the params that might require an update the BridgeGovernance size is around 31.577KB. A couple of approaches were tested by this point:
- extracting params to a separate lib with part of the internal code moved to the lib. This approach helped to some degree, but still, the contract was growing too much.
- `*ChangeInitiated` fields internal with external `getRemaining***UpdateTime` 30.817 vs public `*ChangeInitiated` 30.487 fields. Leaving `public` fields because the size is smaller.


dimpar:
Please hold reviewing this PR..

PR #275, opened 2022-05-11
Frauds fix: Sighash calculated inside submitFraudChallenge

tomaszslabon:
Due to security reasons `sighash` is calculated inside the `submitFraudChallenge` function based on the passed `sha256preimage` parameter which is single SHA-256 hash of the `preimage` used during input signing.

Passing `sighash` calculated outside `submitFraudChallenge` enables ECDSA signature forgery - the wallet could be wrongly accused of a fraud.


eth-r (approved):

solidity/contracts/bridge/Fraud.sol
```diff
@@ -368,8 +379,10 @@ library Fraud {
         BridgeState.Storage storage self,
         bytes calldata walletPublicKey,
         uint32[] calldata walletMembersIDs,
-        bytes32 sighash
+        bytes memory sha256preimage
```
lukasz-zimnoch, line 382: I think we could stick with a `sighash` here since we don't do a signature check in this function.

solidity/contracts/bridge/Fraud.sol
```diff
@@ -109,32 +109,41 @@ library Fraud {
...
     ///      - The signature (represented by r, s and v) must be generated by
-    ///        the wallet behind `walletPublicKey` during signing of `sighash`
+    ///        the wallet behind `walletPubKey` during signing of `sighash`
+    ///        which was calculated from `sha256preimage`
     ///      - Wallet can be challenged for the given signature only once
     function submitFraudChallenge(
         BridgeState.Storage storage self,
         bytes calldata walletPublicKey,
-        bytes32 sighash,
+        bytes memory sha256preimage,
```
lukasz-zimnoch, line 133: I'd name it `preimageSha256` because it will show the order of operations in a better way (preimage -> sha256).

solidity/contracts/bridge/Fraud.sol
pdyraga: I think it makes sense to have `sha256preimage` for the consistency reason you mentioned. The fraud notifier does provides `walletPublicKey`, `sha256preimage` to `submitFraudChallenge`, I think it makes sense to continue with the same set of parameters for `notifyFraudChallengeDefeatTimeout`.

solidity/contracts/bridge/Fraud.sol
pdyraga: Agreed.

solidity/test/data/fraud.ts
```diff
@@ -84,7 +87,8 @@ export const nonWitnessSignSingleInputTx: FraudTestData = {
     "63ac6776a914e257eccafbc07c381642ce6e7e55120fb077fbed8804e0250162b1" +
     "75ac68ffffffff01b8240000000000001600148db50eb52063ea9d98b3eac91489" +
     "a90f738986f60000000001000000",
-  sighash: "0x5d09cd07392c7163335b67eacc999491a3794c15b88e2b59094be5c5b157064b",
+  sha256preimage:
+    "0x90d29d1021c2a1a381c0c8a39f5ffda96770e1c2ffed835f822500033841b2e3",
```
pdyraga, line 91: Wouldn't it be easier to store both `sighash` and `preimageSha256` in test data? We would avoid all `sha256(sha256preimage)` in `Bridge.Frauds.test.ts` and called `data.preimageSha256` or `data.sighash` depending on what is needed.

solidity/contracts/bridge/Fraud.sol
tomaszslabon: Done in 23a59f3e45809064c84f9678bc96e60529bec98e.

solidity/test/data/fraud.ts
tomaszslabon: I added the `sighash` field to test data in 8d5368a082a858912c01bb5097a06544f657229d.

pdyraga:
👀

pdyraga (approved):

Issue #256, opened 2022-05-03
One-transaction TBTC to BTC


pdyraga:
Currently, going from TBTC to BTC requires at least two transactions:
- `TBTC.approveAndCall`,
- `Bridge.requestRedemption`.

We could use the calldata in `receiveApproval` to pass the parameters required by `Bridge.requestRedemption` and allow to go to BTC in one transaction.

PR #288, opened 2022-05-13
Set of updates to Bank contract documentation

pdyraga:
Depends on #287 
Refs https://github.com/keep-network/tbtc-v2/issues/256
Refs #289

Added `@params` to all public functions - this approach is aligned with `Bridge`
contract. Clearly listed all requirements in `@dev` section, and did a set of
grammar updates. Updated `increaseBalanceAndCall` documentation to do not
mention sweeping given it is an implementation detail of the `Bridge` and not
something the `Bank` should not be interested in. Also. the documentation of that
function would get outdated as soon as when we would implement quickminting.


solidity/contracts/bank/Bank.sol
```diff
@@ -22,16 +22,16 @@ import "../vault/IVault.sol";
...
 /// @dev Bank is a governable contract and the Governance can upgrade the Bridge
 ///      address.
 contract Bank is Ownable {
     address public bridge;
 
-    /// @notice The balance of a given account in the Bank. Zero by default.
+    /// @notice The balance of the given account in the Bank. Zero by default.
     mapping(address => uint256) public balanceOf;
 
     /// @notice The remaining amount of balance a spender will be
```
beaushinkle, line 37: In the docstring for `Bank`, we say that balances are transferred between "holders", and in `allowance`, we say that a "spender" can transfer on behalf of an "owner".

When is an "owner" different than a "holder"?

beaushinkle:
I cannot comment on the lines directly since they weren't changed in this PR, but there are a couple of places, like L353 where we reference EIP2612 signatures: `Used to construct EIP2612 signature provided` without an article.

Something like "Used to construct the EIP2612 signature provided..." would be better (and "an EIP2612 signature..." in most other cases).

beaushinkle:
```
    /// @notice Returns the current nonce for EIP2612 permission for the
    ///         provided balance owner for a replay protection. Used to
    ///         construct EIP2612 signature provided to `permit` function.
    mapping(address => uint256) public nonce;
```

I think this should be something like "...balance owner to provide replay attack protection" or "...balance owner to protect against replay attacks".

beaushinkle:
On L50
```
    /// @notice Returns EIP2612 Permit message hash. Used to construct EIP2612
    ///         signature provided to `permit` function.
```
can become
```
    /// @notice Returns EIP2612 Permit message hash. Used to construct an EIP2612
    ///         signature provided to the `permit` function.
```

beaushinkle:
L215
```
    ///         Users can authorize a transfer of their balance with a signature
    ///         conforming EIP712 standard, rather than an on-chain transaction
```
to
```
    ///         Users can authorize a transfer of their balance with a signature
    ///         conforming to the EIP712 standard, rather than an on-chain transaction
```

solidity/contracts/bank/Bank.sol
```diff
@@ -138,8 +145,8 @@ contract Bank is Ownable {
         );
     }
 
-    /// @notice Atomically increases the balance allowance granted to `spender`
-    ///         by the caller by the given `addedValue`.
+    /// @notice Atomically increases the caller's balance allowance granted to
+    ///         `spender` by the given `addedValue`.
     function increaseBalanceAllowance(address spender, uint256 addedValue)
```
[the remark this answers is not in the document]
pdyraga: https://github.com/keep-network/tbtc-v2/pull/288/commits/a41ac06b0cc4049e3dfe31b293fef3a898ab4a0a

pdyraga:
@beaushinkle @lukasz-zimnoch This is ready for another chance. If there are no major blockers, I propose we do not hold this PR but merge, continue the discussion in this PR, and I'll make the necessary changes - if any - in a separate PR. There are two (soon three) PRs waiting in the queue for this changeset to be merged.

lukasz-zimnoch (dismissed):

solidity/contracts/bank/Bank.sol
```diff
@@ -22,33 +22,34 @@ import "../vault/IVault.sol";
 
 /// @title Bitcoin Bank
 /// @notice Bank is a central component tracking Bitcoin balances. Balances can
-///         be transferred between holders and holders can approve their
-///         balances to be spent by others. Balances in the Bank are updated for
-///         depositors who deposit their Bitcoin into the Bridge and only the
-///         Bridge can increase balances.
+///         be transferred between balance owners, and balane owners can approve
```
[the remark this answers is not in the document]
pdyraga: 😬 a32ae1999861ae99392fc13dad88715e65aab032

beaushinkle (approved):

solidity/contracts/bank/Bank.sol
```diff
@@ -138,8 +151,10 @@ contract Bank is Ownable {
         );
     }
 
-    /// @notice Atomically increases the balance allowance granted to `spender`
-    ///         by the caller by the given `addedValue`.
+    /// @notice Atomically increases the caller's balance allowance granted to
```
beaushinkle, line 154: What are we using the word "atomically" to mean here? In something concurrent, I would imagine anything atomic to happen threadsafe, either all or nothing. Is that the same here?

pdyraga: Let's say that balance owner A approved address B to use 100 units of A's balance in the Bank calling `Bank.approveBalance(B, 100)`.

After some time, A changed their mind and decided to change the allowance to 120 units. 

If A makes it by calling `Bank.approveBalance(B, 120)`, malicious B can front-run this transaction and first spend 100 units from the first approval, and then spend 120 units from the second approval, spending 220 total. Definitely not what was A's intention:
```
A: Bank.approveBalance(B, 100)
B: Bank.transferBalanceFrom(A, B, 100)
A: A: Bank.approveBalance(B, 120)
B: Bank.transferBalanceFrom(A, B, 120)
```

If A uses `Bank.increaseBalanceAllowance(B, 20)` instead, malicious B can spend no more than 120. This is why we say it happened *atomically*. It's similar to a database transaction where nothing can happen in between.

Issue #145, opened 2022-02-22
Typescript code for redemption proofs


pdyraga:
We need to add support for redemption proofs in the typescript tBTC v2 library. Needed for smart contract system tests.

PR #291, opened 2022-05-16
Redemption proof typescript

tomaszslabon:
Closes: https://github.com/keep-network/tbtc-v2/issues/145
This PR contains functionalities in typescript for assembling and broadcasting redemption transaction proofs:
- function `submitRedemptionProof` was added to the Bridge interface
- function `proveRedemption` was added to the `TBTC` interface.


typescript/src/index.ts
```diff
@@ -188,6 +189,24 @@ export interface TBTC {
     bitcoinClient: BitcoinClient
   ): Promise<void>
 
+  /**
+   * Prepares the proof of a redemption transaction and submits it to the
+   * Bridge on-chain contract.
+   * @param transactionHash - Hash of the transaction being proven.
+   * @param mainUtxo - Recent main UTXO of the wallet as currently known on-chain.
+   * @param walletPubKeyHash - 20-byte public key hash of the wallet
+   * @param bridge - Interface to the Bridge on-chain contract.
```
lukasz-zimnoch, line 198: `Interface to the Bridge on-chain contract.` -> `Handle to the Bridge on-chain contract.` and let's populate it to `redemption.ts`.

typescript/src/index.ts
```diff
@@ -188,6 +189,24 @@ export interface TBTC {
     bitcoinClient: BitcoinClient
   ): Promise<void>
 
+  /**
+   * Prepares the proof of a redemption transaction and submits it to the
+   * Bridge on-chain contract.
+   * @param transactionHash - Hash of the transaction being proven.
+   * @param mainUtxo - Recent main UTXO of the wallet as currently known on-chain.
+   * @param walletPubKeyHash - 20-byte public key hash of the wallet
```
lukasz-zimnoch, line 197: I think we should use `walletPublicKey` and do the hashing inside (using `computeHash160` from `bitcoin.ts`). This way we'll have the same interface as during deposits so writing system tests may be easier as we'll need to rely just on the PK and don't bother about hashing. Also, when we document such a hex values, let's explicitly put their byte size and the note they are an un-prefixed (or 02/03 prefixed like the compressed keys) hex string.

typescript/src/chain.ts
```diff
@@ -45,6 +46,20 @@ export interface Bridge {
...
+   * @param redemptionTx - Redemption transaction data
+   * @param redemptionProof - Redemption proof data
+   * @param mainUtxo - Data of the wallets main UTXO
+   * @param walletPubKeyHash - 20-byte public key hash of the wallet
+   */
+  submitRedemptionProof(
+    redemptionTx: DecomposedRawTransaction,
+    redemptionProof: Proof,
+    mainUtxo: UnspentTransactionOutput,
+    walletPubKeyHash: BytesLike
```
lukasz-zimnoch, line 60: Just like I mentioned in `index.ts`, let's use plain-text `walletPublicKey`. This will be consistent with what we have in `revealDeposit` where `deposit: Deposit` has a plain-text key of the wallet too. Bridge implementation will do the appropriate hashing under the hood. Also, tying the API with plain-text key of the wallet seems more flexible since we'll also need to deal with frauds in future where the key is needed for signature verification.

typescript/src/redemption.ts
```diff
@@ -264,3 +265,38 @@ export async function createRedemptionTransaction(
...
+  bitcoinClient: BitcoinClient
+): Promise<void> {
+  const confirmations = await bridge.txProofDifficultyFactor()
+  const proof = await createTransactionProof(
+    transactionHash,
+    confirmations,
+    bitcoinClient
+  )
+  // TODO: instead of getting rawTransaction, use transaction part of proof and
+  // convert it to raw transaction.
```
lukasz-zimnoch, line 293: Can we put a note (as well in the same TODO in `proveDepositSweep`) explaining why we can't do it right now?

typescript/test/utils/mock-bridge.ts
```diff
@@ -20,6 +20,13 @@ interface RevealDepositLogEntry {
   deposit: Deposit
 }
 
+interface RedemptionBridgeLogEntry {
```
lukasz-zimnoch, line 23: Maybe just `RedemptionProofLogEntry`? This will be consistent with `DepositSweepProofLogEntry`

lukasz-zimnoch (changes requested):
Some initial comments. Once addressed, I'll do a thorough pass of the code and tests once again.

typescript/src/chain.ts
```diff
@@ -45,6 +46,20 @@ export interface Bridge {
     deposit: Deposit
   ): Promise<void>
 
+  /**
+   * Submits a redemption transaction proof to the on-chain contract.
+   * @param redemptionTx - Redemption transaction data
+   * @param redemptionProof - Redemption proof data
+   * @param mainUtxo - Data of the wallets main UTXO
```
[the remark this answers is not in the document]
tomaszslabon: Done in 3a3773cd762884db129225bae7961b3554ba4eeb.

typescript/src/index.ts
```diff
@@ -188,6 +189,24 @@ export interface TBTC {
...
+   * @param bridge - Interface to the Bridge on-chain contract.
+   * @param bitcoinClient - Bitcoin client used to interact with the network.
+   * @returns Empty promise.
+   */
+  proveRedemption(
```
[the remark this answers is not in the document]
tomaszslabon: Done in cf3c5717a4e516dde69054a5ef4337f73bff6a24.

typescript/src/index.ts
```diff
@@ -255,6 +274,7 @@ const tbtc: TBTC = {
   sweepDeposits,
   createDepositSweepTransaction,
   proveDepositSweep,
+  proveRedemption,
```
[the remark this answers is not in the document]
tomaszslabon: Done in cf3c5717a4e516dde69054a5ef4337f73bff6a24.

typescript/src/chain.ts
tomaszslabon: Done in 3491b959b2105173486d028e91926e68abfeaa6b.

typescript/src/index.ts
tomaszslabon: Done in 3491b959b2105173486d028e91926e68abfeaa6b.

typescript/test/utils/mock-bridge.ts
tomaszslabon: Done in 75b561e1a724388f482970378e5a348d1ce35dbb.

typescript/src/redemption.ts
tomaszslabon: Actually, I don't know why we cannot do that now. Can't we just use `rawTransaction` downloaded from from the `bitcoinClient`? Using the converter may result in fewer network connections though.

lukasz-zimnoch: This is because we don't have the right converter yet. We can use `rawTransaction` for now. My comment was about adding an explanation saying what should be done to consider this TODO as addressed.

tomaszslabon: Done in d4b6523c143327cd52a483bc6d6e73764edf194a.

typescript/test/utils/mock-bridge.ts
```diff
@@ -38,7 +46,11 @@ export class MockBridge implements Bridge {
   }
 
   get revealDepositLogEntry(): RevealDepositLogEntry[] {
```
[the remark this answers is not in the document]
tomaszslabon: Done in fb30e89b5014f0d31fb3b46ea115598f455530aa.

lukasz-zimnoch (approved):

PR #295, opened 2022-05-16
Deployments updates

nkuba:
In this PR we solve problems with deployment using the external `WalletRegistry` deployment script.

With the changes from 821c971bcdeca039aff16e182c45403eef7c881d the `WalletRegistry` contract will be deployed as a dependency, so the deployment cycle will use proper implementation of ecdsa contracts.

Depends on https://github.com/keep-network/keep-core/pull/2993
Closes https://github.com/keep-network/tbtc-v2/issues/267


solidity/deploy/05_deploy_bridge.ts
```diff
@@ -24,7 +24,11 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     log: true,
   })
   const Redemption = await deploy("Redemption", { from: deployer, log: true })
-  const Wallets = await deploy("Wallets", { from: deployer, log: true })
+  const Wallets = await deploy("Wallets", {
+    contract: "contracts/bridge/Wallets.sol:Wallets",
```
pdyraga, line 28: Why do we need it?

nkuba: >Wallets contract name duplicates with @keep-network/ecdsa. We need to be
specific in the deployment script which Wallets we want to deploy.

20ab23db71d5042164e047eab0eca1d6f49c7a75

pdyraga (approved):

PR #290, opened 2022-05-16
TBTCVault is unminting TBTC to Balance instead of redeeming

pdyraga:
Depends on https://github.com/keep-network/tbtc-v2/pull/287
Refs https://github.com/keep-network/tbtc-v2/issues/256
Refs https://github.com/keep-network/tbtc-v2/issues/289

This change cleans up the nomenclature. `Bridge` has `deposit` and `redemption`
actions while `TBTCVault` has `mint` and `unmint` actions. This way, the act of
going from TBTC to `Bank` balance is `unminting`, and the act of going from
`Bank` balance to BTC is `redemption`.

PR #296, opened 2022-05-16
Bridge.receiveBalanceApproval

### LINKED ISSUE #289: One-transaction balance to BTC
Currently, going from Bank balance to BTC requires two transactions:
- `Bank.approve(Bridge)`,
- `Bridge.requestRedemption`.

We need to shorten this path, ideally to a single transaction.

pdyraga:
Depends on #290
Refs #256
Closes #289

Previously, it was needed to execute two separate transactions to redeem `Bank`
balance to BTC: `Bank.approve` and `Bridge.requestRedemption`. The change
extracting `receiveBalanceApproval` out of `IVault` to a separate
`IReceiveApproval` interface allowed to integrate it with `Bridge` contract that
can now receive the approved balance by requesting redemption. This way, redeemer
can call `Bank.approveBalanceAndCall` with `Bridge` contract to request
redemption in a single transaction.

Next to Bitcoin-specific data passed as `bytes extraData` to
`approveBalanceAndCall /receiveBalanceApproval`, we also pass the redeemer
address that can be different from the `balanceOwner` address. A real-world use
of this functionality that will soon be implemented for `TBTCVault`
is when someone wants to go from vault balance straight to BTC. Then, the vault
is the balance owner but someone else is the redeemer who can claim the
balance back in case something goes wrong during the redemption.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -19,6 +19,7 @@ import "@keep-network/random-beacon/contracts/Governable.sol";
 import {IWalletOwner as EcdsaWalletOwner} from "@keep-network/ecdsa/contracts/api/IWalletOwner.sol";
 
 import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/utils/math/SafeCastUpgradeable.sol";
 
 import "./IRelay.sol";
 import "./BridgeState.sol";
@@ -31,6 +32,7 @@ import "./Wallets.sol";
 import "./Fraud.sol";
 import "./MovingFunds.sol";
 
+import "../bank/IReceiveBalanceApproval.sol";
 import "../bank/Bank.sol";
 
 /// @title Bitcoin Bridge
@@ -54,7 +56,12 @@ import "../bank/Bank.sol";
 /// @dev Bridge is an upgradeable component of the Bank. The order of
 ///      functionalities in this contract is: deposit, sweep, redemption,
 ///      moving funds, wallet lifecycle, frauds, parameters.
-contract Bridge is Governable, EcdsaWalletOwner, Initializable {
+contract Bridge is
+    Governable,
+    EcdsaWalletOwner,
+    Initializable,
+    IReceiveBalanceApproval
+{
     using BridgeState for BridgeState.Storage;
     using Deposit for BridgeState.Storage;
     using DepositSweep for BridgeState.Storage;
@@ -386,7 +393,9 @@ contract Bridge is Governable, EcdsaWalletOwner, Initializable {
     }
 
     /// @notice Requests redemption of the given amount from the specified
-    ///         wallet to the redeemer Bitcoin output script.
+    ///         wallet to the redeemer Bitcoin output script. Handles the
+    ///         simplest case in which the redeemer's balance is decreased in
+    ///         the Bank.
     /// @param walletPubKeyHash The 20-byte wallet public key hash (computed
     ///        using Bitcoin HASH160 over the compressed ECDSA public key).
     /// @param mainUtxo Data of the wallet's main UTXO, as currently known on
@@ -394,8 +403,8 @@ contract Bridge is Governable, EcdsaWalletOwner, Initializable {
     /// @param redeemerOutputScript The redeemer's length-prefixed output
     ///        script (P2PKH, P2WPKH, P2SH or P2WSH) that will be used to lock
     ///        redeemed BTC.
-    /// @param amount Requested amount in satoshi. This is also the TBTC amount
-    ///        that is taken from redeemer's balance in the Bank upon request.
+    /// @param amount Requested amount in satoshi. This is also the Bank balance
+    ///        that is taken from the `balanceOwner` upon request.
     ///        Once the request is handled, the actual amount of BTC locked
     ///        on the redeemer output script will be always lower than this value
     ///        since the treasury and Bitcoin transaction fees must be incurred.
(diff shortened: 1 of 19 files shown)


pdyraga:
https://github.com/keep-network/tbtc-v2/pull/290 is merged so I am undrafting this PR.

solidity/contracts/bridge/Redemption.sol
```diff
@@ -258,15 +264,163 @@ library Redemption {
...
+    ///        script (P2PKH, P2WPKH, P2SH or P2WSH) that will be used to lock
+    ///        redeemed BTC,
+    ///        - redeemer: The Ethereum address of the redeemer who will be able
+    ///        to claim Bank balance if anything goes wrong during the redemption.
+    ///        In the most basic case, when someone redeems their Bitcoin
+    ///        balance from the Bank, `balanceOwner` is the same as `redemeer`.
+    ///        However, when a Vault is redeeming part of its balance for some
+    ///        redeemer address (for example, someone who has earlier deposited
+    ///        into that Vault), `balanceOwner` is the Vault, and `redemeer` is
+    ///        the address for which the vault is redeeming its balance to.
```
lukasz-zimnoch, line 330: Can we explain parameters in the same order as they are encoded in the `redemptionData` ?

solidity/contracts/bridge/Bridge.sol
```diff
@@ -423,11 +432,82 @@ contract Bridge is Governable, EcdsaWalletOwner, Initializable {
...
+    ///        script (P2PKH, P2WPKH, P2SH or P2WSH) that will be used to lock
+    ///        redeemed BTC,
+    ///        - redeemer: The Ethereum address of the redeemer who will be able
+    ///        to claim Bank balance if anything goes wrong during the redemption.
+    ///        In the most basic case, when someone redeems their balance
+    ///        from the Bank, `balanceOwner` is the same as `redemeer`.
+    ///        However, when a Vault is redeeming part of its balance for some
+    ///        redeemer address (for example, someone who has earlier deposited
+    ///        into that Vault), `balanceOwner` is the Vault, and `redemeer` is
+    ///        the address for which the vault is redeeming its balance to.
```
lukasz-zimnoch, line 484: Can we explain parameters in the same order as they are encoded in the `redemptionData` ?

solidity/contracts/bridge/Bridge.sol
```diff
@@ -423,11 +432,82 @@ contract Bridge is Governable, EcdsaWalletOwner, Initializable {
...
+    /// @dev Requirements:
```
lukasz-zimnoch, line 485: Worth noting that the caller must be the `Bank`.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -423,11 +432,82 @@ contract Bridge is Governable, EcdsaWalletOwner, Initializable {
...
+    ///      - `mainUtxo` components must point to the recent main UTXO
+    ///        of the given wallet, as currently known on the Ethereum chain,
+    ///      - `redeemerOutputScript` must be a proper Bitcoin script,
+    ///      - `redeemerOutputScript` cannot have wallet PKH as payload,
+    ///      - `amount` must be above or equal the `redemptionDustThreshold`,
+    ///      - Given `walletPubKeyHash` and `redeemerOutputScript` pair can be
+    ///        used for only one pending request at the same time,
+    ///      - Wallet must have enough Bitcoin balance to proceed the request,
+    ///      - Balance owner must make an allowance in the Bank that the Bridge
+    ///        contract can spend the given `amount`.
```
lukasz-zimnoch, line 496: This may suggest that the balance owner needs to make a separate allowance. Maybe it will be better to remove that point as the philosophy behind `receiveBalanceApproval` implies that the allowance was already made?

solidity/contracts/bridge/Redemption.sol
```diff
@@ -258,15 +264,163 @@ library Redemption {
...
+    ///        of the given wallet, as currently known on the Ethereum chain,
+    ///      - `redeemerOutputScript` must be a proper Bitcoin script,
+    ///      - `redeemerOutputScript` cannot have wallet PKH as payload,
+    ///      - `amount` must be above or equal the `redemptionDustThreshold`,
+    ///      - Given `walletPubKeyHash` and `redeemerOutputScript` pair can be
+    ///        used for only one pending request at the same time,
+    ///      - Wallet must have enough Bitcoin balance to proceed the request,
+    ///      - Balance owner must make an allowance in the Bank that the Bridge
+    ///        contract can spend the given `amount`.
+    function requestRedemption(
```
lukasz-zimnoch, line 343: Do we really need that function? I think we can decode the `redemptionData` directly in `Bridge.receiveBalanceApproval` and just call the internal `requestRedemption` function. The whole thing with encoding extra data is the concept related to the receive approval mechanism so it could make sense to keep it there. Also, that change will remove a level of indirection and all the burden related to code and documentation maintenance. I think the same applies to the second `requestRedemption` version.

solidity/test/bridge/Bridge.Redemption.test.ts
```diff
@@ -704,6 +704,232 @@ describe("Bridge - Redemption", () => {
...
+      redemptionTxMaxFee = (await bridge.redemptionParameters())
+        .redemptionTxMaxFee
+
+      balanceOwner = thirdParty
+      // eslint-disable-next-line prefer-destructuring
+      redeemer = (await getUnnamedAccounts())[10]
+
+      // Simulate the balance owner has a twice Bank balance allowing to make the request.
+      await bank.setBalance(balanceOwner.address, requestedAmount.mul(2))
+    })
```
lukasz-zimnoch, line 732: Looks like we modify the contract state here. It would be good to add a `createSnapshot` call before and an `after` section with `restoreSnapshot` to avoid strange failures in the future.

solidity/test/bridge/Bridge.Redemption.test.ts
```diff
@@ -704,6 +704,232 @@ describe("Bridge - Redemption", () => {
...
+            await bridge.setWalletMainUtxo(walletPubKeyHash, mainUtxo)
+          })
+
+          after(async () => {
+            await restoreSnapshot()
+          })
+
+          context("when main UTXO data are valid", () => {
+            context("when redeemer output script is standard type", () => {
+              // Arbitrary standard output scripts.
```
lukasz-zimnoch, line 783: s/scripts/script

solidity/contracts/bridge/Redemption.sol
pdyraga: Yeah, definitely that's how they should be ordered. Fixed in https://github.com/keep-network/tbtc-v2/pull/296/commits/88e7b7571363650eda82a0ffd68db81f2b276a54.

solidity/contracts/bridge/Bridge.sol
pdyraga: https://github.com/keep-network/tbtc-v2/pull/296/commits/88e7b7571363650eda82a0ffd68db81f2b276a54

solidity/contracts/bridge/Bridge.sol
pdyraga: Right. Removed in https://github.com/keep-network/tbtc-v2/pull/296/commits/88e7b7571363650eda82a0ffd68db81f2b276a54.

solidity/test/bridge/Bridge.Redemption.test.ts
pdyraga: https://github.com/keep-network/tbtc-v2/pull/296/commits/5da51b05bcf10d39f0835dbd28f7cae3337b8518

solidity/test/bridge/Bridge.Redemption.test.ts
pdyraga: Good catch. Fixed in https://github.com/keep-network/tbtc-v2/pull/296/commits/5da51b05bcf10d39f0835dbd28f7cae3337b8518.

solidity/contracts/bridge/Redemption.sol
pdyraga: This was my first version but unfortunately, the compiler fails with a known stack too deep error on `SafeCastUpgradeable.toUint64(amount)` line. Decomposing it into two functions helps.

lukasz-zimnoch: Ok, got it!

lukasz-zimnoch (approved):

PR #302, opened 2022-05-18
One-transaction TBTC to BTC

### LINKED ISSUE #256: One-transaction TBTC to BTC

pdyraga:
Closes #256 

Modified `TBTCVault.receiveApproval` function to allow unminting TBTC and
requesting Bank balance redemption in the Bridge in a single transaction via
`TBTC.approveAndCall`. Also, added `TBTCVault.redeem` function allowing to
unmint TBTC and request redemption in the `Bridge` on an already approved TBTC
balance.

During the work on this feature, I have been exploring different options,
including exposing `IBridge` interface that would be called from `Bank` or
`TBTCVault` to redeem balance. The current version won because it does not
tightly-couple vault implementations with the specific `Bridge` implementation.
For example, the second version of the `Bridge` may not even have a concept of
a wallet or a main wallet UTXO. Passing parameters as bytes is more flexible and
uses an existing `Bank` mechanism of `receiveBalanceApproval`.

### DIFF
--- a/solidity/contracts/vault/TBTCVault.sol
+++ b/solidity/contracts/vault/TBTCVault.sol
@@ -136,7 +136,7 @@ contract TBTCVault is IVault, Governable {
         }
     }
 
-    /// @notice Burns `amount` of TBTC from the caller's account and transfers
+    /// @notice Burns `amount` of TBTC from the caller's balance and transfers
     ///         `amount` back to the caller's balance in the Bank.
     /// @dev Caller must have at least `amount` of TBTC approved to
     ///       TBTC Vault.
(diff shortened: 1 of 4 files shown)


solidity/test/fixtures/bridge.ts
```diff
@@ -19,6 +21,10 @@ export default async function bridgeFixture() {
     await helpers.signers.getNamedSigners()
   const [thirdParty] = await helpers.signers.getUnnamedSigners()
 
+  const tbtc: TBTC = await helpers.contracts.getContract("TBTC")
+
+  const tbtcVault: TBTCVault = await helpers.contracts.getContract("TBTCVault")
```
lukasz-zimnoch, line 26: Shouldn't we also give ownership of `TBTC` to the `TBTCVault`? It would be good to mimic the real-world deployment as much as possible. Also, if we do that, we'll be able to skip this step in `TBTCVault.Redemption.test.ts`.

solidity/test/vault/TBTCVault.Redemption.test.ts
```diff
@@ -0,0 +1,580 @@
...
+      })
+    })
+
+    context("when there is a single redeemer", () => {
+      const redeemerOutputScriptP2WPKH =
+        "0x160014f4eedc8f40d4b8e30771f792b065ebec0abaddef"
+      const redeemerOutputScriptP2WSH =
+        "0x220020ef0b4d985752aa5ef6243e4c6f6bebc2a007e7d671ef27d4b1d0db8dcc93bc1c"
+      const redeemerOutputScriptP2PKH =
+        "0x1976a914f4eedc8f40d4b8e30771f792b065ebec0abaddef88ac"
```
lukasz-zimnoch, line 164: Let's also test the `P2SH` case.

solidity/contracts/vault/TBTCVault.sol
```diff
@@ -145,23 +145,47 @@ contract TBTCVault is IVault, Governable {
...
-    ///      execute unminting without an additional approval transaction.
-    ///      The function can be called only via `approveAndCall` of TBTC token.
+    /// @notice Burns `amount` of TBTC from the caller's balance and transfers
+    ///         `amount` of Bank balance to the Bridge requesting redemption
+    ///         based on the provided `redemptionData`.
+    /// @param amount Amount of TBTC to unmint and request to redeem in Bridge.
+    /// @param redemptionData Redemption data in a format expected from
+    ///        `redemptionData` parameter of Bridge's `receiveBalanceApproval`
+    ///        function.
+    function redeem(uint256 amount, bytes calldata redemptionData) external {
```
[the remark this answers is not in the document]
pdyraga: I think it does not because *unminting* is "TBTC to Bank balance" and *redemption* is "Bank balance to BTC". In this case, *unmint* has to happen before *redeem* and we are doing internally.

But... I see your point. I think `unmintAndApprove` is too cryptic. But what about `unmintAndRedeem`?

solidity/test/fixtures/bridge.ts
lukasz-zimnoch: Got it! Thanks for the clarification!

solidity/test/vault/TBTCVault.Redemption.test.ts
pdyraga: Makes sense! See 08228f958a885618b33e435dd7253dd872fe3fe2.

solidity/contracts/vault/TBTCVault.sol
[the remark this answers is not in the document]
pdyraga: See e12b1cf143b7003420b83df4a49a84f1bcd06a5d please.

lukasz-zimnoch (approved):

Issue #250, opened 2022-05-03
Integration tests


pdyraga:
# Wallet creation unit tests

We need to confirm `WalletRegistry` <> `Bridge` interactions work as intended. [`Bridge.requestNewWallet`](https://github.com/keep-network/tbtc-v2/blob/00076d66ab1e4233f9e40817305de7dd30425d19/solidity/contracts/bridge/Bridge.sol#L802) should lead to a new ECDSA wallet being created.

# Slashing integration tests

In #238, #236, #234 we implemented slashing for redemption timeout, fraud, and moving funds timeout. Tests are based on mocked `WalletRegistry`.

We need to implement separate tests working with a real `WalletRegistry` and Threshold `TokenStaking` to confirm the gas cost of slashing and be 100% sure the flow works correctly.

# Deposit -> Redeem path

We should cover deposit reveal -> deposit sweep -> redemption request -> redemption proof happy path using Bitcoin testnet data from unit tests. No integration with Bitcoin node or Electrum should be done. This will be covered by system tests (see #143)

PR #304, opened 2022-05-18
Integration tests: create new wallet

nkuba:
Depends on https://github.com/keep-network/tbtc-v2/pull/300
Refs https://github.com/keep-network/tbtc-v2/issues/250

In this PR we add an initial integration test to create a new wallet.
The test uses a real implementation of the ECDSA Wallet Registry and mocked Random Beacon.

In another PR we will handle Random Beacon deployment to be able to use the real contract.

Tests for other paths (e.g. slashing) will be added in other PRs.


solidity/test/integration/utils/ecdsa-wallet-registry.ts
```diff
@@ -0,0 +1,269 @@
...
+  await walletRegistry.connect(operator).joinSortitionPool()
+
+  const operatorID = await sortitionPool.getOperatorID(
+    await operator.getAddress()
+  )
+
+  return operatorID
+}
+
+export async function produceEcdsaDkgResult(
```
pdyraga, line 40: Maybe `performDkg`? It's not only about producing a result that is later submitted by some other function but also about selecting group members, submitting and approving that result. It's a full DKG path.

solidity/test/integration/utils/ecdsa-wallet-registry.ts
```diff
@@ -0,0 +1,269 @@
...
+}
+
+async function selectGroup(
+  sortitionPool: Contract,
+  seed: BigNumber
+): Promise<Operator[]> {
+  const identifiers = await sortitionPool.selectGroup(
+    WALLET_SIZE,
+    ethers.utils.hexZeroPad(seed.toHexString(), 32)
+  )
```
pdyraga, line 105: Why don't we use [`selectGroup` from `WalletRegistry`](https://github.com/keep-network/keep-core/blob/9cf3620dbd8ccd016dc193b2eb23f0a2509dc87c/solidity/ecdsa/contracts/WalletRegistry.sol#L1150-L1152)? This is what the client will call, it does not need to bother about the `SortitionPool`.

Also, if we use this function I think we should be able to get rid of `calculateDkgSeed` and all the complexity calculating and passing the seed introduces.

solidity/test/integration/WalleCreation.test.ts
```diff
@@ -0,0 +1,119 @@
...
+  before(async () => {
+    ;({ governance, bridge, walletRegistry, randomBeacon } =
+      await waffle.loadFixture(fixture))
+
+    // Update only the parameters that are crucial for this test.
+    await updateWalletRegistryDkgResultChallengePeriodLength(
+      walletRegistry,
+      governance,
+      dkgResultChallengePeriodLength
+    )
```
pdyraga, line 51: And why do we update it in the first place? The test would take too long to execute?

solidity/test/integration/WalleCreation.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    })
+
+    it("should register a new wallet details in the Bridge", async () => {
+      const storedWallet = await bridge.wallets(walletPubKeyHash)
+
+      expect(storedWallet.ecdsaWalletID).to.be.equal(
+        ecdsaWalletTestData.walletID
+      )
+
+      expect(storedWallet.state).to.be.equal(1)
```
pdyraga, line 94: We can use:

https://github.com/keep-network/tbtc-v2/blob/4cd75f5a691a08de4eb56a16a27ec0f09e1008d4/solidity/test/fixtures/index.ts#L37-L44

Example:
https://github.com/keep-network/tbtc-v2/blob/4cd75f5a691a08de4eb56a16a27ec0f09e1008d4/solidity/test/bridge/Bridge.Wallets.test.ts#L608

solidity/test/integration/WalleCreation.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      expect(await bridge.activeWalletPubKeyHash()).to.be.equal(
+        walletPubKeyHash
+      )
+    })
+
+    it("should consume around 93 000 gas for Bridge.requestNewWallet transaction", async () => {
+      await assertGasUsed(requestNewWalletTx, 93_000)
+    })
+
+    // TODO: Should we also validate gas used by Random Beacon and ECDSA Wallet Registry transactions?
```
pdyraga, line 117: At a minimum, the one approving result - it's calling the `Bridge` in a callback.

solidity/hardhat.config.ts
```diff
@@ -55,7 +70,16 @@ const config: HardhatUserConfig = {
...
+      accounts: {
+        // Number of accounts that should be predefined on the testing environment.
+        count:
+          testConfig.nonStakingAccountsCount +
+          testConfig.stakingRolesCount * testConfig.operatorsCount,
+      },
       tags: ["local"],
+      // Ignore contract size on deployment to hardhat network, to be able to
+      // deploy stub contracts in tests.
+      allowUnlimitedContractSize: process.env.TEST_USE_STUBS_TBTC === "true",
```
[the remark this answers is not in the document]
nkuba: Have you also tried running `yarn test`? The unit tests use stubs.

solidity/test/integration/utils/ecdsa-wallet-registry.ts
nkuba: Good suggestion. I'd name it `performEcdsaDkg` to be more precise, as we will have a similar for the random beacon.

solidity/test/integration/utils/ecdsa-wallet-registry.ts
nkuba: Nice catch! Yeah, I missed this function. I copied the code from ecdsa's test utils.

solidity/test/integration/WalleCreation.test.ts
nkuba: Yeah exactly, the test hangs for a really long time.

solidity/test/integration/utils/ecdsa-wallet-registry.ts
nkuba: c8f7d9b60cdd121f41bcac32bf2a2495b668bdad

solidity/test/integration/utils/ecdsa-wallet-registry.ts
pdyraga: Sure, `performEcdsaDkg` works.

nkuba: 006efd4e6d1316b21b322933c16fbff7be5b28e0

solidity/test/integration/WalleCreation.test.ts
nkuba: e438c884aca08cc30d869c7c7591460771bfa1c9

solidity/test/integration/WalleCreation.test.ts
nkuba: 04f418bc529cc75f870c133e53dd0b5ac99a9b40

solidity/hardhat.config.ts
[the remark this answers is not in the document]
pdyraga: Unit tests should not be affected by changes in this PR, right?

I have just tried running both `yarn test` and `yarn test:integration` with the two lines earlier mentioned commented out and everything is green.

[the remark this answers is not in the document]
nkuba: You're right! I changed the approach to the interproject deployments so many times that I got confused.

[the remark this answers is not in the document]
nkuba: 7e5165fef2e9def2af36f7ddbe29c2028a2052a4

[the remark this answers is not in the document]
pdyraga: We should also remove the line at 176:
```
   // Don't check size of contracts used as stubs in tests.
    except: ["contracts/test/.*Stub$"],
```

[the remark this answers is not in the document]
nkuba: a6cee7a712cad1ba39091fe95f9a3eb36e750317

pdyraga (approved):

PR #311, opened 2022-05-19
Allow the governance to recover ERC20 and ERC721 from TBTCVault 

pdyraga:
~~Depends on #309. Keeping as a draft until #309 is not merged.~~
Refs: https://github.com/keep-network/tbtc-v2/issues/39

ERC20 and ERC721 can be sent mistakenly to `TBTCVault`, two functions allow the
governance to recover them. Most importantly though, the governance can use
`recoverERC20` to recover TBTCv1 tokens sent by `VendingMachine` to `TBTCVault`
after the upgrade.

PR #312, opened 2022-05-19
Transfer ownership of TBTCVault to the Governance

pdyraga:
~~Depends on #311. Keeping as a draft until #311 is not merged.~~
Refs: #39

Modified deployment scripts by inserting a step transferring `TBTCVault`
ownership to the Governance address.

Also, I replaced `Governable` with `Ownable` in `TBTCVault`. Note that `TBTCVault`
and `Bank` are not deployed behind a proxy so they can use `Ownable`. What is
more, using `Ownable` vs `Governable` should let us differentiate which contract
is owned by a multisig (or DAO) and which contract is owned by an additional
governance contract:

- `Bank` is `Ownable` and owned by multisig/DAO
- `TBTCVault` is `Ownable` and owned by multisig/DAO
- `Bridge` is `Governable` and is owned by `BridgeGovernance` (work in progress)
  that is owned by multisig/DAO
- `WalletRegistry` is `Governable` and is owned by `WalletRegistryGovernance`
  that is owned by multisig/DAO
- `RandomBeacon` is `Governable` and is owned by `RandomBeaconGovernance`
  that is owned by multisig/DAO


tomaszslabon (approved):

pdyraga:
#311 is merged so I am undrafting this PR.

solidity/test/vault/TBTCVault.test.ts
```diff
@@ -128,7 +128,7 @@ describe("TBTCVault", () => {
             account1.address,
             to1e18(800)
           )
-        ).to.be.revertedWith("Caller is not the governance")
+        ).to.be.revertedWith("Ownable: caller is not the owner")
```
lukasz-zimnoch, line 131: Would be good to change all contexts' names from `when called not by the governance` to `when called not by the owner`

pdyraga: I was thinking about it but the owner of the vault is still governance (DAO/council). It's just not the intermediary governance contract.

lukasz-zimnoch (approved):

PR #314, opened 2022-05-19
Integration tests: slashing

nkuba:


solidity/test/integration/Slashing.test.ts
```diff
@@ -0,0 +1,531 @@
...
+          }
+        })
+
+        it("should close the wallet in the wallet registry", async () => {
+          // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+          expect(await walletRegistry.isWalletRegistered(ecdsaWalletID)).to.be
+            .false
+        })
+
+        it("should close the wallet in the bridge", async () => {
```
lukasz-zimnoch, line 183:
```suggestion
        it("should terminate the wallet in the bridge", async () => {
```

solidity/test/integration/Slashing.test.ts
```diff
@@ -0,0 +1,531 @@
...
+        await updateDepositDustThreshold(10_000) // 0.0001 BTC
+        await updateRedemptionDustThreshold(2_000) // 0.00002 BTC
+
+        await bridge.revealDeposit(fundingTx, reveal)
+
+        // TODO: Replace mocks with the real implementation
+        relay.getCurrentEpochDifficulty.returns(
+          SingleP2SHDeposit.chainDifficulty
+        )
+        relay.getPrevEpochDifficulty.returns(SingleP2SHDeposit.chainDifficulty)
```
lukasz-zimnoch, line 242: See the comment about the relay in `fixture.ts`.

solidity/test/integration/Slashing.test.ts
```diff
@@ -0,0 +1,531 @@
...
+          }
+        })
+
+        it("should not close the wallet in the wallet registry", async () => {
+          // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+          expect(await walletRegistry.isWalletRegistered(ecdsaWalletID)).to.be
+            .true
+        })
+
+        it("should transition the wallet in the bridge to moving funds state", async () => {
```
lukasz-zimnoch, line 337: Let's use the original state name as we usually do in unit tests:
```suggestion
        it("should transition the wallet in the bridge to the MovingFunds state", async () => {
```

Also, worth dropping a comment where we will explain that the wallet was switched to `MovingFunds` state because its BTC balance was non-zero.

solidity/test/integration/Slashing.test.ts
```diff
@@ -0,0 +1,531 @@
...
+          SingleP2SHDeposit.chainDifficulty
+        )
+        relay.getPrevEpochDifficulty.returns(SingleP2SHDeposit.chainDifficulty)
+
+        await bridge.submitDepositSweepProof(
+          SingleP2SHDeposit.sweepTx,
+          SingleP2SHDeposit.sweepProof,
+          SingleP2SHDeposit.mainUtxo,
+          tbtcVault.address
+        )
```
lukasz-zimnoch, line 397: We do deposit reveal and sweep here in order to have a non-zero BTC balance right? If so, can we drop a comment clarifying that?

solidity/test/integration/Slashing.test.ts
```diff
@@ -0,0 +1,531 @@
...
+          }
+        })
+
+        it("should close the wallet in the wallet registry", async () => {
+          // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+          expect(await walletRegistry.isWalletRegistered(ecdsaWalletID)).to.be
+            .false
+        })
+
+        it("should close the wallet in the bridge", async () => {
```
lukasz-zimnoch, line 471:
```suggestion
        it("should terminate the wallet in the bridge", async () => {
```

solidity/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "target": "ES2018",
```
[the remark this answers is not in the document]
nkuba: We use syntax that requires a newer target:
https://github.com/keep-network/tbtc-v2/blob/361ce054cb9e68cb9bb85b20396cade5c6a52cab/solidity/test/integration/utils/ecdsa-wallet-registry.ts#L173-L181

Without the target we get `TypeError: walletMembers.getIds is not a function`.

We could either refactor the code or use the target.

solidity/test/integration/Slashing.test.ts
```diff
@@ -0,0 +1,531 @@
...
+} from "../data/fraud"
+import { walletState } from "../fixtures"
+import { SingleP2SHDeposit } from "../data/deposit-sweep"
+import { UTXOStruct } from "../../typechain/Bridge"
+
+const { increaseTime } = helpers.time
+const { createSnapshot, restoreSnapshot } = helpers.snapshot
+
+const NO_MAIN_UTXO = {
+  txHash: "0x0000000000000000000000000000000000000000000000000000000000000000",
```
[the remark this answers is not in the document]
nkuba: We can even import `NO_MAIN_UTXO` from test data 
962ea2e8c72f8addf444d2adf8f69fa4013343ca

solidity/test/integration/utils/fixture.ts
```diff
@@ -46,6 +71,11 @@ export const fixture = deployments.createFixture(
       await walletRegistry.sortitionPool()
     )
 
+    // TODO: INTEGRATE WITH THE REAL RELAY
```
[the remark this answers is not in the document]
nkuba: a1746f05435fa2d172fb3164c2b1e7f7b6980254

solidity/test/integration/Slashing.test.ts
nkuba: 5a91ee6fdedf3ce4a3e707ffc0c4c1013eca15ea

solidity/test/integration/Slashing.test.ts
```diff
@@ -0,0 +1,531 @@
...
+        // so we need to update the dust threshold to be below it.
+        await updateDepositDustThreshold(10000) // 0.0001 BTC)
+
+        await bridge.revealDeposit(fundingTx, reveal)
+
+        // TODO: Replace mocks with the real implementation
+        relay.getCurrentEpochDifficulty.returns(
+          SingleP2SHDeposit.chainDifficulty
+        )
+        relay.getPrevEpochDifficulty.returns(SingleP2SHDeposit.chainDifficulty)
```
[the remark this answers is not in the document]
nkuba: 5a91ee6fdedf3ce4a3e707ffc0c4c1013eca15ea

solidity/test/integration/Slashing.test.ts
nkuba: d21abed2d94e7f48ec126fefca9149ab2accbe3f

solidity/test/integration/Slashing.test.ts
```diff
@@ -0,0 +1,529 @@
...
+
+    // Update only the parameters that are crucial for this test.
+    await updateWalletRegistryDkgResultChallengePeriodLength(
+      walletRegistry,
+      governance,
+      dkgResultChallengePeriodLength
+    )
+  })
+
+  describe("notifyFraudChallengeDefeatTimeout", async () => {
```
tomaszslabon, line 85: There are four scenarios where slashing happens:
- `notifyFraudChallengeDefeatTimeout`
- `notifyRedemptionTimeout`
- `notifyMovingFundsTimeout`
- `notifyMovedFundsSweepTimeout`

We have integration tests for the first three.
In theory we could add the test for `notifyMovedFundsSweepTimeout`, although it probably will be the most difficult (there are two wallets involved).

@nkuba @pdyraga @lukasz-zimnoch 
Do you think we should add such a test, or is it an overkill?

Slashing on `notifyMovedFundsSweepTimeout` should  use about the same amount of gas as in other cases.

solidity/tsconfig.json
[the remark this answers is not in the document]
lukasz-zimnoch: Ok, got it! Thanks for the clarifiaction!

solidity/test/integration/Slashing.test.ts
```diff
@@ -0,0 +1,529 @@
...
+        const { fundingTx, reveal } = SingleP2SHDeposit.deposits[0]
+        reveal.vault = tbtcVault.address
+
+        // We use a deposit funding bitcoin transaction with a very low amount,
+        // so we need to update the dust threshold to be below it.
+        await updateDepositDustThreshold(10000) // 0.0001 BTC)
+
+        // Reveal and sweep the deposit to set up a main UTXO for the deposit,
+        // so when operator inactivity is reported the deposit is transferred to
+        // the MovingFunds instead of the Closing state.
```
lukasz-zimnoch, line 382:
```suggestion
        // Reveal and sweep the deposit to set up a main UTXO for the wallet,
        // so when operator inactivity is reported the wallet is transferred to
        // the MovingFunds instead of the Closing state.
```

solidity/test/integration/Slashing.test.ts
nkuba: Let's merge it please and consider adding the 4th scenario in another PR

PR #319, opened 2022-05-20
Note on upgradeability for `Bridge.receiveBalanceApproval`

pdyraga:
Closes: #39 

`Bridge` is an upgradeable contract deployed behind a `TransparentUpgradeableProxy`.
Accepting redemption data as bytes provides great flexibility. The `Bridge` is
just like any other contract with a balance approved in the `Bank` and can be
upgraded to another version without being bound to a particular interface
forever. For example, what if 2 years from now, `Bridge` uses a single wallet
with all network operators as key share holders with a weekly key rotation?
This flexibility comes with the cost though - developers integrating their
vaults and dApps with `Bridge` using `approveBalanceAndCall` need to pay extra
attention to `redemptionData` and adjust the code in case the expected structure
of `redemptionData`  changes.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -492,7 +492,19 @@ contract Bridge is
     ///      - `amount` must be above or equal the `redemptionDustThreshold`,
     ///      - Given `walletPubKeyHash` and `redeemerOutputScript` pair can be
     ///        used for only one pending request at the same time,
-    ///      - Wallet must have enough Bitcoin balance to proceed the request,
+    ///      - Wallet must have enough Bitcoin balance to proceed the request.
+    ///
+    ///      Note on upgradeability:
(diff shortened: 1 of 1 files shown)


solidity/contracts/bridge/Bridge.sol
```diff
@@ -492,7 +492,19 @@ contract Bridge is
     ///      - `amount` must be above or equal the `redemptionDustThreshold`,
     ///      - Given `walletPubKeyHash` and `redeemerOutputScript` pair can be
     ///        used for only one pending request at the same time,
-    ///      - Wallet must have enough Bitcoin balance to proceed the request,
+    ///      - Wallet must have enough Bitcoin balance to proceed the request.
```
tomaszslabon, line 495: Maybe `process the request` instead of `proceed the request`?

tomaszslabon (approved):

PR #332, opened 2022-05-24
Mechanism Writeup: Sweeping

beaushinkle:
This PR fills in the Sweeping section of the mechanism writeup!

Tagging @pdyraga for review

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -124,10 +124,9 @@ can be done immediately; no need to wait for confirmations) by calling
 
 `Bridge.revealDeposit` takes in the funding transaction, and then the necessary
 information to reconstruct the Bitcoin script: `depositor`, `blindingFactor`,
-`walletPubKeyHash` and `refundPubkeyHash`. Then, it makes sure that the funding
-transaction happened on bitcoin, reconstructs the script, hashes it, verifies
-that the hashes match, and then stores the deposit as waiting to be
-<sweeping,swept> associated to the provided `depositor`.
+`walletPubKeyHash` and `refundPubkeyHash`. Then it reconstructs the script,
+hashes it, verifies that the hashes match, and then stores the deposit as
+waiting to be <sweeping,swept> associated to the provided `depositor`.
 
 [sweeping]
 ==== Sweeping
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -124,15 +124,34 @@ can be done immediately; no need to wait for confirmations) by calling
...
+`walletPubKeyHash` and `refundPubkeyHash`. Then it reconstructs the script,
+hashes it, verifies that the hashes match, and then stores the deposit as
+waiting to be <sweeping,swept> associated to the provided `depositor`.
 
 [sweeping]
 ==== Sweeping
 
+Periodically, off-chain clients associated to a wallet collect a batch of
+deposits and create a sweep transaction. This transaction includes all of the
+deposit UTXOs as well as the wallet's UTXO (`Wallet.mainUtxoHash`) as inputs
```
pdyraga, line 136: Maybe: "This transaction includes revealed and valid deposit UTXOs" ?

I see two reasons:
1. Wallet is not including a revealed deposit if something is wrong with it (for example, lock time). That's why I think *all* could be confusing.
2. There is no guarantee wallet will include all valid revealed deposits since the last sweep given the problems with proving too large transactions on Ethereum. Wallet may need to process them in batches.

README.adoc
```diff
@@ -161,3 +180,15 @@ the creation of a new active wallet.
...
+
+4+s|Sweeping
+
+|depositTreasuryFeeDivisor
+|The deposit fee divisor of one BTC to take as a treasury fee.
+|Yes
+|`2000 => 1/2000 = 5 BPS`
+
+|depositTxMaxFee
+|The max amount of satoshis the wallet is allowed to pay to miners.
```
pdyraga, line 192: Worth noting this number is per single deposit.

`movedFundsSweepTxMaxTotalFee`, for example, is for the whole move funds transaction.

README.adoc
beaushinkle: [0088980](https://github.com/keep-network/tbtc-v2/pull/332/commits/008898070030619e662700b93d1fd8de90b059e3)

woah difficulty of 2

README.adoc
beaushinkle: [a26f56b](https://github.com/keep-network/tbtc-v2/pull/332/commits/a26f56b4b0c0f07e4e21af2ef668f620e6c19fb6)

pdyraga (approved):

PR #334, opened 2022-05-25
Mechanism Writeup: Redeeming

beaushinkle:
This PR fills in the Redeeming section!

Tagging @pdyraga for review

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -197,7 +197,7 @@ that the hashes match, and then stores the deposit as waiting to be
 [sweeping]
 ==== Sweeping
 
-==== Redemptions
+==== Redeeming
 
 An account with a balance can request a redemption via
 `Bridge.requestRedemption`. We verify that the destination is valid (P2PKH,
--- a/README.adoc
+++ b/README.adoc
@@ -132,7 +132,30 @@ that the hashes match, and then stores the deposit as waiting to be
 [sweeping]
 ==== Sweeping
 
+==== Redemptions
 
+An account with a balance can request a redemption via
+`Bridge.requestRedemption`. We verify that the destination is valid (P2PKH,
+P2WPKH, P2SH or P2WSH), and build a redemption key based on the wallet's PKH
+and destination. There can only be one pending redemption per PKH-destination
+pair. The treasury takes a cut (`Bridge.redemptionTreasuryFeeDivisor`, 5 BPS),
+and then we reduce the account's balance and start a timer.
+
+After `Bridge.redemptionTimeout` (48 hours), anyone may call
+`Bridge.notifyRedemptionTimeout`. This will decrease
+`wallet.pendingRedemptionsValue`, mark the redemption as "timed out", punish
+the operators for `Bridge.redemptionTimeoutSlashingAmount`, and reward the
+notifier for a percentage (`Bridge.redemptionTimeoutNotifierRewardMultiplier`,
+100%) of the slashed stake. The redeemer is reimbursed the balance of the
+redemption, and the wallet begins to move its funds and eventually close via
+`Wallets.notifyWalletTimedOutRedemption`.
+
+To avoid this, the operators must fulfill the redemption by signing a
+transaction off-chain (potentially in a batch), submitting it to the Bitcoin
+chain, and then proving that they did so via `Bridge.submitRedemptionProof`. We
+perform an SPV proof to ensure the transaction occurred, it is well-formed, and
+then we decrease all of the redeemer's balances and increase the treasury's
+balance with its cut. 
 
 == Parameters
 
@@ -161,3 +184,15 @@ the creation of a new active wallet.
 creation of a new active wallet regardless of age.
 |Yes
 |`25 BTC`
+
+4+s|Redemption
+
+|redemptionTreasuryFeeDivisor
+|The redemption fee divisor of one BTC to take as a treasury fee.
+|Yes
+|`2000 => 1/2000 = 5 BPS`
+
+|redemptionTimeout
+|Length of time a wallet has to fulfill a redemption.
+|Yes
+|`48 hours`


beaushinkle:
> I think we should target them to tbtc-v2/docs/index.adoc instead.

Deal! How about I move them after this PR?

README.adoc
```diff
@@ -197,6 +197,31 @@ that the hashes match, and then stores the deposit as waiting to be
...
 ==== Sweeping
 
+==== Redeeming
+
+An account with a balance can request a redemption via
+`Bridge.requestRedemption`. We verify that the destination is valid (P2PKH,
+P2WPKH, P2SH or P2WSH), and build a redemption key based on the wallet's PKH
+and destination. There can only be one pending redemption per PKH-destination
+pair. The treasury takes a cut (`Bridge.redemptionTreasuryFeeDivisor`, 5 BPS),
+and then we reduce the account's balance and start a timer.
```
pdyraga, line 229:
```suggestion
and then we reduce the account's Bank balance and start a timer.
```

README.adoc
```diff
@@ -219,6 +219,31 @@ treasury (5 BPS; governable), and updates the user balances with information
...
+
+After `Bridge.redemptionTimeout` (48 hours), anyone may call
```
pdyraga, line 231: Should we mention 48h is governable?

README.adoc
```diff
@@ -219,6 +219,31 @@ treasury (5 BPS; governable), and updates the user balances with information
 from `DepositSweep.processDepositSweepTxInputs` and
 `DepositSweep.depositSweepTxFeeDistribution`.
 
+==== Redeeming
+
+An account with a balance can request a redemption via
+`Bridge.requestRedemption`. We verify that the destination is valid (P2PKH,
+P2WPKH, P2SH or P2WSH), and build a redemption key based on the wallet's PKH
+and destination. There can only be one pending redemption per PKH-destination
+pair. The treasury takes a cut (`Bridge.redemptionTreasuryFeeDivisor`, 5 BPS),
```
pdyraga, line 228: Should we mention 5 BPS is governable?

README.adoc
```diff
@@ -219,6 +219,31 @@ treasury (5 BPS; governable), and updates the user balances with information
...
+and then we reduce the account's balance and start a timer.
+
+After `Bridge.redemptionTimeout` (48 hours), anyone may call
+`Bridge.notifyRedemptionTimeout`. This will decrease
+`wallet.pendingRedemptionsValue`, mark the redemption as "timed out", punish
+the operators for `Bridge.redemptionTimeoutSlashingAmount`, and reward the
+notifier for a percentage (`Bridge.redemptionTimeoutNotifierRewardMultiplier`,
+100%) of the slashed stake. The redeemer is reimbursed the balance of the
```
pdyraga, line 236: Now I think we should maybe not include 5 BPS, 48h or 100% values in the text at all given they are governable and can get outdated quickly. Having them as default values in the table at the bottom makes sense though.

README.adoc
```diff
@@ -219,6 +219,31 @@ treasury (5 BPS; governable), and updates the user balances with information
...
+redemption, and the wallet begins to move its funds and eventually close via
+`Wallets.notifyWalletTimedOutRedemption`.
```
pdyraga, line 238: Err... not sure what we mean here by being eventually closed. The wallet that moved funds gets closed when they finalize the closing period (MOVING_FUNDS -> CLOSING -> CLOSED)

README.adoc
```diff
@@ -219,6 +219,31 @@ treasury (5 BPS; governable), and updates the user balances with information
 from `DepositSweep.processDepositSweepTxInputs` and
 `DepositSweep.depositSweepTxFeeDistribution`.
 
+==== Redeeming
+
+An account with a balance can request a redemption via
+`Bridge.requestRedemption`. We verify that the destination is valid (P2PKH,
```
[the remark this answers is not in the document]
beaushinkle: > I think the first two are worth being mentioned here, the third one... somewhere else. We also need a documentation for the concept of a Bank and Vault, and TBTCVault, so we could place it there.

Deal! The next concepts were going to be minting / unminting, so I think those fit well there

README.adoc
beaushinkle: I meant that this this kicks the process off where closing is the sort of end destination, but I see how that's confusing. I'll change the wording!

README.adoc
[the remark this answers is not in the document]
beaushinkle: [bf9a09a](https://github.com/keep-network/tbtc-v2/pull/334/commits/bf9a09ac56243c439cd68dddba2c5226ff325193)

README.adoc
beaushinkle: [cdc85fc](https://github.com/keep-network/tbtc-v2/pull/334/commits/cdc85fc2fa5677c6c43d37d358ca258fbb3d53d1)

README.adoc
```diff
@@ -209,16 +209,47 @@ The first is a cost vs time tradeoff. SPV proofs are expensive, so by dividing
...
+pair. The treasury takes a cut (`Bridge.redemptionTreasuryFeeDivisor`), and
+then we reduce the account's Bank balance and start a timer.
+
+[TIP]
+`Bridge.requestRedemption` requires a Bank balance approval to the Bridge. This
+can either be made in a separate transaction first via
+`Bank.approveBalance(Bridge.address, ...)` or in a single transaction via
+`Bank.approveBalanceAndCall(Bridge.address, ...)`.
+
+After `Bridge.redemptionTimeout`, anyone may call
```
pdyraga, line 237:
```suggestion
After the redemption timeout, anyone may call
```

`Bridge.redemptionTimeout` is not a thing. It's [exposed via `Bridge.redemptionParameters()`](https://github.com/keep-network/tbtc-v2/blob/938bf76a89f8b8bfdeb25ef8e322476a7919f2e7/solidity/contracts/bridge/Bridge.sol#L1610).

pdyraga: We could also clarify it a bit:

```suggestion
If the redemption was not performed by the wallet, after the redemption timeout, anyone may call
```

beaushinkle: Gotcha! I've been using the values at https://github.com/keep-network/tbtc-v2/blob/main/solidity/contracts/bridge/Bridge.sol#L263-L292 willy nilly, but I can adjust this for sure. Do I need to also take it out of the params section?

pdyraga: No, having it in parameters sections makes sense. I think we just should not refer to it via `Bridge.`. All other similar references point to functions that can actually be called.

pdyraga (approved):

Issue #144, opened 2022-02-22
TBTC v2 documentation


pdyraga:
#251 is about smart-contract-focused documentation. 

Eventually, we need something like:
https://docs.keep.network/tbtc/
https://docs.keep.network/random-beacon/

That will take what is in multiple RFCs, what is in smart contracts, and present it in one document.

Note: Let's make sure #251 is done first and use it as an opportunity to ensure all needed paths and governable parameters from RFC are implemented in smart contracts.

PR #337, opened 2022-05-25
Describe wallet states in the documentation

pdyraga:
Refs #144 

Described wallet states and added a reference to a diagram with wallet state transitions.

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -32,6 +32,32 @@ without a third-party intermediary.
 
 == System Architecture
 
+=== Wallet States
+
+There are six states for a wallet:
+
+- `UNKNOWN`: The wallet is unknown to the Bridge.
+- `LIVE`: The wallet can sweep deposits and accept redemption requests.
+- `MOVING_FUNDS`: The wallet was deemed unhealthy and is expected to move its
+  outstanding funds to another wallet. The wallet can still fulfill its
+  pending redemption requests although new redemption requests and new deposit
+  reveals are not accepted.
+- `CLOSING`: The wallet moved or redeemed all its funds and is in the closing
+  period where it is still a subject of fraud challenges and must defend against
+  them. This state is needed to protect against deposit frauds on deposits
+  revealed but not swept.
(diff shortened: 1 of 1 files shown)


docs/index.adoc
```diff
@@ -32,6 +32,32 @@ without a third-party intermediary.
...
+  reveals are not accepted.
+- `CLOSING`: The wallet moved or redeemed all its funds and is in the closing
+  period where it is still a subject of fraud challenges and must defend against
+  them. This state is needed to protect against deposit frauds on deposits
+  revealed but not swept.
+- `CLOSED`: The wallet finalized the closing period successfully and can no
+  longer perform any action in the Bridge.
+- `TERMINATED`: The wallet committed a fraud that was reported, did not move
+  funds to another wallet before a timeout, or did not sweep funds moved to it
+  from another wallet before a timeout. The wallet is blocked and can not
```
michalinacienciala, line 53: > The wallet committed a fraud that was reported

The `notifyFraudChallengeDefeatTimeout` leading to TERMINATED state can be also called if fraud wasn't committed, but somebody accused wallet of committing it and the wallet did not defeat the challenge on time. I propose:

_The wallet did not move funds to another wallet before a timeout, did not sweep funds moved to it from another wallet before a timeout, or was accused of a fraud and did not defeat the fraud challenge before a timeout._

beaushinkle:
This is wonderful

beaushinkle:
> Double checking that each function does what it says it does and then I'm good

Okay - went through every one of the listed functions in the state transitions and manually verified that what you see is what you get

docs/index.adoc
michalinacienciala: Yeah, I agree it's an unlikely scenario. I thought that the description I proposed is closer to the actual implementation - but you're right, we don't need to match this description 1-to-1 with what's implemented, it's here more to give a general idea about the role of each state. I'm fine with leaving it as it is.

michalinacienciala (approved):

PR #341, opened 2022-05-25
Create workflow for publishing `tbtc-v2.js` package 

michalinacienciala:
In this PR we add a GH Actions workflow that will publish npm
packages containing compiled `typescript` code and version them
using `-dev.X` suffix.
Workflow will be triggered every time changes to files in the
`./typescript` folder get merged to `main`. It will be also
possible to trigger the workflow manually.

TODO:
- [x] Remove temporary config before merge to `main`.


.github/workflows/npm-typescript.yml
```diff
@@ -0,0 +1,53 @@
+name: NPM
```
lukasz-zimnoch, line 1: Can we name it in a way indicating this is a workflow related to `typescript`? And I think we should do the same for `npm-contracts.yml`

michalinacienciala: Yah, I missed that. Addressed in [663fa76](https://github.com/keep-network/tbtc-v2/pull/341/commits/663fa762352154f8d331786de9ebc969e76c332d).

lukasz-zimnoch (approved):

PR #343, opened 2022-05-25
Mechanism Writeup: Unminting

beaushinkle:
This one is a little shorter - @pdyraga anything else you think is relevant here? I was thinking about allowance stuff, but felt like that wasn't part of the mechanism, and was better suited for talking about the bank and vault in general


pdyraga (approved):

pdyraga:
> I was thinking about allowance stuff, but felt like that wasn't part of the mechanism, and was better suited for talking about the bank and vault in general

Yeah, definitely!

PR #348, opened 2022-05-27
TBTC documentation: Maintainers

pdyraga:
Refs #144 

Described the concept of a maintainer and how it is implemented in
`MaintainerProxy`.

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -359,6 +359,100 @@ entirely, they can save a transaction and gas by calling
 `TBTCVault.unmintAndRedeem` which routes into authorizing the Bridge to
 <<redeeming,redeem>> via `bank.approveBalanceAndCall`.
 
+=== Maintainers
+
+`Bridge` has a set of public knowledge functions that need to be called from
+time to time. A public-knowledge function is where anyone has access to the
+information required to assemble the transaction and the transaction does not
+lead to punishment (slashing).
+
+To compensate for calling these transactions, the caller has the gas spent
+reimbursed by a DAO-funded ETH pool in the same transaction. This works only if
+callers follow the order and do not try to front-run each other. Given that the
+reimbursement code is constructed in such a way that the caller is net-zero or
+minimally positive after the refund, there should be no reason for callers to
+front-run each other. On the other hand, a single malicious front-runner can
+break this scheme. If there is someone who does not follow the order and keeps
+submitting transactions - for which they are reimbursed - everyone else will
+stop doing that because they will only burn gas on reverted transactions.
+
+One approach to address this problem could be to enforce the submission order
+on-chain. Although it sounds easy at first, it is complicated and expensive in
+practice. The code needs to take into account scenarios when someone is
+temporarily offline and missed their submission round or when someone is
+permanently offline and keeps missing their round all the time which causes
+delays and slows down the network. Attempts to address these problems by marking
+someone as ineligible for rewards in case they missed their round are making the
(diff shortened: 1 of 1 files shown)


docs/index.adoc
```diff
@@ -359,6 +359,100 @@ entirely, they can save a transaction and gas by calling
...
+a mechanism for slashing submitters not respecting the order but it is again,
+complex in practice, and makes us reimplement solutions like Gelato Network.
+
+We approach this problem in the easiest way possible by reducing a problem to
+a smaller group and introducing a list of submitters who are getting reimbursed
+for their calls. We call them maintainers. Maintainers are third-party bots who
+are expected to follow an off-chain informal agreement about the submission
+order and do not front-run each other. If they do, they can get deauthorized for
+reimbursements by the DAO. Maintainers are no more trusted than any other
+address on Ethereum. Anyone can submit public-knowledge `Bridge` transactions
```
michalinacienciala, line 402: We could consider adding somewhere here an information about the fact that maintainers are getting authorized by DAO. It's mentioned later ("Only
DAO-authorized maintainers can call `Bridge`...), but it would make sense to say they're authorized before we mention deauthorizing. 
Also, what do you think about mentioning the names of the functions authorizing/deauthorizing the maintainers?

docs/index.adoc
```diff
@@ -359,6 +359,100 @@ entirely, they can save a transaction and gas by calling
...
 
+=== Maintainers
+
+`Bridge` has a set of public knowledge functions that need to be called from
+time to time. A public-knowledge function is where anyone has access to the
+information required to assemble the transaction and the transaction does not
+lead to punishment (slashing).
+
+To compensate for calling these transactions, the caller has the gas spent
+reimbursed by a DAO-funded ETH pool in the same transaction. This works only if
```
michalinacienciala, line 370: I would suggest to change this to "the caller **could have** the gas spent reimbursed... This **would** work only if...". I know it's explained later why we don't reimburse direct calling of the `Bridge` functions and why we use maintainers, but someone just skimming the text could think that the calling `Bridge` functions directly is reimbursed.

michalinacienciala (commented):
Great explanation! I have only two non-blocking comments.

docs/index.adoc
pdyraga: Hmm... I am not sure about it. 

> To compensate for calling these transactions, the caller has the gas spent
reimbursed by a DAO-funded ETH pool in the same transaction.

This is true and it's how it works now in `MaintainerProxy`.

> This works only if
callers follow the order and do not try to front-run each other.

This is true and it's how it works now in `MaintainerProxy`. That's the reason we have `onlyMaintainer` modifier on some functions.

docs/index.adoc
pdyraga: Good idea! See 561ce4ccba8a153feabb24c5cfe6b63b04c9f4df please.

docs/index.adoc
michalinacienciala: Well, yeah, when `Bridge` functions are called via `MaintainerProxy` they get reimbursed. But when they're called directly - they're not. And at this stage of the document we haven't mentioned the `MaintainerProxy` yet, so the assumption is that we're talking about calling the `Bridge` functions directly.

But I know my comment was a bit nitpicky :) I'm ok with leaving the text as it is if you don't think my propositions adds clarity.

michalinacienciala (approved):

Issue #347, opened 2022-05-27
Update construction of redemption key in Typescript library


pdyraga:
In #301 we changed how the redemption key is built. Instead of building it as `keccak256(walletPubKeyHash | redeemerOutputScript)` we now build it as `keccak256(keccak256(redeemerOutputScript) | walletPubKeyHash)`. This change was not ported to Typescript library though. We updated documentation but did not update values in tests:
https://github.com/keep-network/tbtc-v2/blob/de265cf8dd52411f853f01d18c520af473448efd/typescript/test/ethereum.test.ts#L41 (see https://github.com/keep-network/tbtc-v2/commit/73dee0bafb727fcdb4ee499b0fb1c4f73b8db8be)

PR #350, opened 2022-05-27
🚧 😇 🔑  Update construction of a redemption key

michalsmiarowski:
Closes: #347 

There was a change that modified how the redemption key is build. Instead of
building it as `keccak256(walletPubKeyHash | redeemerOutputScript)` we now build
it as `keccak256(keccak256(redeemerOutputScript) | walletPubKeyHash)`. The
documentation in ts-lib was updated, but we did not update the values and
bulding redemptions key itself in tests.

I also tested the function using `walletPubKeyHash` and `redeemerOutputScript` 
used in 
https://github.com/keep-network/tbtc-v2/blob/de265cf8dd52411f853f01d18c520af473448efd/solidity/test/bridge/Bridge.Redemption.test.ts#L252-L255
and the generated key was correct.

### DIFF
--- a/typescript/src/ethereum.ts
+++ b/typescript/src/ethereum.ts
@@ -57,8 +57,11 @@ export class Bridge implements ChainBridge {
     // Build the redemption key by using the 0x-prefixed wallet PKH and
     // prefixed output script.
     const redemptionKey = utils.solidityKeccak256(
-      ["bytes20", "bytes"],
-      [`0x${walletPubKeyHash}`, prefixedRawRedeemerOutputScript]
+      ["bytes32", "bytes20"],
+      [
+        utils.solidityKeccak256(["bytes"], [prefixedRawRedeemerOutputScript]),
+        `0x${walletPubKeyHash}`,
+      ]
     )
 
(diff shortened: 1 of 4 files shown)


typescript/src/ethereum.ts
```diff
@@ -57,8 +57,11 @@ export class Bridge implements ChainBridge {
     // Build the redemption key by using the 0x-prefixed wallet PKH and
     // prefixed output script.
     const redemptionKey = utils.solidityKeccak256(
-      ["bytes20", "bytes"],
-      [`0x${walletPubKeyHash}`, prefixedRawRedeemerOutputScript]
+      ["bytes32", "bytes20"],
```
lukasz-zimnoch, line 60: Can we also do that change in the mock implementation of the `Bridge`? This is located in `test/utils/mock-bridge.ts` file.

michalsmiarowski: 0a6746a6ede9534dbd6fc67b8ffc18dd4791c0ab

lukasz-zimnoch (approved):

PR #351, opened 2022-05-27
Integration test: full flow

tomaszslabon:
Closes: #250 

This PR adds a full deposit-redemption flow integration test.

The flow is as follows:
 1) wallet creation
 2) revealDeposit
 3) submitDepositSweepProof
 4) requestRedemption
 5) submitRedemptionProof
 
 After each step checks were performed.
 
 The test data comes from transactions that were made in Bitcoin testnet.

### DIFF
--- a/solidity/test/integration/FullFlow.test.ts
+++ b/solidity/test/integration/FullFlow.test.ts
@@ -1,0 +1,338 @@
+/* eslint-disable no-await-in-loop */
+/* eslint-disable @typescript-eslint/no-extra-semi */
+import { ethers, helpers, waffle } from "hardhat"
+import type { FakeContract } from "@defi-wonderland/smock"
+import type { BigNumberish } from "ethers"
+import { utils } from "ethers"
+import type { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers"
+import { expect } from "chai"
+import type {
+  TBTC,
+  Bridge,
+  Bank,
+  TBTCVault,
+  TestRelay,
+  IRandomBeacon,
+  WalletRegistry,
+} from "../../typechain"
+import {
+  performEcdsaDkg,
+  updateWalletRegistryDkgResultChallengePeriodLength,
+} from "./utils/ecdsa-wallet-registry"
+import { produceRelayEntry } from "./utils/random-beacon"
+import { UTXOStruct } from "../../typechain/Bridge"
+import {
+  walletPublicKey,
+  walletPubKeyHash,
+  revealDepositData,
+  depositSweepData,
+  redemptionData,
+} from "./data/integration"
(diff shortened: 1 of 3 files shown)


solidity/test/integration/FullFlow.test.ts
```diff
@@ -0,0 +1,338 @@
...
+      })
+
+      describe("when a deposit is revealed", async () => {
+        before(async () => {
+          revealDepositData.reveal.vault = tbtcVault.address
+
+          // We use a deposit funding bitcoin transaction with a very low amount,
+          // so we need to update the dust and redemption thresholds to be below it.
+          await updateDepositDustThreshold(10000) // 0.0001 BTC
+          await updateRedemptionDustThreshold(2000) // 0.00002 BTC
```
pdyraga, line 107: The test starts failing after merging `main` changes. This is because #352 updated TX max fees in `Bridge` constructor:
- Deposit dust threshold is 0.01 BTC 
- Deposit TX max fee is now 0.001 BTC
- Redemption dust threshold is 0.01 BTC
- Redemption TX max fee is now 0.001 BTC

Also, I added a validation rule that the deposit dust threshold must be greater than the per-deposit max allowed Bitcoin network fee. The redemption dust threshold must be greater than the per-redemption max allowed Bitcoin network fee: https://github.com/keep-network/tbtc-v2/pull/352/commits/f9f75faeed956fbce381be8d11017a23097877c1

If we are scaling down dust threshold, we should also scale down the max fees. Let's do something similar as in `Slashing.test.ts` in #352:

```suggestion
          // We use a deposit funding bitcoin transaction with a very low amount,
          // so we need to update the dust and redemption thresholds to be below it.
          // TX max fees need to be adjusted as well given that they need to
          // be lower than dust thresholds.
          await updateDepositDustThresholdAndTxMaxFee(10_000, 2_000) // 0.0001 BTC, 0.00002 BTC
          await updateRedemptionDustThresholdAndTxMaxFee(2_000, 200) // 0.00002 BTC, 0.000002 BTC
```

tomaszslabon: Done in https://github.com/keep-network/tbtc-v2/pull/351/commits/8e0f71e9f2d717fca2e5fe97e0e8056aeef0c9dd.
Although I used a bit higher numbers for thresholds and max fees that fit the data I used when creating the transactions:
```
          // 0.001 BTC, 0.0001 BTC
           await updateDepositDustThresholdAndTxMaxFee(100_000, 10_000)
           // 0.0005 BTC, 0.0001 BTC
           await updateRedemptionDustThresholdAndTxMaxFee(50_000, 10_000)
```

pdyraga (approved):

PR #352, opened 2022-05-27
Revisited initial Bridge parameters

### LINKED ISSUE #258: Revisit initial parameters set in Bridge constructor
https://github.com/keep-network/tbtc-v2/blob/5583abebb8c63bbc57e523d90b6072ec4ffb5d3b/solidity/contracts/bridge/Bridge.sol#L227-L252

pdyraga:
Closes #258 

Parameters are set to be more forgiving for the early days of the network. Timeouts are pretty long and slashing amounts are minimal.

During this work, I realized we should improve validation for deposit and redemption parameters. The deposit dust threshold must be greater than the per-deposit max allowed Bitcoin network fee. The redemption dust threshold must be greater than the per-redemption max allowed Bitcoin network fee. Added the validation in f9f75faeed956fbce381be8d11017a23097877c1.

### DIFF
--- a/solidity/contracts/bridge/BridgeState.sol
+++ b/solidity/contracts/bridge/BridgeState.sol
@@ -356,6 +356,7 @@ library BridgeState {
     ///        such transaction is considered a fraud.
     /// @dev Requirements:
     ///      - Deposit dust threshold must be greater than zero,
+    ///      - Deposit dust threshold must be greater than deposit TX max fee,
     ///      - Deposit treasury fee divisor must be greater than zero,
     ///      - Deposit transaction max fee must be greater than zero.
     function updateDepositParameters(
@@ -367,6 +368,11 @@ library BridgeState {
         require(
             _depositDustThreshold > 0,
             "Deposit dust threshold must be greater than zero"
+        );
+
+        require(
+            _depositDustThreshold > _depositTxMaxFee,
+            "Deposit dust threshold must be greater than deposit TX max fee"
         );
 
         require(
@@ -431,6 +437,8 @@ library BridgeState {
     /// @dev Requirements:
     ///      - Redemption dust threshold must be greater than moving funds dust
     ///        threshold,
+    ///      - Redemption dust threshold must be greater than the redemption TX
+    ///        max fee,
     ///      - Redemption treasury fee divisor must be greater than zero,
     ///      - Redemption transaction max fee must be greater than zero,
     ///      - Redemption timeout must be greater than zero,
@@ -448,6 +456,11 @@ library BridgeState {
         require(
             _redemptionDustThreshold > self.movingFundsDustThreshold,
             "Redemption dust threshold must be greater than moving funds dust threshold"
+        );
+
+        require(
+            _redemptionDustThreshold > _redemptionTxMaxFee,
+            "Redemption dust threshold must be greater than redemption TX max fee"
         );
 
         require(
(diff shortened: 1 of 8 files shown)


beaushinkle:
Compiling the changes in english:

* all max mining fees: `0.0001 BTC => 0.001 BTC`
* all slashing punishments: `10_000 T => 100 T`
* redemption timeout: `48 hours => 5 days`
* failed redemption slashing amount: `10_000 T => 100 T`
* moving funds dust amount: `0.0002 BTC => 0.002 BTC `
* failed moving funds slashing amount: `10_000 T => 100 T`
* fraud challenge deposit amount: `2 eth => 5 eth`
* wallet btc transfer max: `10 btc => 50 btc`

Does that look accurate / was that the intent? 

Bitcoin mining fee looks good. Here's a chart of the last 3 years:

image

Other than that weird peak at $63/tx on 2021-04-21, fees have been *comfortably* below 30$.

In the early days, we want to be really light on punishments, and slashing 100T certainly accomplishes that, and making folks pay 5 eth to challenge does too.

Same idea for giving operators 5 days to redeem - if something is janky this gives us time to release a new client version to try to help them fix their thing.

In this system, I would actually want to keep the `walletMaxBtcTransfer` low; since we don't want BTC to pile up *at all* (since the btc isn't protected by punishments), so I would keep it at 10 rather than 50. If we don't have enough wallets, what ends up happening is that we just divide evenly among all the live wallets anyway :)

solidity/contracts/bridge/BridgeState.sol
```diff
@@ -369,6 +370,11 @@ library BridgeState {
             "Deposit dust threshold must be greater than zero"
         );
 
+        require(
+            _depositDustThreshold > _depositTxMaxFee,
+            "Deposit dust threshold must be greater than deposit TX max fee"
+        );
```
beaushinkle, line 376: very reasonable

pdyraga:
> Compiling the changes in english:
> 
> * all max mining fees: `0.0001 BTC => 0.001 BTC`
> * all slashing punishments: `10_000 T => 100 T`
> * redemption timeout: `48 hours => 5 days`
> * failed redemption slashing amount: `10_000 T => 100 T`
> * moving funds dust amount: `0.0002 BTC => 0.002 BTC `
> * failed moving funds slashing amount: `10_000 T => 100 T`
> * fraud challenge deposit amount: `2 eth => 5 eth`
> * wallet btc transfer max: `10 btc => 50 btc`
> 
> Does that look accurate / was that the intent?

Yes, except that I changed `walletMaxBtcTransfer` back to 10 BTC according to your suggestion.

Slashing amounts are low to do not punish early stakers too much in case there are bugs in the client.

All max mining fees are pretty high - while for most of the time we should have wallet executing transactions using fees not even close to the max ones, I think leaving some margin makes sense, especially early days when we'll have more to worry about.

All timeouts are pretty long and I think they should stay that long. One of the lessons from v1 is that 6 hours timeouts were just too much and operators located in different time zones could not even start coordinating problem resolution before the timeout was hit. This should not be that problematic with 51-of-100 threshold, but I think having enough time margin is important especially early days.

> In this system, I would actually want to keep the `walletMaxBtcTransfer` low; since we don't want BTC to pile up _at all_ (since the btc isn't protected by punishments), so I would keep it at 10 rather than 50. If we don't have enough wallets, what ends up happening is that we divide evenly among all the live wallets anyway :)

You are right, changed back in https://github.com/keep-network/tbtc-v2/pull/352/commits/12c9bf665103d2d2641c568a2dbafc6b606d37ab.

beaushinkle (approved):

PR #353, opened 2022-05-28
TBTC Documentation: Diagrams!

beaushinkle:
This PR adds @michalinacienciala 's wonderful diagrams to the index doc.

I'm still really unsure about where to fit them because they're such show-stealers that trying to weave them into an explanation is unwieldy.

I ended up giving them their own section, but if @pdyraga or @michalinacienciala have better ideas, I'm game!

### DIFF
--- a/docs/index.adoc
+++ b/docs/index.adoc
@@ -479,13 +479,19 @@ wallet.
 
 == Wallet Lifecycle Diagrams
 
+==== Handling Fraud
+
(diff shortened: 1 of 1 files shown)


pdyraga:
> they're such show-stealers that trying to weave them into an explanation is unwieldy.

agreed! 💯

pdyraga (approved):

Issue #143, opened 2022-02-22
System tests


pdyraga:
We need end-to-end system tests with T `TokenStaking` and ECDSA `WalletRegistry` involved. Right now this issue is just a placeholder and it will evolve over time into something more specific.

PR #358, opened 2022-05-31
Expose test utils as Hardhat tasks

lukasz-zimnoch:
Refs: #143 

This pull request exposes some useful test utils (defined in `test/integration/utils`) as Hardhat tasks that can be used from the outside. Those utils will be essential for system test scenarios (developed as part of #339) that must use the local Hardhat network until the Ropsten deployment goes live. 

For now, two new commands are available:
- `npx hardhat test-utils:register-operators` that registers an arbitrary number of pre-defined signers as sortition pool operators. This is required to perform wallet creation.
- `npx hardhat test-utils:create-wallet` that creates a wallet with a given public key and registers it in the bridge. This is required to perform deposits and redemptions.

I recommend exploring the commit history to get the full picture of each change.

### DIFF
--- a/solidity/test/integration/utils/ecdsa-wallet-registry.ts
+++ b/solidity/test/integration/utils/ecdsa-wallet-registry.ts
@@ -2,7 +2,6 @@
 // We should consider exposing them in @keep-network/ecdsa for an external usage.
 
 /* eslint-disable no-await-in-loop */
-import { deployments, ethers, helpers } from "hardhat"
 
 import type { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers"
 import type {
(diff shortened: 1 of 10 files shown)


solidity/test/integration/utils/ecdsa-wallet-registry.ts
```diff
@@ -11,6 +10,7 @@ import type {
   BigNumberish,
   Signer,
 } from "ethers"
+import { HardhatRuntimeEnvironment } from "hardhat/types"
```
nkuba, line 13:
```suggestion
import type { HardhatRuntimeEnvironment } from "hardhat/types"
```

solidity/test/integration/utils/staking.ts
```diff
@@ -1,8 +1,8 @@
-import { helpers } from "hardhat"
-
 import type { BigNumberish, Contract, Signer } from "ethers"
+import { HardhatRuntimeEnvironment } from "hardhat/types"
```
nkuba, line 2:
```suggestion
import type { HardhatRuntimeEnvironment } from "hardhat/types"
```

nkuba (approved):

PR #361, opened 2022-06-03
Tweaks around Bitcoin transaction processing code

lukasz-zimnoch:
Refs: #143 

This pull request introduces some tweaks to functions that work with Bitcoin transactions.

First, we extract a separate type that represents the Bitcoin transaction hash (`TransactionHash`). We also make it explicit that this type represents a transaction hash in the byte order used by the block explorers and wallet software. This way we capture those details in one place which was not possible if we would still use the ordinary `string` type.

Second, all `make*` functions return now the transaction hash and the UTXO relevant for further processing (deposit UTXO or new main UTXO). The same was done for the `create*Transaction` functions. This change should allow all clients to follow up on the transaction after using one of the library functions.

### DIFF
--- a/typescript/src/bitcoin.ts
+++ b/typescript/src/bitcoin.ts
@@ -11,6 +11,15 @@ import { StaticWriter, BufferWriter } from "bufio"
 import { BigNumber } from "ethers"
 
 /**
+ * Represents a transaction hash (or transaction ID) as an un-prefixed hex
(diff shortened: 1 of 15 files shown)


tomaszslabon:
If we search for `transactionHash: string` we will still find some transaction hashes represented by strings.
I think we could use the `TransactionHash` type there as well.

tomaszslabon (approved):

PR #363, opened 2022-06-07
Add light relay

eth-r:
The light relay doesn't record the entire bitcoin blockchain, but instead only the timestamps and PoW targets of each retarget. Chains of headers are accepted if they have sufficient work and their targets match the target/s of the relevant retarget epoch/s. This reduces the storage costs of maintaining the relay by roughly a factor of 500.


solidity/contracts/relay/Relay.sol
```diff
@@ -0,0 +1,507 @@
...
+contract Relay is Ownable, ILightRelay {
+    using BytesLib for bytes;
+    using BTCUtils for bytes;
+    using ValidateSPV for bytes;
+    using RelayUtils for bytes;
+
+    bool public ready;
+    // Number of blocks required for a retarget proof.
+    // Governable
+    // Should be set to a fairly high number (e.g. 20-50) in production.
```
pdyraga, line 88: Something we need to be aware of and something we could make explicit in the documentation is that the higher  `proofLength`, the longer delay with submitting proofs for transactions made just before the epoch ends.

That is, I need to wait for enough headers in the new epoch to accumulate so that the relay can `retarget` and I can prove the work in `validateChain`.

For 50 blocks, it means ~8h delay assuming 6 blocks are mined per hour.

solidity/contracts/relay/Relay.sol
```diff
@@ -0,0 +1,507 @@
...
+                i * 80,
+                previousHeaderDigest
+            );
+
+            currentHeaderTimestamp = headers.extractTimestampAt(i * 80);
+
+            require(
+                currentHeaderTimestamp > previousHeaderTimestamp,
+                "Invalid timestamp in header chain"
+            );
```
pdyraga, line 367: I am not sure if we should validate it this way. 

> A timestamp is accepted as valid if it is greater than the median timestamp of previous 11 blocks, and less than the network-adjusted time + 2 hours. "Network-adjusted time" is the median of the timestamps returned by all nodes connected to you. As a result block timestamps are not exactly accurate, and they do not need to be. Block times are accurate only to within an hour or two.

https://en.bitcoin.it/wiki/Block_timestamp 

If headers form a valid chain (we validate the previous header digest), and we checked the difficulty target for all of them (it's done in `validateHeader`), do we still need to check timestamps?

solidity/contracts/relay/Relay.sol
```diff
@@ -0,0 +1,507 @@
...
+    /// that the difficulty of the new epoch is calculated correctly according
+    /// to the block timestamps, and that the next X blocks would be valid in
+    /// the new epoch.
+    /// We have no information of block heights, so we cannot enforce that
+    /// retargets only happen every 2016 blocks; instead, we assume that this
+    /// is the case if a valid proof of work is provided.
+    /// It is possible to cheat the relay by providing X blocks from earlier in
+    /// the most recent epoch, and then mining X new blocks after them.
+    /// However, each of these malicious blocks would have to be mined to a
+    /// higher difficulty than the legitimate ones.
```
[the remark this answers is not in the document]
eth-r: You can only reuse legitimate blocks that have been mined at the current difficulty D. Let's say the epoch began at time 10001 and ended (legitimately) at time 12016, and that the difficulty stays almost constant at ~D with this retarget. Let's say that blocks were mined every second during this period.

The last 6 blocks of the epoch were produced at 12011, 12012, 12013, 12014, 12015, and 12016.

If you try to reuse the blocks from this epoch, your last block at the given difficulty was produced either _at_ 12016 as in the legitimate chain, or earlier. If you present an earlier block as your last block, it results in a higher difficulty.

To abuse this, you'd need to use the blocks [12012 .. 12016] and then mine an additional block at time 18064, giving you the pre-retarget chain of [12012, 12013, 12014, 12015, 12016, 18064]. The addition of the malicious last block now means that our difficulty drops from D to D/4 because it looks like the epoch lasted 4 times the expected time.

If we add validation to the retarget so the timestamp of the last block of the epoch may not be in the future, this would give the legitimate chain plenty of time to pre-empt attempts to mine such chains (because such a difficulty-reducing chain could only be submitted significantly later than the real chain).

solidity/test/relay/Relay.test.ts
```diff
@@ -0,0 +1,1260 @@
...
+  return {
+    deployer,
+    governance,
+    thirdParty,
+    relay,
+  }
+}
+
+describe("LightRelay", () => {
+  let deployer: SignerWithAddress
```
pdyraga, line 51: Looks that it's unused.

solidity/test/relay/Relay.test.ts
```diff
@@ -0,0 +1,1260 @@
...
+          const proofHeaders = concatenateHexStrings(headerHex.slice(5, 11))
+          await relay.connect(governance).setProofLength(6)
+          const tx = await relay.validateChainGasReport(proofHeaders)
+          const txr = await tx.wait()
+
+          expect(txr.status).to.equal(1)
+        })
+      })
+
+      context("with proof length 18", () => {
```
pdyraga, line 805: Non-blocking this PR but how hard would it be to add a unit test with 50+50 mainnet retarget? This would allow us to assess mainnet gas costs and see if there are no other surprises.

solidity/contracts/relay/LightRelay.sol
```diff
@@ -0,0 +1,595 @@
...
+    /// headers of the previous epoch.
+    ///
+    /// If the difficulty increases significantly between relay genesis and the
+    /// present, creating fraudulent proofs for earlier epochs becomes easier.
+    /// Users of the relay should check the timestamps of valid headers and
+    /// only accept appropriately recent ones.
+    function validateChain(bytes memory headers)
+        external
+        view
+        relayActive
```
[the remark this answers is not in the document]
eth-r: For gas costs I'd prefer the latter.

solidity/test/relay/Relay.test.ts
eth-r: Not too hard, I just need to collect the header data for it

pdyraga: Let's do it but in a separate PR.

pdyraga (approved):

PR #364, opened 2022-06-09
Typescript library improvements pack

lukasz-zimnoch:
This pull request introduces a bunch of improvements for the `typescript` library. I recommend reviewing this PR commit-by-commit. Here is a recap of the changes:

### Reorganization of the library public API

So far, the `typescript` library had a default export that exposed the `TBTC` interface containing all library public functions. This approach turned out to be too complicated to reason about for external clients. As an improvement, we decided to simplify it. The `index.ts` file now looks as follows:
```
const TBTC = {
  calculateDepositAddress,
  suggestDepositWallet,
  revealDeposit,
  getRevealedDeposit,
  requestRedemption,
  getRedemptionRequest,
}

const SpvMaintainer = {
  submitDepositSweepProof,
  submitRedemptionProof,
}

export { TBTC, SpvMaintainer, ElectrumClient, EthereumBridge }
```
The `TBTC` object contains all functions relevant for regular TBTC system clients interested in deposits and redemptions (e.g. tBTC dApp). The `SpvMaintainer` object targets the maintainers of the system that will be responsible for submitting SPV proofs of Bitcoin transactions performed by the wallets. Last but not least, the `ElectrumClient` and `EthereumBridge` are the default implementations of `BitcoinClient` and `ChainBridge` respectively. They are provided by the `typescript` library to facilitate integration but specific clients may want to use their own implementations.

### Introduce `DepositScriptParameters` type

The `typescript` library uses the `Deposit` interface to represent a deposit. However, some use cases related to deposit script construction need only a subset of the `Deposit`'s fields. In order to make the API cleaner, we introduce the `DepositScriptParameters` type and use it in appropriate places. This type looks as follows:
```
  Deposit,
  | "depositor"
  | "blindingFactor"
  | "walletPublicKey"
  | "refundPublicKey"
  | "refundLocktime"
>
```

### Add Ethereum implementation of the chain identifier.

The `Identifier` interface declared in the `chain.ts` file is supposed to represent the chain identifier in a chain-agnostic way. Such a solution should facilitate potential multi-chain support in the future. This pull request adds the Ethereum implementation of the `Identifier` interface by introducing the `Address` class in the `ethereum.ts` file. This class holds all the details specific to Ethereum addresses like the `0x` prefix and checksumming.

### Expose getter for revealed deposits

The `typescript` library now exposes `getRevealedDeposit` function that allows getting deposits that were revealed to the on-chain bridge. The signature of the function looks as follows:
```
getRevealedDeposit(
  utxo: UnspentTransactionOutput,
  bridge: Bridge
): Promise<RevealedDeposit>
```
### Expose getter for pending and timed-out redemption requests

The `typescript` library now exposes `getRedemptionRequest` function that allows getting both pending and timed-out request registered on the on-chain bridge. The signature of the function looks as follows:
```
getRedemptionRequest(
  walletPublicKey: string,
  redeemerOutputScript: string,
  type: "pending" | "timedOut",
  bridge: Bridge
): Promise<RedemptionRequest>
```
### Suggestion of the deposit wallet

The `typescript` library is now able to suggest the preferred wallet that should be used during deposit creation. This is achieved by exposing the `suggestDepositWallet` function whose signature looks as follows:
```
suggestDepositWallet(
  bridge: Bridge
): Promise<string | undefined>
```
Currently, that function always suggests the bridge's active wallet. Note that the return type may be `undefined` if the active wallet is not currently set. The clients using that function must handle that case properly.

### Add external modules typings

The `typescript` library imports several third-party libraries (e.g. `bcoin`, `bcrypto`) that ale plain JS modules and don't declare any typings. TypeScript typically complains about that and so far, we used `@ts-ignore` annotations to suppress those warnings. However, those annotations are erased during TS compilation. That cause build problems in packages that imports `typescript` library as dependency. To overcome that problem, we remove `@ts-ignore` markups and add a `typings.d.ts` file that declares external libraries as proper modules.


tomaszslabon (dismissed):

typescript/src/deposit-sweep.ts
```diff
@@ -36,7 +36,7 @@ import { createTransactionProof } from "./proof"
  *          - the sweep transaction hash,
  *          - the new wallet's main UTXO produced by this transaction.
  */
-export async function sweepDeposits(
+export async function submitDepositSweepTransaction(
```
michalsmiarowski, line 38: Non-blocking:

Similar to above comments. We could change the function description to:

"**Submits a sweep transaction** by combining all the provided **P2(W)SH** UTXOs and broadcasting a Bitcoin P2WPKH deposit sweep transaction."

typescript/src/chain.ts
```diff
@@ -92,16 +105,39 @@ export interface Bridge {
...
+   * Gets a timed-out redemption from the on-chain contract.
+   * @param walletPublicKey Bitcoin public key of the wallet the request is
+   *        targeted to. Must be in the compressed form (33 bytes long with 02
+   *        or 03 prefix).
+   * @param redeemerOutputScript The redeemer output script the redeemed funds
+   *        are supposed to be locked on. Must be un-prefixed and not prepended
+   *        with length.
+   * @returns Promise with the pending redemption.
+   */
+  timedOutRedemptions(
```
michalsmiarowski, line 131: Shouldn't this be called `timedOutRedemption` since it return only one redemption request based on the `redemptionKey`? Or maybe I don't understand something 🤔

typescript/src/ethereum.ts
```diff
@@ -244,4 +317,118 @@ export class Bridge implements ChainBridge {
...
+  /**
+   * @see {ChainBridge#deposits}
+   */
+  async deposits(
+    depositTxHash: TransactionHash,
+    depositOutputIndex: number
+  ): Promise<RevealedDeposit> {
+    const depositKey = this.buildDepositKey(depositTxHash, depositOutputIndex)
+
+    const deposit = await this._bridge.deposits(depositKey)
```
michalsmiarowski, line 331: Similar to above - shouldn't this be caleed `deposit` instead of `deposits`?

typescript/src/deposit.ts
```diff
@@ -344,3 +382,28 @@ export async function revealDeposit(
...
+  bridge: Bridge
+): Promise<RevealedDeposit> {
+  return bridge.deposits(utxo.transactionHash, utxo.outputIndex)
+}
+
+/**
+ * Suggests a wallet that should be used as the deposit target at the given moment.
+ * @param bridge Handle to the Bridge on-chain contract.
+ * @returns Compressed (33 bytes long with 02 or 03 prefix) public key of
+ *          the wallet.
```
michalsmiarowski, line 403: Maybe we should specify in JSDoc also that it returns a promise.

```suggestion
 * @returns Promise object represents compressed (33 bytes long with 02 or 03 prefix) public key of
 *          the wallet.
```

typescript/src/chain.ts
```diff
@@ -92,16 +105,39 @@ export interface Bridge {
...
+    walletPublicKey: string,
     redeemerOutputScript: string
   ): Promise<RedemptionRequest>
+
+  /**
+   * Gets the public key of the current active wallet.
+   * @returns Compressed (33 bytes long with 02 or 03 prefix) active wallet's
```
michalsmiarowski, line 138: Same as above

typescript/src/ethereum.ts
```diff
@@ -244,4 +317,118 @@ export class Bridge implements ChainBridge {
...
+
+  // eslint-disable-next-line valid-jsdoc
+  /**
+   * @see {ChainBridge#activeWalletPublicKey}
+   */
+  async activeWalletPublicKey(): Promise<string | undefined> {
+    const activeWalletPubKeyHash = await this._bridge.activeWalletPubKeyHash()
+
+    if (
+      activeWalletPubKeyHash === "0x0000000000000000000000000000000000000000"
```
michalsmiarowski, line 384: Can we use `AddressZero` const from `ethers`?

typescript/test/ethereum.test.ts
```diff
@@ -266,6 +318,139 @@ describe("Ethereum", () => {
...
+          // that matches the deposit transaction hash and output index used during
+          // the test call.
+          await bridgeContract.mock.deposits
+            .withArgs(
+              "0x01151be714c10edde62a310bf0604c01134450416a0bf8a7bfd43cef90644f0f"
+            )
+            .returns({
+              depositor: "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
+              amount: BigNumber.from(10000),
+              vault: "0x0000000000000000000000000000000000000000",
```
michalsmiarowski, line 377: Similar as above. If it's not possible to use it in tests, then maybe we should extract it to a separate constant, because it is used so often.

typescript/test/ethereum.test.ts
```diff
@@ -266,6 +318,139 @@ describe("Ethereum", () => {
...
+          expect(await bridgeHandle.activeWalletPublicKey()).to.be.equal(
+            "03989d253b17a6a0f41838b84ff0d20e8898f9d7b1a98f2564da4cc29dcf8581d9"
+          )
+        })
+      })
+
+      context("when there is no active wallet", () => {
+        beforeEach(async () => {
+          await bridgeContract.mock.activeWalletPubKeyHash.returns(
+            "0x0000000000000000000000000000000000000000"
```
michalsmiarowski, line 445: Same as above

typescript/src/bitcoin.ts
```diff
@@ -253,7 +247,7 @@ export interface Client {
 export function decomposeRawTransaction(
   rawTransaction: RawTransaction
 ): DecomposedRawTransaction {
-  const toHex = (bufferWriter: StaticWriter | BufferWriter) => {
+  const toHex = (bufferWriter: any) => {
```
michalsmiarowski, line 250: Non-blocking:

Based on the Typescript documentation [we should avoid using any](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#any). Just pointing that out since I don't have a solution for that as of now, so can't propose anything better 🤔

michalsmiarowski (changes requested):
Very good code! Lot of awesome new functions that will help in the future 🚀 

Left some comments to look at before the merge.

typescript/src/deposit-sweep.ts
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/364/commits/e3da6c934d74540d8265dcbd89bbd717385e63cd

typescript/src/chain.ts
lukasz-zimnoch: This name is chosen deliberately to match the name of the function exposed by the `Bridge` function. That should make the lib <-> contract integration easier to understand.

typescript/src/ethereum.ts
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/364#discussion_r895532994

typescript/src/chain.ts
lukasz-zimnoch: And same reply.

typescript/src/ethereum.ts
lukasz-zimnoch: We could but `activeWalletPubKeyHash` is not an address. The only thing that is the same is the 20-byte length. I think it can be confusing if we use `AdressZero`.

typescript/test/ethereum.test.ts
lukasz-zimnoch: The `vault` is actually an address so using `AddressZero` makes perfect sense here. See https://github.com/keep-network/tbtc-v2/pull/364/commits/fdc03efa8b9af17a655f3ed9ef86785e50da0582

typescript/src/chain.ts
michalsmiarowski: I understand, but I was just wondering if the function name really reflects what the function does.

https://github.com/keep-network/tbtc-v2/blob/21eaf4624d160a0d0810f1579b8ac5baf27e9b44/solidity/contracts/bridge/Bridge.sol#L1466-L1472

Maybe we should consider changing it to `timedOutRedemption` in solidity contract (and then do the sime in ts lib), since, like i said, it looks to me that it only returns one redemption request.

typescript/src/ethereum.ts
michalsmiarowski: Ah, I see, you are right. Let's keep it as it is here 👍

typescript/src/chain.ts
lukasz-zimnoch: This is a Solidity convention, we use the same getter name as the object the getter refers to. I don't think we can change it now because the contract codebase is frozen now.

michalsmiarowski (approved):
LGTM 🤖

PR #377, opened 2022-06-29
Add long proof gas cost test for relay

eth-r:
A retarget with proofLength 50 (100 headers total) costs roughly 456k gas, continuing the trend of marginal cost of a retarget being around 3.7k gas per additional header.


solidity/test/relay/Relay.test.ts
```diff
@@ -559,6 +560,86 @@ describe("LightRelay", () => {
         ).to.be.revertedWith("Invalid target in new epoch")
       })
     })
+
+    context("after genesis (long chain)", () => {
+      const longGenesis = longHeaders.epochStart
+      const longGenesisBlock = 739872
+      const longRetargetBlock = 741888
+      const longHeaderStart = 741793
```
pdyraga, line 568: Are they used anywhere?

eth-r: For convenience, can move into comments.

pdyraga (approved):

PR #382, opened 2022-07-11
Add support for deployment on Goerli testnet

michalinacienciala:
Görli became a recommended test network after Ropsten's deprecation
notice (https://blog.ethereum.org/2022/06/21/testnet-deprecation/).
We're modifying GitHub Actions workflow for deploying `tbtc-v2`
contracts to support the deployment on Görli. We're also leaving the
possibility of deployment on Ropsten (this will be removed once we have
have the Görli deployment battle-tested and Ropsten gets shut down).

NOTE: We're temporarily using some testing configuration in the
workflow, which needs to be removed before merge to `main`.

TODO:
- [x] Merge the code with deployment scripts changes and verify the workflow. To do that we need:
  - [x] deployments scripts updated (https://github.com/keep-network/tbtc-v2/pull/387)
  - [x] fresh `@keep-network/ecdsa` package published (will be possible after merge of https://github.com/keep-network/keep-core/issues/3081)
  - [x] rename of CI modules merged and tagged (https://github.com/keep-network/ci/pull/30)
- [x] Remove testing configuration from the workflow

Refs:
https://github.com/keep-network/keep-core/issues/3012
https://github.com/threshold-network/solidity-contracts/pull/114
https://github.com/keep-network/keep-core/pull/3081


michalinacienciala:
@nkuba, ready for review. The workflow was verified here: https://github.com/keep-network/tbtc-v2/runs/7744385930?check_suite_focus=true.

.github/workflows/contracts.yml
```diff
@@ -142,25 +142,30 @@ jobs:
...
+            @keep-network/tbtc@${{ github.event.inputs.environment }}
 
       - name: Configure tenderly
-        if: github.event.inputs.environment == 'ropsten'
         env:
           TENDERLY_TOKEN: ${{ secrets.TENDERLY_TOKEN }}
         run: ./config_tenderly.sh
 
-      - name: Deploy contracts
+      - name: Deploy contract
```
nkuba, line 165: Is this change needed?

nkuba (approved):

.github/workflows/contracts.yml
michalinacienciala: No, that got there by mistake, I'm fixing it as part of https://github.com/keep-network/tbtc-v2/pull/392/files.

PR #391, opened 2022-08-01
Extracted BridgeFraudStub contract

pdyraga:
_“Evil is evil, Stregobor. Lesser, greater, middling, it's all the same. Proportions are negotiated, boundaries blurred. I'm not a pious hermit, I haven't done only good in my life. But if I'm to choose between one evil and another, then I prefer not to choose at all.”_ (_Geralt from Rivia_)

Before we can perform fixes on the target wallets commitment reimbursement, we need to reduce the size of `BridgeStub` contract. Extracting three functions used only for fraud tests to a separate contract is the option with the least changes to the post-audited code from all the options I've been exercising.

Thanks to this change, `BridgeStub` reduces its size from 24.570 to 23.649 (by 0.921). `BridgeFraudStub` is still at capacity, but we can bypass this problem by reducing the number of optimizer runs on `BridgeFraudStub`. I chose not to reduce the number of optimizer runs on `BridgeStub` because it is the contract deployed for the hardhat network used in tests. We need to use the same settings as for mainnet for gas reporter and for unit tests, as much as possible. If we do not do it just for `BridgeFraudStub`, this is the lesser evil that I preferred to choose here.


pdyraga:
Closing in favour of https://github.com/keep-network/tbtc-v2/pull/394/commits/6a60a1ed4f255f7f49b365b69fe3e56a07b2bcc9 from #394. 

The approach from #394 is used [in ECDSA module contracts](https://github.com/keep-network/keep-core/blob/0db388817178962926fd8ceb4a1b763a81e5ab88/solidity/ecdsa/hardhat.config.ts#L182) and does not require changes to unit tests.

PR #387, opened 2022-07-20
Görli contracts deployment

nkuba:
Görli became a recommended test network after Ropsten's deprecation notice ([link](https://blog.ethereum.org/2022/06/21/testnet-deprecation/)).

In this PR we update deployment scripts and configuration to support Görli.

The contracts were already deployed to Görli:
`@keep-network/tbtc-v2@0.1.1-goerli.0`

PR #392, opened 2022-08-02
Add job deploying contracts from `dapp-development` branch

michalinacienciala:
There are situations when team developing T Token Dashboard needs to
locally test some functionalities using modified contracts, for example
ones with shorter authorization decrease delay. We decided to create a
`dapp-development` branch in each of the upstream modules of
`threshold-network/token-dashboard` CI module, which would store the
code of these modified contracts. In this PR we create a
`contracts-dapp-development-deployment-testnet` job which deploys the
contracts, creates an NPM package (with `dappdev<environment>` suffix
and `dapp-development-<environment>` tag) and publishes it to the NPM
registry. At the end of the job we don't have a step informing CI about
completion of the workflow because we don't want to start the deployment
of the T Token Dashboard (which will soon be added as a downstream
module to the CI config) on a public testnet using modified contracts.
The job gets triggered only as a result of `workflow_dispath` event from
a `dapp-development` branch. Currently only `goerli` environment is
supported. We don't run system and unit tests for `dapp-development`
branch, as the tests are not configured to work with the modified
contracts.
Generally, the goal of the changes is to have the full set of
dapp-development-friendly contracts deployed to the NPM registry, so
that the dApp developers could quickly use them by upgrading the
`token-dashboard` dependencies using `yarn upgrade
<package-name>@dapp-development-goerli`.
If the workflow gets dispatched from a different branch than
`dapp-development`, the deploy will behave as it used to, publishing
package with deployed unmodified contracts to the NPM registry under
`<environment>` tag.

TODO:

- [x] Modify the secret for the deployer account in the `contracts-dapp-development-deployment-testnet` job
- [x] Bump actions to latest versions
- [x] Test the workflow. For that we need:
  - [x] [#3121](https://github.com/keep-network/keep-core/pull/3121) merged to `main` and `dapp-development-goerli`-tagged `random-beacon` and `ecdsa` packages published
  - [x] #387 merged to `main`
  - [x] `dapp-development` branch created
- [x] Remove the `dry-run` from the code

Refs:
https://github.com/threshold-network/token-dashboard/issues/136
https://github.com/threshold-network/solidity-contracts/pull/119
https://github.com/keep-network/keep-core/pull/3121


michalinacienciala:
Deployment tested on a branch that combines code from #392 and #403: https://github.com/keep-network/tbtc-v2/actions/runs/3052208927

nkuba (approved):

PR #393, opened 2022-08-02
Validate deposit refund locktime during deposit reveal

lukasz-zimnoch:
Here we introduce the validation of the deposit refund locktime during the deposit reveal. That validation ensures that a given period of time is preserved between the moment of the deposit reveal and the moment when the deposit becomes refundable. The exact duration of this period is determined by the `depositRevealAheadPeriod` governable parameter. This validation is needed to give the wallet enough time to sweep a revealed deposit and avoid potential competition with the depositor doing a deposit refund. The validation is not performed if the value of the `depositRevealAheadPeriod` is `0`.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -287,6 +287,7 @@ contract Bridge is
 
         self.depositDustThreshold = 1000000; // 1000000 satoshi = 0.01 BTC
         self.depositTxMaxFee = 100000; // 100000 satoshi = 0.001 BTC
+        self.depositRevealAheadPeriod = 15 days;
         self.depositTreasuryFeeDivisor = 2000; // 1/2000 == 5bps == 0.05% == 0.0005
         self.redemptionDustThreshold = 1000000; // 1000000 satoshi = 0.01 BTC
         self.redemptionTreasuryFeeDivisor = 2000; // 1/2000 == 5bps == 0.05% == 0.0005
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -188,7 +188,8 @@ contract Bridge is
     event DepositParametersUpdated(
         uint64 depositDustThreshold,
         uint64 depositTreasuryFeeDivisor,
-        uint64 depositTxMaxFee
+        uint64 depositTxMaxFee,
+        uint32 depositRevealAheadPeriod
     );
 
     event RedemptionParametersUpdated(
@@ -1214,19 +1215,26 @@ contract Bridge is
     ///        be incurred by each swept deposit being part of the given sweep
     ///        transaction. If the maximum BTC transaction fee is exceeded,
     ///        such transaction is considered a fraud.
+    /// @param depositRevealAheadPeriod New value of the deposit reveal ahead
+    ///        period parameter in seconds. It defines the length of the period
+    ///        that must be preserved between the deposit reveal time and the
+    ///        deposit refund locktime.
     /// @dev Requirements:
     ///      - Deposit dust threshold must be greater than zero,
     ///      - Deposit treasury fee divisor must be greater than zero,
     ///      - Deposit transaction max fee must be greater than zero.
+    ///      - Deposit reveal ahead period must be grater than zero.
     function updateDepositParameters(
         uint64 depositDustThreshold,
         uint64 depositTreasuryFeeDivisor,
-        uint64 depositTxMaxFee
+        uint64 depositTxMaxFee,
+        uint32 depositRevealAheadPeriod
     ) external onlyGovernance {
         self.updateDepositParameters(
             depositDustThreshold,
             depositTreasuryFeeDivisor,
-            depositTxMaxFee
+            depositTxMaxFee,
+            depositRevealAheadPeriod
         );
     }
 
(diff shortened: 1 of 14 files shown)


solidity/contracts/bridge/Bridge.sol
```diff
@@ -1213,19 +1215,26 @@ contract Bridge is
...
     ///        such transaction is considered a fraud.
+    /// @param depositRevealAheadPeriod New value of the deposit reveal ahead
+    ///        period parameter in seconds. It defines the length of the period
+    ///        that must be preserved between the deposit reveal time and the
+    ///        deposit refund locktime.
     /// @dev Requirements:
     ///      - Deposit dust threshold must be greater than zero,
     ///      - Deposit treasury fee divisor must be greater than zero,
     ///      - Deposit transaction max fee must be greater than zero.
+    ///      - Deposit reveal ahead period must be grater than zero.
```
pdyraga, line 1226: I think this is wrong. We allow for 0 value - it disables the validation.

solidity/contracts/bridge/BridgeState.sol
```diff
@@ -375,16 +383,22 @@ library BridgeState {
...
+    /// @param _depositRevealAheadPeriod New value of the deposit reveal ahead
+    ///        period parameter in seconds. It defines the length of the period
+    ///        that must be preserved between the deposit reveal time and the
+    ///        deposit refund locktime.
     /// @dev Requirements:
     ///      - Deposit dust threshold must be greater than zero,
     ///      - Deposit dust threshold must be greater than deposit TX max fee,
     ///      - Deposit treasury fee divisor must be greater than zero,
     ///      - Deposit transaction max fee must be greater than zero.
+    ///      - Deposit reveal ahead period must be grater than zero.
```
pdyraga, line 395: The description of `depositRevealAheadPeriod` field says that:

> The value equal to zero disables the validation of this parameter.

Also, it does not look like we have a non-zero validation rule for this field.

solidity/hardhat.config.ts
```diff
@@ -206,6 +206,7 @@ const config: HardhatUserConfig = {
     disambiguatePaths: false,
     runOnCompile: true,
     strict: true,
+    except: ["BridgeStub$"],
```
pdyraga, line 209: In #394 I went ahead and added `except: ["^contracts/test"],` but I think this version with only `BridgeStub$` is better. cc @dimpar

solidity/contracts/bridge/BridgeGovernanceParameters.sol
```diff
@@ -456,6 +464,49 @@ library BridgeGovernanceParameters {
         return self.newDepositTxMaxFee;
     }
 
+    /// @param _newDepositRevealAheadPeriod New deposit reveal ahead period.
```
pdyraga, line 467:
```suggestion
    /// @notice Begins the deposit reveal ahead period update process.
    /// @param _newDepositRevealAheadPeriod New deposit reveal ahead period.
```

This piece of documentation is sometimes available in `begin*` functions in this file and sometimes it's not, but I think we should add it to stay consistent with the majority of functions.

solidity/test/bridge/Bridge.MovingFunds.test.ts
```diff
@@ -3067,7 +3067,13 @@ describe("Bridge - Moving funds", () => {
...
                           // Set the max fee to one satoshi less than the fee
                           // used by the transaction.
-                          await bridge.setMovedFundsSweepTxMaxTotalFee(1999)
+                          await bridgeGovernance
+                            .connect(governance)
+                            .beginMovedFundsSweepTxMaxTotalFeeUpdate(1999)
+                          await increaseTime(constants.governanceDelay)
+                          await bridgeGovernance
+                            .connect(governance)
+                            .finalizeMovedFundsSweepTxMaxTotalFeeUpdate()
```
pdyraga, line 3076: With the `BridgeStub` contract size ignored, shouldn't we restore `setMovedFundsSweepTxMaxTotalFee` in the `BridgeStub`? I am fine either way but would be good to have consistency and do not add `setDepositRevealAheadPeriod` if we decide to drop `setMovedFundsSweepTxMaxTotalFee` and always go through the governance contract.

solidity/test/integration/FullFlow.test.ts
```diff
@@ -73,6 +74,14 @@ describeFn("Integration Test - Full flow", async () => {
       governance,
       dkgResultChallengePeriodLength
     )
+
```
pdyraga, line 77: Should we add a note that:
```
// Disable the reveal ahead period since refund locktimes are fixed
// within transactions used in this test suite.
```
?

Same comment to `integration/Slashing.test.ts`.

solidity/contracts/bridge/BridgeState.sol
lukasz-zimnoch: See 95a1f0f9

solidity/test/bridge/Bridge.MovingFunds.test.ts
lukasz-zimnoch: Reverted 20df3905

solidity/test/integration/FullFlow.test.ts
lukasz-zimnoch: See 997d2219

solidity/contracts/bridge/Bridge.sol
```diff
@@ -1213,19 +1215,25 @@ contract Bridge is
...
     ///        such transaction is considered a fraud.
+    /// @param depositRevealAheadPeriod New value of the deposit reveal ahead
+    ///        period parameter in seconds. It defines the length of the period
+    ///        that must be preserved between the deposit reveal time and the
+    ///        deposit refund locktime.
     /// @dev Requirements:
     ///      - Deposit dust threshold must be greater than zero,
     ///      - Deposit treasury fee divisor must be greater than zero,
-    ///      - Deposit transaction max fee must be greater than zero.
+    ///      - Deposit transaction max fee must be greater than zero.x
```
pdyraga, line 1225:
```suggestion
    ///      - Deposit transaction max fee must be greater than zero.
```

pdyraga (approved):

PR #394, opened 2022-08-03
Bridge reimburses the TX cost of submitMovingFundsCommitment

pdyraga:
Given that the Bridge requires `submitMovingFundsCommitment` to be called by one of the wallet members, this function can not be called and reimbursed via `MaintainerProxy`. Instead, `Bridge` must be responsible for reimbursing the cost. This is what is implemented in this PR.

The Bridge requires `submitMovingFundsCommitment` to be called by one of the wallet members, this function can not be called and reimbursed via `MaintainerProxy`.

The implementation in `MaintainerProxy.submitMovingFundsCommitment` was incorrect and the tests were passing only because they had an incorrect assumption of `MaintainerProxy` contract being the wallet member:
```
walletRegistry.isWalletMember
  .whenCalledWith(
    ecdsaWalletTestData.walletID,
    walletMembersIDs,
    maintainerProxy.address,
    walletMemberIndex
  )
.returns(true)
```

`MaintainerProxy` contract is not an operator and can not be a wallet member.


solidity/test/maintainer/MaintainerProxy.test.ts
```diff
@@ -2057,124 +2057,6 @@ describe("MaintainerProxy", () => {
...
-      state: walletState.Unknown,
-      movingFundsTargetWalletsCommitmentHash: ethers.constants.HashZero,
-    }
-
-    // Just an arbitrary main UTXO with value of 26 BTC.
-    const mainUtxo = {
-      txHash:
-        "0xc9e58780c6c289c25ae1fe293f85a4db4d0af4f305172f2a1868ddd917458bdf",
-      txOutputIndex: 0,
-      txOutputValue: to1ePrecision(26, 8),
```
dimpar, line 2078: We can now remove the import at the top.

solidity/contracts/bridge/Bridge.sol
```diff
@@ -209,6 +210,7 @@ contract Bridge is
         uint32 movingFundsTimeout,
         uint96 movingFundsTimeoutSlashingAmount,
         uint32 movingFundsTimeoutNotifierRewardMultiplier,
+        uint16 movingFundsCommitmentGasOffset,
```
dimpar, line 213: I think this should be enough, but just to point out that we limit ourselves to 65535 here.

pdyraga: Yeah. I think this limit is enough. No offset is higher than 27k in `MaintainerProxy` now and we do not have a call to a separate contract in `Bridge` when committing to target wallets.

solidity/contracts/bridge/BridgeState.sol
```diff
@@ -621,6 +628,7 @@ library BridgeState {
         uint32 _movingFundsTimeout,
         uint96 _movingFundsTimeoutSlashingAmount,
         uint32 _movingFundsTimeoutNotifierRewardMultiplier,
+        uint16 _movingFundsCommitmentGasOffset,
```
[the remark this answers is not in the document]
pdyraga: 13121bed84d416c93a9ddd509eb7991d9f42d091

solidity/test/maintainer/MaintainerProxy.test.ts
pdyraga: 4c0301124afcfced638004a9b9b16ee038d5c6e6

dimpar (approved):

PR #396, opened 2022-08-04
Delete entry from `timedOutRedemptions` when it has been processed

pdyraga:
In `BridgeState` we used to say that:

> Timed out requests are stored in this mapping to avoid slashing the wallets multiple times for the same timeout.

This was not true, because when it comes to requesting redemption we say that:
> There is no need to check for existence in timedOutRedemptions since the  wallet's state is changed to other than Live after first time out is reported so making new requests is not possible.

And when the redemption timeout is reported we say that:
> It is worth noting that there is no need to check if timedOutRedemption mapping already contains the given redemption key. There is no possibility to re-use a key of a reported timed-out redemption because the wallet responsible for causing the timeout is moved to a state that prevents it to receive new redemption requests.

Given that the wallet lifecycle prevents the wallet from being slashed twice for the redemption timeout, we can delete an entry from `timedOutRedemptions` mapping when the timed-out redemption has been processed.

Redemption request is deleted from `pendingRedemptions` when it is processed. Pending redemption can be processed in two ways: it can be used in SPV proof and it can be marked as timed-out. Deleting `timedOutRedemptions` when it is processed is consistent with how `pendingRedemptions` are deleted and leaves no dangling state in the Bridge. A timed-out redemption in `timedOutRedemptions` can be processed in just one way: used in SPV proof.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -1517,12 +1517,14 @@ contract Bridge is
     ///         (computed using Bitcoin HASH160 over the compressed ECDSA
     ///         public key) and `redeemerOutputScript` is the Bitcoin script
     ///         (P2PKH, P2WPKH, P2SH or P2WSH) that is involved in the timed
-    ///         out request. Timed out requests are stored in this mapping to
-    ///         avoid slashing the wallets multiple times for the same timeout.
+    ///         out request.
     ///         Only one method can add to this mapping:
     ///         - `notifyRedemptionTimeout` which puts the redemption key
-    ///           to this mapping basing on a timed out request stored
+    ///           to this mapping based on a timed out request stored
     ///           previously in `pendingRedemptions` mapping.
+    ///         Only one method removes entries from this mapping:
+    ///         - `submitRedemptionProof` in case the timed out redemption
+    ///           request was a part of the proven transaction.
     function timedOutRedemptions(uint256 redemptionKey)
         external
         view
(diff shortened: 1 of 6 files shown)


tomaszslabon (approved):

solidity/contracts/bridge/Bridge.sol
```diff
@@ -1517,12 +1517,14 @@ contract Bridge is
...
     ///         (P2PKH, P2WPKH, P2SH or P2WSH) that is involved in the timed
-    ///         out request. Timed out requests are stored in this mapping to
-    ///         avoid slashing the wallets multiple times for the same timeout.
+    ///         out request.
     ///         Only one method can add to this mapping:
     ///         - `notifyRedemptionTimeout` which puts the redemption key
-    ///           to this mapping basing on a timed out request stored
+    ///           to this mapping based on a timed out request stored
     ///           previously in `pendingRedemptions` mapping.
+    ///         Only one method removes entries from this mapping:
```
tomaszslabon, line 1525: Nitpicking, but choose either :
"Only one method can add entries to this mapping:"
...
"Only one method can remove entries from this mapping:"
Or:
"Only one method adds entries to this mapping:"
...
"Only one method removes entries from this mapping:"

pdyraga: https://github.com/keep-network/tbtc-v2/pull/396/commits/078f6f45a59bbe62489de378232961174446bd9d

tomaszslabon (approved):

lukasz-zimnoch (approved):

PR #402, opened 2022-08-12
Execute script initializing WalletOwner after other scripts

michalinacienciala:
The `initializeWalletOwner` function can be called only once per one
instance of `Bridge` contract. This means that once we run the deploy
script calling this function, we cannot successfully call it again using
same `ecdsa` package as a dependency. In our CI we run all the
deployment scripts from the `deploy` folder as a one step (they execute
in alphabetical order). If execution of some deployment script fails, we
re-run all the scripts again. If some script executed after `16_
initialize_wallet_owner.ts` would fail, we would not be able to simple
re-run the job deploying `tbtc-v2` contracts, but we would have to
publish a new `ecdsa` package to get fresh `Bridge` contract.
By renaming the `16_initialize_wallet_owner.ts` to
`90_initialize_wallet_owner.ts`, we move the execution of the script to
end, minimizing the risk of having to deploy a new `ecdsa` package.


nkuba:
I've just found out that I introduced similiar changes but forgot to create a PR (https://github.com/keep-network/tbtc-v2/commit/5cda93de4abf83ece835b49b708a688e0d70ff12) 🤦🏻 Could you also add `func.runAtTheEnd = true` in the script?

nkuba (changes requested):
Please add `func.runAtTheEnd = true` in the `90_initialize_wallet_owner.ts` script.

michalinacienciala:
> Please add `func.runAtTheEnd = true` in the `90_initialize_wallet_owner.ts` script.

@nkuba, addressed, see [bf074d6](https://github.com/keep-network/tbtc-v2/pull/402/commits/bf074d6c38a9e7641ae1dba7e498fcdf3cb94745).

nkuba (approved):

PR #404, opened 2022-09-05
Does not allow updating an existing allowance to a non-zero value

pdyraga:
Does not allow updating an existing allowance to a value that is non-zero to
avoid someone using both the old and the new allowance by unfortunate
transaction ordering. To update an allowance to a non-zero value one should
update to zero first or use `increaseBalanceAllowance` or
`decreaseBalanceAllowance` for an atomic update. We have been warning about this
case in `approveBalance` documentation, but we were not enforcing this behavior.
Now we have a `require` making sure the rule is obeyed.

### DIFF
--- a/solidity/test/bank/Bank.test.ts
+++ b/solidity/test/bank/Bank.test.ts
@@ -399,24 +399,51 @@ describe("Bank", () => {
     context("when the spender had no approved balance before", () => {
       let tx: ContractTransaction
 
-      before(async () => {
-        await createSnapshot()
+      context("when setting approval to non-zero amount", () => {
+        before(async () => {
+          await createSnapshot()
 
-        tx = await bank.connect(owner).approveBalance(spender, amount)
+          tx = await bank.connect(owner).approveBalance(spender, amount)
+        })
+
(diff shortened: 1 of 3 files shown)


solidity/test/bank/Bank.test.ts
```diff
@@ -424,15 +424,33 @@ describe("Bank", () => {
...
+        after(async () => {
+          await restoreSnapshot()
+        })
+
+        it("should replace the previous allowance with zero", async () => {
+          expect(await bank.allowance(owner.address, spender)).to.equal(0)
+        })
+      })
+
+      context("when trying to overwrite with a non-zero value", () => {
```
nkuba, line 448: Can we add a test case where the current approval is zero and the new approval is zero?

pdyraga: Sure, on it.

pdyraga: d782d5800b4da239d1346b33e6ae0c4d89cbfa5c

nkuba (approved):

PR #407, opened 2022-09-06
Warn about 32-bits timestamp representation limitation

pdyraga:
~~Depends on #405~~

The timestamp/locktime is represented using 32 bits for gas efficiency and to be consistent with Bitcoin.

This way, `OperatorRewards` structure in `Rewards` contract takes a single EVM word and the reward withdrawal consumes less gas. `DepositRequest` structure in `Deposit` library of the `Bridge` contract takes two EVM words and the operation revealing a deposit and the operation sweeping a deposit consume less gas. `Wallet` structure takes 4 EVM words and wallet lifecycle operations consume less gas.

`FraudChallenge`, `MovingFundsSweepRequest` ,`Redemtion` libraries of `Bridge` contract represent the timestamp as `uint32` for consistency with the rest of the code.

Note that Bitcoin represents the locktime using 32 bits as well: https://developer.bitcoin.org/devguide/transactions.html#locktime-and-sequence-number

Added a comment to the code to make it clear the `uint32` representation is sufficient only until 2106.

### DIFF
--- a/solidity/contracts/bridge/Deposit.sol
+++ b/solidity/contracts/bridge/Deposit.sol
@@ -85,7 +85,7 @@ library Deposit {
         // Deposit amount in satoshi.
         uint64 amount;
         // UNIX timestamp the deposit was revealed at.
-        // XXX: unsigned 32-bit integer unix seconds, will break around 2106
+        // XXX: Unsigned 32-bit int unix seconds, will break February 7th 2106.
         uint32 revealedAt;
         // Address of the Bank vault the deposit is routed to.
         // Optional, can be 0x0.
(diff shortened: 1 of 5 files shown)


solidity/contracts/bridge/Deposit.sol
```diff
@@ -85,6 +85,7 @@ library Deposit {
         // Deposit amount in satoshi.
         uint64 amount;
         // UNIX timestamp the deposit was revealed at.
+        // XXX: unsigned 32-bit integer unix seconds, will break around 2106
```
nkuba, line 88: Let's be specific that it's about `the year 2016`, as it's unclear what the number `2016` means. Also, it will happen not `around 2016` but `in 2016` (on Sunday, 7 February 2106).

```suggestion
        // XXX: unsigned 32-bit integer unix seconds, will break in the year 2106
```

nkuba (changes requested):

solidity/contracts/bridge/Deposit.sol
pdyraga: 48b54c8bebdaca20defc8d6c986c8102ead443a6

PR #414, opened 2022-10-11
RFC 7: Sweeping coordination

pdyraga:
Before generating tECDSA signature, the off-chain client software needs to agree in the signing group on the message being signed. In tBTC v2, the information needed to evaluate the signed message comes from multiple sources that might not be ideally synced with each other. Moreover, it is not possible to establish a precise point in time based on which the information should be retrieved.

This document aims at describing a protocol for reaching an agreement about the Bitcoin sweep transaction between the signing group members.

The protocol should tolerate small differences in the state of the world, as seen by individual signing group members. It should be compatible with a light relay, and support signing retries. All transactions must be sufficiently confirmed and the transaction being signed must adhere to the recent Bitcoin network fees.

During the work on this RFC, I had to adjust some parts of RFC 1. Specifically, I removed the "early sweeping" section that we are going to replace with a market-based solution in the future, and I replaced `sweep_max_btc` with `sweep_max_deposits`. The time it takes to sign a sweep transaction input and Ethereum block gas limits put a cap on how many deposits can be swept at once. It makes more sense to trigger sweep early, given the number of deposits rather than look at the amount of BTC deposited because it does not affect the submit SPV proof transaction that much.


docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,201 @@
...
+The Bitcoin sweep transaction must have a fee that allows it to be mined in a
+reasonable time. If the Bitcoin network conditions change, there must be a
+mechanism to increase the fee and replace transactions. The wallet must finish
+the current operation before attempting another one.
+
+==== Retries
+
+The tECDSA protocol used for signing must be able to retry in case some signing
+group members are offline or corrupted. The tBTC v2 client software should join
+the pending signing after the client restart.
```
lukasz-zimnoch, line 99: Just to be clear, we are not aiming to join the current attempt but we are just waiting for the new attempt proposal from the Coordinator right?

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+If the current Coordinator does not propose a sweep within the given time
+window, the attempt is skipped.
+
+
+==== Sweep transaction
+
+The Coordinator chooses sufficiently confirmed deposits for the sweep. To avoid
+synchronization problems with deposits on the brink of being sufficiently
+confirmed, the Coordinator may choose deposits with more than 6/20 confirmations.
```
lukasz-zimnoch, line 129: We pick up deposits whose Ethereum deposit reveal transaction has more than 20 confirmations and their corresponding Bitcoin P2(W)SH deposit transaction has more than 6 confirmations right? If so, I think we should make it super clear here as this is not obvious at the first sight. Additionally, we should mention that the Coordinator must make sure that the UTXO produced by the P2(W)SH deposit transaction can be unlocked using the wallet's public key.

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,201 @@
...
+To avoid problems with a maximum size of a transaction, the Coordinator should
+not pick more than 500 deposits. The Coordinator chooses Bitcoin transaction fee
```
lukasz-zimnoch, line 131: Is 500 a safe value? Won't we run out of gas trying to submit an SPV proof of such a sweep transaction to the `Bridge` contract?

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,201 @@
...
+
+The Coordinator must not propose a sweep if there is another transaction from
+the wallet in the Bitcoin mempool.
+
+The Coordinator proposes a sweep over the wallet's broadcast channel.
+
+All signing group members must confirm that all deposits in the proposed sweep
+are sufficiently confirmed, that the maximum number of deposits is not exceeded,
+and that the proposed fee reflects the current situation in the Bitcoin network
+plus/minus some margin.
```
lukasz-zimnoch, line 142: All signing group members should also make sure that all UTXOs produced by the selected P2(W)SH deposit transactions can be unlocked using the wallet's public key.

docs/rfc/rfc-7.adoc
pdyraga: Say there are 10 group members currently offline:

1. The Coordinator proposes a sweep transaction.
2. 90 group members who are online validate this transaction, all of them agree to sign, and inform the rest of the group about the successful validation.
3. 10 group members are back online
4. For some reason, the signing attempt fails. 
5. 90 group members who performed the first attempt, broadcast information about the sweep transaction the are signing.
6. 10 members see that all 90 members (> t+1) agree to do the signing, they validate the transaction as well, and everything is fine, they join signing.

In this scheme, the Coordinator does not have to be online for every retry. The current state is propagated as long as there are t+1 members online.

docs/rfc/rfc-7.adoc
pdyraga: I do not know, this number is something I've "just picked" eyeballing. There is an Open Questions section at the bottom informing that this assumption needs to be validated.

docs/rfc/rfc-7.adoc
lukasz-zimnoch: Ok got it! Thanks!

docs/rfc/rfc-7.adoc
lukasz-zimnoch: Sure I'm asking because `Open Questions` section focuses only on the fraud-related side of this value. We should remember that we can have troubles also on the happy-path side i.e. SPV proof submission.

pdyraga: Ah, right. I'll improve it.

docs/rfc/rfc-7.adoc
pdyraga: See c0d2950d0b96fab1fdaf0e51231ff660e3985976, please.

docs/rfc/rfc-7.adoc
pdyraga: See c0d2950d0b96fab1fdaf0e51231ff660e3985976, please.

docs/rfc/rfc-7.adoc
pdyraga: See c0d2950d0b96fab1fdaf0e51231ff660e3985976, please.

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,205 @@
...
+
+=== Implementation
+
+==== Coordinator
+
+We introduce a rotating role of Coordinator responsible for proposing a sweep
+Bitcoin transaction. The role of Coordinator rotates every 6 hours across all
+operators in the group. For example, if the group consist of the following
+members: `[0xA, 0xB, 0xC, 0xC, 0xB, 0xD`] the rotation rounds are: `0xA, 0xB,
+0xC, 0xD`. The rotation schedule begins at UNIX epoch (00:00:00 UTC on 1 January
```
eth-r, line 109: I'm not sure whether we'd prefer to rotate by members or seats; rotating by members means that a relatively small number of uncooperative operators can disproportionately affect a group consisting mostly of large operators. On the other hand, if we rotate by seats a single large uncooperative operator can clog the group for excessive amounts of time. We may want to make this reasoning and the chosen tradeoff explicit.

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,205 @@
...
+the light relay but one day we may switch to the sparse relay. This would allow
+establishing a reference between a point in time on Ethereum and Bitcoin chains.
+The reference point in time on Ethereum is a fixed block number - for example,
+we sweep every 100th block - and the reference point in time on Bitcoin is the
+last Bitcoin block recorded on Ethereum's sparse relay at the given Ethereum's
+block number.
+
+== Open Questions
+
+Is the maximum of 500 transactions in a batch enough to not risk not being able
```
lukasz-zimnoch, line 201: One important thing just came to my mind. We must remember that when we talk about signing a sweep transaction, we are actually talking about signing the N inputs (i.e. unlocking the deposits' UTXOs) of the sweep transaction. That means we need N independent instances of the signing protocol to be done. This is a major performance challenge we must be aware of. It also opens some complex scenarios like, what should we do if only part of N signatures were produced?

pdyraga: That's a good point, I reworked the approach in https://github.com/keep-network/tbtc-v2/pull/414/commits/b057fefa249900c5e31b7166316bca3cd0839d50.

Every input of the Bitcoin sweep transaction needs to be signed separately. Given that each message being signed contains information about all other inputs and the output, the wallet must produce all signatures. Instead of giving up
after 2 hours, the wallet should block all other operations and try for at least a week. All signatures are done simultaneously, in separate instances of the signing state machine. This requires limiting the number of inputs being swept to 20 initially.

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,205 @@
...
+The tECDSA protocol used for signing must be able to retry in case some signing
+group members are offline or corrupted. The tBTC v2 client software should join
+the pending signing after the client restart.
+
+=== Implementation
+
+==== Coordinator
+
+We introduce a rotating role of Coordinator responsible for proposing a sweep
+Bitcoin transaction. The role of Coordinator rotates every 6 hours across all
```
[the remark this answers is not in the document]
pdyraga: My bad, they should rotate every 8 hours. Updated in https://github.com/keep-network/tbtc-v2/pull/414/commits/067cda0821dd329e2786412e70e53848841d5be7.

docs/rfc/rfc-7.adoc
pdyraga: The reasoning behind rotating per operator was that if a large operator is offline and they have more than one seat one after another, they will block sweeping altogether for a long time. If we rotate by operator, we will lose one sweep but the next operator, which is hopefully online, will propose the sweep in the very next time window.

On the other hand, based on experience with v1, large operators are usually running nodes better than small operators...

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,220 @@
...
+which takes 2h 30 min.
+- With 4 malicious members in a signing group, we need 20 retries of the
+protocol in the worst case (`P = (96 choose 51) / (100 choose 51) = 0.05403311465`),
+which takes 5h.
+- With 5 malicious members in a signing group, we need 50 retries of the
+protocol in the worst case (`P = (95 choose 51) / (100 choose 51) = 0.02532802249`),
+which takes 12h 30 min.
+
+Given the fact, the first few retries will focus on inactivity and the
+probability of having more than 2 malicious members in a signing group is low,
```
lukasz-zimnoch, line 189: I don't think this is true (https://github.com/keep-network/keep-core/pull/3352#issuecomment-1274670323). As we discussed, we plan to use the announcement phase before the signature attempt to filter out inactive members early.

pdyraga:
Please hold on with reviewing this RFC. There are some performance problems with simultaneous signing of inputs and I need to rethink the approach.

docs/rfc/rfc-7.adoc
pdyraga: Addressed in the most recent version.

pdyraga:
Reworked the RFC after incorporating the changes described in https://github.com/keep-network/keep-core/issues/3366 to the client. This is now ready for review!

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,237 @@
...
+
+=== Implementation
+
+==== Coordinator
+
+We introduce a rotating role of Coordinator responsible for proposing a sweep
+Bitcoin transaction. The role of Coordinator rotates every 8 hours across all
+operators in the group. For example, if the group consist of the following
+members: `[0xA, 0xB, 0xC, 0xC, 0xB, 0xD`] the rotation rounds are: `0xA, 0xB,
+0xC, 0xD`. The rotation schedule begins at UNIX epoch (00:00:00 UTC on 1 January
```
michalinacienciala, line 108: Just to make sure - the order of the operators on the list we're iterating on will be random, it won't be alphabetically-ordered, right? I'm assuming that if the wallet members are [0xA, 0xC, 0xB, 0xC, 0xE, OxA, 0xD, ...] the order of operators on the list will be [0xA, 0xC, 0xB, 0xE, 0xD, ...]?

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,237 @@
...
+- 16:00 UTC - 16:30 UTC
+
+If the current Coordinator does not propose a sweep within the given time
+window, the attempt is skipped.
+
+The Coordinator must not propose a sweep if there is currently a pending signing
+operation being executed by the wallet or if the last signed transaction from
+the wallet does not have 6 Bitcoin confirmations yet.
+
+The coordinator may slightly delay broadcasting the proposition to give all
```
michalinacienciala, line 127:
```suggestion
The Coordinator may slightly delay broadcasting the proposition to give all
```

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,237 @@
...
+clients in the group a chance to properly recognize a Coordinator rotation.
```
michalinacienciala, line 128: Is that worry about the possible problem with recognizing a Coordinator rotation dictated by a concern that the clients may operate on the systems which may have a slightly different understanding of what a `00:00 UTC` is?

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,237 @@
...
+everyone needs to wait for another sweep window.
+
+==== Increasing fee
+
+At any point in time, the current Coordinator (who is not necessarily the same
+Coordinator who proposed the batch) can propose increasing the Bitcoin fee for
+the sweep transaction that is in the mempool. The signing group members sign the
+transaction if the original transaction is in the mempool for at least 30
+minutes and the fee increase does not exceed the maximum one allowed by the
+Bridge. Given that the wallet is blocked with any other action until the sweep
```
michalinacienciala, line 223: What do you think of referencing here the names of the governable parameters controlling these value? I.e. `sweeping_fee_bump_period`, `btc_fee_max`.
Or maybe linking to https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-1.adoc#234-sweeping for more details?

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,237 @@
...
+wallet does not have 6 Bitcoin confirmations yet.
+
+The Coordinator proposes a sweep over the wallet's broadcast channel.
+
+All signing group members must confirm that all deposits in the proposed sweep
+are sufficiently confirmed, that the maximum number of deposits is not exceeded,
+that the proposed fee reflects the current situation in the Bitcoin network
+plus/minus some margin, that all deposit UTXOs can be unlocked with the
+wallet's public key, and that there is enough time to unlock UTXO before the
+refund.
```
michalinacienciala, line 153: Should members also make sure that operator who proposed the sweep is a rightful Coordinator?

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,237 @@
...
+
+When proposing a sweep, the Coordinator sets the Ethereum block number at which
+the signing protocol should start. All clients validate if that block's number
+is +-2 blocks from the current one.
+
+If all these requirements are met, the last transaction executed by the wallet
+is sufficiently confirmed, and there is no other signing by the wallet in
+progress, the signing group members proceed with tECDSA signing protocol.
```
michalinacienciala, line 161: Let's say the Coordinator proposed a sweep at block `x` and set the time of signing protocol start to `x+2`. Is there a possibility that some separate signing would start at `x+1`? For example a signing of redemption of some older deposits?

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,237 @@
...
+The Coordinator must not propose a sweep if there is currently a pending signing
+operation being executed by the wallet or if the last transaction from the
+wallet does not have 6 Bitcoin confirmations yet.
+
+The Coordinator proposes a sweep over the wallet's broadcast channel.
+
+All signing group members must confirm that all deposits in the proposed sweep
+are sufficiently confirmed, that the maximum number of deposits is not exceeded,
+that the proposed fee reflects the current situation in the Bitcoin network
+plus/minus some margin, that all deposit UTXOs can be unlocked with the
```
michalinacienciala, line 151: Regarding the proposed fee I think members should not only check if it reflects the current situation in Bitcoin, but also check that it's not higher than `base_btc_fee_max`.

docs/rfc/rfc-7.adoc
lukasz-zimnoch: Just a note, maybe we should make it clear that always we say "operator" we basically mean the client instance and when we say "member" we talk about group seat. Looks like this is a common point of confusion.

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,237 @@
...
+
+==== Sweep transaction
+
+The Coordinator chooses sufficiently confirmed deposits for the sweep. To avoid
+synchronization problems with deposits on the brink of being sufficiently
+confirmed, the Coordinator may choose deposits with more than 6 confirmations
+on Bitcoin and 20 confirmations on Ethereum. To avoid problems with proving
+large transactions and too many signatures to be executed, the Coordinator
+should not pick more than 10 deposits. The Coordinator chooses the Bitcoin
+transaction fee reflecting the current situation in the Bitcoin network. To
```
lukasz-zimnoch, line 138: And it should probably do it with respect to the maximum BTC fee accepted by the Bridge (just like Michalina noted below https://github.com/keep-network/tbtc-v2/pull/414/files#r1012844438)

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,237 @@
...
+The signing that keeps failing is retried for no more than 48 hours. Since no
+interim signatures were produced, it is safe to stop the signing without exposing
+wallet for risk of being slashed for signature fraud.
+
+No state management logic is required for retries other than keeping information
+in-memory about the sweep transaction being signed. If more than t+1 clients get
+restarted during the sweep transaction signing, the sweep attempt is skipped and
+everyone needs to wait for another sweep window.
+
+==== Increasing fee
```
lukasz-zimnoch, line 216: Worth mentioning that the wallet members need to hold the preimages of all transactions that were not mined in order to be able to defeat potential fraud challenges.

docs/rfc/rfc-7.adoc
pdyraga: Correct, they are ordered in the way they were selected to the wallet signing group.

docs/rfc/rfc-7.adoc
pdyraga: Yes! Syncing local time with an NTP server feels like an industry standard but we can't guarantee enough group members are in sync. It does not necessarily mean we need to sleep for a second or so. The time it takes to compute a batch may be a delay long enough.

docs/rfc/rfc-7.adoc
pdyraga: Done!

docs/rfc/rfc-7.adoc
pdyraga: Yeah, the RFC says:

> Sweep can only be proposed by the current Coordinator.

If this is not clear from this sentence, I can make it even more explicit.

docs/rfc/rfc-7.adoc
pdyraga: Nope. Only one signing may be in progress and the operators announce their readiness to sign something.

docs/rfc/rfc-7.adoc
pdyraga: Yes, clarified!

solidity/contracts/bridge/Bridge.sol
```diff
@@ -322,8 +322,8 @@ contract Bridge is
         self.fraudChallengeDefeatTimeout = 7 days;
         self.fraudSlashingAmount = 100 * 1e18; // 100 T
         self.fraudNotifierRewardMultiplier = 100; // 100%
-        self.walletCreationPeriod = 1 weeks;
-        self.walletCreationMinBtcBalance = 1e8; // 1 BTC
+        self.walletCreationPeriod = 0;
+        self.walletCreationMinBtcBalance = 0;
```
pdyraga, line 326: This is an accident, I'll rebase.

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,288 @@
...
+
+- `sweep_period`: The amount of time we wait between scheduled sweeps on a wallet.
+- `sweep_max_deposits`: The number of non-dust unswept revealed bitcoin deposits
+  that will trigger an early sweep on a wallet.
+- `sweeping_fee_bump_period`: The amount of time we wait to see if a sweeping
+  transaction is mined before increasing the fee.
+- `sweeping_fee_multiplier_increment`: The amount we add to the sweeping fee
+  multiplier each time a sweeping transaction is not mined within the
+  `sweeping_fee_bump_period`. For example, if this param is set to 0.2 and we
+  are currently at 1.6x, then the next time we would try 1.8x.
```
michalinacienciala, line 288: Is it intentional that we don't mention `sweeping_refund_safety_time`, `dust_threshold`, `base_btc_fee_max`, `sweeping_fee_max_multiplier` and `btc_fee_max` here? I know we don't mention them in the document explicitly, but we refer to some of them indirectly (for example when mentioning `non-dust unswept revealed bitcoin deposits` or `the fee exceeds the maximum one allowed by the Bridge`).

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,288 @@
...
+operators in the group. For example, if the group consist of the following
+members: `[0xA, 0xB, 0xC, 0xC, 0xB, 0xD`] the rotation rounds are: `0xA, 0xB,
+0xC, 0xD`. The rotation schedule begins at UNIX epoch (00:00:00 UTC on 1 January
+1970).
+
+==== Sweeping schedule
+
+Sweep can only be proposed by the current Coordinator.
+
+The Coordinator may propose a sweep at any time if either `sweep_period` seconds
```
beaushinkle, line 115: Does this mean that the sweeping schedule starts looking funky? As in

Say that we launch the wallet without early sweeping and schedule the first sweep for 12:00pm.

We would always have sweeps at

+ 12:00pm
+ 8:00pm
+ 4:00am

every day. If we introduce early sweeping then the following happens:

+ Scheduled Sweep at 12:00pm
+ Early Sweep at 3:00pm
+ Scheduled Sweep at 11:00pm
+ Scheduled Sweep at 7:00am
+ Early Sweep at 10:00am
+ Early Sweep at 1:00pm
+ Scheduled Sweep at 9:00pm

If instead we change it to "we always attempt to sweep every 8 hours, and can sweep early if we have enough deposits", then we have:

+ Scheduled Sweep at 12:00pm
+ Early Sweep at 3:00 pm
+ Scheduled Sweep at 8:00pm
+ Scheduled Sweep at 4:00am
+ Early Sweep at 10:00am
+ Scheduled Sweep at 12:00pm
+ Scheduled Sweep at 8:00pm

The second one keeps us on the 12->8->4->12 loop, with early sweeps sprinkled in between. Definitely a UX/product decision though!

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,288 @@
...
+The Coordinator chooses sufficiently confirmed deposits for the sweep. To avoid
+synchronization problems with deposits on the brink of being sufficiently
+confirmed, the Coordinator may choose deposits with more than 6 confirmations
+on Bitcoin and 20 confirmations on Ethereum. To avoid problems with proving
+large transactions and too many signatures to be executed, the Coordinator
+should not pick more than 15 deposits. The Coordinator chooses the Bitcoin
+transaction fee reflecting the current situation in the Bitcoin network and the
+maximum possible fee accepted by the Bridge. To leave enough time for retries,
+deposits for which Bitcoin transactions unlock for a refund earlier than two
+weeks from now are skipped.
```
beaushinkle, line 152: Two weeks is a lot of buffer! What's the attack vector here?

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,288 @@
...
+In the worst case, it takes approximately 60 hours to sign 15 inputs.
+
+This attack slows down the sweeping schedule but given that the active wallet is
+not performing redemptions or moving funds, it does not affect the funds already
+bridged.
+
+In the first release, the client should assume no more than 3% of the network is
+corrupted and retry the signing of a single input no more than 10 times. If the
+signing of an input failed 10 times, the entire sweep is stopped. This puts the
+group in danger of a fraud challenge from one of the group members but the group
```
beaushinkle, line 247: what's the fraud challenge here? I thought that operators didn't have to sweep deposits

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,288 @@
...
+will have a chance to sweep the deposit during the next sweep window. To defeat
+the fraud challenge, it is enough that the deposit has been swept, and it does
+not matter in which particular transaction.
```
beaushinkle, line 250: Can you elaborate on the fraud challenge here?

docs/rfc/rfc-7.adoc
pdyraga: The last option is definitely a goal. This is explained in line 131: 

> If we sweep early because a lot of deposits have been revealed, then we don't
"push back" our scheduled `sweep_period` sweep. Rather, that sweep continues
as planned.

I will add "we always attempt to sweep every 8 hours, and can sweep early if we have enough deposits" to the beginning of this paragraph as a clarification.

docs/rfc/rfc-7.adoc
pdyraga: Frauds work in a challenge and defeat-the-challenge mode. The challenger [provides a signature and `preimageSha256`](https://github.com/keep-network/tbtc-v2/blob/50267dbf9df4427f475544e459e7ce5a8aa0284a/solidity/contracts/bridge/Fraud.sol#L134-L139). If the wallet signed a transaction, submitted it to the mempool, and after some time decided to bump up the miner fee providing signature over another transaction, the first signature can be used for a challenge. Since that transaction does not exist on the chain, we can't look it up there. Actually, I am not even sure if such a lookup would work for mined transaction. In other words, the wallet must store all successful signatures locally, regardless if they were for transactions that got mined or for transactions that got replaced.

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,288 @@
...
+
+==== Retries
+
+
+The signing protocol is non-attributable and we need to retry in case of
+corrupted data:
+
+- With 1 malicious member in a signing group, we need 2 attempts of the protocol
+in the worst case (`P = (99 choose 51) / (100 choose 51) = 0.49`).
+In the worst case, it takes approximately 2 hours to sign 15 inputs.
```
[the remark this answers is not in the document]
pdyraga: See e6e08163b9834babe89905f98a925b9d38e617c5 please.

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,288 @@
...
+In the first release, the client should assume no more than 3% of the network is
+corrupted and retry the signing of a single input no more than 10 times. If the
+signing of an input failed 10 times, the entire sweep is stopped. This puts the
+group in danger of a fraud challenge from one of the group members but the group
+will have a chance to sweep the deposit during the next sweep window. To defeat
+the fraud challenge, it is enough that the deposit has been swept, and it does
+not matter in which particular transaction.
+
+In the first release, the group maintains the signing state in memory, without
+persisting it to disk. It means that if more than t+1 members restarted their
```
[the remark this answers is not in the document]
pdyraga: But isn't it exactly what this sentence says? If 51 or more members restarted their clients, the in-memory state is gone and the signing will not complete.

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,288 @@
...
+
+==== Sweeping schedule
+
+Sweep can only be proposed by the current Coordinator.
+
+The Coordinator may propose a sweep at any time if either `sweep_period` seconds
+passed since the proof of the last sweep was accepted by the Bridge or
+`sweep_max_deposits` waits in the queue for sweeping. The time it takes to sign
+a sweep transaction input and Ethereum block gas limits put a cap on how many
+deposits can be swept at once. Given that in normal circumstances, the active
```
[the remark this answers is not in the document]
pdyraga: Let's make it more clear! https://github.com/keep-network/tbtc-v2/pull/414/commits/d552aeeaa9a7b82daff2070847ccc43dea5dde9a

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,288 @@
...
+0xC, 0xD`. The rotation schedule begins at UNIX epoch (00:00:00 UTC on 1 January
+1970).
+
+==== Sweeping schedule
+
+Sweep can only be proposed by the current Coordinator.
+
+The Coordinator may propose a sweep at any time if either `sweep_period` seconds
+passed since the proof of the last sweep was accepted by the Bridge or
+`sweep_max_deposits` waits in the queue for sweeping. The time it takes to sign
```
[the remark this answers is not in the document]
pdyraga: Much better! https://github.com/keep-network/tbtc-v2/pull/414/commits/d552aeeaa9a7b82daff2070847ccc43dea5dde9a

docs/rfc/rfc-7.adoc
pdyraga: I added `sweeping_refund_safety_time` to the list and explicitly mentioned initial values for the parameters, see 45406ccdc677c9922bbe23ecf6bef3db937d840b. Regarding the rest of the parameters, they are already included in the `Bridge` contract... under different names that matched better the naming scheme we used. I did not include them to not cause confusion and just described them in English.

docs/rfc/rfc-7.adoc
pdyraga: Please see cf6e0b3d52f8f9256d1b687bfc9d884c66a85b0e.

docs/rfc/rfc-7.adoc
beaushinkle: Roger - I remember us talking about this a few months ago. I don't think it's worth explaining this inline, but I do think it would be useful as a footnote or something. What do you think?

beaushinkle:
Left a non-blocking suggestion for a footnote: I'm good after @michalinacienciala / @lukasz-zimnoch is good!

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,297 @@
...
+  becomes eligible for a refund where we will not include it in a sweeping
+  transaction.
+- `sweep_period`: The amount of time we wait between scheduled sweeps on a wallet.
+  Set to 8 hours initially.
+- `sweep_max_deposits`: The number of non-dust unswept revealed bitcoin deposits
+  that will trigger an early sweep on a wallet.
+  Set to 15 initially.
+- `sweeping_fee_bump_period`: The amount of time we wait to see if a sweeping
+  transaction is mined before increasing the fee.
+  Set to 1 hours initially.
```
michalinacienciala, line 292:
```suggestion
  Set to 1 hour initially.
```

docs/rfc/rfc-7.adoc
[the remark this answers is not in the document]
michalinacienciala: Hmm... The sentence says "more than t+1", which is 52 and more (assuming `t` is `50`, which is also not clear from the document).
Also, would the signing be fine if 50 members got restarted? Don't we need 51 not-restarted members?

docs/rfc/rfc-7.adoc
pdyraga: Sure! Please see bf716a4ff0b4dab3f19974df41498d69ef74e6fd.

docs/rfc/rfc-7.adoc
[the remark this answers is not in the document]
pdyraga: Right. It should say "if t+1 or more". If all 100 members are active and willing to cooperate, and 49 of them restarted their clients, the remaining 51 can still sign. If 50 clients lost their state, the next sweep needs to happen.

Please see 964cf80fbeb48b1a9e63a820de207585abe358a4.

docs/rfc/rfc-7.adoc
pdyraga: Fixed in 964cf80fbeb48b1a9e63a820de207585abe358a4.

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,304 @@
...
+is sufficiently confirmed, and there is no other signing by the wallet in
+progress, the signing group members proceed with tECDSA signing protocol.
+
+Each UTXO being an input to the sweep transaction is unlocked sequentially in
+a separate signing session. Each signing session begins with an announcement
+phase allowing to exclude offline operators. There is a 30-blocks (~6 minutes)
+timeout for each signing attempt.
+
+If the given signing attempt fails for any reason (error or timeout), the next
+attempt starts exactly 30 blocks after the previous one started.
```
lukasz-zimnoch, line 183: I think this is not "exactly" as we have the announcement phase and a small cooldown period.

docs/rfc/rfc-7.adoc
```diff
@@ -0,0 +1,304 @@
...
+is +-2 blocks from the current one.
+
+If all these requirements are met, the last transaction executed by the wallet
+is sufficiently confirmed, and there is no other signing by the wallet in
+progress, the signing group members proceed with tECDSA signing protocol.
+
+Each UTXO being an input to the sweep transaction is unlocked sequentially in
+a separate signing session. Each signing session begins with an announcement
+phase allowing to exclude offline operators. There is a 30-blocks (~6 minutes)
+timeout for each signing attempt.
```
lukasz-zimnoch, line 180: Worth emphasizing that the 30 blocks are the timeout for the actual protocol cryptographic computations.

beaushinkle:
Is this good to merge?

mhluongo:
Ha I almost asked the same last night 😅

beaushinkle:
I'm going to merge it! If we want to go back and touch it up, I think that can be a separate PR, and I think having rfc-7 in main is more valuable than having this as a PR

beaushinkle (approved):

docs/rfc/rfc-7.adoc
pdyraga: Done in https://github.com/keep-network/tbtc-v2/pull/571.

PR #415, opened 2022-10-25
Updated relay deployment script

tomaszslabon:
This PR updates the bitcoin relay deployment scripts, so that they use the `LightRelay` contract.

PR #417, opened 2022-10-27
Further relay deployment updates

nkuba:
This PR is a follow-up for https://github.com/keep-network/tbtc-v2/pull/415.

We are going to use `LightRelay` contract as `IRelay` implementation.
The contract requires a relay maintainer to be working on the given
environment to provide the Bitcoin chain data.

We're not ready yet with the maintainer, so we can't use the
`LightRelay` on goerli. Hence, we keep on using `TestRelay` which is a
dummy implementation of the `IRelay` interface.

Once the maintainer is ready we will switch goerli to use `LightRelay`.

Another thing is that we want just one instance of the `LightRelay`
contract to be deployed on the given environment. We don't want to
redeploy the `LightRelay` on goerli to update the maintainer and lose
all previously submitted bitcoin chain data. To cover that we will expect
the `LightRelay.json` artifact will be committed to the repository after it's
deployed. This way we will reuse the existing instance of the contract.


tomaszslabon (approved):

tomaszslabon:
Looks good to me.
Leaving it to @michalinacienciala and @pdyraga.

michalinacienciala (approved):

PR #418, opened 2022-10-31
Call authorization of MaintainerProxy in ReimursementPool by governance

michalinacienciala:
This PR introduces following changes:

1. Change of the caller in the scripts authorizing the `MaintainerProxy` and `Bridge` in the `ReimbursemetPool` (from `deployer` to `governance`) and removal of script changing the ownership of the `ReimbursementPool` contract. The motivation for this change is the fact that we already change(d) the ownership of the `ReimbursementPool` to `governance` during execution of `ecdsa`'s deployment scripts. So by the time we execute `tbtc-v2` scripts, the  
The `ReimbursementPool` contract is already owned by the `governance`. The authorization function needs to be called by that owner.

2. Upgrade of `tbtc-v2`'s `@keep-network/random-beacon` and `@keep-network/ecdsa` dependencies to the latest development versions (what we had before in the `yarn.lock` got outdated once we published `2.1.x` versions of those dependencies). Without this change, we were getting failures in our unit tests and in deployment on `hardhat`.


michalinacienciala:
I've run a workflow checking the deployment after the `deployer` -> `governance` change and it looks the problem still occurs: https://github.com/keep-network/tbtc-v2/actions/runs/3337632239/jobs/5524223787.

We still get following error:
```
Error: ERROR processing /home/runner/work/tbtc-v2/tbtc-v2/solidity/deploy/19_authorize_bridge_in_reimbursement_pool.ts:
Error: VM Exception while processing transaction: reverted with reason string 'Ownable: caller is not the owner'
    at <UnrecognizedContract>.<unknown> (0xafe1b5bdebd4ae65af2024738bf0735fbb65d44b)
```

The address `0xafe1b5bdebd4ae65af2024738bf0735fbb65d44b` mentioned in the error message is an address of the `ReimbursementPool.json` artifact from `@keep-network/ecdsa@2.1.0-dev.0` package used during deployment of `tbtc-v2` contracts on `hardhat` network.
I'm not sure why the script is complaining about it 🤔

pdyraga (approved):

PR #419, opened 2022-11-07
Add a possibility to use a public key hash when calculating a deposit address

michalsmiarowski:
Related issue: https://github.com/threshold-network/token-dashboard/issues/305

Main changes:
- adds the possibility to calculate the deposit script address using public key hash (PKH)
- adds the possibility to convert public key hash into a bitcoin address and vice versa
- adds function that checks if a given string is 20-bytes long, which is the proper length of a public key hash

# Change details 

### - adds the possibility to calculate the deposit script address using publick key hash (PKH)

There is new function added called `getDepositScriptParameters` which basically takes a `Deposit` object and converts it to `DepositScriptParameters` object. What it does is mainly converts `walletPublicKey` and `refundPublicKey`, which are compressed 33-bytes long public keys,  to a public key hashes. We also remove that hashing from `assembleDepositScript` so it can now use wallet public key hashes directly.

There was also a change in validation of the deposit script parameters inside `validateDepositScriptParameters`. Instead of checking if a `walletPublicKey` and `refundPublicKey` are a compressed public keys we check if they are public key hashes. For more information how it works please see third change details.

### - adds the possibility to convert public key hash into a bitcoin address and vice versa

Added two new functions (`encodeToBitcoinAddress` and `decodeBitcoinAddress`) that can encode a public key hash into a bitcoin address and also decode a bitcoin address into a public key hash. This will be helpful form the UI perspective as the user probably provide a Bitcoin address that we will have to convert into a public key hash to create a deposit script.

Both funcion use already build functionality from `bcoin` lib.

### - adds function that checks if a given string is 20-bytes long, which is the proper length of a public key hash

Added new function - `isPublicKeyHashLength` - that checks if the given string is 20-bytes long, which is the proper length of a public key hash. We can use it to validate public key hashes. 

<details>
<summary>Deprecated PR description</summary>

**Warning: This description is deprecated. It describes a solution that no longer exist in this PR. Keeping it here just in case.** 

Main change:
- adds the possibility to calculate the deposit script address using publick key hash (PKH)

# Change details 

### - adds the possibility to calculate the deposit script address using public key hash (PKH)

The UI will require a refund Bitcoin Address from and not compressed public key. Therefore, we are adding the ability to assemble deposit address using both - compressed public key OR publick key hash (which would be a PKH address given from a user).

The `DepositScriptParameters` type was refactored and it can now be one of the two types:
- DepositScriptParametersWithRefundPublicKey:
```
{
  depositor: Identifier,
  blindingFactor: string,
  walletPublicKey: string,
  refundPublicKey: string,
  refundLocktime: string,
  refundPKHAddress?: never,
}
```
- DepositScriptParametersWithRefundPKHAddress:
```
{
  depositor: Identifier,
  blindingFactor: string,
  walletPublicKey: string,
  refundPublicKey?: never,
  refundLocktime: string,
  refundPKHAddress: string,
}
```

Notice the usage of [never](https://www.tutorialsteacher.com/typescript/typescript-never) data type which indicates that the object can't have both `refundPublicKey` and `refundPKHAddress`. Only one or another.

The hashing from `assembleDepositScript` function was removed so it can use only PKH address. The validation and the hashing (if needed) was extracted one level up to `buildDepositScript` function:

```
/**
 * Builds a Bitcoin locking script for P2(W)SH deposit transaction based on
 * either refundPublicKey or refundPKHAddress.
 * 
 * @param deposit - Details of the deposit with either refundPublicKey or
 *        refundPKHAddress
 * @returns Script as an un-prefixed hex string.
 */
export async function buildDepositScript(
  deposit: DepositScriptParameters
): Promise<string> {
  validateDepositScriptParameters(deposit)

  if (deposit.refundPKHAddress) return await assembleDepositScript(deposit)

  // If refundPKHAddress is not defined then it means that it has
  // refundPublicKey property. We hash it into PKH Address to assemble deposit
  // script.
  const { refundPublicKey, ...depositScriptRestParams } = deposit
  const depositWithRefundPKHAddress = {
    ...depositScriptRestParams,
    refundPKHAddress: computeHash160(refundPublicKey!)
  }

  return await assembleDepositScript(depositWithRefundPKHAddress)
}
```

Here we do the validation of the parameters and also check if `deposit` parameter has `refundPKHAddress` (if yes then we just call `assembleDepositScript`) or `refundPublicKey`. In the second case we compute the PKH Address from the publick key and then call `assembleDepositScript` function with a new `deposit` object that is using `refundPKHAddress`.

In tests I extracted the validation of the deposit script to `shouldReturnProperDepositScript` function since it was used in multiple places.

</details>


typescript/src/deposit.ts
```diff
@@ -67,15 +67,51 @@ export interface Deposit {
 /**
```
lukasz-zimnoch, line 67: The presented solution should work but has severe flaws regarding code cleanliness:
- It complicates the domain types declared in `deposit.ts` by adding new ones that are here just to capture the caveats of a single field (`refundPublicKey`). That complicates the entire library and makes it less generic just to conform to a single use case, i.e. the ability to pass the refund public key as a hash from the dApp.
- Introduces ambiguous functions, i.e. `assembleDepositScript` and `buildDepositScript`. Their names say nothing about the differences between them and which one should be used in which context.

I think we should change the approach. In the beginning, we should note that the `DepositScriptParameters` is a specialized type that is used only during deposit script construction, and `walletPublicKey` and `refundPublicKey` are not used as plain-text public keys in this context but always as public key hashes. That means we can change `DepositScriptParameters` to be:
```
  Deposit,
  | "depositor"
  | "blindingFactor"
  | "refundLocktime"
> & {
    /** Documentation */
    walletPublicKeyHash: string

    /** Documentation */
    refundPublicKeyHash: string 
}
```

Once we do that, we gain the possibility to pass PKH's directly to the `calculateDepositAddress` function. On the other hand,  we can no longer pass `Deposit` as `DepositScriptParameters` in the places we did it so far. However, this can be easily fixed using `computeHash160` to convert the plain-text public key to a public key hash where applicable. The presented solution can involve more changes in the codebase but is much cleaner and should keep the library API consistent. What do you think @michalsmiarowski ?

lukasz-zimnoch (changes requested):

typescript/src/deposit.ts
michalsmiarowski: I agree! Now I also think that my solution introduces unnecessary complications just for a single, pretty simple, use and even good comments won't compensate that as it still might be initially confusing.

Fixing that right now!

typescript/src/bitcoin.ts
```diff
@@ -359,3 +361,40 @@ export function createKeyRing(
...
+export function decodeBase58(text: string): string {
+  return base58.decode(text).toString("hex")
+}
+
+/**
+ * Encodes public key hash into a P2PKH address
+ * @param publicKeyHash - public key hash that will be encoded
+ * @returns P2PKH address encoded from the given public key hash
+ */
+export function generateP2PKHAddress(publicKeyHash: string) {
```
lukasz-zimnoch, line 397: The logic is wrong as the checksum is used in a wrong way. Only 4 first bytes should be taken. Additionally, the `00` is not universal and is network-specific. It will be different for mainnet and testnet addresses. See https://www.oreilly.com/library/view/mastering-bitcoin/9781491902639/ch04.html for reference. 

However, you don't have to do the encoding manually as we can use:
```
bcoin.Address.fromPubkeyhash(Buffer.from(publicKeyHash, "hex"))`
```
to construct a bcoin's `Address` object and then get the string version using `toString(network)`. The `publicKeyHash` must be a HEX string, non-prefixed with `0x`. Please note that in the parameter documentation.
The `network` argument accepted by `toString` implies that this function should accept the `network` parameter as well, likewise the `calculateDepositAddress` function. 

I'd also rename `generateP2PKHAddress` to `encodeBitcoinAddress`. We don't "generate" anything here.

lukasz-zimnoch: One important note, this function should also accept `witness` parameter and use `fromWitnessPubkeyhash` if `witness` is true and `fromPubkeyhash` otherwise. The address is encoded differently for SegWit hence that parameter.

lukasz-zimnoch: Good point, `encodeToBitcoinAddress` will be a good option!

typescript/src/bitcoin.ts
```diff
@@ -359,3 +359,62 @@ export function createKeyRing(
...
+  try {
+    const addressObject = new bcoin.Address()
+    if (witness) {
+      addressObject.fromWitnessPubkeyhash(Buffer.from(publicKeyHash, "hex"))
+    } else {
+      addressObject.fromPubkeyhash(Buffer.from(publicKeyHash, "hex"))
+    }
+    return witness
+      ? addressObject.isWitnessPubkeyhash()
+      : addressObject.isPubkeyhash()
```
michalsmiarowski, line 416: When creating this `isPubkeyhash` function I was not sure if we need a `witness` parameter here 🤔 I've decided to put it anyway and just ask what you think.

As far as I understand from the publick key hash we can create either P2PKH address or P2WPKH address, but the public key hash itself does not separate into witness PKH and non-witness PKH 🤔 Looks to me that it's more related to the lib itself and it needs separate functions based on how the `addressObject` was created (`formPubkeyHash` and `fromWitnessPubkeyHash`) which will determinate how we will decode works (different for `P2PKH` address created from `fromPubKeyHash()` and differnt for `P2WPKH` address created from `fromWitnessPubKeyHash()`).

This probably means that we don't need `witness` parameter here at all.

typescript/test/bitcoin.test.ts
```diff
@@ -45,4 +50,145 @@ describe("Bitcoin", () => {
...
+            expect(() => decodeBitcoinAddress(bitcoinAddress)).to.throw()
+          })
+        })
+      })
+    })
+  })
+
+  describe("isPubkeyhash", () => {
+    const publicKeyHash = "3a38d44d6a0c8d0bb84e0232cc632b7e48c72e0e"
+    const witnessPublicKeyHash = "3a38d44d6a0c8d0bb84e0232cc632b7e48c72e0e"
```
michalsmiarowski, line 164: Additional to above comments here is a good example of what i mean. We use exactly the same public key hash to check if `isPubkeyhash` works properly with `witness` option set to `true` and to `false` and in both scenarios it returns a `true` value.

michalsmiarowski (commented):
@lukasz-zimnoch PR is ready for a re-review. I left some comments for the part of code that I'm not sure of

typescript/src/bitcoin.ts
```diff
@@ -359,3 +359,62 @@ export function createKeyRing(
...
+  witness: boolean,
+  network: string
+): string {
+  return witness
+    ? bcoin.Address.fromWitnessPubkeyhash(
+        Buffer.from(publicKeyHash, "hex")
+      ).toString(network)
+    : bcoin.Address.fromPubkeyhash(Buffer.from(publicKeyHash, "hex")).toString(
+        network
+      )
```
lukasz-zimnoch, line 384: Let's make it a little bit cleaner:
```suggestion
  const buffer = Buffer.from(publicKeyHash, "hex")
  const address = witness
    ? bcoin.Address.fromWitnessPubkeyhash(buffer)
    : bcoin.Address.fromPubkeyhash(buffer)
  return address.toString(network)
```

typescript/src/bitcoin.ts
lukasz-zimnoch: Actually, that function doesn't make sense 😄 A public key hash is just a 20-byte hash, nothing more. Moreover, you never know whether an actual ECDSA public key was used to produce that hash. You can produce a 20-byte hash from literally anything. Here you construct a `bcoin` address using the `publicKeyHash` argument. The validation done by `fromWitnessPubkeyhash` and `fromPubkeyhash` just checks if the size is actually 20-byte so, as I said above, any 20-byte hash will pass it and the `bcoin` address will be constructed. The constructed `bcoin` address holds the information about how it was constructed so, if you used `fromWitnessPubkeyhash`/`fromPubkeyhash` then `isWitnessPubkeyhash`/`isPubkeyhash` will always return true.

I understand why you added that function but the only thing we can validate within `validateDepositScriptParameters` is whether the passed `walletPublicKeyHash`/`refundPublicKeyHash` is a 20-byte value. That said, we can remove `isPubkeyhash` and introduce `isPublicKeyHashLength` instead.

typescript/test/bitcoin.test.ts
```diff
@@ -45,4 +50,145 @@ describe("Bitcoin", () => {
       })
     })
   })
+
+  describe("P2PKH <-> public key hash conversion", () => {
+    const publicKeyHash = "3a38d44d6a0c8d0bb84e0232cc632b7e48c72e0e"
+    const P2WPKHAddress = "bc1q8gudgnt2pjxshwzwqgevccet0eyvwtswt03nuy"
+    const P2PKHAddress = "16JrGhLx5bcBSA34kew9V6Mufa4aXhFe9X"
+
+    const publicKeyHashTestnet = "3a38d44d6a0c8d0bb84e0232cc632b7e48c72e0e"
```
lukasz-zimnoch, line 59: No need to use `publicKeyHashTestnet` as you can reuse `publicKeyHash`. The public key hash does not depend on the specific network and will remain the same. Only the address is network-specific.

typescript/test/deposit.test.ts
```diff
@@ -105,6 +107,112 @@ describe("Deposit", () => {
     testnetAddress: "2MwGJ12ZNLJX3qWqPmqLGzh3EfdN5XAEGQ8",
   }
 
+  /**
+   * Checks if script given in argument is correct
+   * @param script - script as an un-prefixed hex string.
+   * @returns void
+   */
+  function shouldReturnProperDepositScript(script: string): void {
```
lukasz-zimnoch, line 115: Maybe:
```suggestion
  function assertValidDepositScript(script: string): void {
```

lukasz-zimnoch (commented):
Last part of comments. Once they are addressed, I'm good to merge!

typescript/test/deposit.test.ts
michalsmiarowski: Fixed in ebf24cdfc368b558167a7beaa08003375ad1121f

typescript/test/bitcoin.test.ts
michalsmiarowski: Fixed in c271ec5f1a205a117f7932c1d511bc466658c7fb

typescript/src/bitcoin.ts
michalsmiarowski: Fixed in 064f5d9ed9b352a1518b20acf237507c33b7e0b2

typescript/src/bitcoin.ts
michalsmiarowski: Fixed in f52b614cf9ce40c95f10574c4a1d501765880180

lukasz-zimnoch (approved):

PR #424, opened 2022-12-07
Optimistic minting

pdyraga:
# Optimistic minting

There are two permissioned sets in the system: Minters and Guardians, both set up in 1-of-n mode. Minters observe the revealed deposits and request minting TBTC. Any single Minter can perform this action. There is a 3 hours delay between the time of the request from a Minter to the time TBTC is minted. During the time of the delay, any Guardian can cancel the minting.

# The mechanism

For any revealed, unswept deposit, any Minter can request minting using `requestOptimisticMint(bytes32 fundingTxHash, uint32 fundingOutputIndex) external onlyMinter` on `TBTCVault`.

Any Guardian can cancel the minting by calling `function cancelOptimisticMint(bytes32 fundingTxHash, uint32 fundingOutputIndex) external onlyGuard`.

If none of the Guardians canceled minting, any Minter can finalize the request by calling `function finalizeOptimisticMint(bytes32 fundingTxHash, uint32 fundingOutputIndex) external onlyMinter` to mint TBTC.

When TBTC is minted, depositor's optimistic minting debt balance is updated in `optimisticMintingDebt`. When a deposit is swept for the depositor, the vault updates `optimisticMintingDebt` and if this amount is non-zero, pay back the debt and mint only the amount of TBTC that is left after paying the debt.

# Future potential

The optimistic minting mechanism has the potential to be integrated as a quick-minting solution. The current implementation - as is - should fulfill all requirements of a quick-minting solution, including a fee for faster processing of deposits. If the depositor does not want to wait for the Bridge to pick up their deposit during a scheduled sweep, they can pay an additional fee and have the Minters / Guardians make the decision. The wallet will sweep that deposit later. I think the only problem we will need to solve is coordination between Minters, Wallets, and Guardians. A wallet should not pick deposits to sweep for which the optimistic minting was requested. If there was a synchronization issue and that already happened, the existing smart contract will handle it gracefully. If the wallet was first, `finalizeOptimisticMint` will revert for a swept deposit. If the Minter was first to finalize, `submitDepositSweepProof -> receiveBalanceIncrease` will process the swept deposit as an optimistic minting debt payment. 

# Fees

There are two fees in the system:
- Deposit fee - cut by the Bridge when sweeping a deposit.
- Optimistic Minting fee - cut by `TBTCVault` when finalizing optimistic mint, computed after the Deposit fee is cut.

The Optimistic Minting fee has an important role. The Bitcoin miner sweep transaction fee is covered by all depositors. The fee is split between all deposits in the sweep transaction, evenly. When tokens are optimistically minted, it is not possible to say what the Bitcoin miner sweeping fee will be. The fee is ignored and the intention is that the imbalance between the wallet balance and the amount of TBTC minted will be covered by a DAO donation (see `DonationVault`) from the fees cut by the system, especially the Optimistic Minting fee. This is exactly the same situation as with the move funds transaction between wallets.

One of the fees can be 0, both fees can be 0, and both fees can be non-zero. The amount of TBTC minted to the depositor's address has both the Deposit fee and the Optimistic Minting fee cut.

The fees are computed as follows:
```
depositFee = depositedAmount / depositFeeDivisor
optimisticMintingFee = (depositedAmount - depositFee) / optimiticMintingFeeDivisor
```

Say that the deposited amount is 100. If the deposit fee is 2%, and the optimistic minting fee is 1%, the amounts will be:
```
depositFee = 100 * 0.02 = 2
optimisticMintingFee = (100 - 2)* 0.01 = 98 * 0.01 = 0.98
```

This scheme allows the Optimistic Minting mechanism to cut a fee based on the amount the vault will actually receive. This wouldn't be guaranteed if we were calculating both fees based on the original deposited amount.

CoinList charges a flat 0.25% trading fee to wrap or unwrap WBTC (https://coinlist.co/help/what-are-the-fees-to-wrap-and-unwrap-wbtc). If we want to maintain the same fee, for the January release, we have two options:
- If the treasury should receive the fee at the moment of sweeping, set the Deposit fee to 0.25% and the Optimistic Minting fee to 0%.
- If the treasury should receive the fee at the moment of finalizing the Optimistic Mint, set the Optimistic Mint fee to 0.25% and the Deposit Fee to 0%.

# Governable parameters

The owner of the `TBTCVault` can add and remove Minters and Guardians. Additionally, there is an option that Minter can be removed by any Guardian. This is to allow dealing with an errant Minter. 

The owner of the contract can additionally update the delay between Optimistic Minting request and finalization. It can also update the Optimistic Minting fee.

# Note on the separation of concerns

So far we kept vaults completely separate from the `Bridge`. The `Bridge` was managing `Bank` balances. Vaults - including `TBTCVault` - were just applications working on top of the `Bank`. Neither `Bank` nor the vaults were interested in how the `Bridge` worked. They knew nothing about revealed deposits and sweeping.

Given that Minter should be only able to request minting for a revealed, unswept deposit, this separation may no longer be kept. At the same time, the goal is to minimize the impact and keep this "abstraction breaking change" local to the optimistic minting component the `TBTCVault` extends. This will let us easily unplug this mechanism in the future if needed.

Also, the approach where `TBTCVault` asks the `Bridge` for revealed deposits allows us to avoid passing deposit ID from the `Bridge`, through `Bank`, to `TBTCVault` which would break the abstractions and require a lot of changes in a quite complicated, gas-optimized `Bridge` contract. Last but not least, we can track - looking only at Ethereum - the amount of BTC that is in the queue for sweeping and has already been optimistically minted into TBTC ERC20.

### DIFF
--- a/solidity/contracts/vault/TBTCOptimisticMinting.sol
+++ b/solidity/contracts/vault/TBTCOptimisticMinting.sol
@@ -20,7 +20,7 @@ import "../bridge/Deposit.sol";
 
 abstract contract TBTCOptimisticMinting is Ownable {
     // TODO: make it governable?
-    uint256 public constant optimisticMintingDelay = 3 hours;
+    uint256 public constant OPTIMISTIC_MINTING_DELAY = 3 hours;
 
     Bridge public bridge;
 
@@ -112,10 +112,13 @@ abstract contract TBTCOptimisticMinting is Ownable {
         );
 
         uint256 requestedAt = pendingOptimisticMints[depositKey];
-        require(requestedAt != 0, "Optimistic minting not requested");
+        require(
+            requestedAt != 0,
+            "Optimistic minting not requested or already finalized"
+        );
         require(
             /* solhint-disable-next-line not-rely-on-time */
-            block.timestamp - requestedAt > optimisticMintingDelay,
+            block.timestamp - requestedAt > OPTIMISTIC_MINTING_DELAY,
             "Optimistic minting delay has not passed yet"
         );
 
@@ -146,6 +149,11 @@ abstract contract TBTCOptimisticMinting is Ownable {
         uint256 depositKey = calculateDepositKey(
             fundingTxHash,
             fundingOutputIndex
+        );
+
+        require(
+            pendingOptimisticMints[depositKey] > 0,
+            "Optimistic minting not requested of already finalized"
         );
 
         delete pendingOptimisticMints[depositKey];
--- a/solidity/contracts/vault/TBTCOptimisticMinting.sol
+++ b/solidity/contracts/vault/TBTCOptimisticMinting.sol
@@ -1,0 +1,112 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+// ██████████████     ▐████▌     ██████████████
+// ██████████████     ▐████▌     ██████████████
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+// ██████████████     ▐████▌     ██████████████
+// ██████████████     ▐████▌     ██████████████
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+
+pragma solidity 0.8.17;
+
+import "../bridge/Bridge.sol";
+import "../bridge/Deposit.sol";
+
+abstract contract TBTCOptimisticMinting {
+
+    // TODO: make it governable?
+    uint256 public constant optimisticMintingDelay = 3 hours;
+
+    Bridge public bridge;
+
+    mapping(address => bool) public isMinter;
+    mapping(address => bool) public isWatchman;
+
+    mapping(uint256 => uint256) public pendingOptimisticMints;
+
+    mapping(address => uint256) public optimisticMintingDebt;
+
+    modifier onlyMinter() {
+        require(isMinter[msg.sender], "Caller is not a minter");
+        _;
+    }
+
+    modifier onlyWatchman() {
+        require(isWatchman[msg.sender], "Caller is not a watchman");
+        _;
+    }
+
+    constructor(Bridge _bridge) {
+        require(
+            address(_bridge) != address(0),
+            "Bridge can not be the zero address"
+        );
+
+        bridge = _bridge;
+    }
+
+    function _mint(address minter, uint256 amount) virtual internal;
+
+    function optimisticMint(uint256 depositKey) external onlyMinter {
+        Deposit.DepositRequest memory deposit = bridge.deposits(depositKey);
+
+        require(deposit.revealedAt != 0, "The deposit has not been revealed");
+        require(deposit.sweptAt == 0, "The deposit is already swept");
+        require(deposit.vault == address(this), "Unexpected vault address");
+
+        pendingOptimisticMints[depositKey] = block.timestamp;
+
+        // TODO: emit an event
+    }
+
+    function finalizeOptimisticMint(uint256 depositKey) external onlyMinter {
+        uint256 requestedAt = pendingOptimisticMints[depositKey];
+        require(requestedAt != 0, "Optimistic minting not requested");
+        require(
+            block.timestamp - requestedAt > optimisticMintingDelay,
+            "Optimistic minting delay has not passed yet"
+        );
+
+        Deposit.DepositRequest memory deposit = bridge.deposits(depositKey);
+        require(deposit.sweptAt == 0, "The deposit is already swept");
+
+        // TODO: deal with the minting fee
+        _mint(deposit.depositor, deposit.amount);
+        optimisticMintingDebt[deposit.depositor] += deposit.amount;
+
+        delete pendingOptimisticMints[depositKey];
+
+        // TODO: emit an event
+    }
+
+    // TODO: Is this function convenient enough to block minting at 3AM ?
+    //       Do we want to give watchment a chance to temporarily disable
+    //       finalizeOptimisticMint ?
+    function cancelOptimisticMint(uint256 depositKey) external onlyWatchman {
+        delete pendingOptimisticMints[depositKey];
+
+        // TODO: emit an event
+    }
+
+    function repayOptimisticMintDebt(address depositor, uint256 amount) internal returns (uint256) {
+        uint256 debt = optimisticMintingDebt[depositor];
+
+        if (amount > debt) {
+            optimisticMintingDebt[depositor] = 0;
+            return amount - debt;
+        } else {
+            optimisticMintingDebt[depositor] -= amount;
+            return 0;
+        }
+
+        // TODO: emit an event
+    }
+
+    // TODO: functions to manipulate the optimistic minter/watchmen set.
+}
--- a/solidity/contracts/vault/TBTCOptimisticMinting.sol
+++ b/solidity/contracts/vault/TBTCOptimisticMinting.sol
@@ -37,6 +37,18 @@ abstract contract TBTCOptimisticMinting is Ownable {
         uint32 fundingOutputIndex,
         uint256 depositKey
     );
+    event OptimisticMintingFinalized(
+        address indexed minter,
+        bytes32 fundingTxHash,
+        uint32 fundingOutputIndex,
+        uint256 depositKey
+    );
+    event OptimisticMintingCancelled(
+        address indexed guard,
+        bytes32 fundingTxHash,
+        uint32 fundingOutputIndex,
+        uint256 depositKey
+    );
     event MinterAdded(address indexed minter);
     event MinterRemoved(address indexed minter);
     event GuardAdded(address indexed guard);
@@ -116,7 +128,12 @@ abstract contract TBTCOptimisticMinting is Ownable {
 
         delete pendingOptimisticMints[depositKey];
 
-        // TODO: emit an event
+        emit OptimisticMintingFinalized(
+            msg.sender,
+            fundingTxHash,
+            fundingOutputIndex,
+            depositKey
+        );
     }
 
     // TODO: Is this function convenient enough to block minting at 3AM ?
@@ -130,9 +147,15 @@ abstract contract TBTCOptimisticMinting is Ownable {
             fundingTxHash,
             fundingOutputIndex
         );
+
         delete pendingOptimisticMints[depositKey];
 
-        // TODO: emit an event
+        emit OptimisticMintingCancelled(
+            msg.sender,
+            fundingTxHash,
+            fundingOutputIndex,
+            depositKey
+        );
     }
 
     function addMinter(address minter) external onlyOwner {
--- a/solidity/contracts/vault/TBTCOptimisticMinting.sol
+++ b/solidity/contracts/vault/TBTCOptimisticMinting.sol
@@ -98,6 +98,14 @@ abstract contract TBTCOptimisticMinting is Ownable {
         _;
     }
 
+    modifier onlyOwnerOrGuardian() {
+        require(
+            owner() == msg.sender || isGuardian[msg.sender],
+            "Caller is not the owner or guardian"
+        );
+        _;
+    }
+
     constructor(Bridge _bridge) {
         require(
             address(_bridge) != address(0),
@@ -265,9 +273,6 @@ abstract contract TBTCOptimisticMinting is Ownable {
         );
     }
 
-    // TODO: Find a convenient way for a Guardian to block minting at 3AM and
-    //       deal with an errant Minter.
-
     /// @notice Adds the address to the Minter set.
     function addMinter(address minter) external onlyOwner {
         require(!isMinter[minter], "This address is already a minter");
@@ -276,7 +281,7 @@ abstract contract TBTCOptimisticMinting is Ownable {
     }
 
     /// @notice Removes the address from the Minter set.
-    function removeMinter(address minter) external onlyOwner {
+    function removeMinter(address minter) external onlyOwnerOrGuardian {
         require(isMinter[minter], "This address is not a minter");
         delete isMinter[minter];
         emit MinterRemoved(minter);
--- a/solidity/contracts/vault/TBTCOptimisticMinting.sol
+++ b/solidity/contracts/vault/TBTCOptimisticMinting.sol
@@ -58,12 +58,13 @@ abstract contract TBTCOptimisticMinting is Ownable {
     ///         optimistically minted deposit and transferred to the treasury
     ///         upon finalization of the optimistic mint. This fee is computed
     ///         as follows: `fee = amount / optimisticMintingFeeDivisor`.
-    ///         For example, if the fee needs to be 2% of each deposit,
-    ///         the `optimisticMintingFeeDivisor` should be set to `50` because
+    ///         For example, if the fee needs to be 2%, the
+    ///         `optimisticMintingFeeDivisor` should be set to `50` because
     ///         `1/50 = 0.02 = 2%`.
-    ///         Note that the optimistic minting fee does not replace the
-    ///         deposit treasury fee cut by the Bridge. Both fees are deducted
-    ///         from the minted token amount.
+    ///         The optimistic minting fee does not replace the deposit treasury
+    ///         fee cut by the Bridge. The optimistic fee is a percentage AFTER
+    ///         the treasury fee is cut:
+    ///         `optimisticMintingFee = (depositAmount - treasuryFee) / optimisticMintingFeeDivisor`
     uint32 public optimisticMintingFeeDivisor;
 
     /// @notice The time that needs to pass between the moment the optimistic
@@ -99,7 +100,7 @@ abstract contract TBTCOptimisticMinting is Ownable {
     ///         not.
     mapping(address => uint256) public optimisticMintingDebt;
 
-    /// @notice New optimistic minting fee deivisor value. Set only when the
+    /// @notice New optimistic minting fee divisor value. Set only when the
     ///         parameter update process is pending. Once the update gets
     //          finalized, this will be the value of the divisor.
     uint32 public newOptimisticMintingFeeDivisor;
--- a/solidity/contracts/vault/TBTCOptimisticMinting.sol
+++ b/solidity/contracts/vault/TBTCOptimisticMinting.sol
@@ -370,8 +370,6 @@ abstract contract TBTCOptimisticMinting is Ownable {
             return 0;
         }
 
-        // TODO: emit an event
-
-        // TODO: cover with unit tests
+        // TODO: emit an event ?
     }
 }
--- a/solidity/contracts/vault/TBTCOptimisticMinting.sol
+++ b/solidity/contracts/vault/TBTCOptimisticMinting.sol
@@ -75,6 +75,11 @@ abstract contract TBTCOptimisticMinting is Ownable {
     ///         request optimistic minting.
     mapping(address => bool) public isMinter;
 
+    /// @notice List of all Minters.
+    /// @dev May be used to establish an order in which the Minters should
+    ///      request for an optimistic minting.
+    address[] public minters;
+
     /// @notice Indicates if the given address is a Guardian. Only Guardians can
     ///         cancel requested optimistic minting.
     mapping(address => bool) public isGuardian;
@@ -190,6 +195,11 @@ abstract contract TBTCOptimisticMinting is Ownable {
     /// @dev Mints the given amount of TBTC to the given depositor's address.
     ///      Implemented by TBTCVault.
     function _mint(address minter, uint256 amount) internal virtual;
+
+    /// @notice Allows to fetch a list of all Minters.
+    function getMinters() external view returns (address[] memory) {
+        return minters;
+    }
 
     /// @notice Allows a Minter to request for an optimistic minting of TBTC.
     ///         The following conditions must be met:
@@ -388,17 +398,29 @@ abstract contract TBTCOptimisticMinting is Ownable {
         emit OptimisticMintingCancelled(msg.sender, depositKey);
     }
 
-    /// @notice Adds the address to the Minter set.
+    /// @notice Adds the address to the Minter list.
     function addMinter(address minter) external onlyOwner {
         require(!isMinter[minter], "This address is already a minter");
         isMinter[minter] = true;
+        minters.push(minter);
         emit MinterAdded(minter);
     }
 
-    /// @notice Removes the address from the Minter set.
+    /// @notice Removes the address from the Minter list.
     function removeMinter(address minter) external onlyOwnerOrGuardian {
         require(isMinter[minter], "This address is not a minter");
         delete isMinter[minter];
+
+        // We do not expect too many Minters so a simple loop is safe.
+        for (uint256 i = 0; i < minters.length; i++) {
+            if (minters[i] == minter) {
+                minters[i] = minters[minters.length - 1];
+                // slither-disable-next-line costly-loop
+                minters.pop();
+                break;
+            }
+        }
+
         emit MinterRemoved(minter);
     }
 
--- a/solidity/contracts/vault/TBTCOptimisticMinting.sol
+++ b/solidity/contracts/vault/TBTCOptimisticMinting.sol
@@ -20,14 +20,14 @@ import "../bridge/Deposit.sol";
 
 /// @title TBTC Optimistic Minting
 /// @notice The Optimistic Minting mechanism allows to mint TBTC before
-///         TBTCVault receives the Bank balance. There are two permissioned sets
-///         in the system: Minters and Guardians, both set up in 1-of-n mode.
-///         Minters observe the revealed deposits and request minting TBTC.
+///         `TBTCVault` receives the Bank balance. There are two permissioned
+///         sets in the system: Minters and Guardians, both set up in 1-of-n
+///         mode. Minters observe the revealed deposits and request minting TBTC.
 ///         Any single Minter can perform this action. There is a 3 hours delay
 ///         between the time of the request from a Minter to the time TBTC is
 ///         minted. During the time of the delay, any Guardian can cancel the
 ///         minting.
-/// @dev This functionality is a part of TBTCVault. It is implemented in
+/// @dev This functionality is a part of `TBTCVault`. It is implemented in
 ///      a separate abstract contract to achieve better separation of concerns
 ///      and easier-to-follow code.
 abstract contract TBTCOptimisticMinting is Ownable {
@@ -64,18 +64,30 @@ abstract contract TBTCOptimisticMinting is Ownable {
 
     /// @notice Collection of all revealed deposits for which the optimistic
     ///         minting was requested. Indexed by a deposit key computed as
-    ///         keccak256(fundingTxHash | fundingOutputIndex).
+    ///         `keccak256(fundingTxHash | fundingOutputIndex)`.
     mapping(uint256 => OptimisticMintingRequest)
         public optimisticMintingRequests;
 
     /// @notice Optimistic minting debt value per depositor's address. The debt
     ///         represents the total value of all depositor's deposits revealed
     ///         to the Bridge that has not been yet swept and led to the
-    ///         optimistic minting of TBTC. When TBTCVault sweeps a deposit,
+    ///         optimistic minting of TBTC. When `TBTCVault` sweeps a deposit,
     ///         the debt is fully or partially paid off, no matter if that
     ///         particular swept deposit was used for the optimistic minting or
     ///         not.
     mapping(address => uint256) public optimisticMintingDebt;
+
+    /// @notice Divisor used to compute the treasury fee taken from each
+    ///         optimistically minted deposit and transferred to the treasury
+    ///         upon finalization of the optimistic mint. This fee is computed
+    ///         as follows: `fee = amount / optimisticMintingFeeDivisor`.
+    ///         For example, if the fee needs to be 2% of each deposit,
+    ///         the `optimisticMintingFeeDivisor` should be set to `50` because
+    ///         `1/50 = 0.02 = 2%`.
+    ///         Note that the optimistic minting fee does not replace the
+    ///         deposit treasury fee cut by the Bridge. Both fees are deducted
+    ///         from the minted token amount.
+    uint256 public optimisticMintingFeeDivisor;
 
     event OptimisticMintingRequested(
         address indexed minter,
@@ -105,6 +117,7 @@ abstract contract TBTCOptimisticMinting is Ownable {
     event GuardianRemoved(address indexed guardian);
     event OptimisticMintingPaused();
     event OptimisticMintingUnpaused();
+    event OptimisticMintingFeeUpdated(uint256 newOptimisticMintingFeeDivisor);
 
     modifier onlyMinter() {
         require(isMinter[msg.sender], "Caller is not a minter");
@@ -212,14 +225,14 @@ abstract contract TBTCOptimisticMinting is Ownable {
     ///         - The optimistic minting has been requested for the given
     ///           deposit.
     ///         - The deposit has not been swept yet.
-    ///         - At least OPTIMISTIC_MINTING_DELAY passed since the optimistic
+    ///         - At least `OPTIMISTIC_MINTING_DELAY` passed since the optimistic
     ///           minting was requested for the given deposit.
     ///         - The optimistic minting has not been finalized earlier for the
     ///           given deposit.
     ///         - The optimistic minting request for the given deposit has not
     ///           been canceled by a Guardian.
     ///         - The optimistic minting is not paused.
-    ///         This function mints TBTC and increases optimisticMintingDebt
+    ///         This function mints TBTC and increases `optimisticMintingDebt`
     ///         for the given depositor. The finalized optimistic minting
     ///         request is removed from the contract.
     function finalizeOptimisticMint(
@@ -264,8 +277,24 @@ abstract contract TBTCOptimisticMinting is Ownable {
         //
         // This imbalance is supposed to be solved by a donation to the Bridge.
         uint256 amountToMint = deposit.amount - deposit.treasuryFee;
-        _mint(deposit.depositor, amountToMint);
+
+        // The Optimistic Minting mechanism may additionally cut a fee from the
+        // amount that is left after deducting the Bridge deposit treasury fee.
+        // Think of this fee as an extra payment for faster processing of
+        // deposits. One does not need to use the Optimistic Minting mechanism
+        // and they may wait for the Bridge to sweep their deposit if they do
+        // not want to pay the Optimistic Minting fee.
+        uint256 optimisticMintFee = optimisticMintingFeeDivisor > 0
+            ? amountToMint / optimisticMintingFeeDivisor
+            : 0;
+        amountToMint -= optimisticMintFee;
+
         optimisticMintingDebt[deposit.depositor] += amountToMint;
+
+        _mint(deposit.depositor, amountToMint);
+        if (optimisticMintFee > 0) {
+            _mint(bridge.treasury(), optimisticMintFee);
+        }
 
         /* solhint-disable-next-line not-rely-on-time */
         request.finalizedAt = uint64(block.timestamp);
@@ -375,9 +404,25 @@ abstract contract TBTCOptimisticMinting is Ownable {
         emit OptimisticMintingUnpaused();
     }
 
+    // TODO: governance delay
+
+    /// @notice Updates the optimistic minting fee. The fee is computed
+    ///         as follows: `fee = amount / optimisticMintingFeeDivisor`.
+    ///         For example, if the fee needs to be 2% of each deposit,
+    ///         the `optimisticMintingFeeDivisor` should be set to `50` because
+    ///         `1/50 = 0.02 = 2%`.
+    /// @dev See the documentation for optimisticMintingFeeDivisor.
+    function updateOptimisticMintingFee(uint256 newOptimisticMintingFeeDivisor)
+        external
+        onlyOwner
+    {
+        optimisticMintingFeeDivisor = newOptimisticMintingFeeDivisor;
+        emit OptimisticMintingFeeUpdated(newOptimisticMintingFeeDivisor);
+    }
+
     /// @notice Calculates deposit key the same way as the Bridge contract.
     ///         The deposit key is computed as
-    ///         keccak256(fundingTxHash | fundingOutputIndex).
+    ///         `keccak256(fundingTxHash | fundingOutputIndex)`.
     function calculateDepositKey(
         bytes32 fundingTxHash,
         uint32 fundingOutputIndex
@@ -388,14 +433,14 @@ abstract contract TBTCOptimisticMinting is Ownable {
             );
     }
 
-    /// @notice Used by TBTCVault.receiveBalanceIncrease to repay the optimistic
+    /// @notice Used by `TBTCVault.receiveBalanceIncrease` to repay the optimistic
     ///         minting debt before TBTC is minted. When optimistic minting is
     ///         finalized, debt equal to the value of the deposit being
-    ///         a subject of the optimistic minting is incurred. When TBTCVault
+    ///         a subject of the optimistic minting is incurred. When `TBTCVault`
     ///         sweeps a deposit, the debt is fully or partially paid off, no
     ///         matter if that particular deposit was used for the optimistic
     ///         minting or not.
-    /// @dev See TBTCVault.receiveBalanceIncrease
+    /// @dev See `TBTCVault.receiveBalanceIncrease`
     /// @param depositor The depositor whose balance increase is received.
     /// @param amount The balance increase amount for the depositor received.
     /// @return The TBTC amount that should be minted after paying off the
--- a/solidity/contracts/vault/TBTCOptimisticMinting.sol
+++ b/solidity/contracts/vault/TBTCOptimisticMinting.sol
@@ -79,25 +79,25 @@ abstract contract TBTCOptimisticMinting is Ownable {
 
     event OptimisticMintingRequested(
         address indexed minter,
-        address depositor,
+        uint256 indexed depositKey,
+        address indexed depositor,
         uint256 amount,
         bytes32 fundingTxHash,
-        uint32 fundingOutputIndex,
-        uint256 depositKey
+        uint32 fundingOutputIndex
     );
     event OptimisticMintingFinalized(
         address indexed minter,
-        address depositor,
+        uint256 indexed depositKey,
+        address indexed depositor,
         uint256 amount,
         bytes32 fundingTxHash,
-        uint32 fundingOutputIndex,
-        uint256 depositKey
+        uint32 fundingOutputIndex
     );
     event OptimisticMintingCancelled(
         address indexed guardian,
+        uint256 indexed depositKey,
         bytes32 fundingTxHash,
-        uint32 fundingOutputIndex,
-        uint256 depositKey
+        uint32 fundingOutputIndex
     );
     event MinterAdded(address indexed minter);
     event MinterRemoved(address indexed minter);
@@ -199,11 +199,11 @@ abstract contract TBTCOptimisticMinting is Ownable {
 
         emit OptimisticMintingRequested(
             msg.sender,
+            depositKey,
             deposit.depositor,
             deposit.amount,
             fundingTxHash,
-            fundingOutputIndex,
-            depositKey
+            fundingOutputIndex
         );
     }
 
@@ -272,11 +272,11 @@ abstract contract TBTCOptimisticMinting is Ownable {
 
         emit OptimisticMintingFinalized(
             msg.sender,
+            depositKey,
             deposit.depositor,
             amountToMint,
             fundingTxHash,
-            fundingOutputIndex,
-            depositKey
+            fundingOutputIndex
         );
     }
 
@@ -322,9 +322,9 @@ abstract contract TBTCOptimisticMinting is Ownable {
 
         emit OptimisticMintingCancelled(
             msg.sender,
+            depositKey,
             fundingTxHash,
-            fundingOutputIndex,
-            depositKey
+            fundingOutputIndex
         );
     }
 
--- a/solidity/contracts/vault/TBTCOptimisticMinting.sol
+++ b/solidity/contracts/vault/TBTCOptimisticMinting.sol
@@ -245,7 +245,7 @@ abstract contract TBTCOptimisticMinting is Ownable {
 
         require(
             /* solhint-disable-next-line not-rely-on-time */
-            block.timestamp - request.requestedAt > OPTIMISTIC_MINTING_DELAY,
+            block.timestamp > request.requestedAt + OPTIMISTIC_MINTING_DELAY,
             "Optimistic minting delay has not passed yet"
         );
 
@@ -317,7 +317,7 @@ abstract contract TBTCOptimisticMinting is Ownable {
         );
 
         // Delete it. It allows to request optimistic minting for the given
-        // deposit again. Useful in case of an errant Guarian.
+        // deposit again. Useful in case of an errant Guardian.
         delete optimisticMintingRequests[depositKey];
 
         emit OptimisticMintingCancelled(
--- a/solidity/contracts/vault/TBTCOptimisticMinting.sol
+++ b/solidity/contracts/vault/TBTCOptimisticMinting.sol
@@ -46,6 +46,12 @@ abstract contract TBTCOptimisticMinting is Ownable {
     ///         cancel requested optimistic minting.
     mapping(address => bool) public isGuardian;
 
+    /// @notice Indicates if the optimistic minting has been paused. Only the
+    ///         Governance can pause optimistic minting. Note that the pause of
+    ///         the optimistic minting does not stop the standard minting flow
+    ///         where wallets sweep deposits.
+    bool public isOptimisticMintingPaused;
+
     /// @notice Collection of all revealed deposits for which the optimistic
     ///         minting was requested. Indexed by a deposit key computed as
     ///         keccak256(fundingTxHash | fundingOutputIndex). The value is
@@ -87,6 +93,8 @@ abstract contract TBTCOptimisticMinting is Ownable {
     event MinterRemoved(address indexed minter);
     event GuardianAdded(address indexed guardian);
     event GuardianRemoved(address indexed guardian);
+    event OptimisticMintingPaused();
+    event OptimisticMintingUnpaused();
 
     modifier onlyMinter() {
         require(isMinter[msg.sender], "Caller is not a minter");
@@ -103,6 +111,11 @@ abstract contract TBTCOptimisticMinting is Ownable {
             owner() == msg.sender || isGuardian[msg.sender],
             "Caller is not the owner or guardian"
         );
+        _;
+    }
+
+    modifier whenOptimisticMintingNotPaused() {
+        require(!isOptimisticMintingPaused, "Optimistic minting paused");
         _;
     }
 
@@ -125,6 +138,7 @@ abstract contract TBTCOptimisticMinting is Ownable {
     ///           and output index has been revealed to the Bridge.
     ///         - The deposit has not been swept yet.
     ///         - The deposit is targeted into the TBTCVault.
+    ///         - The optimistic minting is not paused.
     ///         After calling this function, the Minter has to wait for
     ///         OPTIMISTIC_MINTING_DELAY before finalizing the mint with a call
     ///         to finalizeOptimisticMint.
@@ -148,6 +162,7 @@ abstract contract TBTCOptimisticMinting is Ownable {
     function optimisticMint(bytes32 fundingTxHash, uint32 fundingOutputIndex)
         external
         onlyMinter
+        whenOptimisticMintingNotPaused
     {
         uint256 depositKey = calculateDepositKey(
             fundingTxHash,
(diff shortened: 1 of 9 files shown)


solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,112 @@
...
+        optimisticMintingDebt[deposit.depositor] += deposit.amount;
+
+        delete pendingOptimisticMints[depositKey];
+
+        // TODO: emit an event
+    }
+
+    // TODO: Is this function convenient enough to block minting at 3AM ?
+    //       Do we want to give watchment a chance to temporarily disable
+    //       finalizeOptimisticMint ?
```
pdyraga, line 90: Thinking of a 3 AM situation for some watchman to block minting. We could have a mechanism allowing to pause optimistic minting for 24h by any watchmen. This pause could be disabled by the owner of this contract. 

If the given watchman proves to be DoSing this way, the owner can remove the watchman from the permissioned set. 

What do you think?

pdyraga: Let's say that one of the minters gets compromised and they start requesting mints in a loop. With just `cancelOptimisticMint`, watchmen would have to run the same loop and this is becoming a game of speed and who has more ether. Options are, for example:
- `pauseOptimisticMints` executed by one watchman, pausing all mints for 24h. The given watchman can pause mints only one time. The owner can `resumeOptimisticMints` at any time.
- `pauseOptimisticMints` executed by at least >= 2 watchman together, pausing all mints indefinitely, until the owner `resumeOptimisticMints`.

solidity/contracts/vault/TBTCVault.sol
```diff
@@ -129,7 +130,9 @@ contract TBTCVault is IVault, Ownable {
     ) external override onlyBank {
         require(depositors.length != 0, "No depositors specified");
         for (uint256 i = 0; i < depositors.length; i++) {
-            _mint(depositors[i], depositedAmounts[i]);
+            address depositor = depositors[i];
+            uint256 amount = depositedAmounts[i];
+            _mint(depositor, repayOptimisticMintDebt(depositor, amount));
```
dimpar, line 135: In case `repayOptimisticMintDebt(...)` returns 0, do we still want to call `_mint(...)` function? I'm not 100% sure, but it looks like we can mint 0 tokens which consume some gas. Maybe we can skip the minting when there's nothing to mint?

mhluongo (commented):
Left some quick comments. Planning to keep reviewing as you go

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,188 @@
...
+
+    event OptimisticMintingRequested(
+        address indexed minter,
+        bytes32 fundingTxHash,
+        uint32 fundingOutputIndex,
+        uint256 depositKey
+    );
+    event MinterAdded(address indexed minter);
+    event MinterRemoved(address indexed minter);
+    event GuardAdded(address indexed guard);
```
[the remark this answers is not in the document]
pdyraga: Yeah, definitely! I started with "watchmen" in the initial docs, but then I realized everyone else is referring to this role as "guardians"/"guards" so I used this term in the Solidity code. I will revisit the initial docs once we are done with the contract.

Is "guard" the same as "guardian"? What's your preference?

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,188 @@
...
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+
+pragma solidity 0.8.17;
+
+import "../bridge/Bridge.sol";
+import "../bridge/Deposit.sol";
+
+abstract contract TBTCOptimisticMinting is Ownable {
+    // TODO: make it governable?
```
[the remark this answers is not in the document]
pdyraga: Agreed, I'll remove this TODO. Regarding the upgrade, this is already covered by `TBTCVault` which is inheriting from `TBTCOptimisticMinting`, so I think we are good:

https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/solidity/contracts/vault/TBTCVault.sol#L206-L223

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: `TBTCVault`'s owner can:
- [`initiateUpgrade(address _newVault) external onlyOwner`](https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/solidity/contracts/vault/TBTCVault.sol#L197)
- [`finalizeUpgrade() external onlyOwner onlyAfterUpgradeGovernanceDelay`](https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/solidity/contracts/vault/TBTCVault.sol#L212)
- [`recoverERC20FromToken(IERC20 token, address recipient, uint256 amount) external onlyOwner`](https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/solidity/contracts/vault/TBTCVault.sol#L230)
- [`recoverERC721FromToken(IERC721 token, address recipient, uint256 tokenId, bytes calldata data) external onlyOwner`](https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/solidity/contracts/vault/TBTCVault.sol#L244)
- [`recoverERC20(IERC20 token, address recipient, uint256 amount) external onlyOwner`](https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/solidity/contracts/vault/TBTCVault.sol#L260)
- [`recoverERC721(IERC721 token, address recipient, uint256 tokenId, bytes calldata data) external onlyOwner`](https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/solidity/contracts/vault/TBTCVault.sol#L274)

In other words, it has upgrade capabilities and emergency token recovery functionalities with a caveat that `recoverERC20` may also be used for withdrawing tBTC v1 if it's not withdrawn from v1<>v2 vending machine [before the upgrade](https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/solidity/contracts/bridge/VendingMachine.sol#L242).

This feels like a set of functionalities that should be assigned to the DAO/large multisig.

On the other hand, functions to remove an errant minter or unpause optimistic minting are not necessarily something we want to delegate to the DAO or large multisigs given the voting time.

We could introduce a `leadGuard` (?) role, assign it to a developer multisig, and organize ACLs in the following way:
- `addMinter onlyOwner` (DAO/council multisig)
- `removeMinter onlyLeadGuardOrOwner` (DAO/council multisig of developer multisig)
- `addGuard onlyLeadGuardOrOwner`
- `removeGuard onlyLeadGuardOrOwner`
- `unpauseMinting onlyLeadGuardOrOwner`
- `pauseMinting onlyGuard` - pauses until `unpauseMinting` is called

In the scenario when a minter gets compromised, any single guard can `pauseMinting`. Only the lead guard or owner (DAO) can unpause, previously removing the minter or guard.

Alternatively, we could not implement `pauseMinting` and just give any guard an option to `removeMinter`. The risk of DoS is the same as with the pause.

What do you think?

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,219 @@
...
+            fundingTxHash,
+            fundingOutputIndex
+        );
+        Deposit.DepositRequest memory deposit = bridge.deposits(depositKey);
+
+        // TODO: validate when it was revealed?
+
+        require(deposit.revealedAt != 0, "The deposit has not been revealed");
+        require(deposit.sweptAt == 0, "The deposit is already swept");
+        require(deposit.vault == address(this), "Unexpected vault address");
```
jMyles, line 92: Perhaps a more intuitive error message here?  "Vault address cannot be bridge address.  Please copy and paste more carefully.  :-)"

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,316 @@
...
+        Deposit.DepositRequest memory deposit = bridge.deposits(depositKey);
+        require(deposit.sweptAt == 0, "The deposit is already swept");
+
+        // TODO: Deal with the minting fee. Right now we mint amount of TBTC
+        //       equal to the output value of the Bitcoin deposit transaction.
+        //       Bridge, when sweeping, will cut a sweeping fee and Bitcoin TX
+        //       fee. If we do not take it into account here, we will create
+        //       an imbalance in the system. Worth noting that even if we do not
+        //       decide to cut fees here, it is possible to fix this imbalance
+        //       later by a donation to the Bridge (see DonationVault).
```
pdyraga, line 201: There are two cuts on the deposited value the Bridge is doing during the sweeping:
- The deposit treasury fee.
- The Bitcoin sweep TX fee, split equally between all deposits in a sweep.

Handling the first one feels straightforward. We have the fee from the moment of the reveal [stored](https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/solidity/contracts/bridge/Deposit.sol#L94-L95) so we can subtract it from the output value when minting TBTC.

Handling the second one is more complicated. We do not know what the Bitcoin minter fee will be at the moment of sweeping so we can either ignore it or... cut the maximum which [is currently set to 0.001 BTC](https://github.com/keep-network/tbtc-v2/blob/8b9d2629bf4333e650a54f32a4da7cf86bf6785e/solidity/contracts/bridge/Bridge.sol#L300) and has to stay higher than the average to leave some flexibility for the wallet.

For optimistic minting, I'd personally prefer to cut the treasure minting fee but ignore the Bitcoin miner fee. This will create an imbalance in the system but it is the same imbalance as with the move funds transactions between wallets and we have a way to deal with it: an external donation to the Bridge targeted to the [`DonationVault`](https://github.com/keep-network/tbtc-v2/blob/main/solidity/contracts/vault/DonationVault.sol).

solidity/contracts/vault/TBTCOptimisticMinting.sol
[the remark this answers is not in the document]
mhluongo: "Guard" is close, but let's just call it "guardian" to avoid any misunderstanding.

solidity/contracts/vault/TBTCOptimisticMinting.sol
dimpar: If we were to introduce a `LeadGuard`, then why don't we make `pauseMinting` also callable by a `LeadGuard`, ie. `pauseMinting onlyLeadGuardOrGuard`?

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,316 @@
...
+    function removeGuard(address guard) external onlyOwner {
+        require(isGuard[guard], "This address is not a guard");
+        delete isGuard[guard];
+        emit GuardRemoved(guard);
+    }
+
+    /// @notice Calculates deposit key the same way as the Bridge contract.
+    ///         The deposit key is computed as
+    ///         keccak256(fundingTxHash | fundingOutputIndex).
+    function calculateDepositKey(
```
dimpar, line 281: Since we have this public function that can be used by anyone, do you think we can pass only `depositKey` to `optimisticMint`, `finalizeOptimisticMint`, and `cancelOptimisticMint()`?

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,316 @@
...
+    ///         sweeps a deposit, the debt is fully or partially paid off, no
+    ///         matter if that particular deposit was used for the optimistic
+    ///         minting or not.
+    function repayOptimisticMintDebt(address depositor, uint256 amount)
+        internal
+        returns (uint256)
+    {
+        uint256 debt = optimisticMintingDebt[depositor];
+
+        if (amount > debt) {
```
dimpar, line 304:
```
        // TODO: emit an event

        if (amount > debt) {
            optimisticMintingDebt[depositor] = 0;
            return amount - debt;
        }

        optimisticMintingDebt[depositor] -= amount;
        return 0;
```
?

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: We are passing funding TX hash and funding output index on purpose - this lets to capture them on-chain. Guardian needs only Etherscan and Blockstream.info to see if the given deposit happened and how many confirmations it has.

solidity/contracts/vault/TBTCOptimisticMinting.sol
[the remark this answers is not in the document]
pdyraga: 60186dbdced4f02de2649f7e22f6aecfc5f5414c

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,316 @@
...
+            requestedAt != 0,
+            "Optimistic minting not requested or already finalized"
+        );
+        require(
+            /* solhint-disable-next-line not-rely-on-time */
+            block.timestamp - requestedAt > OPTIMISTIC_MINTING_DELAY,
+            "Optimistic minting delay has not passed yet"
+        );
+
+        Deposit.DepositRequest memory deposit = bridge.deposits(depositKey);
```
dimpar, line 192: There's this comment https://github.com/keep-network/tbtc-v2/blob/main/solidity/contracts/bridge/Bridge.sol#L1504-L1506 about a `deposits` map containing valid and invalid deposits that a wallet is responsible to validate them. Are we going to use the same validation here?

pdyraga: Not in the contract but Minters and Guardians are supposed to perform this validation, yes. We do not know if the deposit revealed to Ethereum happened at all on Bitcoin.

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: There is one more thing we could do. Ignore Bitcoin fee as proposed here and in https://github.com/keep-network/tbtc-v2/commit/5ed6239a8a8472d9ae90a94eeb80a066ba12d31f but add an optimistic-minting-specific fee. This fee can be 0 initially but it can also be increased by the Governance later. As long as that fee > the actual average Bitcoin mining fee, the treasury will have a source of Bitcoin to refill the gap. I think that fee is the last missing piece to consider the mechanism from this contract as a candidate to replace the market-based quickminting.

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: I approached the implementation and the idea with the lead guard is too complicated.

That's what I would do:
```
addMinter onlyOwner
removeMinter onlyOwnerOrGuardian
addGuardian onlyOwner
removeGuardian onlyOwner

pauseOptimisticMinting onlyOwner
unpauseOptimisticMinting onlyOwner
```

In case a Guardian gets compromised, it can remove all Minters and DoS the contract until the owner removes the compromised Guardian and re-adds Minters.

If Minter gets compromised, it can keep requesting minting for non-existing deposits, Guardians will have to cancel them and remove the errant Minter.

Pause functionality should be used for long-term pauses, not to deal with errant Minters.

pdyraga: See 78d580c53b1057c9c07365e21f66fd3c7118738e and 9745ab844a70fc6a3b07b9500b6ac89141f98d42, please.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,377 @@
...
+        uint32 fundingOutputIndex
+    ) external onlyGuardian {
+        uint256 depositKey = calculateDepositKey(
+            fundingTxHash,
+            fundingOutputIndex
+        );
+
+        require(
+            pendingOptimisticMints[depositKey] > 0,
+            "Optimistic minting not requested of already finalized"
```
dimpar, line 279:
```suggestion
            "Optimistic minting was not requested or it has been finalized"
```

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,377 @@
...
+
+    /// @notice Indicates if the given address is a Guardian. Only Guardians can
+    ///         cancel requested optimistic minting.
+    mapping(address => bool) public isGuardian;
+
+    /// @notice Indicates if the optimistic minting has been paused. Only the
+    ///         Governance can pause optimistic minting. Note that the pause of
+    ///         the optimistic minting does not stop the standard minting flow
+    ///         where wallets sweep deposits.
+    bool public isOptimisticMintingPaused;
```
dimpar, line 53: What do you think if we change this to `bool public isOptimisticMintingAllowed = true`? This way we would use fewer negations in the code which makes reading easier. E.g
Instead of 
```
modifier whenOptimisticMintingNotPaused() {
        require(!isOptimisticMintingPaused, "Optimistic minting paused");
        _;
    }
```
we can write
```
 modifier whenOptimisticMintingAllowed() {
        require(isOptimisticMintingAllowed, "Optimistic minting paused"); // paused -> disallowed?
        _;
    }
```
and use `whenOptimisticMintingAllowed` where applicable.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,377 @@
...
+    function removeGuardian(address guardian) external onlyOwner {
+        require(isGuardian[guardian], "This address is not a guardian");
+        delete isGuardian[guardian];
+        emit GuardianRemoved(guardian);
+    }
+
+    /// @notice Pauses the optimistic minting. Note that the pause of the
+    ///         optimistic minting does not stop the standard minting flow
+    ///         where wallets sweep deposits.
+    function pauseOptimisticMinting() external onlyOwner {
```
dimpar, line 323: If you agree with the prev. comment, then we can simplify this function to
```
function pauseOptimisticMinting() external onlyOwner whenOptimisticMintingAllowed {
        isOptimisticMintingAllowed = false;
        emit OptimisticMintingPaused();
    }
```

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,377 @@
...
+        require(
+            !isOptimisticMintingPaused,
+            "Optimistic minting already paused"
+        );
+        isOptimisticMintingPaused = true;
+        emit OptimisticMintingPaused();
+    }
+
+    /// @notice Unpauses the optimistic minting.
+    function unpauseOptimisticMinting() external onlyOwner {
```
dimpar, line 333: This is the only place we would negate our flag
```
function allowOptimisticMinting() external onlyOwner {
        require(!isOptimisticMintingAllowed, "Optimistic minting is already allowed");
        isOptimisticMintingAllowed = true;
        emit OptimisticMintingUnpaused();
    }
```

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,375 @@
...
+        require(isOptimisticMintingPaused, "Optimistic minting is not paused");
+        isOptimisticMintingPaused = false;
+        emit OptimisticMintingUnpaused();
+    }
+
+    /// @notice Calculates deposit key the same way as the Bridge contract.
+    ///         The deposit key is computed as
+    ///         keccak256(fundingTxHash | fundingOutputIndex).
+    function calculateDepositKey(
```
nkuba, line 342: Could we add a unit test for `calculateDepositKey` function?

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    /// @notice Used by TBTCVault.receiveBalanceIncrease to repay the optimistic
+    ///         minting debt before TBTC is minted. When optimistic minting is
+    ///         finalized, debt equal to the value of the deposit being
+    ///         a subject of the optimistic minting is incurred. When TBTCVault
+    ///         sweeps a deposit, the debt is fully or partially paid off, no
+    ///         matter if that particular deposit was used for the optimistic
+    ///         minting or not.
+    function repayOptimisticMintDebt(address depositor, uint256 amount)
```
nkuba, line 359: For consistency with the map name:
```suggestion
    function repayOptimisticMintingDebt(address depositor, uint256 amount)
```

nkuba, line 359: Could we describe in doc what is this function expected to return and what exactly is the `amount` parameter?

nkuba (commented):
First review round (contracts only).

solidity/contracts/vault/TBTCOptimisticMinting.sol
nkuba: I kind of prefer the term "paused" instead of "allowed", as this path is something extraordinary, and pausing minting is exactly what it is meant to do.
What is more, the variable will be by default initialized with `false`, so no need to initiate it at the contract deployment.

solidity/contracts/vault/TBTCOptimisticMinting.sol
mhluongo: Reiterating from our live discussion...

> add an optimistic-minting-specific fee.

yes please! Let's not start it at 0 though, but instead set it to whatever WBTC's current deposit fee is.

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: b70fb9a5b68ffcb6829fd15fc380d36b3b18923c

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,377 @@
...
+        emit OptimisticMintingCancelled(
+            msg.sender,
+            fundingTxHash,
+            fundingOutputIndex,
+            depositKey
+        );
+    }
+
+    /// @notice Adds the address to the Minter set.
+    function addMinter(address minter) external onlyOwner {
```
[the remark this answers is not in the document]
pdyraga: `removeMinter` has `onlyOwnerOrGuardian` so that a Guardian can remove an errant minter, see [this](https://github.com/keep-network/tbtc-v2/pull/424#discussion_r1042313896) discussion. We may have a situation when an errant Minter keeps requesting optimistic minting for deposits that did not happen on Bitcoin in a loop. Then, Guardians need to keep canceling these deposit requests in a loop and it becomes a game of who is faster and who has more ether. `removeMinter` will hopefully never be used by Guardians.

In the case of `addMinter`, we do not want any of the Guardians to be able to add Minters. The only role of the Guardian is to watch what Minters are doing and potentially react, not to decide who is a Minter. It is best for security when these two sets are separated.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,377 @@
...
+    /// @notice Allows a Guardian to cancel optimistic minting request. The
+    ///         following conditions must be met:
+    ///         - The optimistic minting request for the given deposit exists.
+    ///         - The optimistic minting request for the given deposit has not
+    ///           been finalized yet.
+    /// @dev Guardians must validate the following conditions for every deposit
+    ///      for which the optimistic minting was requested:
+    ///      - The deposit happened on Bitcoin side and it has enough
+    ///        confirmations.
+    ///      - The optimistic minting has been requested early enough so that
```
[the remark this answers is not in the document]
pdyraga: Depends on the locktime, how many deposits are there to be swept for the wallet, and how fast the wallet can sweep. That's something we'll have to decide when implementing Guardian's bot. For a 6-month lock, I would expect that the optimistic minting should be requested no later than one month after the deposit is revealed. It leaves 5 months for a sweep.

solidity/contracts/vault/TBTCOptimisticMinting.sol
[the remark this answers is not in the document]
pdyraga: 2823919eae18055e1ca13451511e43d401c9a79e

pdyraga: This conflicts with [this](https://github.com/keep-network/tbtc-v2/pull/424#discussion_r1047236280) comment. @nkuba @dimpar let's settle on one version. Both options work for me but I'd probably prefer the current (shorter) one from 2823919eae18055e1ca13451511e43d401c9a79e.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,377 @@
...
+            fundingOutputIndex
+        );
+        Deposit.DepositRequest memory deposit = bridge.deposits(depositKey);
+
+        require(deposit.revealedAt != 0, "The deposit has not been revealed");
+        require(deposit.sweptAt == 0, "The deposit is already swept");
+        require(deposit.vault == address(this), "Unexpected vault address");
+
+        /* solhint-disable-next-line not-rely-on-time */
+        pendingOptimisticMints[depositKey] = block.timestamp;
```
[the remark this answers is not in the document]
pdyraga: That's a good one. `pendingOptimisticMints` is intentionally cleared in `cancelOptimisticMint` to cover the case of an errant Guardian who could cancel optimistic minting for a valid revealed deposit. The idea was to leave the responsibility of canceling optimistic mints if they were already finalized in the past to Guardians, but given we can fit two timestamps (and more) into a single EVM word, we can have this validation on-chain and make the Guardian implementation way lighter. See 5d2541179268bfdf181ff39907505f9a9d0aa1a3 please.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,377 @@
...
+            fundingOutputIndex,
+            depositKey
+        );
+    }
+
+    /// @notice Allows a Minter to finalize previously requested optimistic
+    ///         minting. The following conditions must be met:
+    ///         - The optimistic minting has been requested for the given
+    ///           deposit.
+    ///         - The deposit has not been swept yet.
```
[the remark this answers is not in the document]
pdyraga: Let's separate how this will work in January and how this will - potentially - work a couple of months later.

In January, wallets are not yet sweeping. The 3h delay is the time for the deposit transaction to be revealed on Ethereum, mined on Bitcoin, and become sufficiently confirmed on both Ethereum and Bitcoin. (20 + 6 confirmations?).

Once we activate sweeping, we will have to decide if the optimistic minting gets paused or if it stays as a quickminting solution. If we pause it, we have no problem here. If we use it as a quickminting solution, we will have kind of a race between the optimistic minting and wallets sweeping.

If the optimistic minting gets requested, and the wallet sweeps first, the finalization will not be possible. The request may stay dangling or potentially be canceled by Guardians to clean up the optimistic minting state. I am not sure if it is worth the gas expenditure though.

If the optimistic minting gets requested, and it gets finalized before the wallet sweeps, the `optimisticMintingDebt` will get incurred. When the wallet finalizes the sweep of the deposit, it will be partially paid off. This will also be a scenario when we activate the sweeping of deposits from January. Please see the unit tests added in 6d5eddeaa3f63bbf1fd4d5ba5a7d02ef572d0b38 - I worked on them while you were reviewing.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,380 @@
...
+        // The Bitcoin miner fee is ignored. When sweeping, the miner fee is
+        // subtracted so the optimisticMintingDebt may stay non-zero after the
+        // deposit is swept.
+        //
+        // This imbalance is supposed to be solved by a donation to the Bridge.
+        uint256 amountToMint = deposit.amount - deposit.treasuryFee;
+        _mint(deposit.depositor, amountToMint);
+        optimisticMintingDebt[deposit.depositor] += amountToMint;
+
+        delete pendingOptimisticMints[depositKey];
```
lukasz-zimnoch: Deleting can be problematic as the same deposit key can be reused for optimistic minting if the deposit was not yet swept. If we decide to not delete, we should also prevent `pendingOptimisticMints` duplicates in the `optimisticMint` function. Unless we want to push that responsibility to the guardians? If so, we should clearly note that requirement.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,377 @@
...
+    ///      minting for this deposit because the wallet will not be able to
+    ///      sweep it. The on-chain optimistic minting code does not perform any
+    ///      validation for gas efficiency: it would have to perform the same
+    ///      validation as `validateDepositRefundLocktime` and expect the entire
+    ///      `DepositRevealInfo` to be passed to assemble the expected script
+    ///      hash on-chain. Guardians must validate if the deposit happened on
+    ///      Bitcoin, that the script hash has the expected format, and that the
+    ///      wallet is an active one so they can also validate the time left for
+    ///      the refund.
+    function optimisticMint(bytes32 fundingTxHash, uint32 fundingOutputIndex)
```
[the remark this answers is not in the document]
pdyraga: Good idea. See dbe3c7e8d5768bb866247be6822c46cb51b5547d, please. This way, the naming is aligned with:
between `requestOptimisticMint`, `finalizeOptimisticMint`, and `cancelOptimisticMint`.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,419 @@
...
+            request.requestedAt != 0,
+            "Optimistic minting not requested for the deposit"
+        );
+        require(
+            request.finalizedAt == 0,
+            "Optimistic minting already finalized for the deposit"
+        );
+
+        // Delete it. It allows to request optimistic minting for the given
+        // deposit again. Useful in case of an errant Guarian.
```
tomaszslabon, line 320:
```suggestion
        // deposit again. Useful in case of an errant Guardian.
```

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,377 @@
...
+    ///         to the Bridge that has not been yet swept and led to the
+    ///         optimistic minting of TBTC. When TBTCVault sweeps a deposit,
+    ///         the debt is fully or partially paid off, no matter if that
+    ///         particular swept deposit was used for the optimistic minting or
+    ///         not.
+    mapping(address => uint256) public optimisticMintingDebt;
+
+    event OptimisticMintingRequested(
+        address indexed minter,
+        address depositor,
```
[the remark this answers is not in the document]
pdyraga: Yes, please see [this](https://github.com/keep-network/tbtc-v2/pull/424#discussion_r1048386483) comment.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,320 @@
...
+///         minted. During the time of the delay, any Guardian can cancel the
+///         minting.
+/// @dev This functionality is a part of TBTCVault. It is implemented in
+///      a separate abstract contract to achieve better separation of concerns
+///      and easier-to-follow code.
+abstract contract TBTCOptimisticMinting is Ownable {
+    /// @notice The time that needs to pass between the moment the optimistic
+    ///         minting is requested and the moment optimistic minting is
+    ///         finalized with minting TBTC.
+    uint256 public constant OPTIMISTIC_MINTING_DELAY = 3 hours;
```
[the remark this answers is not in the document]
pdyraga: Via contract upgrade, see [this](https://github.com/keep-network/tbtc-v2/pull/424#discussion_r1043828782) discussion please.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,380 @@
...
+
+    /// @notice Adds the address to the Minter set.
+    function addMinter(address minter) external onlyOwner {
+        require(!isMinter[minter], "This address is already a minter");
+        isMinter[minter] = true;
+        emit MinterAdded(minter);
+    }
+
+    /// @notice Removes the address from the Minter set.
+    function removeMinter(address minter) external onlyOwnerOrGuardian {
```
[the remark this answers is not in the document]
pdyraga: Yes. We may have a situation when an errant Minter keeps requesting optimistic minting for deposits that did not happen on Bitcoin in a loop. Then, Guardians need to keep canceling these deposit requests in a loop and it becomes a game of who is faster and who has more ether. `removeMinter` will hopefully never be used by Guardians. Please see [this](https://github.com/keep-network/tbtc-v2/pull/424#discussion_r1042313896) discussion.

solidity/contracts/vault/TBTCVault.sol
```diff
@@ -70,7 +71,11 @@ contract TBTCVault is IVault, Ownable {
         _;
     }
 
-    constructor(Bank _bank, TBTC _tbtcToken) {
+    constructor(
+        Bank _bank,
+        TBTC _tbtcToken,
+        Bridge _bridge
+    ) TBTCOptimisticMinting(_bridge) {
```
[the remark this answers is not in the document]
pdyraga: Yeah, see 2e84c185c28b4dc7731ea9df2f879d5a1819de81, please. We used `bank.bridge()` in `TBTCVault` to have a single update point in case replacing the entire Bridge is needed. Before optimistic minting, `TBTCVault` worked only with `Bank` and was not interested in how the `Bridge` worked. Hence, it was possible to do `bank.bridge()` to obtain the `Bridge` address for the redemption request. Nothing more was needed. With the optimistic minting, `TBTCVault` must inspect the deposit state, so updating Bridge is no longer that easy. That being said, we can refer to the `bridge` field from `TBTCOptimisticMinting` the `TBTCVault` inherits from.

solidity/contracts/vault/TBTCVault.sol
```diff
@@ -70,7 +71,11 @@ contract TBTCVault is IVault, Ownable {
         _;
     }
 
-    constructor(Bank _bank, TBTC _tbtcToken) {
+    constructor(
+        Bank _bank,
+        TBTC _tbtcToken,
+        Bridge _bridge
```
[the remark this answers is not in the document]
pdyraga: `TBTCVault` no longer calls `bank.bridge()`, see [this](https://github.com/keep-network/tbtc-v2/pull/424#discussion_r1047134331) discussion, please.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,380 @@
...
+        );
+
+        uint256 requestedAt = pendingOptimisticMints[depositKey];
+        require(
+            requestedAt != 0,
+            "Optimistic minting not requested or already finalized"
+        );
+        require(
+            /* solhint-disable-next-line not-rely-on-time */
+            block.timestamp - requestedAt > OPTIMISTIC_MINTING_DELAY,
```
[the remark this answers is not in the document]
pdyraga: Yep! See dc7eab2f59fe7c9e62761a23208c05d827292a14, please.

solidity/contracts/vault/TBTCVault.sol
pdyraga: I have mixed feelings about it. Our `ERC20WithPermit` implementation [allows to mint 0 tokens](https://github.com/thesis/solidity-contracts/blob/79f658f544232876f70da4bb4b15f365bad6372b/contracts/token/ERC20WithPermit.sol#L158-L170). Similarly, [OpenZeppelin ERC20](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3d7a93876a2e5e1d7fe29b5a0e96e222afdc4cfa/contracts/token/ERC20/ERC20.sol#L259-L272) implementation. Here is OpenZeppelin thread about it: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/513

Minting 0 will cost some gas but this is quite a rare case and checking if the value is non-zero will also cost some gas while this check will return false in the vast majority of cases. Also, having `Minted` event with 0 value could help in debugging the transaction/building graph.

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: This address is not something the user is entering or copy-pasting. The vault's address is set during the deposit reveal, and only deposits targeted to the `TBTCVault` can be requested for optimistic minting. 

> "Vault address cannot be bridge address. Please copy and paste more carefully. :-)"

The user is not copy-pasting the value for this call. If they made a mistake during the deposit reveal, they must wait for the P2SH unlock. 

> "The deposit's vault doesn't match the current vault"

There is only one `TBTCVault`, and "current" is not changing.

Is the current revert message unintuitive? 🤔

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: @nkuba @lukasz-zimnoch @mhluongo Thoughts on this one? The current version feels more intuitive to me.

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: +1 to @nkuba's answer. Additionally, I'd like to note that the current solution is similar to how OpenZeppelin [handles it](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/security/Pausable.sol).

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: See [this answer](https://github.com/keep-network/tbtc-v2/pull/424#discussion_r1047585100) please.

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: See [this answer](https://github.com/keep-network/tbtc-v2/pull/424#discussion_r1047585100) please.

solidity/contracts/vault/TBTCOptimisticMinting.sol
dimpar: ok

solidity/contracts/vault/TBTCOptimisticMinting.sol
dimpar: No biggy, a shorter version is fine by me. There was a typo that Kuba also noticed. But I think we miss `was`, ie `Optimistic minting was not requested or already finalized`

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: That's true but I think the current revert message is a bit more intuitive. Note that `pauseOptimisticMinting` says `Optimistic minting already paused` and `requestOptimisticMint` says `Optimistic minting paused`.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,518 @@
...
+    /// @notice Divisor used to compute the treasury fee taken from each
+    ///         optimistically minted deposit and transferred to the treasury
+    ///         upon finalization of the optimistic mint. This fee is computed
+    ///         as follows: `fee = amount / optimisticMintingFeeDivisor`.
+    ///         For example, if the fee needs to be 2% of each deposit,
+    ///         the `optimisticMintingFeeDivisor` should be set to `50` because
+    ///         `1/50 = 0.02 = 2%`.
+    ///         Note that the optimistic minting fee does not replace the
+    ///         deposit treasury fee cut by the Bridge. Both fees are deducted
+    ///         from the minted token amount.
```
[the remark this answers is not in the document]
pdyraga: We could apply the optimistic minting fee based on the original deposit amount, but that would lead to inconsistent amounts if the optimistic minting fee would be high. It is easiest to explain it with an example. Say that the deposit amount is 1000 and that the Bridge fee is 20. 980 is what lands in TBTCVault. If the optimistic deposit fee was anything > 98%, we would be printing tokens because only 980 out of the original 1000 lands in the vault. In other words, the vault can cut a fee only from what it is actually receiving. We could probably introduce some kind of validation between fees but that would break the separation between vaults (applications) and the Bridge.

Please see 22e18b7c1efb262eeda1be7ead8e865c69b8b14c for documentation improvements.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,518 @@
...
+    /// @notice Optimistic minting debt value per depositor's address. The debt
+    ///         represents the total value of all depositor's deposits revealed
+    ///         to the Bridge that has not been yet swept and led to the
+    ///         optimistic minting of TBTC. When `TBTCVault` sweeps a deposit,
+    ///         the debt is fully or partially paid off, no matter if that
+    ///         particular swept deposit was used for the optimistic minting or
+    ///         not.
+    mapping(address => uint256) public optimisticMintingDebt;
+
+    /// @notice New optimistic minting fee deivisor value. Set only when the
```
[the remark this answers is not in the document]
pdyraga: Done in 22e18b7c1efb262eeda1be7ead8e865c69b8b14c.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,420 @@
...
+        emit OptimisticMintingUnpaused();
+    }
+
+    /// @notice Calculates deposit key the same way as the Bridge contract.
+    ///         The deposit key is computed as
+    ///         keccak256(fundingTxHash | fundingOutputIndex).
+    function calculateDepositKey(
+        bytes32 fundingTxHash,
+        uint32 fundingOutputIndex
+    ) public view returns (uint256) {
```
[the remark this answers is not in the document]
pdyraga: > Shouldn't this be a pure function?

Yes, done!

> Apart from this could this function be abstracted to the Deposit library, so deposit keys are always computed this way? 

This is true but I'd prefer to do it separately. This PR is already huge.

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,518 @@
...
+        isOptimisticMintingPaused = false;
+        emit OptimisticMintingUnpaused();
+    }
+
+    /// @notice Begins the process of updating optimistic minting fee.
+    ///         The fee is computed as follows:
+    ///         `fee = amount / optimisticMintingFeeDivisor`.
+    ///         For example, if the fee needs to be 2% of each deposit,
+    ///         the `optimisticMintingFeeDivisor` should be set to `50` because
+    ///         `1/50 = 0.02 = 2%`.
```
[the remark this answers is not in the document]
pdyraga: Hard to say what the "right" values are so I'd say not to handcuff ourselves and leave them in the hands of the DAO. I think all the edge cases are properly handled by the contract.

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: CoinList charges a flat 0.25% trading fee to wrap or unwrap WBTC (https://coinlist.co/help/what-are-the-fees-to-wrap-and-unwrap-wbtc). If we want to maintain the same fee, for the January release, we have two options:

- If the treasury should receive the fee at the moment of sweeping, set the Deposit fee to 0.25% and the Optimistic Minting fee to 0%.
- If the treasury should receive the fee at the moment of finalizing the Optimistic Mint, set the Optimistic Mint fee to 0.25% and the Deposit Fee to 0%.

solidity/contracts/vault/TBTCOptimisticMinting.sol
mhluongo: I prefer the current version

solidity/contracts/vault/TBTCOptimisticMinting.sol
dimpar: I'd just say `Optimistic minting was paused` in both cases because they happened in the past and use a modifier here. Up to you.

pdyraga:
There is one more thing worth discussing which I believe is a product question cc @mhluongo. 

There are two ways to mint TBTC:
- Reveal the deposit and route it to `TBTCVault` in the same TX. (Deposit reveal info has an `address vault` field.) When sweeping, the wallet is calling the vault. The `TBTCVault` has the balance in the Bank increase and mints TBTC.
- Reveal deposit with `0x0` `vault` address. The depositor's balance in the `Bank` is increased. Bitcoin crossed the Bridge but no token minting happened. After some time, the depositor can call the `TBTCVault` to mint TBTC based on the depositor's Bank balance. They can do it in one or two transactions: `approveBankBalance` and then `mint` or `receiveBalanceApproval`. `TBTCVault` will received the depositor's balance in the Bank and will mint TBTC.

When TBTC is optimistically minted, the optimistic debt in the TBTC vault is increased for the depositor. That debt is paid off - fully or partially - during the sweeping of that depositor's next deposit.

However, the debt is not paid off when someone is using the second option - deposit to Bank with no minting and then mint using Bank balance. My reasoning was that is not an optimistic minting path - for an optimistic minting deposit has to be revealed and routed to `TBTCVault` so there is no point in complicating the other part and making it more gas-expensive to use.

If we can cover both parts, I can do it here.

mhluongo:
I suppose we'd also like to support "optimistic credits" to make sure people can still deposit + attain a BTC balance if they don't want to mint in the same tx.

manumonti (commented):
Requesting some minor changes and giving some suggestions as well as asking some questions.

mhluongo:
Thinking more about the bank fee issue, we can push resolving it until after the launch as we won't support the non-minting path initially. So let's punt on both of these for now.

solidity/contracts/vault/TBTCOptimisticMinting.sol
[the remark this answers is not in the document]
mhluongo: Agreed with both of y'all, let's do it in the next version.

mhluongo (approved):
I'm for merging here and addressing outstanding concerns in new issues when appropriate and, ideally, in the next revision of this contract after this lands with the chaosnet launch.

Note, I've kept my review to Solidity and only briefly read over test coverage. I'm looking for one more +1 before merge if anyone is down for it cc @nkuba @lukasz-zimnoch

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,557 @@
...
+import "../bridge/Bridge.sol";
+import "../bridge/Deposit.sol";
+import "../GovernanceUtils.sol";
+
+/// @title TBTC Optimistic Minting
+/// @notice The Optimistic Minting mechanism allows to mint TBTC before
+///         `TBTCVault` receives the Bank balance. There are two permissioned
+///         sets in the system: Minters and Guardians, both set up in 1-of-n
+///         mode. Minters observe the revealed deposits and request minting TBTC.
+///         Any single Minter can perform this action. There is a 3 hours delay
```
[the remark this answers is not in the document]
pdyraga: 4d1770c026bf5e0c30ab38cda11ec77697a75a45

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,557 @@
...
+    ///         - At least `optimisticMintingDelay` passed since the optimistic
+    ///           minting was requested for the given deposit.
+    ///         - The optimistic minting has not been finalized earlier for the
+    ///           given deposit.
+    ///         - The optimistic minting request for the given deposit has not
+    ///           been canceled by a Guardian.
+    ///         - The optimistic minting is not paused.
+    ///         This function mints TBTC and increases `optimisticMintingDebt`
+    ///         for the given depositor. The finalized optimistic minting
+    ///         request is removed from the contract.
```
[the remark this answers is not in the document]
pdyraga: 4d1770c026bf5e0c30ab38cda11ec77697a75a45

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,557 @@
...
+        }
+
+        /* solhint-disable-next-line not-rely-on-time */
+        request.finalizedAt = uint64(block.timestamp);
+
+        emit OptimisticMintingFinalized(
+            msg.sender,
+            depositKey,
+            deposit.depositor,
+            newDebt
```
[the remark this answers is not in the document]
pdyraga: Would it be possible/problematic to read it from the `Minted` event to the treasury address?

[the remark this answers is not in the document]
manumonti: Well, if I understand correctly, reading `Minted` events would help to know how much fees are receiving the treasury, but it is not easy to know how much of these fees are `OptimisticMintFee`.

Anyway, I'm not sure if tracking the generation of `OptimisticMintFees` is interesting enough for adding a specific parameter to the event.

nkuba (dismissed):
I reviewed contracts only.

solidity/test/vault/TBTCVault.OptimisticMinting.test.ts
```diff
@@ -0,0 +1,1867 @@
...
+        return {
+          mockBank,
+          mockBridge,
+          tbtc,
+          tbtcVault,
+        }
+      }
+
+      context("when both deposits were optimistically minted", () => {
+        let f: Fixture
```
lukasz-zimnoch, line 1709: Maybe just `fixture`? The `f` says nothing.

solidity/test/vault/TBTCVault.OptimisticMinting.test.ts
```diff
@@ -0,0 +1,1867 @@
...
+    it("should calculate the same key as the Bridge", async () => {
+      expect((await bridge.deposits(depositKey)).revealedAt).to.equal(
+        await lastBlockTime()
+      )
+    })
+  })
+
+  describe("receiveBalanceIncrease", () => {
+    context(
+      "when the deposit for which optimistic minting was requested gets swept",
```
lukasz-zimnoch, line 1600:
```suggestion
      "when the deposit for which optimistic minting was requested gets swept after finalization",
```

solidity/test/vault/TBTCVault.OptimisticMinting.test.ts
```diff
@@ -0,0 +1,1867 @@
...
+
+        it("should emit an event", async () => {
+          await expect(tx)
+            .to.emit(tbtcVault, "OptimisticMintingDebtRepaid")
+            .withArgs(depositRevealInfo.depositor, 1500)
+        })
+      }
+    )
+
+    context("when multiple deposits gets swept", () => {
```
lukasz-zimnoch, line 1658:
```suggestion
    context("when multiple deposits gets swept after finalization", () => {
```

lukasz-zimnoch (commented):
Last comments and I'm good to merge!

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -0,0 +1,557 @@
...
+    }
+
+    /// @dev Mints the given amount of TBTC to the given depositor's address.
+    ///      Implemented by TBTCVault.
+    function _mint(address minter, uint256 amount) internal virtual;
+
+    /// @notice Allows to fetch a list of all Minters.
+    function getMinters() external view returns (address[] memory) {
+        return minters;
+    }
```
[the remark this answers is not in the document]
pdyraga: It is needed so that we can fetch all minters in one call with `await tbtcVault.getMinters()`. Without this function, we'd have to fetch them index by index: `await tbtcVault.minters(0)`, `await tbtcVault.minters(1)`, ...

solidity/test/vault/TBTCVault.OptimisticMinting.test.ts
pdyraga: I started with `fixture` but then all the function calls in `before` and in assertions got quite long and IMO harder to follow. My reasoning was that the `f: Fixture` is so close to the code using `f` that it's easier to find what `f` is.

solidity/test/vault/TBTCVault.OptimisticMinting.test.ts
pdyraga: b9a9a2c5b87a65be1d2062e763101605cd976ef0

solidity/test/vault/TBTCVault.OptimisticMinting.test.ts
pdyraga: b9a9a2c5b87a65be1d2062e763101605cd976ef0

solidity/test/vault/TBTCVault.OptimisticMinting.test.ts
```diff
@@ -0,0 +1,1867 @@
...
+
+      context("when the deposit has been revealed", () => {
+        context("when the deposit has been swept", () => {
+          before(async () => {
+            await createSnapshot()
+
+            await bridge.revealDeposit(fundingTx, depositRevealInfo)
+
+            relay.getPrevEpochDifficulty.returns(chainDifficulty)
+            relay.getCurrentEpochDifficulty.returns(chainDifficulty)
```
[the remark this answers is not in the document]
pdyraga: 3af08f40ae65713cc02723e0d00a3ed28ac37ff0

lukasz-zimnoch (approved):

solidity/contracts/vault/TBTCOptimisticMinting.sol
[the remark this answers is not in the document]
cygnusv: I think it's interesting enough, but it should be possible to track via the Minted events that belong to the same TX.

PR #425, opened 2022-12-10
Use public key hashes in assembleDepositScript()

michalsmiarowski:
For the dashboard we want to be able to mock the deposit transaction with custom mocked bitcoin client. This is needed to be able ot test the dashboard without the need to connect to the real bitcoin client.

For that we wanted to use `assembleDepositScript` function but the problem was that it used `Deposit` object as a parameter, which had `walletPublicKey` and `refundPublicKey`. We then hash it into `walletPublicKeyHash` and `refundPublicKeyHash` inside that function, so we didn't use `refundPublicKey` and `walletPublicKey` there at all besides that. From the dashboard perspective we only have the `refundPublicKeyHash` and we can`t really convert it to a `refundPublicKey` since the convertion goes as below:

`private key -> public key -> compressed public key -> public key hash (PKH) <-> base58 encoded PKH (address)`

This commit changes that in a way that we pass `DepositScriptParameters` instead of `Deposit` object + `amount` as a separate property (before it was placed inside `Deposit` parameters and `DepositScriptParameters` doesn't have that).

### DIFF
--- a/typescript/src/deposit.ts
+++ b/typescript/src/deposit.ts
@@ -150,9 +150,12 @@ export async function submitDepositTransaction(
     })
   }
 
+  const depositScriptParameters = getDepositScriptParameters(deposit)
+
   const { transactionHash, depositUtxo, rawTransaction } =
     await assembleDepositTransaction(
-      deposit,
+      depositScriptParameters,
+      deposit.amount,
       utxosWithRaw,
       depositorPrivateKey,
       witness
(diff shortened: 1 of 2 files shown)


typescript/src/deposit.ts
```diff
@@ -179,7 +183,8 @@ export async function submitDepositTransaction(
  *          - the deposit transaction in the raw format
  */
 export async function assembleDepositTransaction(
-  deposit: Deposit,
+  deposit: DepositScriptParameters,
+  amount: BigNumber,
```
lukasz-zimnoch, line 187: I'm not a huge fan of the detached amount but I don't see a better way 🤷 

Also, a new question arises: Can we refactor the whole `Deposit` interface to operate on wallet and refund key hashes instead of plain text? Of course, this is a huge change and not for this PR but this may be worth exploring if we observe that we have a constant need to make changes like the one presented in this PR. After all, the initial decision to base the `Deposit` interface on plain text public keys could be wrong as we were lacking a lot of information back in the day.

lukasz-zimnoch (approved):

typescript/src/deposit.ts
michalsmiarowski: Refactor of the `Deposit` interface: https://github.com/keep-network/tbtc-v2/pull/441

PR #427, opened 2022-12-15
Zero deposit and redemption fees

pdyraga:
During the work on the optimistic minting fee in #424, I realized the validation of the deposit and redemption treasury fee parameter in the Bridge does not allow setting the fee to zero, while the Bridge supports the zero-case for deposit and redemption.

We have a condition in `Deposit.sol` in the `revealDeposit` function allowing to properly handle the case when the Bridge's deposit treasury fee is zero:

```
deposit.treasuryFee = self.depositTreasuryFeeDivisor > 0
    ? fundingOutputAmount / self.depositTreasuryFeeDivisor
    : 0;
```

Similarly, we have a condition in `Redemption.sol` in the `requestRedemption` function allowing to properly handle the case when the Bridge's redemption treasury fee is zero:

```
uint64 treasuryFee = self.redemptionTreasuryFeeDivisor > 0
    ? amount / self.redemptionTreasuryFeeDivisor
    : 0;
```

This has been fixed and zero `depositTreasuryFeeDivisor` and `redemptionTreasuryFeeDivisor ` are allowed now. I have
also added unit tests ensuring:

- revealing a deposit with a zero treasury fee works fine,
- sweeping a deposit with a zero treasury fee works fine,
- requesting a redemption with a zero treasury fee works fine. 
 
The case of providing a redemption proof when the fee is zero is already covered, see all the `bridge.setRedemptionTreasuryFeeDivisor(0)` calls across `Bridge.Redemption.test.ts`.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -1240,7 +1240,7 @@ contract Bridge is
     ///        deposit refund locktime.
     /// @dev Requirements:
     ///      - Deposit dust threshold must be greater than zero,
-    ///      - Deposit treasury fee divisor must be greater than zero,
+    ///      - Deposit dust threshold must be greater than deposit TX max fee,
     ///      - Deposit transaction max fee must be greater than zero.
     function updateDepositParameters(
         uint64 depositDustThreshold,
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -1302,7 +1302,8 @@ contract Bridge is
     /// @dev Requirements:
     ///      - Redemption dust threshold must be greater than moving funds dust
     ///        threshold,
-    ///      - Redemption treasury fee divisor must be greater than zero,
+    ///      - Redemption dust threshold must be greater than the redemption TX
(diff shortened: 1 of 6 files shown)


dimpar (approved):
lgtm

solidity/contracts/bridge/Bridge.sol
```diff
@@ -1302,7 +1302,8 @@ contract Bridge is
     /// @dev Requirements:
     ///      - Redemption dust threshold must be greater than moving funds dust
     ///        threshold,
-    ///      - Redemption treasury fee divisor must be greater than zero,
+    ///      - Redemption dust threshold must be greater than the redemption TX
+    ///        max fee,
     ///      - Redemption transaction max fee must be greater than zero,
```
tomaszslabon, line 1307: We are missing one requirement in the description of `updateRedemptionParameters` functions:
```
- redemption transaction max total fee must be greater or equal to redemption transaction max fee
```

pdyraga: Eagle eye! 0b16550757d9b19d018e68d74864949a593b7e17

tomaszslabon (approved):

PR #428, opened 2022-12-20
TS: Resolve contract details from artifacts (address, deployment block)

nkuba:
In this PR we introduce a couple of changes to the typescript library:

### Use the contract's address stored in the contract artifact - d96285e767e01a302df47bb4013462bf4fe74826

The artifacts contain addresses of the contracts deployed for the network the package dependency is used. This address will be used by default to initiate the contract, to overwrite the address user can provide it in the `ContractConfig`.

### Read `WalletRegistry` artifact from `@keep-network/ecdsa` package - cd51ed0e45d9ed1fadde80d4acee0005dcb246e6

The `@keep-network/ecdsa` is a correct source of the `WalletRegistry` contract. As the contract is deployed as part of the ECDSA contracts deployment. The previous solution of getting this artifact from `@keep-network/tbtc-v2` worked only for the `development` environment, as this is a specific of the `development` package to contain contract artifacts deployed locally in CI.

### Define contract's deployment block number - d683dc5bb79129d7aaeffa837877681d7ef05f7a

To support events querying we need to determine the contract's deployment block number. The value is stored in the contract artifact, so we just read it and set it in a private property. To overwrite the value user can provide it in the `ContractConfig`.


typescript/src/ethereum.ts
```diff
@@ -39,11 +47,15 @@ export interface ContractConfig {
   /**
    * Address of the Ethereum contract as a 0x-prefixed hex string.
    */
-  address: string
+  address?: string
```
lukasz-zimnoch, line 50: Let's make it clear in the docs what happens if the `address` is not provided.

typescript/src/ethereum.ts
```diff
@@ -39,11 +47,15 @@ export interface ContractConfig {
...
   /**
    * Signer that will sign all contract transactions.
    */
   signer: Signer
+  /**
+   * Number of a block in which the contract was deployed.
+   */
+  deployedAtBlockNumber?: number
```
lukasz-zimnoch, line 58: Let's make it clear in the docs what happens if the `deployedAtBlockNumber ` is not provided.

lukasz-zimnoch (approved):

PR #441, opened 2022-12-23
Full refactor of `Deposit` interface to operate on public key hashes

michalsmiarowski:
Related discussion: https://github.com/keep-network/tbtc-v2/pull/425#discussion_r1050519278

This is a refactor of a `Deposit` interface to operate on public key hashes instead of plain text public keys. Before the changes the `Deposit` interface looked like this:

```
{
  depositor: Identifier
  amount: BigNumber
  blindingFactor: string
  walletPublicKey: string
  refundPublicKey: string
  refundLocktime: string
  vault?: Identifier
}
```

and this PR changes it to:

```
{
  depositor: Identifier
  amount: BigNumber
  blindingFactor: string
  walletPubKeyHash: string /* pubKeyHash instead of publicKey */
  refundPubKeyHash: string /* pubKeyHash instead of publicKey */
  refundLocktime: string
  vault?: Identifier
}
```

All the methods and it tests, that use an object of a type `Deposit`, were refactored properly so that it operates on pub key hashes.

Few changes worth noticing:

### - Add `vault` as a separate parameter to `revealDeposit` method

The `revealDeposit` method was not using the `amount` parameter from the `Deposit` interface at all so it was not needed. Because of that I've decided to change the `deposit` parameter type to `DepositScriptParameters` and add an optional `vault` parameter to the `revealDeposit` method.

### - Change return type of the `revealDeposit` method

I've changed the return type of the `revealDeposit` method from `void` to `string` - it now returns the transaction hash. We need to return either that or the whole `ContractTransaction` to be able to see if the transaction was successfull/error and to display the proper info in the dapp. Decided to return transaction hash which is just a `string` so that we don't have to return a type from an external library (`ContractTransaction` from `ethers`).

**TODO:** Same change will be required for all methods that fire up a transaction:
- requestRedemption
- submitDepositSweepProof
- submitRedemptionProof
- (...and maybe more...)

### - Change pub key validation in `prepareInputSignData`

Since we are now operating on pub key hashes instead of plain text public keys I've changed the validation in the `prepareInputSignData` to make sure the pub key hash is correct and same that the pub key hash assembled from wallet key ring passed as a parameter.


typescript/src/chain.ts
```diff
@@ -43,12 +43,14 @@ export interface Bridge {
    * @param depositOutputIndex - Index of the deposit transaction output that
    *        funds the revealed deposit
    * @param deposit - Data of the revealed deposit
+   * @returns Transaction hash of the reveal deposit transaction as string
```
lukasz-zimnoch, line 46: Just flagging: The type of this parameter is an important design decision in the context of potential multi-chain support. A `string` makes sense for Ethereum but may be not suitable for other chains. That said, I think we are good for now but let's keep that in mind while making that change for other methods of the `Bridge` interface.

typescript/src/deposit.ts
```diff
@@ -42,16 +41,20 @@ export interface Deposit {
...
   /**
-   * Compressed (33 bytes long with 02 or 03 prefix) Bitcoin public key of
-   * the wallet that is meant to receive the deposit.
+   * Public key hash of the wallet that is meant to receive the deposit. Must
+   * be an unprefixed hex string (without 0x prefix).
+   *
+   * You can use `computeHash160` function to get the hash from a plain text public key.
    */
-  walletPublicKey: string
+  walletPubKeyHash: string
```
lukasz-zimnoch, line 49: Let's be consistent with naming and avoid abbreviations:
```suggestion
  walletPublicKeyHash: string
```
Remember about changing all references as well

typescript/src/deposit.ts
```diff
@@ -42,16 +41,20 @@ export interface Deposit {
...
   /**
-   * Compressed (33 bytes long with 02 or 03 prefix) Bitcoin public key that
-   * is meant to be used during deposit refund after the locktime passes.
+   * Public key hash that is meant to be used during deposit refund after the
+   * locktime passes. Must be an unprefixed hex string (without 0x prefix).
+   *
+   * You can use `computeHash160` function to get the hash from a plain text public key.
    */
-  refundPublicKey: string
+  refundPubKeyHash: string
```
lukasz-zimnoch, line 57: Let's be consistent with naming and avoid abbreviations:
```suggestion
  refundPublicKeyHash: string
```
Remember about changing all references as well

lukasz-zimnoch:
This refactoring looks really good and I think we can go even further and finally obtain a uniform deposit interface suitable for the entire library. After the changes made so far we have:
```
export interface Deposit {
  depositor: Identifier
  amount: BigNumber
  blindingFactor: string
  walletPublicKeyHash: string
  refundPublicKeyHash: string
  refundLocktime: string
  vault?: Identifier
}

  Deposit,
  | "depositor"
  | "blindingFactor"
  | "refundLocktime"
  | "walletPubKeyHash"
  | "refundPubKeyHash"
> & {}
```

That means the only fields that are in `Deposit` but not in `DepositScriptParameters` are `amount` and `vault`. What if we remove `DepositScriptParameters` and have only `Deposit` in the following form:
```
export interface Deposit {
  depositor: Identifier
  blindingFactor: string
  walletPublicKeyHash: string
  refundPublicKeyHash: string
  refundLocktime: string
}
```
If we do that, we can use the `Deposit` interface for all functions and add `amount` and optional `vault` arguments where applicable. The `amount` will be added for all functions that are responsible for producing the deposit transaction. The `vault` field for code dealing with deposit reveals. Both fields will also become part of the `RevealedDeposit` type.

At the first sight, we should be able to introduce the aforementioned changes smoothly. The only "problematic" part is:
```
[deposit-sweep.ts]

if (previousOutput.value != deposit.amount.toNumber()) {
    throw new Error("Mismatch between amount in deposit and deposit tx")
}
```
This validation is here because `Deposit` contains the `amount` field. Once we remove it, that validation can be safely removed as well. The deposit amount should be just taken from the funding UTXO. This is how the `Bridge` contract behaves [here](https://github.com/keep-network/tbtc-v2/blob/ec671324c17c195aa8604eb2f4b12e34a065c947/solidity/contracts/bridge/Deposit.sol#L246)

Please think about it in detail and double-check if it makes sense. If so, we can make it either in this PR or in a separate one, as you prefer.

typescript/src/deposit.ts
michalsmiarowski: @lukasz-zimnoch I would prefer to stick to `PubKeyHash` instead of `PublicKeyHash` because we have that abbreviation in the contracts. For example we have `activeWalletPubKeyHash` in the `BridgeState.sol` or we have `reveal` parameter of the `revealDeposit` method in the `Deposit.sol` which is a structure that contains `walletPubKeyHash` and `refundPubKeyHash`. The inconsistnency betweent the contracts and the ts lib may be inconvinient.

What do you think? Changin this is not a problem at all so please confirm if you still want to do it and I'll do it 😉

michalsmiarowski:
> This refactoring looks really good and I think we can go even further and finally obtain a uniform deposit interface suitable for the entire library. After the changes made so far we have:
> 
> ```
> export interface Deposit {
>   depositor: Identifier
>   amount: BigNumber
>   blindingFactor: string
>   walletPublicKeyHash: string
>   refundPublicKeyHash: string
>   refundLocktime: string
>   vault?: Identifier
> }
> 
> export type DepositScriptParameters = Pick<
>   Deposit,
>   | "depositor"
>   | "blindingFactor"
>   | "refundLocktime"
>   | "walletPubKeyHash"
>   | "refundPubKeyHash"
> > & {}
> ```
> 
> That means the only fields that are in `Deposit` but not in `DepositScriptParameters` are `amount` and `vault`. What if we remove `DepositScriptParameters` and have only `Deposit` in the following form:
> 
> ```
> export interface Deposit {
>   depositor: Identifier
>   blindingFactor: string
>   walletPublicKeyHash: string
>   refundPublicKeyHash: string
>   refundLocktime: string
> }
> ```
> 
> If we do that, we can use the `Deposit` interface for all functions and add `amount` and optional `vault` arguments where applicable. The `amount` will be added for all functions that are responsible for producing the deposit transaction. The `vault` field for code dealing with deposit reveals. Both fields will also become part of the `RevealedDeposit` type.
> 
> At the first sight, we should be able to introduce the aforementioned changes smoothly. The only "problematic" part is:
> 
> ```
> [deposit-sweep.ts]
> 
> if (previousOutput.value != deposit.amount.toNumber()) {
>     throw new Error("Mismatch between amount in deposit and deposit tx")
> }
> ```
> 
> This validation is here because `Deposit` contains the `amount` field. Once we remove it, that validation can be safely removed as well. The deposit amount should be just taken from the funding UTXO. This is how the `Bridge` contract behaves [here](https://github.com/keep-network/tbtc-v2/blob/ec671324c17c195aa8604eb2f4b12e34a065c947/solidity/contracts/bridge/Deposit.sol#L246)
> 
> Please think about it in detail and double-check if it makes sense. If so, we can make it either in this PR or in a separate one, as you prefer.

I agree 👍  I will address it in a separate PR to not prolong the review process of this one

typescript/src/deposit.ts
lukasz-zimnoch: Actually, we don't have to be super consistent with what we have in the contracts. The mapping is done in one place, i.e.`ethereum.ts` file. We should rather seek to have everything consistent in the library itself. If we have `PubKey` everywhere in the library, that works for me. What I'm trying to say here is to avoid the case where we have both `PubKey` and `PublicKey` in the library.

michalsmiarowski: Changed it to `PublicKey`:
fcc0e42e55b0fde1659bac5dd7ac081241f0cdd1
69dfd10cfee4a7b56dae13a76a9600da11ff171f

typescript/src/deposit.ts
michalsmiarowski: Changed it to `PublicKey`:
fcc0e42e55b0fde1659bac5dd7ac081241f0cdd1
69dfd10cfee4a7b56dae13a76a9600da11ff171f

michalsmiarowski:
@lukasz-zimnoch Redy for next round of review

lukasz-zimnoch (approved):

Issue #439, opened 2022-12-22
Allow only the depositor to reveal a deposit


pdyraga:
Consider this scenario:
- Alice reveals her Bitcoin deposit and routes it to vault 0xABC
- Eve notices Alice's TX in the mempool and frontruns it with the same data but the vault 0xDEF

As a result, Alice's deposit lands in vault 0xDEF and not in vault 0xABC. Both 0xDEF and 0xABC have to be trusted vaults but it is still an issue given it is not what Alice was asking for.

https://github.com/keep-network/tbtc-v2/blob/main/solidity/contracts/bridge/Deposit.sol#L152

mhluongo:
A little surprised this wasn't caught in audit

pdyraga:
Reopening as this has also be addressed in `tbtc-v2.ts` library, see [this](https://github.com/keep-network/tbtc-v2/pull/442#pullrequestreview-1232438145) comment.

PR #442, opened 2022-12-23
Allow only the depositor to reveal a deposit

pdyraga:
Closes: #439 

Consider this scenario:
- Alice reveals her Bitcoin deposit and routes it to vault 0xABC
- Eve notices Alice's TX in the mempool and front-runs it with the same data but the vault 0xDEF

As a result, Alice's deposit lands in vault 0xDEF and not in vault 0xABC. Both 0xDEF and 0xABC have to be trusted vaults but it is still an issue given it is not what Alice was asking for.

Here we change the validation requirements of Bridge.revealDeposit function. From now on, this function must be called by the same Ethereum address as the one used in the P2(W)SH BTC deposit transaction as a depositor.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -347,12 +347,12 @@ contract Bridge is
     /// @param fundingTx Bitcoin funding transaction data, see `BitcoinTx.Info`.
     /// @param reveal Deposit reveal data, see `RevealInfo struct.
     /// @dev Requirements:
+    ///      - This function must be called by the same Ethereum address as the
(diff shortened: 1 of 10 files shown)


pdyraga:
@r-czajkowski @michalsmiarowski Bringing your attention to this PR in case if affects anything in the dApp.

lukasz-zimnoch (approved):

lukasz-zimnoch (approved):
We also need to adjust the `tbtc-v2.ts` library [here](https://github.com/keep-network/tbtc-v2/blob/0b8b556e3513b5219f309373a73192dcf823ec2a/typescript/src/ethereum.ts#L158) and remember to use the `Bridge` handle with the depositor as signer while revealing the deposit.

Issue #435, opened 2022-12-21
Script to genesis the LightRelay


pdyraga:
The `LightRelay` contract [must be initialized with a call to `genesis`](https://github.com/keep-network/tbtc-v2/blob/9d392bf984e3f4ec9e1ba1876eee00e04ad7ed54/solidity/contracts/relay/LightRelay.sol#L123-L163) after the deployment so that `getCurrentEpochDifficulty()` and `getPrevEpochDifficulty()` return non-zero values. To achieve it, we must `genesis` the relay with a previous Bitcoin epoch and immediately perform at least one `retarget`.

PR #443, opened 2022-12-28
Relay genesis and maintainer authorization

tomaszslabon:
Closes https://github.com/keep-network/tbtc-v2/issues/435.

This PR adds deployment scripts that:
* perform the `genesis` on the `LightRelay` contract,
* perform the first retarget on the `LightRelay` contract,
* set the requirement of authorisation to the `LightRelay` contract and set the address of the relay maintainer.

The scripts were tested: they set the state of `LightRelay` (current and previous difficulties, authorisation) 
appropriately.


solidity/deploy/23_authorize_relay_maintainer.ts
```diff
@@ -0,0 +1,27 @@
...
+    { from: deployer, log: true, waitConfirmations: 1 },
+    "setAuthorizationStatus",
+    true
+  )
+
+  await execute(
+    "LightRelay",
+    { from: deployer, log: true, waitConfirmations: 1 },
+    "authorize",
+    deployer
```
tomaszslabon, line 20: @nkuba @pdyraga 
Do we know the address (or addresses) of the accounts that will be maintaining the relay?
For now I just put the deployer himself here.

solidity/deploy/21_genesis_relay.ts
```diff
@@ -0,0 +1,31 @@
...
+    genesisHeader,
+    genesisHeight,
+    genesisProofLength
+  )
+}
+
+export default func
+
+func.tags = ["GenesisLightRelay"]
+func.dependencies = ["LightRelay"]
```
nkuba, line 31: We should use `func.skip` property to determine networks for which scripts from this PR should be executed.
See the example here: https://github.com/keep-network/tbtc-v2/blob/6ff3d7b1c7d0a421507037acf3a421c4261dfcdd/solidity/deploy/90_initialize_wallet_owner.ts#L20-L21

I guess we don't want to run them for `goerli` as we use `TestRelay` stub instead of the `LightRelay` there:
https://github.com/keep-network/tbtc-v2/blob/6ff3d7b1c7d0a421507037acf3a421c4261dfcdd/solidity/deploy/00_resolve_relay.ts#L21-L22

solidity/deploy/23_authorize_relay_maintainer.ts
pdyraga: Let's not authorize anyone now. We will authorize later, when setting up the off-chain maintainer.

solidity/deploy/22_retarget_relay.ts
```diff
@@ -0,0 +1,115 @@
...
+    { from: deployer, log: true, waitConfirmations: 1 },
+    "retarget",
+    headers
+  )
+}
+
+export default func
+
+func.tags = ["RetargetLightRelay"]
+func.dependencies = ["LightRelay"]
```
pdyraga, line 115: Shouldn't it depend on `GenesisLightRelay`?

solidity/deploy/23_authorize_relay_maintainer.ts
```diff
@@ -0,0 +1,27 @@
...
+  )
+}
+
+export default func
+
+func.tags = ["AuthorizeLightRelayMaintainer"]
+func.dependencies = ["LightRelay"]
```
pdyraga, line 27: Shouldn't it depend on `RetargetLightRelay`? We do not want to enable authorization before we do the first retarget.

solidity/deploy/23_authorize_relay_maintainer.ts
tomaszslabon: Removed authorization of the deployer account in 8bd7fa56dc286cf22ca36d6dbe7f34d8a9867fa6.
However, I left the requirement that the maintainers must be authorized (see `setAuthorizationStatus`).
The deployer will need to authorize some account in the future, so retargets can be submitted.

solidity/deploy/22_retarget_relay.ts
tomaszslabon: It should.
Fixed in e652e7e61f070fa573232761b15a87adddb5adc8.

solidity/deploy/23_authorize_relay_maintainer.ts
tomaszslabon: It should.
Fixed in e652e7e61f070fa573232761b15a87adddb5adc8.

solidity/deploy/21_genesis_relay.ts
tomaszslabon: Indeed. 
In ed8983798ba6d06892072267ad970e9263fbc254 I added to all the scripts that modify `LightRelay`: 
```
func.skip = async (hre: HardhatRuntimeEnvironment): Promise<boolean> =>
  hre.network.name === "goerli"
 ```

**Comment by pdyraga** [comment]
👀

File: solidity/deploy/21_genesis_relay.ts [hunk 5]
```diff
@@ -0,0 +1,34 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types"
```
pdyraga, line 1: `21_genesis_relay.ts` prefix is now conflicting with `21_authorize_maintainer_proxy_in_bridge.ts`. Since `LightRelay` genesis is something we want to execute only for mainnet, I propose to prefix it with `91_`.

solidity/deploy/21_genesis_relay.ts
pdyraga: I think we should rather use:
```
hre.network.name !== "mainnet"
```

We not only want to skip this step for Goreli but also for local deployment that also uses a stub. This step is only for mainnet.

solidity/deploy/21_genesis_relay.ts
tomaszslabon: Done in c8f8bfb3bab35edf9d7e8ac8b2e863b59db9cf2a:
```
        renamed:    21_genesis_relay.ts -> 91_genesis_relay.ts
        renamed:    22_retarget_relay.ts -> 92_retarget_relay.ts
        renamed:    23_authorize_relay_maintainer.ts -> 93_authorize_relay_maintainer.ts
```

pdyraga:
I tested it by deploying `LightRelay` on a local geth instance, performing genesis and retarget. Scripts work as expected:
```
❯ yarn deploy --tags LightRelay --network development
deploying "LightRelay" (tx: 0x759bc7ef3bc3a5a3a8001cf9d53b061510fb1a9eaaf1b1c879a088466374f63b)...: deployed at 0xbbC2813bE1723aF6021c96F29b4DfaF1C1b2d0db with 1977380 gas

❯ yarn deploy --tags RetargetLightRelay --network development
reusing "LightRelay" at 0xbbC2813bE1723aF6021c96F29b4DfaF1C1b2d0db
executing LightRelay.genesis (tx: 0x68ba75d71a420a464550676fc50445ac51a2c786751afc8dcc199f16b2f10b84) ...: performed with 98727 gas
executing LightRelay.retarget (tx: 0x8b4a6dfdca35384fea4697164ca8facefbb38ddd58037ca71c7a64676b15b783) ...: performed with 234196 gas
```

The difficulty returned by the contract after those two operations is `34244331613176` for the previous epoch and 
`35364065900457` for the current epoch. This is in line with tBTC v1 [mainnet relay](https://etherscan.io/address/0x1531b6e3d51BF80f634957dF81A990B92dA4b154#readContract) contract returning `35364065900457` as the previous epoch difficulty, checked on block `16340164`.

There is one unresolved comment - I believe `func.skip` [condition is wrong](https://github.com/keep-network/tbtc-v2/pull/443#discussion_r1062307614). Other than that, I am good to merge.

solidity/deploy/21_genesis_relay.ts
tomaszslabon: Done in d13fe6ff3af383afd56e3fa7c13b6fef8f2c252f.

pdyraga (approved):

PR #421, opened 2022-11-21
Fixing tBTCv2 system tests

dimpar:
This PR updates/fixes tBTCv2 integration and system tests. 

- Fix prerequisites by adding the correct test setup to run the tests. This includes some changes in the `system-test.yml` and the deployment scripts
- Introducing an actual Random Beacon contract with available groups to perform work for the integration test
- Run the full circle of deposit and redemption (no TBTC Vault / no minting)

PR #444, opened 2022-12-29
Reorganize System Test CI workflow

nkuba:
Here we extract scripts to separate steps for a better view of the job execution in the CI.

Depends on https://github.com/keep-network/tbtc-v2/pull/421


michalinacienciala (commented):
Looks good to me. The only comment I have (which is not strictly related to this PR) is that I don't see why do we need the `Cleanup for fresh installation` step. Each GH Actions job is run in a separate runner so there shouldn't be anything to clean at that early stage of the job...

dimpar:
> Looks good to me. The only comment I have (which is not strictly related to this PR) is that I don't see why do we need the `Cleanup for fresh installation` step. Each GH Actions job is run in a separate runner so there shouldn't be anything to clean at that early stage of the job...

I don't remember if there was a specific error that resulted in adding this step, but frequently dealing with some outdated cached data or dependencies in `node_modules` it might be that this step was added simply to avoid any errors that can be potentially thrown because of that. However, you made a good point that if this is run in a separate new runner, then I think we can try removing it and see what happens :)

dimpar:
Latest run of system tests passed https://github.com/keep-network/tbtc-v2/actions/workflows/system-tests.yml

michalinacienciala (approved):

PR #445, opened 2022-12-29
TransactionHash as Hex class

nkuba:
Handling TransactionHash as a string causes a lot of confusion when integrating with the `tbtc-v2.ts` library. To be more explicit we introduce a class that expects a hexadecimal string (either prefixed or not). The class exposes functions to output the value as a prefixed or unprefixed string and gives a way to easily reverse the hex.

### DIFF
--- a/typescript/src/hex.ts
+++ b/typescript/src/hex.ts
@@ -4,7 +4,7 @@
 export class Hex {
   protected readonly _hex: Buffer
 
-  constructor(value: Buffer | string) {
+  protected constructor(value: Buffer | string) {
     if (typeof value === "string") {
       if (!value.match(/^(0x|0X)?[0-9A-Fa-f]*$/)) {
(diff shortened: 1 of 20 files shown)


typescript/src/hex.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * Represents a hexadecimal value.
+ */
+export class Hex {
+  protected readonly _hex: Buffer
+
+  constructor(value: Buffer | string) {
```
lukasz-zimnoch, line 7: Looks like we use the `from` factory method to create new `Hex` instances across the codebase. Maybe we can make the constructor `private` to force a uniform way of creating `Hex` instances for the future?

nkuba: a987ecfe39ab79e9168a6f330c4f142a48f59243

lukasz-zimnoch (approved):

Issue #437, opened 2022-12-21
Disable redemptions for the initial release


pdyraga:
For the January release based on optimistic minting, the off-chain client will not have the functionality allowing to perform a redemption. For this reason, redemption should be disabled in smart contracts.

Probably, the easiest way to achieve it is to set the redemption dust threshold to the max value allowed by the type.

Note that it is just about confirming that the proposed solution works as expected and preparing an extra deployment step for the mainnet deployment.

Issue #438, opened 2022-12-21
v2 Bridge contracts deployment


pdyraga:
Team | Address
------------ | -------------
**Deployment Team** | `0x123694886DBf5Ac94DDA07135349534536D14cAf`
**Threshold Council** | `0x9f6e831c8f8939dc0c830c6e492e7cef4f9c2f5f`

# Prerequisites
- #431
- #433
- #434
- #436
- #437
- [ ] Transfer `VendingMachine@0x6590DFF6abEd7C077839E8227A4f12Ec90E6D85F` ownership from Keep Community multisig to Threshold Council (tx: `0x...`)

# Tag the new version
All steps done by the Deployment Team
- [ ] Tag `solidity/v1.0.0` in `keep-network/tbtc-v2`
- [ ] Sign `solidity/v1.0.0` commit hash

# Deploy and initialize the difficulty relay
All steps done by the Deployment Team
- [ ] Deploy `LightRelay` contract (address: `0x...`)
- [ ] Perform the `genesis` with a previous Bitcoin epoch using `genesisProofLength` of 20 blocks (tx: `0x...`)
- [ ] `retarget` the relay with the current Bitcoin epoch (tx: `0x...`)
- [ ] Set the authorization status to `true` (tx: `0x...`)

# Deploy v2 bridge contracts
All steps done by the Deployment Team
- [ ] Deploy `Bank` contract (address: `0x...`)
- [ ] Deploy `Bridge` implementation contract (address: `0x...`)
- [ ] Deploy `Bridge` proxy contract (address: `0x...`)
- [ ] Deploy `Bridge` `ProxyAdmin` contract (address: `0x...`)
- [ ] Initialize `Bridge` proxy contract (tx: `0x...`) with the following parameters: 
  - `_bank`: `Bank.address`
  - `_relay`: `LightRelay.address`
  - `_treasury`: `0x9f6e831c8f8939dc0c830c6e492e7cef4f9c2f5f` (Threshold Council or Treasury Guild?)
  - `_ecdsaWalletRegistry`: `0x46d52E41C2F300BC82217Ce22b920c34995204eb`
  - `_reimbursementPool`: `0x8adF3f35dBE4026112bCFc078872bcb967732Ea8`
  - `_txProofDifficultyFactor`: `6`
- [ ] Deploy `TBTCVault` contract (address: `0x...`) with the following parameters:
  - `_bank`: `Bank.address`
[...]

PR #448, opened 2023-01-02
Deployment step for disabling redemptions on mainnet

lukasz-zimnoch:
Depends on: #446 
Closes: #437

For the initial release based on optimistic minting, the off-chain client will not have the functionality allowing to perform redemptions. For this reason, we need to temporarily disable redemptions in the `Bridge` contract. We achieve that by adding an additional step to the deployment scripts. This step is invoked for `mainnet` network only and sets the redemption dust threshold to the maximum value of the `uint64` type, the redemption timeout to the maximum value of the `uint32` type, and the redemption slashing amount to zero.


solidity/deploy/07_disable_redemptions.ts
```diff
@@ -0,0 +1,40 @@
...
+  const { execute, read } = deployments
+  const { deployer } = await getNamedAccounts()
+
+  deployments.log("disabling redemptions in the Bridge")
+
+  // To disable redemptions, we need to make the redemption dust threshold
+  // extremely high. The redemption dust threshold value is determined by the
+  // redemptionDustThreshold governance parameter. This parameter is uint64.
+  // The maximum value for the uint64 is 2^64-1 = 18446744073709551615.
+  const redemptionDustThreshold = ethers.BigNumber.from("18446744073709551615")
```
pdyraga, line 15: I do not think this is enough, `2^64-1` in e18 precision is just 18 TBTC. It is not a tiny amount but it's also not that high amount. We need one more thing: what about setting `redemptionTimeout` to the max possible value of `2^32-1 = 4294967295` which is 136 years?

solidity/deploy/07_disable_redemptions.ts
```diff
@@ -0,0 +1,40 @@
...
+  await execute(
+    "Bridge",
+    { from: deployer, log: true, waitConfirmations: 1 },
+    "updateRedemptionParameters",
+    redemptionDustThreshold,
+    redemptionParameters.redemptionTreasuryFeeDivisor,
+    redemptionParameters.redemptionTxMaxFee,
+    redemptionParameters.redemptionTxMaxTotalFee,
+    redemptionParameters.redemptionTimeout,
+    redemptionParameters.redemptionTimeoutSlashingAmount,
```
pdyraga, line 29: Per _Adjust parameters for Chaosnet phase 1_ section of https://github.com/keep-network/tbtc-v2/issues/438, let's set the slashing amount to 0.

solidity/deploy/07_disable_redemptions.ts
pdyraga: I've just merged https://github.com/keep-network/tbtc-v2/pull/446 to unblock this work, but if we decide to set `redemptionTimeout` to the max value allowed by the type, I think we should do the same for `fraudChallengeDefeatTimeout` in `06_disable_fraud_challenges.ts`, for consistency.

solidity/deploy/07_disable_redemptions.ts
lukasz-zimnoch: See d372066d

solidity/deploy/07_disable_redemptions.ts
lukasz-zimnoch: See d372066d and f236d3cc

pdyraga (approved):

PR #449, opened 2023-01-02
Allow only the depositor to reveal a deposit- typescript

r-czajkowski:
Closes: #439 

Adjust the Ethereum implementation of the `revealDeposit` function to changes from https://github.com/keep-network/tbtc-v2/pull/442. The `depositor` field has been removed from `DepositRevealInfo` struct and the `revealDeposit` function must be called by the same Ethereum address as the one used in the `P2(W)SH` BTC deposit transaction as a depositor.


nkuba (approved):
I tested it locally in system tests to reveal a deposit and it worked fine.

Issue #434, opened 2022-12-21
Disable moving funds between wallets for the initial release


pdyraga:
For the January release based on optimistic minting, the off-chain client will not have the functionality allowing to move funds to another wallet. For this reason, the moving funds feature should be disabled in smart contracts.

Not confirmed but worth checking if this would be enough: set `walletMaxAge` to a high value and `walletClosureMinBtcBalance` to zero.

Note that it is just about confirming that the proposed solution works as expected and preparing an extra deployment step for the mainnet deployment.

pdyraga:
Not confirmed but worth checking if enough: set `walletMaxAge` to a high value and `walletClosureMinBtcBalance` to a low value.

pdyraga:
Note that it may be just about implementing a unit test confirming the expected behavior and setting the values during the deployment.

PR #450, opened 2023-01-02
Deployment step for disabling moving funds on mainnet

lukasz-zimnoch:
Depends on: #448
Closes: #434

For the initial release based on optimistic minting, the off-chain client will not have the functionality allowing to perform moving funds. For this reason, we need to temporarily disable moving funds in the `Bridge` contract. We achieve that by adding an additional step to the deployment scripts. This step is invoked for `mainnet` network only and sets:
- The `walletMaxAge` to the maximum value of the `uint32` type
- The `walletClosureMinBtcBalance` to the lowest possible value (`1`)
- The `movingFundsTimeout` to the maximum value of the `uint32` type
- The `movingFundsTimeoutSlashingAmount` to zero


solidity/deploy/08_disable_moving_funds.ts
```diff
@@ -0,0 +1,87 @@
...
+  // very low probability.
+  //
+  // The third case is the only one relevant for this script. A wallet cannot
+  // become "closeable" if both of the following is true:
+  // - walletMaxAge is set to the maximum value allowed for uint32 type (2^32-1 = 4294967295)
+  // - walletClosureMinBtcBalance is zero
+  //
+  // In practice, walletClosureMinBtcBalance's minimum value enforced by the
+  // contract is 1. That should do the trick because only non-active wallets
+  // with 0 BTC (no funds to move) can become "closeable" in that case.
```
pdyraga, line 34: ...and we can close them manually by calling `notifyMovingFundsBelowDust` without moving funds because they have less than `movingFundsDustThreshold`, correct?

Would it make sense to write it down here?

pdyraga (approved):
Just one comment/question, we can address it in another PR.

Issue #431, opened 2022-12-21
Determine minting and redemption fee structure


pdyraga:
CoinList charges a flat 0.25% trading fee to wrap or unwrap WBTC (https://coinlist.co/help/what-are-the-fees-to-wrap-and-unwrap-wbtc). If we want to maintain the same fee, for the January release, we have two options:

- If the treasury should receive the fee at the moment of sweeping, set the Deposit fee to 0.25% and the Optimistic Minting fee to 0%.
- If the treasury should receive the fee at the moment of finalizing the Optimistic Mint, set the Optimistic Mint fee to 0.25% and the Deposit Fee to 0%.

To be done here:
- [x] @mhluongo and @pdyraga to determine the fee structure
- [x] @lukasz-zimnoch and @tomaszslabon to prepare an extra mainnet deployment script step to update fees

mhluongo:
My default: Let's start with an optimistic minting fee of 0.2% (rather than 0.25%) and a 0% deposit fee.

If this means we'll be slightly undercollateralized on the BTC side after a sweep due to Bitcoin network fees, we can burn the excess tBTC to make up the difference when sweeping launches.

PR #451, opened 2023-01-03
Deployment step for setting initial deposit fees on mainnet

lukasz-zimnoch:
Depends on: #450
Closes: #431

For the initial release based on optimistic minting, we want to use the following fees:
- optimistic minting fee of `0.2%`
- deposit treasury fee of `0%`


lukasz-zimnoch:
The only way of updating the optimistic fee divisor in the `TBTCVault` is through `beginOptimisticMintingFeeUpdate` and `finalizeOptimisticMintingFeeUpdate` calls. That means the finalization must be done manually once the governance delay elapses. @pdyraga let me know if we are good with that and if so, I'll update https://github.com/keep-network/tbtc-v2/issues/438 to reflect the manual finalization requirement.

pdyraga:
> The only way of updating the optimistic fee divisor in the `TBTCVault` is through `beginOptimisticMintingFeeUpdate` and `finalizeOptimisticMintingFeeUpdate` calls. That means the finalization must be done manually once the governance delay elapses. @pdyraga let me know if we are good with that and if so, I'll update #438 to reflect the manual finalization requirement.

Can we instead set the default value in `TBTCOptimisticMinting`? I think it shouldn't break bridge tests because they do not use the optimistic minting. The only tests that may need an update are those in `TBTCVault.OptimisticMinting.test.ts`.

lukasz-zimnoch:
> > The only way of updating the optimistic fee divisor in the `TBTCVault` is through `beginOptimisticMintingFeeUpdate` and `finalizeOptimisticMintingFeeUpdate` calls. That means the finalization must be done manually once the governance delay elapses. @pdyraga let me know if we are good with that and if so, I'll update #438 to reflect the manual finalization requirement.
> 
> Can we instead set the default value in `TBTCOptimisticMinting`? I think it shouldn't break bridge tests because they do not use the optimistic minting. The only tests that may need an update are those in `TBTCVault.OptimisticMinting.test.ts`.

See 013b0622

pdyraga (approved):

Issue #432, opened 2022-12-21
Determine and adjust deposit lock time


pdyraga:
We need to determine and update the deposit lock time. 6mo feels like a good value but we could consider increasing it to 9mo/12mo in case of any unexpected issues after the deployment.

Steps that need to be taken:
- [x] Determine the lock time (product question) - @mhluongo @pdyraga 
- [x] Update the lock time in smart contracts - @lukasz-zimnoch @tomaszslabon (#452)
- [x] Update the lock time in the typescript library - @michalsmiarowski @lukasz-zimnoch (#452)
- [x] Reflect the change in the dashboard if needed (update copy) - @michalsmiarowski

mhluongo:
Let's launch with a 9 month lock time in case the backlog of sweeps is larger than we're anticipating.

lukasz-zimnoch:
@r-czajkowski @michalsmiarowski heads-up regarding changes on the dApp front. In #452, we modified the `calculateDepositRefundLocktime` function by adding an optional `depositRefundLocktimeDuration`. Please keep in mind that you need to provide a proper `depositRefundLocktimeDuration` value while computing the `refundLocktime` hex for `calculateDepositAddress`. The proper value in the context of contract changes presented in #452 is `21772800`. This is 9 months in seconds, assuming 1 month = 4 weeks = 28 days. Then, you must reveal a deposit within 14 days, counting from the timestamp you use as deposit creation.

PR #452, opened 2023-01-03
Set the initial deposit locktime

lukasz-zimnoch:
Depends on: #451
Refs: #432

For the initial release with optimistic minting, we want to use a deposit lock time of 9 months. Here we adjust that on the contract and typescript library side.

On the typescript front, we adjust the utility function `calculateDepositRefundLocktime` to optionally accept other lock time duration than the default 30 days. This function along with `calculateDepositAddress` will allow the dApp to produce deposit addresses that embed 9 months as the lock time. 

On the contract side, we set the `depositRevealAheadPeriod` to 8 months and 2 weeks (assuming 1 month = 30 days). This parameter enforces the latest possible time to make the reveal. Assuming the dApp produces deposit addresses with 9 months as the lock time, that means there are 2 weeks to reveal the deposit. After that time, the deposit must wait until the deposit becomes refundable, without the possibility to mint/sweep it earlier. This requirement must be emphasized by the dApp.


typescript/src/deposit.ts
```diff
@@ -14,11 +14,11 @@ import { Bridge, Identifier } from "./chain"
...
 
 /**
- * Duration of the deposit refund locktime in seconds. After that time, the
- * depositor can make a refund of an unswept deposit using the refund public
+ * Default duration of the deposit refund locktime in seconds. After that time,
+ * the depositor can make a refund of an unswept deposit using the refund public
  * key.
  */
-export const DepositRefundLocktimeDuration = 2592000 // 30 days
+export const DefaultDepositRefundLocktimeDuration = 2592000 // 30 days
```
pdyraga, line 21: Shouldn't we set the default to 90 days instead?

lukasz-zimnoch: A lot of tests rely on 30 days. Using an optional value is also more flexible. Last but not least, 90 days is rather temporary.

pdyraga: How big is the impact on tests? 

I think we can use an optional value but we should set the default to what is currently expected by the system. The smart contract will not allow revealing anything that is 8 months and 2 weeks ahead of the unlock. 

8 months and 2 weeks is a value for the next few months, it's is not that temporary.

If we leave this as 30 days and optional, external devs using `tbtc-v2.ts` will produce non-revealable deposits from their applications. The same will happen in case we forget to use a non-default value in the dApp.

solidity/deploy/10_set_deposit_params.ts
```diff
@@ -10,6 +10,11 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
 
   const depositTreasuryFeeDivisor = ethers.BigNumber.from("0")
 
+  // We set the deposit reveal ahead period to 8 months and two weeks, assuming
+  // 1 month = 4 weeks = 28 days. That gives 238 days which translates to
+  // 238 * 24 * 60 * 60 = 20563200 seconds
+  const depositRevealAheadPeriod = ethers.BigNumber.from("20563200")
```
pdyraga, line 16: Can we assume one month has 30 days? This will give us two weeks extra (254 days)
```
(30 * 8 + 14) * 24 * 60 * 60 = 21945600
```

typescript/src/deposit.ts
pdyraga: >  In that case, I think there should not be any default in `tbtc-v2.ts` and we should always expect an explicit value.

Makes sense!

pdyraga (commented):
Just one comment and I am good to go @lukasz-zimnoch.

solidity/deploy/10_set_deposit_params.ts
lukasz-zimnoch: See 5ef8f99c

pdyraga (approved):

PR #453, opened 2023-01-04
Optimistic minting debt calculation fix

pdyraga:
The optimistic debt is explained in the documentation in the following way:

> The debt represents the total value of all depositor's deposits revealed to the Bridge that has not been yet swept and led to the optimistic minting of TBTC. (...)

This explanation is perfectly correct but it is not how the debt was calculated for a non-zero optimistic minting fee. Previously, the optimistic minting fee was deducted from the debt. This is not correct - the fee is also optimistically minted and goes to the treasury address. All TBTC that is optimistically minted: the part that goes to the depositor and the part that goes to the treasury should be added to the optimistic minting debt. When the deposit is swept, it is paying off both the depositor's part and the treasury's part.

### DIFF
--- a/solidity/contracts/vault/TBTCOptimisticMinting.sol
+++ b/solidity/contracts/vault/TBTCOptimisticMinting.sol
@@ -334,13 +334,12 @@ abstract contract TBTCOptimisticMinting is Ownable {
         uint256 optimisticMintFee = optimisticMintingFeeDivisor > 0
             ? amountToMint / optimisticMintingFeeDivisor
             : 0;
-        amountToMint -= optimisticMintFee;
 
         uint256 newDebt = optimisticMintingDebt[deposit.depositor] +
             amountToMint;
         optimisticMintingDebt[deposit.depositor] = newDebt;
 
-        _mint(deposit.depositor, amountToMint);
+        _mint(deposit.depositor, amountToMint - optimisticMintFee);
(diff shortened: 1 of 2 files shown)


solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -334,13 +334,12 @@ abstract contract TBTCOptimisticMinting is Ownable {
         uint256 optimisticMintFee = optimisticMintingFeeDivisor > 0
             ? amountToMint / optimisticMintingFeeDivisor
             : 0;
-        amountToMint -= optimisticMintFee;
```
lukasz-zimnoch, line 337: I think it's worth making it crystal-clear that the OM fee is actually optimistically minted as well so it must be added to the debt. This is not obvious when reading the code.

lukasz-zimnoch (approved):

solidity/contracts/vault/TBTCOptimisticMinting.sol
pdyraga: Done in  653406ab32b95b8c0c520fb38dccb7d9e64a5e90. Please take a look and let me know if this explanation is enough.

lukasz-zimnoch: LGTM!

PR #457, opened 2023-01-04
TS: Query for DepositRevealed events

nkuba:
In this PR we add a function to query for `DepositRevealed` events emitted by the `Bridge` contract.

Depends on https://github.com/keep-network/tbtc-v2/pull/454


typescript/src/chain.ts
```diff
@@ -18,6 +18,24 @@ export interface Identifier {
   identifierHex: string
 }
 
+/**
+ * Represents a generic chain event.
+ */
+export interface Event {
+  /**
+   * Ethereum block number of the event emission.
```
lukasz-zimnoch, line 26: This is a generic chain event placed in a generic `chain.ts` file. I think we should remove all mentions about "Ethereum":
```suggestion
   * Block number of the event emission.
```

typescript/src/chain.ts
```diff
@@ -18,6 +18,24 @@ export interface Identifier {
...
+   */
+  blockNumber: number
+  /**
+   * Ethereum block hahs of the event emission.
```
lukasz-zimnoch, line 30:
```suggestion
   * Block hash of the event emission.
```

typescript/src/chain.ts
```diff
@@ -18,6 +18,24 @@ export interface Identifier {
...
+   */
+  blockHash: string
+  /**
+   * Ethereum transaction hash within which the event was emitted.
```
lukasz-zimnoch, line 34:
```suggestion
   * Transaction hash within which the event was emitted.
```

typescript/src/deposit.ts
```diff
@@ -98,6 +98,19 @@ export type RevealedDeposit = Pick<
...
+ */
+export type DepositRevealedEvent = Pick<
+  Deposit,
+  "depositor" | "amount" | "blindingFactor" | "refundLocktime" | "vault"
+> & {
+  fundingTxHash: TransactionHash
+  fundingOutputIndex: number
+  walletPubKeyHash: string
+  refundPubKeyHash: string
+} & Event
```
lukasz-zimnoch, line 112: This can be simplfied:
```suggestion
export type DepositRevealedEvent = Deposit & {
  fundingTxHash: TransactionHash
  fundingOutputIndex: number
} & Event
```
Note that `*publicKeyHash` is widely used in the library. The contract uses an abbreviation (`*pubKeyHash`) but this can be easily mapped in `ethereum.ts` and not affect the library.

typescript/src/ethereum.ts
```diff
@@ -126,6 +157,49 @@ export class Bridge extends EthereumContract implements ChainBridge {
...
 
+  /**
+   * Query emitted DepositRevealed events.
+   * @param fromBlock Block number from which events should be queried.
+   * @param toBlock Block number to which events should be queried.
+   * @param filterArgs Arguments for events filtering.
+   * @returns Found DepositRevealed events.
+   * @see queryEvents
+   */
+  async queryDepositRevealedEvents(
```
lukasz-zimnoch, line 168: How do we plan to expose that outside? Note that we already have that in the `Bridge` interface:
```
  deposits(
    depositTxHash: TransactionHash,
    depositOutputIndex: number
  ): Promise<RevealedDeposit>
```
We have also the `RevealedDeposit` type in `deposit.ts`. If we plan to enhance the `Bridge` with:
```
queryDepositRevealedEvents(...): Promise<DepositRevealedEvent[]>
```
Then, the `Bridge` interface will become unclear because we will not have clarity when use the`deposits` method and when `queryDepositRevealedEvents`. The difference is basically an implementation detail: the Ethereum `Bridge` stores more information in the event to be more gas-efficient. Shouldn't we hide that detail in `ethereum.ts` implementation and expose a single method in the `Bridge` interface that will return all the required information fetched directly from the contract state and event using `RevealedDeposit` with additional fields? In other words: does the outside world that uses the `Bridge` interface needs to know that we store the deposit data in two different places (storage and events)? What do you think?

r-czajkowski: > Shouldn't we hide that detail in ethereum.ts implementation and expose single method in the Bridge interface that will return all the required information fetched directly from the contract state and event using RevealedDeposit with additional fields?

I agree but would be great if we could get all deposits by depositor address(we need to find all deposits by a depositor in T dapp). So maybe we could expose 2 methods in the `Bridge` interface:
- `deposits(depositor: Identifier): Promise<RevealedDeposit[]>`- to get all deposits by depositor,
- `deposit( depositTxHash: TransactionHash, depositOutputIndex: number):  Promise<RevealedDeposit>` -to get a single deposit.

```ts
deposit(
 depositTxHash: TransactionHash,
 depositOutputIndex: number
): Promise<RevealedDeposit> {
 const depositKey = Bridge.buildDepositKey(depositTxHash, depositOutputIndex)
 
 const deposit = await this._instance.deposits(depositKey)
 
 // It will return all `DepositRevealed` events so we must find the exact one
 // by comparing `depositTxHash` and `depositOutputIndex` and then parse data
 // to include fields from the event.
 const event = (await this.queryEvents(
  "DepositRevealed",
  undefined,
  undefined,
  [deposit.depositor]
 )).find(...)
 
 return this.parseRevealedDeposit(deposit)
}

deposits(depositor: Identifier): Promise<RevealedDeposit[]> {
 const events = await this.queryEvents(
  "DepositRevealed",
  undefined,
  undefined,
  [deposit.depositor]
 )
 
 // Get the all data from `BridgeContract.deposits` function for each deposit
 const deposits = await Promise.all(
      events.map((event) =>
        Bridge.buildDepositKey(
          event.args.fundingTxHash,
          event.args.fundingOutputIndex
        )
      ).map(this._instance.deposits)
    )
  
 // parse all deposits to RevealedDeposit.
 return parsedDeposits
}
```

The problem could probably be that we can't find the `DepositRevealed` event by `depositTxHash` and `depositOutputIndex` in the `deposit` function because those params are not indexed so we need to iterate through all events to find that exact one. See:
https://github.com/keep-network/tbtc-v2/blob/88bf26f91a232dcd2dd1e7707666dc7674572629/solidity/contracts/bridge/Deposit.sol#L105-L106

Another problem may be in the `deposits` function because of the call to the contract(`BridgeContract.deposits`) for each deposit to get fields from the [`DepositRequest` struct](https://github.com/keep-network/tbtc-v2/blob/main/solidity/contracts/bridge/Deposit.sol#L81).

lukasz-zimnoch: Sure, we need to take filtering into account. I just wanted to propose a general idea. However, I think we should strongly lean towards keeping the interface as simple as possible. In the case mentioned by @r-czajkowski it would be better to have a single method in the `Bridge`:
```
deposits(filter Filter): Promise<RevealedDeposit[]>
```
In that case, the `Filter` object should allow the construction of an arbitrary filter to either fetch a single deposit using its funding transaction data or multiple deposits for the given depositor (or other combinations supported by the implementation). Having two separate methods:
```
deposits(depositor: Identifier): Promise<RevealedDeposit[]>
deposit( depositTxHash: TransactionHash, depositOutputIndex: number):  Promise<RevealedDeposit>
```
is not an optimal solution IMO. If we need to add an additional filter or change the existing one, we need to change the interface. It is always better to encapsulate such details in the implementation so they do not affect the client code.

typescript/src/chain.ts
nkuba: f816ce56c496c34b79d7fab46bf555c1688dc046

typescript/src/chain.ts
nkuba: f816ce56c496c34b79d7fab46bf555c1688dc046

typescript/src/deposit.ts
nkuba: d855431d690746b7b0e60ab57b12d376890010c7

typescript/src/ethereum.ts
nkuba:
```
  deposits(
    depositTxHash: TransactionHash,
    depositOutputIndex: number
  ): Promise<RevealedDeposit>
```
`deposits` function requires `depositKey` to be provided and it returns only a single specific `RevealedDeposit`. I find the name of the function `deposits` confusing and I'd like to propose we rename it to `getDeposit`.

We want to fetch all the events emitted by the contract in one call.
Additionally, we want to give a developer using this library the flexibility to filter events based on indexed fields (what @r-czajkowski mentioned) or based on a range of blocks (e.g. to allow incremental data gathering over time).

lukasz-zimnoch: > I find the name of the function deposits confusing and I'd like to propose we rename it to getDeposit.

Yeah, that's a good idea!

Regarding the rest, I think we need to stop and decide what we expect from the `Bridge` interface (and its Ethereum implementation) first. We basically have two possible approaches. 

**Approach 1**

This is the one I mentioned above. In this approach, the implementation of the `Bridge` handles as much of details as possible. The outside world does not know about the internals of the contract, for example, it uses both storage and events to hold deposit data. In this case, the `Bridge` exposes either
```
getDeposits(filter: Filter): Promise<RevealedDeposit[]>
```
or:
```
getDepositByTx(depositTxHash: TransactionHash, depositOutputIndex: number): Promise<RevealedDeposit>
getDepositsByDepositor(depositor: Identifier): Promise<RevealedDeposit[]>
// etc.
```
whichever is easier to implement. This approach should not talk about events as we treat them as an implementation detail. We talk about "deposits". If we need to listen for revealed deposits, we should expose something like:
```
onDeposit(handler: (deposit: RevealedDeposit) => void)
```
This approaches pushes all the complexity to the Ethereum `Bridge` implementation but produces a simpler chain API that can be used by clients.

**Approach 2** 

This is an approach similar to what we have in Go. We consider the `Bridge` interface as a mapping of the contract API and the Ethereum implementation is something similar to the Go contract binding. The `Bridge` interface should expose the same functions as the solidity contract and additional ones used to work with events:
```
getDeposit(depositTxHash: TransactionHash, depositOutputIndex: number): Promise<RevealedDeposit>
getDepositRevealedEvents(...): Promise<DepositRevealedEvent[]>
onDepositRevealedEvent(handler: (event: DepositRevealedEvent) => void)
// etc.
```
This approach is more low-level and it exposes some additional complexity outside. The client code must know when to fetch deposits from storage (using `getDeposit`) and when from events (using `getDepositRevealedEvents`). It must also combine both when needed on its own. If we want to be better, we can add a thin layer in the `tbtc-v2.ts` library itself (for example, additional functions in the `deposit.ts` file) that will handle that complexity and will expose a simpler API using `index.ts`. After all the `Bridge` interface is not supposed to be used directly by external clients but should be injected as an argument to the public `tbtc-v2.ts` library functions exposed by `index.ts`.

Please think about it for a while as this is an important design decision that will affect the future development of the library and client code based on it. Let's choose one approach and be consistent with it.

nkuba: I like Approach 2 as it gives flexibility to a developer using the library. This is similar to the tbtc.js library we developed for V1. The library will be a middle layer simplifying communication with the contracts.

lukasz-zimnoch: @r-czajkowski please let us know if approach 2 works for you in the context of dApp development. Do you see any potential problems?

r-czajkowski: Yep approach 2 works for me and I don't see any problems-  at least for now.

lukasz-zimnoch: @nkuba, I'm wondering if we can consider this discussion resolved. Do you plan to align changes presented in this PR to approach 2 or that will be done in the next PRs?

nkuba: Yes, thank you gents for the discussion. I'll incorporate the changes into this PR.

typescript/src/ethereum.ts
```diff
@@ -126,6 +157,49 @@ export class Bridge extends EthereumContract implements ChainBridge {
...
+      toBlock,
+      ...filterArgs
+    )
+
+    return events.map<DepositRevealedEvent>((event) => {
+      return {
+        blockNumber: BigNumber.from(event.blockNumber).toNumber(),
+        blockHash: event.blockHash,
+        transactionHash: event.transactionHash,
+        fundingTxHash: TransactionHash.from(event.args!.fundingTxHash),
```
lukasz-zimnoch, line 185: I think we should reverse that hash. The hash in the emitted event is in the native Bitcoin byte order but `TransactionHash` represents the opposite order (https://github.com/keep-network/tbtc-v2/blob/2b8297e3991b73f10a9923546d1fcb66752535d8/typescript/src/bitcoin.ts#L15) and this is how the client code interprets it.

typescript/src/ethereum.ts
nkuba: Two changes:
9caf61bb7d2809ce58fdeadede8371895c21b4b3
daa60edf7bd4e190d8d6f91aebe1e652c06bbc4c

typescript/src/ethereum.ts
nkuba: Great catch! 5986a62d46d0fd0325d3d0b5c4cd5dbc10db1d29

typescript/src/ethereum.ts
```diff
@@ -132,6 +163,47 @@ export class Bridge
...
+        depositor: new Address(event.args!.depositor),
+        amount: BigNumber.from(event.args!.amount),
+        blindingFactor: event.args!.blindingFactor,
+        walletPublicKeyHash: event.args!.walletPubKeyHash,
+        refundPublicKeyHash: event.args!.refundPubKeyHash,
+        refundLocktime: event.args!.refundLocktime,
+        vault:
+          event.args!.vault === constants.AddressZero
+            ? undefined
+            : new Address(event.args!.vault),
```
lukasz-zimnoch, line 202: Just to make sure: Are we sure that all values that come from the event conform to the specification of the `Deposit` fields as is? For example, are the byte values unprefixed?

lukasz-zimnoch: Yep, we need a uniform way to represent those values. We started with unprefixed strings as this was the best way to deal with such values in the code related to Bitcoin (deposit script etc.). Now, we have multiple parts that need different formats so we should definitely improve that. Can we leave a TODO near the `Deposit` type to track this?

lukasz-zimnoch: Ah ok, you already did it in https://github.com/keep-network/tbtc-v2/commit/13508d21680cc22c62bacefe5c6fc05069816aef. Good!

lukasz-zimnoch (approved):

PR #460, opened 2023-01-04
TS: Optimistic Minting

nkuba:
This PR introduces changes necessary for the TBTCv2 Maintainer.

Depends on:
https://github.com/keep-network/tbtc-v2/pull/454
https://github.com/keep-network/tbtc-v2/pull/457
https://github.com/keep-network/tbtc-v2/pull/458
https://github.com/keep-network/tbtc-v2/pull/459


typescript/src/ethereum.ts
```diff
@@ -562,3 +571,86 @@ class WalletRegistry extends EthereumContract {
     return publicKey.substring(2)
   }
 }
+
+/**
+ * Implementation of the Ethereum TBTCVault handle.
+ */
+export class TBTCVault extends EthereumContract implements ChainTBTCVault {
```
r-czajkowski, line 578: I believe we should refactor the file structure. It's hard to navigate through code if we keep all implementations in one file. I think we could create a dir for each chain and then create dirs for each handle. For example:
```
└── ethereum/
    ├── index.ts // Export all from bridge, vault etc.
    ├── bridge/
    │   └── index.ts // Ethereum implementation of the Bridge handle
    ├── contract/
    │   └── index.ts // Common impl of the Ethereum contract
    └── vault/
        └── index.ts // Ethereum implementation of the Vault handle
```

Ofc we can address it in a separate PR just flagging.

typescript/src/optimistic-minting.ts
```diff
@@ -0,0 +1,52 @@
...
+  depositKey: BigNumber
+  depositor: Identifier
+  amount: BigNumber
+  fundingTxHash: TransactionHash
+  fundingOutputIndex: number
+} & Event
+
+export type OptimisticMintingRequest = {
+  requestedAt: number
+  finalizedAt: number
```
r-czajkowski, line 16: Can we explain in a comment what `0` value means?

https://github.com/keep-network/tbtc-v2/blob/15be27edf4d852a284cf0a9ae5b9f422ef8d2cab/solidity/contracts/vault/TBTCOptimisticMinting.sol#L40-L42

typescript/src/ethereum.ts
nkuba: Yeah, we need refactoring!

lukasz-zimnoch: We should think about refactoring but I'm not sure if the proposed file structure with multiple nestings will be easier to follow than one file with several hundreds of lines of code. I think we can group files but aim for a more flat and simpler structure:
```
ethereum
    bridge.ts
    vault.ts
    index.ts
```

typescript/src/ethereum.ts
```diff
@@ -562,3 +571,86 @@ class WalletRegistry extends EthereumContract {
...
+    request: any
+  ): OptimisticMintingRequest {
+    return {
+      requestedAt: BigNumber.from(request.requestedAt).toNumber(),
+      finalizedAt: BigNumber.from(request.finalizedAt).toNumber(),
+    }
+  }
+
+  // TODO: Add docs
+  // TODO: Consider exposing in optimistic-minting.ts and in chain.ts
```
lukasz-zimnoch, line 627: Just a note that we need to make it in the same way as agreed during [this discussion](https://github.com/keep-network/tbtc-v2/pull/457#discussion_r1062288531)

typescript/src/index.ts
```diff
@@ -20,11 +25,19 @@ const TBTC = {
   getRevealedDeposit,
   requestRedemption,
   getRedemptionRequest,
+  requestOptimisticMint,
+  getOptimisticMintingRequest,
```
lukasz-zimnoch, line 29: I'm wondering about exposing all optimistic-minting-related functions in a separate namespace. Initially, the plan was:
- `TBTC` namespace exposes all functions relevant for regular TBTC system clients interested in deposits and redemptions (e.g. tBTC dApp)
- `SpvMaintainer` targets the maintainers of the system that will be responsible for submitting SPV proofs of Bitcoin transactions performed by the wallets

The `requestOptimisticMint` and `getOptimisticMintingRequest` will be used by Minters and Guardians so they don't quite match the `TBTC` namespace. I think we have options like:

 **Option 1**
```
const OptimisticMinting = {
    requestOptimisticMint
    getOptimisticMintingRequest
    // other functions specific to both Minters and Guardians
}
```

**Option 2**
```
const Minter = {
    requestOptimisticMint
    getOptimisticMintingRequest
    // other functions specific to Minters
}

const Guardian = {
    // functions specific to Guardians
}
```
What do you think?

nkuba: It makes sense. I prefer option #1.

typescript/src/optimistic-minting.ts
nkuba: 695ef2d30742d6a3910607fc54ea8f2c9a430203

typescript/src/ethereum.ts
```diff
@@ -562,3 +571,86 @@ class WalletRegistry extends EthereumContract {
...
+  /**
+   * Parses a optimistic minting request using data fetched from the on-chain contract.
+   * @param request Data of the optimistic minting request.
+   * @returns Parsed optimistic minting request.
+   */
+  private parseOptimisticMintingRequest(
+    request: any
+  ): OptimisticMintingRequest {
+    return {
+      requestedAt: BigNumber.from(request.requestedAt).toNumber(),
```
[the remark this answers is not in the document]
nkuba: We wrap it to make sure the type returned from the chain is exactly what we expect.

typescript/src/ethereum.ts
lukasz-zimnoch: @nkuba what about this one? Do we plan to address it in this PR or separately? In the case of the latter, can be leave a TODO?

typescript/src/index.ts
```diff
@@ -27,4 +31,13 @@ const SpvMaintainer = {
...
+export {
+  Bitcoin,
+  TBTC,
+  SpvMaintainer,
+  OptimisticMinting,
+  ElectrumClient,
+  Ethereum,
+  EthereumBridge,
+  EthereumTBTCVault,
+}
```
lukasz-zimnoch, line 43: Although it seems convenient now, I'm afraid that exposing the whole modules (for example, `Bitcoin` and `Ethereum` here) without any control will make the API of the library unusable in the long term and will cause a lot of confusion for users. We should rather aim for a concise and well-defined interface oriented around specific use cases. 

So far the approach was:
- `TBTC` exposes everything that is needed to interact with the tbtc contracts as an user (dApp)
- `SpvMaintainer` exposes everything needed by SPV maintainers
- `ElectrumClient` is the reference Electrum-based implementation of the `bitcoin.Client`
- `EtheremBridge` is the reference Ethereum implementation of the `chain.Bridge

If we start to expose the whole `ethereum` and `bitcoin` modules, one day we will also expose the entire `deposit` module, `redemption` module, and so on. This will produce a lot of mess as we will expose the same methods by multiple namespaces. For example, we will have `TBTC. suggestDepositWallet` and `Deposit.suggestDepositWallet` - library users won't know which one should they choose.

On the other hand, it is clear that the library grows and we need to expose more than we expose now, especially some types in order to avoid importing them from `dist` on the library client side. Assuming we want to keep the API oriented around use cases and expose types in a clear way maybe we should do the following:

```
// We avoid the 'import * as X from "module.ts"' pattern

// functions for TBTC users
const TBTC = {
  calculateDepositAddress,
  suggestDepositWallet,
  revealDeposit,
  getRevealedDeposit,
  requestRedemption,
  getRedemptionRequest,
}

// functions for SPV maintainers
const SpvMaintainer = {
  submitDepositSweepProof,
  submitRedemptionProof,
}

// functions for optimistic minting maintainers
const MintingMaintainer = {
 (...)
}

// only types, no functions! We can also name it as 'Primitives' or so
const Types = {
    Bitcoin: {
        Client,
        Transaction,
        // other relevant Bitcoin primitives
    }

    Chain: {
        Bridge,
        Identifier
        // other Chain primitives that can be useful for writing implementations
    }

    Electrum: {
        Credentials,
        Client
        // other relevant Electrum types
     }

     Ethereum: {
         Address,
         Bridge,
         TBTCVault,
         // other relevant Ethereum types
     }
}

export { 
    TBTC, 
    SpvMaintainer, 
    MintingMaintainer,
    Types
}
```
What do you think @nkuba? Also curious about @pdyraga thoughts as he originally proposed the approach we use currently.

typescript/src/wallet.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/**
+ * Represents a deposit.
+ */
+export interface Wallet {
+  /**
+   * Identifier of a ECDSA Wallet registered in the ECDSA Wallet Registry.
+   */
+  ecdsaWalletID: string
```
lukasz-zimnoch, line 60: We should probably use `Hex` as type.

typescript/src/wallet.ts
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Compressed public key of the ECDSA Wallet.
+   */
+  walletPublicKey: Hex
+  /**
+   * Latest wallet's main UTXO hash.
+   */
+  mainUtxoHash: string
```
lukasz-zimnoch, line 68: We can use `Hex` here.

typescript/src/wallet.ts
```diff
@@ -0,0 +1,98 @@
...
+   */
+  closingStartedAt: number
+  /**
+   * Total count of pending moved funds sweep requests targeting this wallet.
+   */
+  pendingMovedFundsSweepRequestsCount: number
+  /**
+   * Current state of the wallet.
+   */
+  state: number
```
lukasz-zimnoch, line 93: Shouldn't we use the `WalletState` enum as type?

typescript/src/wallet.ts
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Moving funds target wallet commitment submitted by the wallet.
+   */
+  movingFundsTargetWalletsCommitmentHash: string
```
lukasz-zimnoch, line 97: Can be `Hex`

lukasz-zimnoch (changes requested):
Not looked at the tests yet. Will make a pass over it once all comments regarding the code are addressed.

typescript/src/ethereum.ts
nkuba: 1edcb3960228d1313fea8341b28c8f773a8ae21b

lukasz-zimnoch: Ok, leaving this discussion unresolved for future reference.

typescript/src/index.ts
nkuba: Unfortunatelly, I found the current approach confusing from both developer and user perspectives.

By exporting the whole modules we export only the parts that are marked explicitly with `export`, they are also accessible by direct importing from `dist` files.

I've seen dapp was using direct import from `dist`, as the index was not exporting all necessary things. Also, I've noticed code duplicating in https://github.com/threshold-network/token-dashboard/tree/main/src/threshold-ts or new functions that I would expect to be implemented in this library.

It's hard to decide in a library what an actual user would need, as we cannot predict all the use cases. We should give some flexibility to a user.

I expected this library to work as its' counterpart from V1 (`tbtc.js`) - a proxy to communicate with contracts, wrapping only complex logic or helpers like retries.

I don't want to get into a broad discussion about refactoring in this PR. I'll try to update this PR to match the approach that was previously established and leave the changes for the future.

nkuba: I would also like to hear @r-czajkowski and @michalsmiarowski expectations from the dapp development perspective.

lukasz-zimnoch: > I don't want to get into a broad discussion about refactoring in this PR. I'll try to update this PR to match the approach that was previously established and leave the changes for the future.

Yep definitely. However, we need to discuss it in the near future. There is a strong conflict of interest between multiple use cases and we need to agree on a single approach that will be optimal for everyone. As you said, the best we can do for now is stick to the current approach as much as we can without causing a revolution and return here later. Can be leave an explicit TODO about that in the `index.ts`?

Also, leaving my thoughts regarding other parts for future discussion:

> By exporting the whole modules we export only the parts that are marked explicitly with export, they are also accessible by direct importing from dist files.

Yeah, that's true. However, the fact that something is exported outside of its module does not imply that it should be also exported from the `index.ts` by default. The `index.ts`, just as its name implies, is the place that should define a concise entry point for library users. Exposing everything through `index.ts` by default does not serve that purpose and, as I said above, will likely produce a mess that will cause even more confusion as result.

> I've seen dapp was using direct import from dist, as the index was not exporting all necessary things. Also, I've noticed code duplicating in https://github.com/threshold-network/token-dashboard/tree/main/src/threshold-ts or new functions that I would expect to be implemented in this library.

The `dist` directory is basically a container that holds all public artifacts of the library. We cannot prevent importing from it in any way. This is why I think we should create a proper `index.ts` that will emphasize the most important things. Developers can import from `dist` what they want anyway. In that context, what is the point of exporting the entire `dist` through `index.ts`? In that case, the `index.ts` is just a path shortener.

> It's hard to decide in a library what an actual user would need, as we cannot predict all the use cases. We should give some flexibility to a user.

Totally agreed. But, isn't that flexibility provided by `dist`?

nkuba: aeea14f022e3e27570434a49c2d07a98d155c60c

lukasz-zimnoch: Leaving that discussion open for future reference.

typescript/src/ethereum.ts
```diff
@@ -567,3 +607,162 @@ class WalletRegistry extends EthereumContract<ContractWalletRegistry> {
...
+  // eslint-disable-next-line valid-jsdoc
+  /**
+   * @see {ChainTBTCVault#requestOptimisticMint}
+   */
+  async requestOptimisticMint(
+    depositTxHash: TransactionHash,
+    depositOutputIndex: number
+  ): Promise<string> {
+    const tx = await this._instance.requestOptimisticMint(
+      depositTxHash.toPrefixedString(),
```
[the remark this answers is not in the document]
nkuba: 56f5f3352bd55a32b44c0779adff3ea4eff60dfc

typescript/src/ethereum.ts
```diff
@@ -567,3 +607,162 @@ class WalletRegistry extends EthereumContract<ContractWalletRegistry> {
...
+  // eslint-disable-next-line valid-jsdoc
+  /**
+   * @see {ChainTBTCVault#cancelOptimisticMint}
+   */
+  async cancelOptimisticMint(
+    depositTxHash: TransactionHash,
+    depositOutputIndex: number
+  ): Promise<string> {
+    const tx = await this._instance.cancelOptimisticMint(
+      depositTxHash.toPrefixedString(),
```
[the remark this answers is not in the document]
nkuba: 56f5f3352bd55a32b44c0779adff3ea4eff60dfc

typescript/src/ethereum.ts
```diff
@@ -567,3 +607,162 @@ class WalletRegistry extends EthereumContract<ContractWalletRegistry> {
...
+  // eslint-disable-next-line valid-jsdoc
+  /**
+   * @see {ChainTBTCVault#finalizeOptimisticMint}
+   */
+  async finalizeOptimisticMint(
+    depositTxHash: TransactionHash,
+    depositOutputIndex: number
+  ): Promise<string> {
+    const tx = await this._instance.finalizeOptimisticMint(
+      depositTxHash.toPrefixedString(),
```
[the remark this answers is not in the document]
nkuba: 56f5f3352bd55a32b44c0779adff3ea4eff60dfc

typescript/src/ethereum.ts
```diff
@@ -567,3 +607,162 @@ class WalletRegistry extends EthereumContract<ContractWalletRegistry> {
...
+
+    return events.map<OptimisticMintingRequestedEvent>((event) => {
+      return {
+        blockNumber: BigNumber.from(event.blockNumber).toNumber(),
+        blockHash: Hex.from(event.blockHash),
+        transactionHash: Hex.from(event.transactionHash),
+        minter: new Address(event.args!.minter),
+        depositKey: BigNumber.from(event.args!.depositKey),
+        depositor: new Address(event.args!.depositor),
+        amount: BigNumber.from(event.args!.amount),
```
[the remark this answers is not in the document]
nkuba: I'm wondering what would be the best way to handle this in the library to not get confused. To compare this value with the actual satoshi amount in the bitcoin transaction we will need to convert it, right?

Should we add a special type for such value?

typescript/src/wallet.ts
nkuba: It's not `TransactionHash`, right?

lukasz-zimnoch: Yeah, it's not.

typescript/src/ethereum.ts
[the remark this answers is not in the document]
lukasz-zimnoch: > I'm wondering what would be the best way to handle this in the library to not get confused. To compare this value with the actual satoshi amount in the bitcoin transaction we will need to convert it, right?

Yep indeed. 

> Should we add a special type for such value?

Hard to say what will be the best way right now. We can start by making it crystal-clear in the field docstring and leaving a TODO to address it later. If we want to introduce a special type for 1e18, we will probably need to introduce such a type for satoshis amounts as well.

typescript/src/index.ts
nkuba: a0d79b289825c321cb34108518c51a9b3faa16b5

typescript/src/wallet.ts
nkuba: c83a463a81160e27fd4eb4d3763f71a2ee3a989a

typescript/src/wallet.ts
nkuba: c83a463a81160e27fd4eb4d3763f71a2ee3a989a

typescript/src/wallet.ts
nkuba: c83a463a81160e27fd4eb4d3763f71a2ee3a989a

typescript/src/wallet.ts
nkuba: d885dfd5cbafe659c42ec60ad020f1abc37d6b80

typescript/src/ethereum.ts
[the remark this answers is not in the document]
nkuba: 2b19c412e683d222c13540dd573f59a78996d1c1

typescript/src/ethereum.ts
```diff
@@ -567,3 +607,162 @@ class WalletRegistry extends EthereumContract<ContractWalletRegistry> {
...
+        fundingTxHash: TransactionHash.from(event.args!.fundingTxHash),
```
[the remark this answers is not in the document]
nkuba: 51b793123df503965453d523f18692b580e83847

lukasz-zimnoch (approved):

lukasz-zimnoch (approved):

Issue #456, opened 2023-01-04
Optimize deployment scripts for testnet


pdyraga:
For testnet we may want to temporarily use other parameter values than for mainnet. For example, we may want to start with a shorter challenge period to run several key generations and then switch it back to the value equal to the mainnet one.

The easiest option could be to assign the governance to some address controlled by dev team on testnet and to decrease the governance delay there. Some other options may also be explored during the work on this issue.

PR #464, opened 2023-01-06
Adjust deployment scripts for testnet

lukasz-zimnoch:
Closes: #456 

Here we make some adjustments to increase the flexibility of testnet deployments. The main goal is to have a short test feedback cycle that will allow checking multiple scenarios using different combinations of the contract's governance parameters. In order to achieve that, we must be able to change governable parameters quickly. This change sets a 60-second governance delay for testnet and passes the governance over testnet contracts to an address controlled by the dev team.

### DIFF
--- a/solidity/hardhat.config.ts
+++ b/solidity/hardhat.config.ts
@@ -172,35 +172,35 @@ const config: HardhatUserConfig = {
     //       Inspect usages and rename.
     governance: {
       default: 2,
-      goerli: 0,
+      goerli: "0xCac19049825F370dB0836cB0d8E4D024F78eb2eB", // Dev team
     },
     chaosnetOwner: {
       default: 3,
-      goerli: 0,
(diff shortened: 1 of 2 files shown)


solidity/hardhat.config.ts
```diff
@@ -172,35 +172,35 @@ const config: HardhatUserConfig = {
...
     },
     treasury: {
       default: 7,
-      goerli: 0,
+      goerli: "0xCac19049825F370dB0836cB0d8E4D024F78eb2eB", // Dev team
     },
     spvMaintainer: {
       default: 8,
-      goerli: 0,
+      goerli: "0xCac19049825F370dB0836cB0d8E4D024F78eb2eB", // Dev team
```
pdyraga, line 203: I am not sure about this one, once we activate sweeping, this should probably be some separate bot address running on testnet. Anyway, it is not something we need to worry about right now.

pdyraga (approved):

Issue #455, opened 2023-01-04
Inspect and refresh deployment scripts


pdyraga:
This issue covers the work needed for a final inspection of tBTC deployment scripts before the mainnet deployment.
- [ ] Verify the deployment using Goerli network.
- [ ] Make sure Etherscan verification plugin is used correctly and covers all contracts.
- [ ] Make sure Tenderly plugin is used correctly and covers all contracts.
- [ ] Check the mainnet deployment against the scenario from #438 and make sure all steps are properly covered.
- [ ] Check the Hardhat versions of Hardhat plugins.

PR #469, opened 2023-01-10
Treasury address made governable

pdyraga:
Refs #455 

The treasury address was not governable before. Making it governable now via `BridgeGovernance` contract, with the same delay as other `Bridge` parameters.

The Treasury will be initially set to the Token Holder DAO. If it proves unwieldy to make use of the fees via DAO voting, the DAO may decide to transfer the Treasury address to the Council or one of the Guilds.

### DIFF
--- a/solidity/test/bridge/Bridge.Governance.test.ts
+++ b/solidity/test/bridge/Bridge.Governance.test.ts
@@ -4329,7 +4329,7 @@ describe("Bridge - Governance", () => {
     )
   })
 
-  describe("beingTreasuryUpdate", () => {
+  describe("beginTreasuryUpdate", () => {
     const newTreasury = "0x8A71228c19A3531384FC203F56290D3aF01B16bD"
 
     context("when the caller is not the owner", () => {
--- a/solidity/test/bridge/Bridge.Governance.test.ts
+++ b/solidity/test/bridge/Bridge.Governance.test.ts
@@ -298,6 +298,8 @@ describe("Bridge - Governance", () => {
       })
     })
 
+    // FIXME: This test is not covering the behavior described!
+    // The expected revert message should be "Change not initiated".
     context("when the update process is not initialized", () => {
       it("should revert", async () => {
         await expect(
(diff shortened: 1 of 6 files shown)


solidity/test/bridge/Bridge.Parameters.test.ts
```diff
@@ -1769,4 +1771,65 @@ describe("Bridge - Parameters", () => {
       })
     })
   })
+
+  describe("updateTreasury", () => {
+    const newTreasury = "0x4EDCe37af1c6e2CDfe086D7940C5a8AAde9c72e3"
+
+    context("when caller is the contract guvnor", () => {
+      before(async () => {
+        // TODO: We transfer the ownership of the Bridge governance from the
```
dimpar, line 1780: But wouldn't updating through the BridgeGovernance be better aligned with a real-world scenario where the owner of the Bridge is BridgeGovernance https://github.com/keep-network/tbtc-v2/blob/main/solidity/deploy/17_transfer_bridge_governance.ts#L10 ?

solidity/test/bridge/Bridge.Governance.test.ts
```diff
@@ -4297,6 +4329,125 @@ describe("Bridge - Governance", () => {
     )
   })
 
+  describe("beingTreasuryUpdate", () => {
```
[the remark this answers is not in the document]
pdyraga: 2abfd17a1cf4e729f88d83ee4ffda63e9624d123

solidity/test/bridge/Bridge.Parameters.test.ts
pdyraga: This is a real-world scenario only until the `BridgeGovernance` is the owner of `Bridge` :) There is no guarantee this will always be the case. So I think we should test `Bridge` directly in unit tests for `Bridge` and test `BridgeGovernance + Bridge` combo in unit tests for `BridgeGovernance`. This way we can be sure that the validation is in the `Bridge.updateTreasury(...)` and not in `BridgeGovernance.finalizeTreasuryUpdate()`.

dimpar (approved):

PR #475, opened 2023-01-13
Set wallet closure minimum BTC balance to zero

lukasz-zimnoch:
Port of #474 to `releases/mainnet/solidity/v1.0.0` branch.

For the initial release, we want to prevent wallets from entering into the `MOVING_FUNDS`, `CLOSING` and `CLOSED` states. Wallets must remain `LIVE` to be able to sweep funds when the time comes. So far, the `walletClosureMinBtcBalance` governable parameter was set to its minimal value of `1` satoshi. That value means a wallet can be directly moved to the `CLOSING` state if their balance in the `Bridge` is `0` satoshi and the wallet is no longer the latest one (i.e. active). This is a problem for optimistic minting because all wallets will have `0` satoshi in the `Bridge` until sweeping is implemented. Here we fix that by allowing to set `0` as a value of the `walletClosureMinBtcBalance` parameter and use that value as default in the deployment scripts.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -1449,7 +1449,6 @@ contract Bridge is
     /// @dev Requirements:
     ///      - Wallet maximum BTC balance must be greater than the wallet
     ///        minimum BTC balance,
-    ///      - Wallet closure BTC balance must be greater than zero,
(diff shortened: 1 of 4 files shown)


pdyraga:
Some tests are failing because they can't resolve `*Stub` artifacts for mainnet deployment. I'll cherry-pick this change manually.

PR #478, opened 2023-01-16
Remove the `etherscan` tag for `goerli` environment

michalinacienciala:
Currently in the CI we use `hardhat-deploy` to verify contracts on Goerli, for which we don't need to use tags. Tags are used for verification using `@nomiclabs/hardhat-etherscan` - we would need to adjust CI and hardhat config to support it.


pdyraga (approved):

nkuba:
Please remember to also align other repos, e.g.:
https://github.com/keep-network/keep-core/blob/e14d15b1b9058508cf6dc5f2a3d2a9c48d57a881/solidity/ecdsa/hardhat.config.ts#L121

https://github.com/keep-network/keep-core/blob/e14d15b1b9058508cf6dc5f2a3d2a9c48d57a881/solidity/random-beacon/hardhat.config.ts#L109

PR #483, opened 2023-01-18
Exclude typechain script from postinstall

nkuba:
We use typechain script to generate TS bindings for contracts that we install as dependencies. With postinall script the bindings were generated always when we installed the dependencies. Unfortunatelly if this package was being used in another project as dependency this postinstall script was also executed, which doesn't make sense as we publish the generated bindings as part of the package.

With this change here we move the typechain bindings geneartion to the build script so they are generated when building the package before publication.

### DIFF
--- a/.github/workflows/npm-typescript.yml
+++ b/.github/workflows/npm-typescript.yml
@@ -35,11 +35,6 @@ jobs:
       - name: Resolve latest contracts
         run: yarn upgrade @keep-network/tbtc-v2 --exact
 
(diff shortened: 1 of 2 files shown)


michalinacienciala (commented):
We also need to remove this: https://github.com/keep-network/tbtc-v2/blob/main/.github/workflows/npm-typescript.yml#L38-41

michalinacienciala (approved):

PR #486, opened 2023-01-19
Add a possibility to pass options to the Electrum client

michalsmiarowski:
We need to have a possibility to add `options` object to Electrum client and we couldn't do that through tbtc-v2.ts lib. This is why we introduce additional and optional property `options` in `Credentials` interface.

### DIFF
--- a/typescript/src/electrum.ts
+++ b/typescript/src/electrum.ts
@@ -30,11 +30,12 @@ export interface Credentials {
    * Protocol used by the Electrum server.
    */
   protocol: "tcp" | "tls" | "ssl" | "ws" | "wss"
-  /**
-   * Additional options used by the Electrum server.
-   */
-  options?: any
 }
+
+/**
+ * Additional options used by the Electrum server.
+ */
+type Options = object
 
 /**
  * Represents an action that makes use of the Electrum connection. An action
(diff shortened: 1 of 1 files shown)


typescript/src/electrum.ts
```diff
@@ -30,6 +30,10 @@ export interface Credentials {
    * Protocol used by the Electrum server.
    */
   protocol: "tcp" | "tls" | "ssl" | "ws" | "wss"
+  /**
+   * Additional options used by the Electrum server.
+   */
+  options?: any
```
nkuba, line 36: I'm not sure about adding this property to the `Credentials`. Shouldn't it be a separate Client's property?

```ts
type Options = object

export class Client implements BitcoinClient {
  private credentials: Credentials
  private options?: Options

  constructor(credentials: Credentials, options?: Options) {
    this.credentials = credentials
    this.options = options
  }


  static fromUrl(url: string, options?: Options): Client {
    const credentials = this.parseElectrumCredentials(url)
    return new Client(credentials, options)
  }

```

CC @lukasz-zimnoch

nkuba (changes requested):

typescript/src/electrum.ts
lukasz-zimnoch: Yep, I think @nkuba is right. Those are two separate beings.

michalsmiarowski: @nkuba d2bef3175c872710f687be1339c8bbd119239163

nkuba (approved):

PR #487, opened 2023-01-19
Export ClientOptions type

michalsmiarowski:
We want to use that type in the dApp and to be able to do that we need this variable to be exported.

We also change the name of it from `Options` to `ClientOptions` as the `Options` seems to generic.

### DIFF
--- a/typescript/src/electrum.ts
+++ b/typescript/src/electrum.ts
@@ -49,9 +49,9 @@ type Action<T> = (electrum: any) => Promise<T>
  */
 export class Client implements BitcoinClient {
   private credentials: Credentials
-  private options?: Options
+  private options?: ClientOptions
 
-  constructor(credentials: Credentials, options?: Options) {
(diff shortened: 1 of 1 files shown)


typescript/src/electrum.ts
```diff
@@ -35,7 +35,7 @@ export interface Credentials {
 /**
  * Additional options used by the Electrum server.
  */
-type Options = object
+export type ClientOptions = object
```
nkuba, line 38: We need to update the usage of this type in the `Client` class as well.

nkuba (changes requested):

typescript/src/electrum.ts
michalsmiarowski: Sorry, wanted to do it too fast 🤦‍♂️ Fixed in e7b1090c82638ab4009e27cbaeff75e11e76f4d4

nkuba (approved):

PR #488, opened 2023-01-20
Fix the TBTC token ownership update in deployment scripts for Goerli

pdyraga:
Unit tests cover the path of the VendingMachine upgrade to TBTCVault. We transfer the ownership of TBTC token to the VendingMachine for Hardhat network used for unit tests to enable this path. For all other environments (e.g. local dev, Goerli), we want the TBTCVault to be the owner of TBTC token. This simulates the final 1.0 version mainnet state.


michalinacienciala (commented):
Looks good to me, but I wouldn't rely on just my opinion. Let's have somebody else review on thi as well.

michalinacienciala (approved):

lukasz-zimnoch (approved):

PR #491, opened 2023-01-23
Convert Bitcoin precision to Ethereum precision in TBTC Vault

pdyraga:
Depends on #489.

This is an in-progress work and unit tests need to be fixed.

When minting TBTC we should multiply the Bank balance expressed in satoshis by `SATOSHI_MULTIPLIER = 10**10` to convert from Bitcoin precision to Ethereum precision. `mint`, `unmint`, `unmintAndRedeem` and `receiveApproval` functions of `TBTCVault` operate in Ethereum world so they accept 1e18 precision numbers. `receiveBalanceApproval` and `receiveBalanceIncrease` are called by the Bank which operates in satoshis so they accept Bitcoin precision. An important requirement when converting between precisions is to strip the amount so that the change can not be used for printing tokens.

`strip1e18Amount` view functions allows to check the stripped amount, remainder, and the Bank balance.


solidity/contracts/vault/TBTCVault.sol
```diff
@@ -74,68 +74,82 @@ contract TBTCVault is IVault, Ownable, TBTCOptimisticMinting {
...
         );
-        _mint(msg.sender, amount);
-        bank.transferBalanceFrom(msg.sender, address(this), amount);
+        _mint(msg.sender, strippedAmount);
+        bank.transferBalanceFrom(msg.sender, address(this), balanceToTransfer);
     }
 
-    /// @notice Transfers the given `amount` of the Bank balance from the caller
-    ///         to TBTC Vault and mints `amount` of TBTC to the caller.
+    /// @notice Transfers `sathoshis` of the Bank balance from the caller
```
tomaszslabon, line 98:
```suggestion
    /// @notice Transfers `satoshis` of the Bank balance from the caller
```

solidity/contracts/vault/TBTCVault.sol
```diff
@@ -74,68 +74,82 @@ contract TBTCVault is IVault, Ownable, TBTCOptimisticMinting {
...
-        uint256[] calldata depositedAmounts
+        uint256[] calldata depositedSatoshiAmounts
     ) external override onlyBank {
         require(depositors.length != 0, "No depositors specified");
         for (uint256 i = 0; i < depositors.length; i++) {
             address depositor = depositors[i];
-            uint256 amount = depositedAmounts[i];
-            _mint(depositor, repayOptimisticMintingDebt(depositor, amount));
+            uint256 satoshis = depositedSatoshiAmounts[i];
+            _mint(depositor, repayOptimisticMintingDebt(depositor, satoshis)); // TODO: SATOSHI_MULTIPLIER
```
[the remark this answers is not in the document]
pdyraga: It should be:
```
            _mint(
                depositor,
                repayOptimisticMintingDebt(
                    depositor,
                    satoshis * SATOSHI_MULTIPLIER
                )
            );
```

`optimisticMintingDebt` is already in 1e18 and `repayOptimisticMintingDebt` should operate in 1e18 precision.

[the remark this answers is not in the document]
pdyraga: It is updated in the most recent version.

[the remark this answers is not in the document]
lukasz-zimnoch: Ok, so we are assuming the `TBTCOptimisticMintining` is oriented around ERC 1e18 precision and its public API operates on it. That's good but we still have the `OptimisticMintingRequested` event that contains an amount in 1e8. I think we should make it super clear in `TBTCOptimisticMintining`: 
- all amounts are 1e18 ERCs
- every piece of code that takes the amount from the Bridge's deposit request must convert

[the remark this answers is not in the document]
pdyraga: Good catch. See f0d9686f69a90911236fd0cbd670375051dc223b, please. I think this event was the only place where we had to do the conversion. I have also added a note to the `optimisticMintingDebt` to make it clear the values are in 1e18 precision.

solidity/contracts/vault/TBTCVault.sol
pdyraga: Fixed in the most recent version.

solidity/contracts/vault/TBTCVault.sol
```diff
@@ -74,68 +74,92 @@ contract TBTCVault is IVault, Ownable, TBTCOptimisticMinting {
...
+    ///         Bank balance.
+    /// @dev TBTC Vault must have an allowance for caller's balance in the
+    ///      Bank for at least `amount / SATOSHI_MULTIPLIER`.
     /// @param amount Amount of TBTC to mint.
     function mint(uint256 amount) external {
+        (
+            uint256 strippedAmount,
+            ,
+            uint256 balanceToTransfer
+        ) = amountToSatoshis(amount);
```
lukasz-zimnoch, line 90: Shouldn't we use names returned by `amountToSatoshis `? Here and everywhere else where `amountToSatoshis` is used?

pdyraga: We should, my fault. See db846c182e000762b8a577a8935f29517072a0f7, please.

solidity/test/vault/TBTCVault.test.ts
```diff
@@ -1121,4 +1245,51 @@ describe("TBTCVault", () => {
       })
     })
   })
+
+  describe("amountToSatoshis", () => {
+    context("when the amount is convertible without a remainder", () => {
```
lukasz-zimnoch, line 1250:
```suggestion
    context("when the amount is convertible with a remainder", () => {
```

dimpar:
I managed to look only at contracts and they looked good to me.

solidity/test/vault/TBTCVault.test.ts
pdyraga: 7e0dfdff40106455f6abc366ae785296821894f8

lukasz-zimnoch (approved):

Issue #422, opened 2022-11-25
TS: getTransactionConfirmations won't work with Electrs


nkuba:
The current implementation of `getTransactionConfirmations` won't work with Electrs, as verbose transaction details are not supported there (see: https://github.com/Blockstream/electrs/pull/36#issuecomment-936850568).
https://github.com/keep-network/tbtc-v2/blob/a9424d18287c1958fbfd14e0df6621bf3578a38c/typescript/src/electrum.ts#L190-L205

We should implement a workaround to cover Electrs. 
An example workaround was implemented in the Keep Client:
https://github.com/keep-network/keep-core/blob/d69e971fc734031111699f95eccc3bbbafe4345f/pkg/bitcoin/electrum/electrum.go#L163-L206

Issue #485, opened 2023-01-19
Implement retries for Electrum in tbtc-v2.ts


michalinacienciala:
Currently we don't have retries, which sometimes causes the System Tests to fail.

Example of a failing System Tests workflow: https://github.com/keep-network/tbtc-v2/actions/runs/3954068027/attempts/1.

It's not only the system tests problem, but we should do it for the tbtc-v2.ts library, so that optimistic minting maintainer and dapp can benefit from that as well.

PR #501, opened 2023-01-25
TS: Support Blockstream's Esplora/Electrs servers for electrum protocol integration

nkuba:
Closes: https://github.com/keep-network/tbtc-v2/issues/422

In this PR we update the implementation of TypeScript Electrum integration to support Esplora/Electrs servers.

As per https://github.com/Blockstream/electrs/pull/36 verbose transactions are not supported by Esplora/Electrs.
This affects our implementation of `getTransaction` and `getTransactionConfirmations` functions.

For a consistent code in the client without alternative paths for different Electrum servers implementations I decided to not use verbose transactions at all.

### [getTransactionConfirmations](https://github.com/keep-network/tbtc-v2/commit/26d4f0198fc964f49abbcf997777f89576d86b19)

1. Get the raw transaction
2. Deserialize the raw transaction
3. Find transaction block height by finding it in a history of
   transactions for the output script included in the transaction.
4. Get the latest block height
5. Calculate number of confirmations by subtracting the transaction
   block height from the latest block height and adding one.

### [getTransaction](https://github.com/keep-network/tbtc-v2/commit/a7aedd11fbfc7b26cbda5376a33207c7966f48bf)

We get a raw transaction and deserialize it with `bcoin`. This lets us define a consistent type for returned transactions (093d4ece7d2cd83d3d554a97d0985a67b7f4d154). Before these changes, I observed that Electrum server implementations are not consistent with data returned in verbose JSON.

### Electrum tests

We can test electrum integration against different kinds of servers.
The most popular implementations are:
- ElectrumX
- Fulcrum
- Electrs/Esplora

We can find a list of public servers here: https://1209k.com/bitcoin-eye/ele.php?chain=tbtc

The electrs-esplora server seems pretty unstable, so we don't want to enable it in tests until we add retries (https://github.com/keep-network/tbtc-v2/issues/485).

### DIFF
--- a/typescript/src/electrum.ts
+++ b/typescript/src/electrum.ts
@@ -149,30 +149,48 @@ export class Client implements BitcoinClient {
    */
   getTransaction(transactionHash: TransactionHash): Promise<Transaction> {
     return this.withElectrum<Transaction>(async (electrum: any) => {
-      const transaction = await electrum.blockchain_transaction_get(
+      // We cannot use `blockchain_transaction_get` with `verbose = true` argument
+      // to get the the transaction details as Esplora/Electrs doesn't support verbose
+      // transactions.
+      // See: https://github.com/Blockstream/electrs/pull/36
+      const rawTransaction = await electrum.blockchain_transaction_get(
         transactionHash.toString(),
-        true
+        false
       )
 
-      const inputs = transaction.vin.map(
+      if (!rawTransaction) {
+        throw new Error(`Transaction not found`)
+      }
+
+      // Decode the raw transaction.
+      const transaction = bcoin.TX.fromRaw(rawTransaction, "hex")
+
+      const inputs = transaction.inputs.map(
         (input: any): TransactionInput => ({
-          transactionHash: TransactionHash.from(input.txid),
-          outputIndex: input.vout,
-          scriptSig: input.scriptSig,
+          transactionHash: TransactionHash.from(input.prevout.hash).reverse(),
+          outputIndex: input.prevout.index,
+          scriptSig: {
+            asm: input.script.toASM(true),
+            hex: input.script.toRaw().toString("hex"),
+            type: input.script.getType(),
+          },
         })
       )
 
-      const outputs = transaction.vout.map(
-        (output: any): TransactionOutput => ({
-          outputIndex: output.n,
-          // The `output.value` is in BTC so it must be converted to satoshis.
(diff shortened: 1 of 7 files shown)


nkuba:
An alternative approach that I would also like to investigate would be to use verbose transactions as before and fallback to the logic proposed in this PR only in case verbose is not supported by the server.

typescript/src/electrum.ts
```diff
@@ -203,15 +214,81 @@ export class Client implements BitcoinClient {
...
+
+      // Initialize txBlockHeigh with minimum int32 value to identify a problem when
+      // a block height was not found in a history of any of the script hashes.
+      //
+      // The history is expected to return a block height for confirmed transaction.
+      // If a transaction is unconfirmed (is still in the mempool) the height will
+      // have a value of `0` or `-1`.
+      let txBlockHeight: number = Math.min()
+      for (const output of transaction.outputs) {
+        const scriptHash: Buffer = output.script.sha256()
```
lukasz-zimnoch, line 248: The `sha256` will not work in all cases. It produces a 32-byte script hash that is specific only to P2WSH. Other types P2SH, P2PKH, and P2WPKH use a 20-byte script computed using `hash160`.

lukasz-zimnoch (changes requested):

nkuba:
> An alternative approach that I would also like to investigate would be to use verbose transactions as before and fallback to the logic proposed in this PR only in case verbose is not supported by the server.

I tested it and it did not result in any improvement in test execution time. This is probably as if we get verbose transactions the processing has to be done on the server side. Let's leave it as is and update it in the future if necessary.

lukasz-zimnoch (approved):

shesek:
> I tested it and it did not result in any improvement in test execution time

Interesting, I would've expected verbose transactions to perform better than fetching the entire address history and looking it up there.

Which electrum server implementation did you benchmark this against? Did you try with transactions that use reused addresses, such that looking up the address returns other historical transactions too?

(For example [`3ae54f00613ca0a39bbb202e49dc7f2abba9e5ed4c3fcdaed768c7effeeca403`](https://blockstream.info/tx/3ae54f00613ca0a39bbb202e49dc7f2abba9e5ed4c3fcdaed768c7effeeca403), which sends to an address with nearly 700k historical txs -- I guess it might not work at all if there are that many of them?)

PR #504, opened 2023-01-26
TS: Define BitcoinNetwork enum

nkuba:
We added `BitcoinNetwork` enum to reduce the possibility of confusion on network-related operations like generating addresses with the use of `bcoin` library.
The `bcoin` library expects mainnet to be provided in a string argument as `main`, which I find very confusing and easy to miss. It is mentioned in the `encodeToBitcoinAddress`'s JSDoc, but who reads the docs? 🤷🏻

The enum will also be used to determine the network supported by the electrum server the client connects to.

Refs: https://github.com/keep-network/optimistic-minting/issues/27

### DIFF
--- a/typescript/src/bitcoin.ts
+++ b/typescript/src/bitcoin.ts
@@ -391,13 +391,21 @@ export function encodeToBitcoinAddress(
 }
 
 /**
- * Decodes P2PKH or P2WPKH address into a public key hash.
+ * Decodes P2PKH or P2WPKH address into a public key hash. Throws if the
+ * provided address is not PKH-based.
  * @param address - P2PKH or P2WPKH address that will be decoded.
  * @returns Public key hash decoded from the address. This will be an unprefixed
  *        hex string (without 0x prefix).
  */
 export function decodeBitcoinAddress(address: string): string {
   const addressObject = new bcoin.Address(address)
+
+  const isPKH =
+    addressObject.isPubkeyhash() || addressObject.isWitnessPubkeyhash()
+  if (!isPKH) {
+    throw new Error("Address must be P2PKH or P2WPKH")
+  }
+
   return addressObject.getHash("hex")
 }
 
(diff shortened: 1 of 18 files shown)


typescript/src/bitcoin/network.ts
```diff
@@ -0,0 +1,66 @@
+import { TransactionHash } from "../bitcoin"
```
lukasz-zimnoch, line 1: Can we keep a flat file structure for now? I propose to name this file `bitcoin-network.ts` and place it in `src` dir directly (same for the test file). We will probably refactor the file structure anyway soon so it would be good to start from a consistent approach.

typescript/src/bitcoin.ts
```diff
@@ -373,19 +374,20 @@ export function computeHash160(text: string): string {
...
- *        For example, `main` or `testnet`.
+ * @param network - Network the address should be encoded for.
  * @returns P2PKH or P2WPKH address encoded from the given public key hash
+ * @throws Throws an error if network is not supported.
  */
 export function encodeToBitcoinAddress(
   publicKeyHash: string,
   witness: boolean,
-  network: string
+  network: BitcoinNetwork
```
lukasz-zimnoch, line 384: There are more places where we expect `network: string` (e.g `calculateDepositAddress`). Shouldn't we replace them with `network: BitcoinNetwork` as well?

lukasz-zimnoch (changes requested):

typescript/src/bitcoin/network.ts
nkuba: 3f1b7b3f2cfd43a894970a5f167c449ecbb9f228

typescript/src/bitcoin.ts
nkuba: Good catch! 435e93ae2735bcfdb08da9b69d72fed988a36430

I can't find any other places though.

nkuba:
@lukasz-zimnoch can we merge it? All comments were addressed.

lukasz-zimnoch (approved):

PR #510, opened 2023-01-30
Make `decodeBitcoinAddress` throw on non-PKH addresses

lukasz-zimnoch:
Refs: https://github.com/threshold-network/token-dashboard/issues/393

The `decodeBitcoinAddress` documentation says that this function decodes P2PKH and P2WPKH addresses into a public key hash. However, this requirement is not enforced by the function code. Here we add a validation that makes the function throw if the passed address is not PKH-based. That means the function will throw an exception if the passed address is, for example, P2SH or P2WSH.

### DIFF
--- a/typescript/src/bitcoin.ts
+++ b/typescript/src/bitcoin.ts
@@ -389,13 +389,21 @@ export function encodeToBitcoinAddress(
 }
 
 /**
- * Decodes P2PKH or P2WPKH address into a public key hash.
+ * Decodes P2PKH or P2WPKH address into a public key hash. Throws if the
+ * provided address is not PKH-based.
  * @param address - P2PKH or P2WPKH address that will be decoded.
  * @returns Public key hash decoded from the address. This will be an unprefixed
  *        hex string (without 0x prefix).
  */
 export function decodeBitcoinAddress(address: string): string {
   const addressObject = new bcoin.Address(address)
+
+  const isPKH =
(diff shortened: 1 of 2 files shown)


typescript/src/bitcoin.ts
```diff
@@ -389,13 +389,21 @@ export function encodeToBitcoinAddress(
...
  * @returns Public key hash decoded from the address. This will be an unprefixed
  *        hex string (without 0x prefix).
  */
 export function decodeBitcoinAddress(address: string): string {
   const addressObject = new bcoin.Address(address)
+
+  const isPKH =
+    addressObject.isPubkeyhash() || addressObject.isWitnessPubkeyhash()
+  if (!isPKH) {
+    throw new Error("address must be P2PKH or P2WPKH")
```
michalsmiarowski, line 404: Non-blocking:

We usually start the error text with a capital letter in this project. 

Remember to also update the tests if you decide to fix that.

michalsmiarowski (approved):
LGTM 🔥 Left one non-blocking comments but overall looks good!

typescript/src/bitcoin.ts
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/510/commits/b446878f9dbe0a8bfaaf89b7d2c6508cb221d9f1

michalsmiarowski (approved):
LGTM 🔥 🚒

PR #514, opened 2023-01-30
Introduce temporary retry mechanism to system tests Electrum calls

lukasz-zimnoch:
The testnet Electrum instances are often unreliable and cause failures of system test scenarios. Here we aim to mitigate that problem by wrapping all Electrum calls done by system tests with a temporary retry mechanism. Eventually, this mechanism should be replaced with a full-fledged retry support built-in in the `ElectrumClient` at the `tbtc-v2.ts` level.


nkuba:
Could this be replaced with https://github.com/keep-network/tbtc-v2/pull/517?

lukasz-zimnoch:
> Could this be replaced with #517?

Yeah, definitely. I'll make this PR draft until #517 lands

lukasz-zimnoch:
Closing in favor of #541

Issue #499, opened 2023-01-25
TS: Implement retries for Ethereum transactions


nkuba:
We need retries for Ethereum transactions submission. Similar to what we had in tbtc.js:
https://github.com/keep-network/tbtc.js/blob/9838c06f21082cc050fab5f1304b95d6fa3db6a7/src/EthereumHelpers.js#L379

PR #516, opened 2023-02-01
Retries for Ethereum requests

nkuba:
We noticed that requests to ethereum endpoints are failing to due server errors. In this PR we wrap all ethereum requests with a backoff retrier.

The backoff retrier is an implementation used in `tbtc.js` library (https://github.com/keep-network/tbtc.js/blob/9838c06f21082cc050fab5f1304b95d6fa3db6a7/src/lib/backoff.js).

Closes: https://github.com/keep-network/tbtc-v2/issues/499

### DIFF
--- a/typescript/src/backoff.ts
+++ b/typescript/src/backoff.ts
@@ -11,13 +11,35 @@ export function retryAll(error: any): true {
 }
 
 /**
+ * A matcher to specify list of error messages that should abort the retry loop
+ * and throw immediately.
+ * @param matchers List of patterns for error matching.
+ * @returns Matcher function that returns false if error matches one of the patterns.
+ *          True is returned if no matches are found and retry loop should continue
+ */
+export function skipRetryWhenMatched(
+  matchers: Array<string | RegExp>
+): ErrorMatcherFn {
+  return (err: unknown): boolean => {
+    if (err instanceof Error) {
+      for (const matcher of matchers) {
+        if (err.message.match(matcher)) {
+          return false
+        }
+      }
+    }
+    return true
+  }
+}
+
+/**
  * @callback ErrorMatcherFn A function that returns true if the passed error
  *           matches and false otherwise. Used to determine if a given error is
  *           eligible for retry in `withBackoffRetries`.
  * @param {Error} error The error to check for eligibility.
  * @return {boolean} True if the error matches, false otherwise.
  */
-type ErrorMatcherFn = (err: unknown) => boolean
+export type ErrorMatcherFn = (err: unknown) => boolean
 
 /**
  * @callback RetrierFn A function that can retry any function passed to it a
@@ -30,7 +52,7 @@ type ErrorMatcherFn = (err: unknown) => boolean
 /**
  * A function that is called with execution status messages.
  */
-type ExecutionLoggerFn = (msg: string) => void
+export type ExecutionLoggerFn = (msg: string) => void
 
 /**
  * Returns a retrier that can be passed a function to be retried `retries`
@@ -92,6 +114,6 @@ export function backoffRetrier<T>(
     }
 
     // Last attempt, unguarded.
-    return await fn()
+    return fn()
   }
 }
--- a/typescript/src/backoff.ts
+++ b/typescript/src/backoff.ts
@@ -76,6 +76,8 @@ export type ExecutionLoggerFn = (msg: string) => void
  *
  * @param {number} retries The number of retries to perform before bubbling the
  *        failure out.
+ * @param {number} backoffStepMs Initial backoff step in milliseconds that will
+ *        be increased exponentially for subsequent retry attempts. (default = 1000 ms)
  * @param {ExecutionLoggerFn} logger A logger function to pass execution messages.
  * @param {ErrorMatcherFn} [errorMatcher=retryAll] A matcher function that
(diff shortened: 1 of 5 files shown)


typescript/src/ethereum.ts
```diff
@@ -367,7 +393,11 @@ export class Bridge
       vault: vault ? `0x${vault.identifierHex}` : constants.AddressZero,
     }
 
-    const tx = await this._instance.revealDeposit(depositTxParam, revealParam)
+    const tx = await backoffRetrier<ContractTransaction>(
+      this._totalRetryAttempts
+    )(async () => {
+      return await this._instance.revealDeposit(depositTxParam, revealParam)
+    })
```
pdyraga, line 400: Will we retry in case it is not an unexpected failure (e.g. broken connection) but the gas estimation fails because the deposit has been already revealed? Ideally, I think we shouldn't retry for the latter but it's hard to implement and the current solution is an improvement anyway so I do not find it blocking.

typescript/src/backoff.ts
```diff
@@ -0,0 +1,97 @@
...
+        logger(`making attempt number ${attempt}`)
+
+        return await fn()
+      } catch (error: unknown) {
+        if (!errorMatcher(error)) {
+          // If the matcher doesn't match this error, rethrow and stop.
+          throw error
+        }
+
+        const backoffMillis = Math.pow(2, attempt) * 1000
```
lukasz-zimnoch, line 81: Can we make the backoff multiplier configurable? While working on #514, I realized that `1000 ms` is often not enough for Electrum retries, especially when errors are caused by the node being out of sync and, for example, not yet seeing the requested transaction. We can use `1000 ms` by default but I'd love to pass anther value if needed.

typescript/src/backoff.ts
```diff
@@ -0,0 +1,97 @@
...
+          `attempt ${attempt} failed: ${error}; ` +
+            `retrying after ${waitMillis} milliseconds`
+        )
+
+        await new Promise((resolve) => setTimeout(resolve, waitMillis))
+      }
+    }
+
+    // Last attempt, unguarded.
+    return await fn()
```
lukasz-zimnoch, line 95: Is there any sense to `await` here? We need to return a promise anyway. Also, the ESLint complains: "Returning an awaited promise is not allowed in this context."

typescript/src/ethereum.ts
nkuba: I pushed a couple of commits that should improve it and cover the situation you mentioned:
ccd4d0a9b2a6be5392151cd5c7dee259ef89f69c, 5e19011be61bf8b2009e3d67ad4765bc37440a8e, 0cda5e7a587432774e53624a839a4abb153531f6

typescript/src/backoff.ts
nkuba: Can't we just add more retries if we need a longer backoff?

typescript/src/backoff.ts
nkuba: Niche catch. I updates it as part of https://github.com/keep-network/tbtc-v2/pull/516/commits/ccd4d0a9b2a6be5392151cd5c7dee259ef89f69c#diff-6f0abbdd421b0e7a7890b8e6ede7a16fb44fc7dd491c59dcd21e9888583d2476R117

typescript/src/backoff.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+export type ErrorMatcherFn = (err: unknown) => boolean
+
+/**
+ * @callback RetrierFn A function that can retry any function passed to it a
+ *           set number of times, returning its result when it succeeds. Created
+ *           using `backoffRetrier`.
+ * @param {function(): Promise<T>} fn The function to be retried.
+ * @return {Promise<T>}
+ */
```
tomaszslabon, line 50: This piece of documentation is probably not needed.

tomaszslabon, line 50: Shouldn't we define:
```
export type RetrierFn = ....
```
below this piece of documentation?

typescript/src/backoff.ts
```diff
@@ -0,0 +1,119 @@
...
+ *        receives the error when an exception is thrown, and returns true if
+ *        the error should lead to a retry. A false return will rethrow the
+ *        error and terminate the retry loop.
+ * @returns {RetrierFn<T>} A function that can retry any function.
+ */
+export function backoffRetrier<T>(
+  retries: number,
+  logger: ExecutionLoggerFn = console.debug,
+  errorMatcher: ErrorMatcherFn = retryAll
+) {
```
tomaszslabon, line 90:
```suggestion
): RetrierFn<T> {
```

typescript/src/backoff.ts
lukasz-zimnoch: We can, but in some cases, it may make more sense to wait a little longer than bombarding the target service with new requests. This change should not be a big deal but can add great flexibility.

nkuba: a20f04ada97c204b1b5a1e6e8b36a13c55819195

typescript/src/backoff.ts
nkuba: c26032cd5fe315f00c53727f35494651bdbe5fa9

tomaszslabon (approved):

PR #524, opened 2023-02-07
Rerun System Tests on failure

michalinacienciala:
We've seen that from time to time System Tests workflow fails due to problems with connectivity to Electrum or due to some other randomly occurring issues (like the problems with installing dependencies). We now add a job to the System Tests workflow which is started only if one of the frevious jobs failed. The job will trigger a Rerun workflow, which will then trigger the System Tests again. The reason why we need to rerun via a separate workflow is because reruns cannot be called on the workflows/jobs which are still running.


nkuba:
IMHO we should hold with adding general job reruns. We should try to fix all the problems we're experiencing on installations and test execution. 
Recent updates to tbtc-v2.ts library implementing ethereum and electrum requests retries (https://github.com/keep-network/tbtc-v2/pull/541) should improve the stability of test execution.
We should also determine other problems and try fixing them one by one.

pdyraga:
> IMHO we should hold with adding general job reruns. We should try to fix all the problems we're experiencing on installations and test execution. Recent updates to tbtc-v2.ts library implementing ethereum and electrum requests retries (#541) should improve the stability of test execution. We should also determine other problems and try fixing them one by one.

This work is captured in https://github.com/keep-network/tbtc-v2/issues/531 and is on our sprint board.

PR #526, opened 2023-02-08
Chain events monitoring tool

lukasz-zimnoch:

Here we provide a tBTC v2 monitoring tool that is meant to observe on-chain contracts and raise alerts/notifications on several platforms. 

### Architecture

The high-level architecture of the monitoring tool looks as follows:
```
       +------------------+
       |                  |
       |  tBTC v2 system  |
       |                  |
       +------------------+
               ^  ^
               |  |
        +------+  +------+
        |                |
        |                |
 +------+------+  +------+------+
 |             |  |             |
 |  Monitor 1  |  |  Monitor 2  |
 |             |  |             |
 +------+------+  +------+------+
        |                |
        |                |
        +------+  +------+
               |  |
               v  v
          +------------+      +---------------+
          |            |      |               |
          |  Manager   +----->|  Persistence  |
          |            |      |               |
          +----+--+----+      +---------------+
               |  |
               |  |
        +------+  +------+
        |                |
        v                v
+--------------+  +--------------+
|              |  |              |
|  Receiver 1  |  |  Receiver 2  |
|              |  |              |
+--------------+  +--------------+
```

Specific components are:

- **tBTC v2 system**: On-chain smart contracts
- **Monitors**: Components that observe the **tBTC v2 system** and generate appropriate system events according to their internal logic
- **Manager**: The main component that manages the given monitoring run. Gathers the system events generated by the **Monitors** and dispatches them to registered **Receivers**
- **Persistence**: Component that persists **Manager**'s processing data
- **Receivers**: Components that receive system events from the **Manager** and decide about handling or ignoring them

Such a structure gives great flexibility and allows to easily extend the monitoring tool as new **Monitors** and **Receivers** can be implemented without hassle. Also, the **Persistence** component can be swapped anytime.

### The flow

The monitoring tool is meant to be used in a script-like way, i.e. triggered in a loop with fixed delays between subsequent runs. A single run always focuses on a specific chain block window. The window's starting block is determined based on the previous run's end block held by the **Persistence** component (the so-called checkpoint block). The starting block is also decremented by a constant factor of 12 to cover possible chain reorganizations. The window's ending block is always the latest mined block (i.e. the chain tip). For example, assuming that the initial checkpoint block is 100, the loop will look as follows:

- Run 1: Start block 100 - 12 = 88, end block 150
- Run 2: Start block 150 - 12 = 138, end block 155
- Run 3: Start block 155 - 12 = 143, end block 200

This way we are achieving a sliding block window that covers the entire chain. Worth noting, the checkpoint block is updated only if the given run completes with success. A run is considered successful when no single error was thrown by any of the components. If the run is not successful, the checkpoint block is not updated and the next run will cover the failed range again:

- Run 1: Start block 100 - 12 = 88, end block 150 - failed
- Run 2: Start block 100 - 12 = 88, end block 170 - failed
- Run 3: Start block 100 - 12 = 88, end block 200 - success
- Run 4: Start block 200 - 12 = 188, end block 250 - success

This way the monitoring tool guarantees to not miss any system events. However, such a solution can produce duplicated system events. To prevent that, the **Manager** component contains a deduplication logic that leverages the **Persistence** component to filter out already handled system events.

### Usage

To trigger a single run for development purposes, you need to install dependencies first:
```
yarn install
```
Then, build the package:
```
yarn build
```
Finally, run it using Node runtime:
```
node .
```
Keep in mind there are environment variables that must be set. Please refer: https://github.com/keep-network/tbtc-v2/tree/tbtc-monitoring/monitoring#run for details.

### Next steps

Future plans are described in https://github.com/keep-network/pm/issues/33#issuecomment-1427983334


monitoring/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@keep-network/tbtc-v2-monitoring",
+  "version": "1.0.0-dev",
+  "license": "MIT",
```
pdyraga, line 4: Nitpicking but let's use `GPL-3.0-only` for all the new stuff, this is aligned with [tbtc-v2 code](https://github.com/keep-network/tbtc-v2/blob/main/solidity/package.json#L4).

monitoring/src/contracts.ts
```diff
@@ -0,0 +1,42 @@
...
+  let packageName: string
+
+  switch (context.environment) {
+    case Environment.Mainnet: {
+      packageName = "@keep-network/tbtc-v2-mainnet"
+      break
+    }
+    case Environment.Testnet: {
+      packageName = "@keep-network/tbtc-v2-testnet"
+      break
```
pdyraga, line 16: I am confused about this part. Are we referencing the smart contracts package? We are not publishing them under those names.

monitoring/src/deposit-monitor.ts
```diff
@@ -0,0 +1,52 @@
...
+})
+
+export class DepositMonitor implements IncidentMonitor {
+  private bridge: Bridge
+
+  constructor(bridge: Bridge) {
+    this.bridge = bridge
+  }
+
+  async check(fromBlock: number, toBlock: number): Promise<Incident[]> {
```
pdyraga, line 31: I would consider separating *Notifications* from *Incidents* and have the severity only for *Incidents*.

I think an industry standard is to treat *Incidents* as something major. For example, in [Google metrics docs](https://cloud.google.com/monitoring/alerts/incidents-events) we have:

> An incident, also called an alert, is a record of the triggering of an alerting policy. Unless an alerting policy is snoozed or disabled, Cloud Monitoring opens an incident when a condition of an alerting policy is triggered. The incident contains information that you can use to investigate the cause of the notification.

Another example from [PagerDuty docs](https://support.pagerduty.com/docs/incidents):
> An incident represents a problem or an issue that needs to be addressed and resolved. Incidents trigger on services, and a service’s escalation policy prompts notifications to go out to on-call responders to remediate the issue.

In the context of tBTC v2, deposit revealed and large deposit revealed are notifications and they may have different importance. An incident would be, for example, optimistic minting requested too early.

pdyraga (commented):
I like how you modeled monitors and receivers and how the code is organized. I would consider separating incidents into notifications and incidents. Other than that, it all looks good to me.

monitoring/src/contracts.ts
lukasz-zimnoch: Look at the `package.json`, those are aliases over specific package versions.

monitoring/src/deposit-monitor.ts
lukasz-zimnoch: My reasoning: I wanted to have a single representation of a monitoring "event" and do the differentiation using severity (for now it's "minor", "major" and "critical" but I'm thinking about adding "notification" or "informational" as well). A single type is much easier to handle for monitors and receivers and helps keep the code simple. It looks like the name `Incident` is just unfortunate but I wanted to avoid naming it `Event` as it seems to be too broad and generic. I'll re-think the approach.

lukasz-zimnoch: Sure, I'll just call it `SystemEvent` as we also have chain events and the JS built-in `Event` type.

lukasz-zimnoch: See 053036454b1cd94f250698d6c438b439cd3ee593

monitoring/package.json
lukasz-zimnoch: See 2d54deee

monitoring/src/discord-receiver.ts
```diff
@@ -0,0 +1,53 @@
...
+      value: entry[1],
+    }))
+
+    fields.push({ name: "block", value: `${systemEvent.block}` })
+
+    const discordEmbeds = [
+      {
+        title: systemEvent.title,
+        color: 0x003399, // Blue as everything is informational.
+        fields,
```
pdyraga, line 41: Would be great to add an additional field that would format `amountSat` into human-readable value in BTC.

monitoring/src/file-persistence.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (!(await this.db.exists(handledSystemEventsPath))) {
+      return {}
+    }
+
+    return this.db.getObject<Record<SystemEventReceiverId, SystemEvent[]>>(
+      handledSystemEventsPath
+    )
+  }
+
+  // TODO: Consider deleting old events to optimize database file size.
```
pdyraga, line 47: +1

pdyraga (approved):
Awesome work!

PR #533, opened 2023-02-14
Expose additional event getters required by the monitoring system

lukasz-zimnoch:

Here we expose some additional getters in the `typescript` library: `TBTCVault.getOptimisticMintingCancelledEvents` and `Bridge.getNewWalletRegisteredEvents`. The monitoring system is interested in these events but the library has not exposed them so far.

### DIFF
--- a/typescript/src/optimistic-minting.ts
+++ b/typescript/src/optimistic-minting.ts
@@ -40,6 +40,22 @@ export type OptimisticMintingRequestedEvent = {
 } & Event
 
 /**
+ * Represents an event that is emitted when an optimistic minting request
+ * is cancelled on chain.
+ */
+export type OptimisticMintingCancelledEvent = {
+  /**
+   * Guardian's chain identifier.
+   */
+  guardian: Identifier
+  /**
+   * Unique deposit identifier.
+   * @see Bridge.buildDepositKey
+   */
+  depositKey: BigNumber
+} & Event
+
+/**
  * Represents optimistic minting request for the given deposit revealed to the
  * Bridge.
  */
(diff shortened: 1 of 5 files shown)


typescript/src/optimistic-minting.ts
```diff
@@ -39,6 +39,22 @@ export type OptimisticMintingRequestedEvent = {
...
+export type OptimisticMintingCancelledEvent = {
+  /**
+   * Guardian's chain identifier.
+   */
+  guardian: Identifier
+  /**
+   * Unique deposit identifier.
+   * @see Bridge.buildDepositKey
+   */
+  depositKey: BigNumber
```
nkuba, line 55: Could this be a `Hex`? `buildDepositKey` returns a string with `0x` prefix.

nkuba (changes requested):

typescript/src/optimistic-minting.ts
lukasz-zimnoch: A deposit key is an `uint256` in the `Bridge` contract so we should rather make `buildDepositKey` return a `BigNumber`. Left a TODO about that: d6e9352b

nkuba: The value of the deposit key is a result of `solidityKeccak256`. It feels natural to store the digest as hex.

lukasz-zimnoch: Honestly, I don't see any serious difference in how we handle it within typescript, we can always display both representations according to the given case. I just wanted to keep it aligned with the contract-side representation but if you think it will be better to keep it as `Hex`, here it is: 30587e04

nkuba: Thanks!

nkuba (approved):

PR #534, opened 2023-02-14
Electrum connection fallbacks

nkuba:
In this PR we add a possibility to define a list of electrum servers that the client should use for a connection.
The client will try to connect to the servers in the same order as they are provided, so it's the consumer's responsibility to shuffle the fallback servers if needed.


typescript/src/electrum.ts
```diff
@@ -120,19 +138,51 @@ export class Client implements BitcoinClient {
...
+      try {
+        electrum = await connect(credentials)
+      } catch (err) {
+        console.warn(
+          `failed to connect to electrum server: [${credentials.protocol}://${credentials.host}:${credentials.port}]: ${err}`
+        )
+        continue
+      }
+      break
+    }
```
tomaszslabon, line 182: Would a loop like this be a bit simpler to read?
```
    for (const credentials of this.credentials) {
      try {
        electrum = await connect(credentials)
        break
      } catch (err) {
        console.warn(
          `failed to connect to electrum server: [${credentials.protocol}://${credentials.host}:${credentials.port}]: ${err}`
        )
      }
    }
```

There would only be a `break` without `continue`.

nkuba: Thanks for the suggestion! Done in e3056d8627a887693d69104b501983861d831453

typescript/src/electrum.ts
```diff
@@ -120,19 +138,50 @@ export class Client implements BitcoinClient {
...
-      throw new Error(`Electrum server connection failure: [${error}]`)
+
+      return electrum
+    }
+
+    let electrum: Electrum | undefined = undefined
+    for (const credentials of this.credentials) {
+      try {
+        electrum = await connect(credentials)
+        break
```
dimpar, line 175: Do we always throw an error because of any trouble with the connection? If a connection returns because of some other reason and will not **throw** an error, then I think this `break` would prevent taking another credential from the credentials array.

nkuba: Within `connect` function we try to establish connection to the server and send a `server_ping` request to test it. In case of problems or timeout we throw an error:
https://github.com/keep-network/tbtc-v2/blob/feb7cfe1c98a68a2ec8f19749926e12d9f39ad63/typescript/src/electrum.ts#L149-L168

dimpar: Ok, as long as the error is thrown and not just simply returned, then I think we're fine.

tomaszslabon (approved):

dimpar (approved):

PR #535, opened 2023-02-14
Dockerfile and CI workflow for chain events monitoring tool

lukasz-zimnoch:
Depends on: #526

Here we Dockerize the monitoring component and provide a CI workflow to automate builds and check formatting.

### Dockerfile

The Dockerfile is a standard one for Node-based applications. We used some tricks to reduce the final image size:
- Multi-stage build
- Copy only essential files
- Install only production dependencies by pruning the dev ones
- Additional `node_modules` pruning using `node-prune` tool

This reduced the image size from the initial ~3.5 GB to 1.41 GB (virtual size is 200 MB). 

### GitHub Actions workflow

We added the `Monitoring` workflow that consists of the following jobs:
- `monitoring-detect-changes`: triggered for pull request events in order to detect changes in the `./monitoring` directory
- `monitoring-format`: triggered for pull requests if `monitoring-detect-changes` returns true or for pushes to `main` branch that includes changes to `./monitoring` directory. Checks the formatting in the `./monitoring` module
- `monitoring-build`: triggered in same cases as `monitoring-format` and also for workflow dispatch events and scheduled runs. Builds the typescript code in `./monitoring` module
- `monitoring-docker`: triggered only for workflow dispatch events (i.e. manually). Builds the Docker image for `./monitoring` module and can push it to a Docker registry. The exact behavior depends on the `environment` input variable value:
    - `environment=local` (default case): just builds the Docker image without pushing ([example run](https://github.com/keep-network/tbtc-v2/actions/runs/4185037539))
    - `environment=goerli`: builds the Docker image and pushes it to `keep-test` GCR ([example run](https://github.com/keep-network/tbtc-v2/actions/runs/4185304489)) 
    - `environment=mainnet`: builds the Docker image and pushes it to `keep-prd` GCR (not configured yet, will be possible once `mainnet.env` is added to https://github.com/keep-network/ci/tree/main/config/env and `KEEP_PRD_GCR_JSON_KEY` secret is set)


monitoring/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+RUN yarn build
+
+# Prune development dependencies.
+RUN rm -rf ./node_modules
+RUN rm -rf ./external
+RUN yarn install --production --frozen-lockfile --ignore-scripts
+RUN yarn run postinstall
+
+# Prune other unnecessary files from dependencies using node-prune.
+RUN node-prune
```
nkuba, line 39: Out of curiosity what's the image size reduction we achieved thanks to the `node-prune`?

lukasz-zimnoch: IIRC smth around 300MB

monitoring/.dockerignore
```diff
@@ -0,0 +1,20 @@
...
+docs/
+out/
+README.adoc
+
+# Node
+external/
+node_modules/
+
+# Make
+Makefile
```
pdyraga, line 20: Should we cleanup this file a bit? For example:
- there is no `Makefile` for this module,
- we have neither `example` nor `config`.

.github/workflows/monitoring.yml
```diff
@@ -0,0 +1,147 @@
+name: Monitoring
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
```
pdyraga, line 5: Why do we need this schedule?

lukasz-zimnoch: It's a common pattern in this repository, each submodule is built every night in order to have up-to-date status badges in the README file.

monitoring/.dockerignore
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/535/commits/48c3768fd09a58d896b4a50f63e37b2fc34b0050

pdyraga (approved):
🚀

PR #537, opened 2023-02-15
Validating Bitcoin transaction proof

tomaszslabon:
Refs: https://github.com/keep-network/optimistic-minting/issues/6

This PR adds functionalities for validating transaction inclusion in the Bitcoin blockchain:
- the data for the proof is taken from the electrum client
- the Merkle tree is validated. This ensures the transaction was included in the first block
- the chain of block headers are validated. This ensures the block in which  the transaction entered the blockchain has confirmations. 

Testnet transaction are treated specially: when the passed previous/current difficulties are both set to `1`, we skip
difficulty checks. This is due to difficulty in Testnet frequently dropping to `1`.

One thing that can be problematic is using `LightRelay` as the source of current/previous difficulties.
It takes some time before `LightRelay` is updated after a new Bitcoin epoch begins.
Therefore the current/previous difficulties in `LightRelay` sometimes lag the real ones.


typescript/src/bitcoin.ts
```diff
@@ -374,6 +481,27 @@ export function computeHash160(text: string): string {
   return hash160.digest(Buffer.from(text, "hex")).toString("hex")
 }
 
+/**
+ * Computes the double SHA256 for the given text.
+ * @param text - Text the double SHA256 is computed for.
+ * @returns Hash as a 32-byte un-prefixed hex string.
+ */
+export function computeHash256(text: string): string {
```
nkuba, line 489: How about returning `Hex` type here?

typescript/src/bitcoin.ts
```diff
@@ -374,6 +481,27 @@ export function computeHash160(text: string): string {
...
+}
+
+/**
+ * Converts a hash in hex string in little endian to a BigNumber.
+ * @param hash - Hash in hex-string format.
+ * @returns BigNumber representation of the hash.
+ */
+export function hashLEToBigNumber(hash: string): BigNumber {
+  return BigNumber.from(
+    "0x" + Buffer.from(hash, "hex").reverse().toString("hex")
```
nkuba, line 501:
```suggestion
    Hex.from(hash, "hex").reverse().toString()
```

nkuba (commented):
First review round

typescript/src/bitcoin.ts
```diff
@@ -172,6 +173,112 @@ export interface TransactionMerkleBranch {
...
+  nonce: number
+}
+
+/**
+ * Deserializes a block header in the raw representation to BlockHeader.
+ * @param rawBlockHeader - BlockHeader in the raw format.
+ * @returns Block header as a BlockHeader.
+ */
+export function deserializeBlockHeader(rawBlockHeader: string): BlockHeader {
+  const buffer = Buffer.from(rawBlockHeader, "hex")
```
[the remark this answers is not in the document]
tomaszslabon: Done in 072614a49f7d9c2f96b1d51f6f5f4cff2b8ae64d.

typescript/src/bitcoin.ts
tomaszslabon: Done in 072614a49f7d9c2f96b1d51f6f5f4cff2b8ae64d.

typescript/src/proof.ts
```diff
@@ -70,3 +79,253 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+
+    // Save the current block header hash to compare it with the next block
+    // header's previous block header hash.
+    previousBlockHeaderHash = currentBlockHeaderHash
+
+    // Check if the stored block difficulty is equal to previous or current
+    // difficulties.
+    const difficulty = targetToDifficulty(difficultyTarget)
+
+    if (previousEpochDifficulty.eq(1) && currentEpochDifficulty.eq(1)) {
```
tomaszslabon, line 287: This is a special case for Bitcoin Testnet.
When both difficulties are set to `1` we skip difficulties check.
I assumed the relay used in our testnet environment always returns `1` for current and previous difficulties.

We could also temporarily pass `1` as previous and current difficulties for Mainnet, unit we have the `LightRelay` up and running.

typescript/src/bitcoin.ts
```diff
@@ -172,6 +173,112 @@ export interface TransactionMerkleBranch {
...
+}
+
+/**
+ * Converts a block header's bits into difficulty target.
+ * @param bits - bits from block header.
+ * @returns Difficulty target as a BigNumber.
+ */
+export function bitsToDifficultyTarget(bits: number): BigNumber {
+  const exponent = ((bits >>> 24) & 0xff) - 3
+  const mantissa = bits & 0xffffff
```
lukasz-zimnoch, line 262: I think it deserves a detailed explanation:
```
// A serialized 80-byte block header stores the `bits` value as a 4-byte little-endian hexadecimal value in a slot
// including bytes 73, 74, 75, and 76. This function's input argument is expected to be a numerical representation 
// of that 4-byte value reverted to the big-endian order. For example, if the `bits` little-endian value in the header 
// is `0xcb04041b`, it must be reverted to the big-endian form `0x1b0404cb` and turned to a decimal number 
// `453248203` in order to be used as this function's input.
//
// The `bits` 4-byte big-endian representation is a compact value that works like a base-256 version of 
// scientific notation. It encodes the target exponent in the first byte and the target mantissa in the last three 
// bytes. Referring to the previous example, if `bits = 453248203`, the hexadecimal representation is 
// `0x1b0404cb` so the exponent is `0x1b` while the mantissa is `0x0404cb`.
//
// To extract the exponent, we need to shift right by 3 bytes (24 bits), extract the last byte of the result, and
// subtract 3 (because of the mantissa length):
// - 0x1b0404cb >>> 24 = 0x0000001b
// - 0x0000001b & 0xff = 0x1b
// - 0x1b - 3 = 24 (decimal)
//
// To extract the mantissa, we just need to take the last three bytes:
// - 0x1b0404cb & 0xffffff = 0x0404cb = 263371 (decimal)
// 
// The final difficulty can be computed as mantissa * 256^exponent:
// - 263371 * 256^24 = 
//
// Sources:
// - https://developer.bitcoin.org/reference/block_chain.html#target-nbits
// - https://wiki.bitcoinsv.io/index.php/Target

const exponent = ((bits >>> 24) & 0xff) - 3
const mantissa = bits & 0xffffff

const difficultyTarget = BigNumber.from(mantissa).mul(
    BigNumber.from(256).pow(exponent)
)
return difficultyTarget
```

typescript/src/bitcoin.ts
```diff
@@ -374,6 +481,26 @@ export function computeHash160(text: string): string {
   return hash160.digest(Buffer.from(text, "hex")).toString("hex")
 }
 
+/**
+ * Computes the double SHA256 for the given text.
+ * @param text - Text the double SHA256 is computed for.
+ * @returns Hash as a 32-byte un-prefixed hex string.
+ */
+export function computeHash256(text: string): Hex {
```
lukasz-zimnoch, line 489: Worth making `text` input a `Hex`. We are interpreting it as `Buffer.from(text, "hex")` anyway.

typescript/src/bitcoin.ts
```diff
@@ -374,6 +481,26 @@ export function computeHash160(text: string): string {
...
+  const secondHash = sha256.digest(firstHash)
+  return Hex.from(secondHash)
+}
+
+/**
+ * Converts a hash in hex string in little endian to a BigNumber.
+ * @param hash - Hash in hex-string format.
+ * @returns BigNumber representation of the hash.
+ */
+export function hashLEToBigNumber(hash: string): BigNumber {
```
lukasz-zimnoch, line 500: Can't we just make `hash: Hex`?

typescript/src/proof.ts
```diff
@@ -38,7 +47,7 @@ export async function assembleTransactionProof(
 
   const headersChain = await bitcoinClient.getHeadersChain(
     txBlockHeight,
-    requiredConfirmations
+    requiredConfirmations - 1
```
lukasz-zimnoch, line 50: Worth clarifying why we need to subtract one here.

typescript/src/proof.ts
```diff
@@ -70,3 +79,253 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+ *        should compute to.
+ * @param intermediateNodeHashes The Merkle tree intermediate node hashes,
+ *        concatenated as a single string.
+ * @param transactionIndex The index of the transaction being validated within
+ *        the block, used to determine the path to traverse in the Merkle tree.
+ * @throws {Error} If the Merkle tree is not valid.
+ * @returns An empty return value.
+ */
+function validateMerkleTree(
+  transactionHash: string,
```
lukasz-zimnoch, line 153: We can do
```suggestion
  transactionHash: TransactionHash,
```
and reverse the hash at the very bottom of the call chain. (`validateMerkleTreeHashes`).

typescript/src/proof.ts
```diff
@@ -70,3 +79,253 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+  merkleRootHash: string,
```
lukasz-zimnoch, line 154:
```suggestion
  merkleRootHash: Hex,
```

typescript/src/proof.ts
```diff
@@ -70,3 +79,253 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+  intermediateNodeHashes: string,
```
lukasz-zimnoch, line 155: How about:
```suggestion
  intermediateNodeHashes: Hex,
```

typescript/src/proof.ts
```diff
@@ -70,3 +79,253 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+  transactionIndex: number
+) {
+  // Shortcut the empty-block case
+  if (
+    transactionHash == merkleRootHash &&
+    transactionIndex == 0 &&
+    intermediateNodeHashes.length == 0
+  ) {
+    return
+  }
```
lukasz-zimnoch, line 165: Worth elaborating on why we check those specific cases.

typescript/src/proof.ts
```diff
@@ -70,3 +79,253 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+ * @param merkleRootHash The Merkle root hash that the intermediate nodes should
+ *        compute to.
+ * @param transactionIndex The index of the transaction in the block, used
+ *        to determine the path to traverse in the Merkle tree.
+ * @throws {Error} If the intermediate nodes are of an invalid length or if the
+ *         computed root hash does not match the merkle root hash parameter.
+ * @returns An empty return value.
+ */
+function validateMerkleTreeHashes(
+  transactionHash: string,
```
lukasz-zimnoch, line 192: We can do
```suggestion
  transactionHash: TransactionHash,
```
and reverse the hash within this function, right in the place of use.

typescript/src/proof.ts
```diff
@@ -70,3 +79,253 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+  intermediateNodesHashes: string,
```
lukasz-zimnoch, line 193:
```suggestion
  intermediateNodesHashes: Hex,
```

typescript/src/proof.ts
```diff
@@ -70,3 +79,253 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+  merkleRootHash: string,
```
lukasz-zimnoch, line 194:
```suggestion
  merkleRootHash: Hex,
```

typescript/src/proof.ts
```diff
@@ -70,3 +79,253 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+  ) {
+    return
+  }
+
+  validateMerkleTreeHashes(
+    transactionHash,
+    intermediateNodeHashes,
+    merkleRootHash,
+    transactionIndex
+  )
```
lukasz-zimnoch, line 172: Let's try to keep the same argument order between `validateMerkleTree` and `validateMerkleTreeHashes`

typescript/src/proof.ts
```diff
@@ -70,3 +79,253 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+      current = computeHash256(
+        intermediateNodesHashes.slice(i, i + 64) + current
+      ).toString()
+    } else {
+      current = computeHash256(
+        current + intermediateNodesHashes.slice(i, i + 64)
+      ).toString()
+    }
+    idx = idx >> 1
+  }
```
lukasz-zimnoch, line 219: I'd love to have a detailed explanation of this loop. Looks really cryptic.

typescript/src/proof.ts
```diff
@@ -70,3 +79,253 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+
+    // Check if the current block header stores the hash of the previously
+    // processed block header. Skip the check for the first header.
+    if (index !== 0) {
+      if (
+        !previousBlockHeaderHash.equals(currentHeader.previousBlockHeaderHash)
+      ) {
+        throw new Error("Invalid headers chain")
+      }
+    }
```
lukasz-zimnoch, line 264: We are basically checking the chain integrity here, right?

typescript/src/proof.ts
```diff
@@ -70,3 +79,253 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+        "Header difficulty not at current or previous Bitcoin difficulty"
+      )
+    }
+
+    // Additionally, require the header to be at current difficulty if some
+    // headers at current difficulty have already been seen. This ensures
+    // there is at most one switch from previous to current difficulties.
+    if (requireCurrentDifficulty && !difficulty.eq(currentEpochDifficulty)) {
+      throw new Error("Header must be at current Bitcoin difficulty")
+    }
```
lukasz-zimnoch, line 307: I see we are doing it differently than the SPV solidity lib does (https://github.com/keep-network/tbtc-v2/blob/8b2b2b97c856b4bed42a97788cfa2af8b6aba8d8/solidity/contracts/bridge/BitcoinTx.sol#L227). We are checking headers separately while the lib checks the accumulated difficulty. Could you explain the exact reasons?

typescript/test/data/proof.ts
```diff
@@ -336,3 +336,265 @@ export const multipleInputsProofTestData: ProofTestData = {
...
+    previousDifficulty: BigNumber.from(37590453655497),
+    currentDifficulty: BigNumber.from(39350942467772),
+  },
+}
+
+/**
+ * Test data that is based on a random Bitcoin testnet transaction
+ * https://live.blockcypher.com/btc-testnet/tx/b78636ae08e6c17261a9f3134109c13c2eb69f6df52e591cc0e0780f5ebf6472/
+ */
+export const testnetTransactionData: TransactionProofData = {
```
lukasz-zimnoch, line 534: How that differs from `transactionConfirmationsInOneEpochData`?

lukasz-zimnoch (changes requested):

typescript/src/bitcoin.ts
```diff
@@ -172,6 +173,112 @@ export interface TransactionMerkleBranch {
...
+   */
+  nonce: number
+}
+
+/**
+ * Serializes a BlockHeader to the raw representation.
+ * @param blockHeader - block header.
+ * @returns Serialized block header.
+ */
+export function serializeBlockHeader(blockHeader: BlockHeader): string {
```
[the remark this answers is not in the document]
tomaszslabon: Done in dde5ac56613658b93e2f4f67c0474b5399717ff4.

typescript/src/bitcoin.ts
```diff
@@ -172,6 +173,112 @@ export interface TransactionMerkleBranch {
...
+    nonce: nonce,
+  }
+}
+
+/**
+ * Converts a block header's bits into difficulty target.
+ * @param bits - bits from block header.
+ * @returns Difficulty target as a BigNumber.
+ */
+export function bitsToDifficultyTarget(bits: number): BigNumber {
```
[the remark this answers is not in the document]
tomaszslabon: Makes sense. Done in 742b44f0937ebedb302335d8e32331a4699f6921.

typescript/src/bitcoin.ts
```diff
@@ -172,6 +173,112 @@ export interface TransactionMerkleBranch {
   position: number
 }
 
+/**
+ * BlockHeader represents the header of a Bitcoin block. For reference, see:
+ * https://developer.bitcoin.org/reference/block_chain.html#block-headers.
+ */
+export interface BlockHeader {
```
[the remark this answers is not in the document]
tomaszslabon: Done in c09301bab376c3fba3c6c8670972ea1ad5d27a29.

typescript/src/bitcoin.ts
tomaszslabon: Added the description in c41eb0627aaced1f9014990e4f3425dc35d49590.

typescript/src/proof.ts
tomaszslabon: Yes.
The block headers need to be "connected": each block should point to the previously processed block header (except for the first one).

typescript/src/bitcoin.ts
tomaszslabon: Done in 2b0b653d5d4131cc0c7b60bf2a42396a6af2eb0a.

typescript/src/proof.ts
tomaszslabon: Done in 57c55f4e3fc5cb0446dafeb566e274134ba9d77c.

typescript/src/proof.ts
tomaszslabon: Done in 9fea7d9b2acb15fd1b4974b949c6c3c08ef7bda9.

typescript/src/proof.ts
tomaszslabon: Done in 57c55f4e3fc5cb0446dafeb566e274134ba9d77c.

typescript/src/proof.ts
tomaszslabon: Done in 43cc758b27fb9c0c77ee7cb16d91fdd97fe6cde0.

typescript/src/proof.ts
tomaszslabon: Done in 7fdc5c0618bb05d9e1a915e6fa27210a494efb36.

typescript/src/proof.ts
tomaszslabon: Done in 3488910e67c7c1147fbf8ee0cb34a2c23bd269b9.

typescript/src/proof.ts
tomaszslabon: In the solidity implementation there can be a situation when the current difficulty is lower than the previous difficulty and the first header is in the previous difficulty epoch and subsequent headers are in the current difficulty epoch.
In that case, the total accumulated difficulty will be lower than the required difficulty and the validation will fail.
Therefore, we had to either ensure all the headers in the chain are in the same epoch or we must provide a longer headers chain.

This typescript implementation is stricter and more precise. We simply verify the difficulty of each header separately: the difficulty must be at previous or current difficulty. Additionally, while iterating over the headers we ensure that the possible switch from previous difficulty headers and current difficulty ones happens at most one time. We can also precisely verify that the provided headers chain's length equals exactly the required number of confirmations.

typescript/src/proof.ts
tomaszslabon: I added explanatory comment in a1e54032d0e30095c9a329f2f6fb93a55f81cd18.

typescript/src/proof.ts
lukasz-zimnoch: Makes sense! Thanks!

typescript/test/data/proof.ts
lukasz-zimnoch: Ok, fine!

lukasz-zimnoch (approved):

typescript/src/proof.ts
tomaszslabon: Added explanation in c673dca748ec6ba23bf54b260b4e7cd083e2a2c5.

typescript/src/proof.ts
```diff
@@ -70,3 +85,315 @@ function createMerkleProof(txMerkleBranch: TransactionMerkleBranch): string {
...
+ *         computed root hash does not match the merkle root hash parameter.
+ * @returns An empty return value.
+ */
+function validateMerkleTreeHashes(
+  transactionHash: TransactionHash,
+  merkleRootHash: Hex,
+  intermediateNodeHashes: Hex[],
+  transactionIndex: number
+) {
+  // To prove the transaction inclusion in a block we only need the hashes that
```
mhluongo, line 207: This would make a good blog post 👀

PR #540, opened 2023-02-16
Wallet and minting monitors for chain events monitoring tool

lukasz-zimnoch:
Depends on: https://github.com/keep-network/tbtc-v2/pull/539

Here we add two new monitors that generate the following system events:
- `Wallet registered (informational)`: Informs about a new wallet being registered on the `Bridge`
- `Optimistic minting cancelled (warning)`: Warns about a cancelled optimistic minting request
- `Optimistic minting requested too early (critical)`: Warns about an optimistic minting request that was issued too early regarding its funding BTC transaction confirmation status
- `Optimistic minting requested for undetermined Bitcoin transaction (critical)`: Warns about an optimistic minting request the funding transaction info cannot be determined for


monitoring/src/wallet-monitor.ts
```diff
@@ -1,11 +1,41 @@
...
+
+    const chainEvents = await this.bridge.getNewWalletRegisteredEvents({
+      fromBlock,
+      toBlock,
+    })
+
+    // eslint-disable-next-line no-console
+    console.log("completed wallet monitor check")
+
+    return chainEvents.map(WalletRegistered)
```
pdyraga, line 39: This is a good start but I think we should also monitor:
- DKG started,
- DKG result submitted,
- DKG result challenged (critical).

DKG result approved is the same as this one.

monitoring/src/minting-monitor.ts
```diff
@@ -1,11 +1,181 @@
...
+      const confirmation = confirmations[index]
+
+      switch (confirmation.status) {
+        case "fulfilled": {
+          const actualConfirmations = confirmation.value
+          const requiredConfirmations = this.requiredConfirmations(
+            ce.amount.div(satoshiMultiplier)
+          )
+
+          if (actualConfirmations < requiredConfirmations) {
```
pdyraga, line 141: Note that this logic is valid for the time being but may require updates.

Right now, @nkuba is working on an update to Minter to request optimistic minting earlier and Guardian to cancel if conditions are not fulfilled by the end of the optimistic minting delay.

lukasz-zimnoch: Yep, I'm aware. We must remember that the monitoring must evolve along with the monitored components and adjust things when the time comes.

monitoring/src/wallet-monitor.ts
lukasz-zimnoch: Yeah, this is the initial version. We need to plan the next steps regarding monitoring work and maintain a backlog of things we want to add.

dimpar (commented):
lgtm, just one minor comment

monitoring/src/minting-monitor.ts
```diff
@@ -1,11 +1,181 @@
...
+  title: "Optimistic minting requested for undetermined Bitcoin transaction",
+  type: SystemEventType.Critical,
+  data: {
+    minter: `0x${chainEvent.minter.identifierHex}`,
+    depositKey: chainEvent.depositKey.toPrefixedString(),
+    depositor: `0x${chainEvent.depositor.identifierHex}`,
+    amountSat: chainEvent.amount.div(satoshiMultiplier).toString(),
+    btcFundingTxHash: chainEvent.fundingTxHash.toString(),
+    btcFundingOutputIndex: chainEvent.fundingOutputIndex.toString(),
+    btcClientResponse,
```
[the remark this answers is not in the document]
lukasz-zimnoch: This is ES6 feature called object shorthand, see https://eslint.org/docs/latest/rules/object-shorthand

dimpar (approved):

Issue #509, opened 2023-01-30
TS: yarn install freezes the system | typescript


Revantark:
Trying to integrate tbtc-v2.ts in my project, but when i do 
`yarn install`
after a while, the system freezes.

It is happening in the postInstall script `npm rebuild`

To reproduce :

1. clone tbtc-v2 repo;
2. cd typescript
3. run `yarn install`

this just hangs forever

pdyraga:
What is the Node and Yarn version you are using? Please check [the build prerequisites](https://github.com/keep-network/tbtc-v2/tree/main/solidity#prerequisites) section in the docs.

Revantark:
> What is the Node and Yarn version you are using? Please check [the build prerequisites](https://github.com/keep-network/tbtc-v2/tree/main/solidity#prerequisites) section in the docs.

Node : v18.12.1
yarn: v1.22.19

lukasz-zimnoch:
@Revantark, thanks for reporting! The problem you observed sometimes happens when using Node versions greater than 14. We need to investigate it and develop a solution. As a short-term workaround, I recommend you use Node 14. Sorry for the inconvenience!

Revantark:
> 

`node-gyp` rebuild fails on `bcrypto` package.

Using node 14

Revantark:
> @Revantark, please make sure you have all OS-specific prerequisites of `node-gyp` listed here: https://github.com/nodejs/node-gyp#installation

@lukasz-zimnoch 
I build the package successfully thanks.

But when using it in a project, I'm getting this error
`Cannot find module 'bcrypto.node'` from tbtc-v2.ts

I'm using yarn to link the local package to a project.

Revantark:
@lukasz-zimnoch Yes, I was doing it. But still i get `Cannot find module 'bcrypto.node'` error in a node project.

How to reproduce :

yarn: 1.22.19
node: v14

1. create an empty node project using `npm init`
2. run `npm install @keep-network/tbtc-v2.ts`
3. run `npm rebuild`
4. run  `npm start`

Error: Cannot find module 'bcrypto.node'

pdyraga:
@lukasz-zimnoch In which places do we need `bcrypto` in `tbtc-v2.ts`? In the worst case, we can copy only the code needed.

lukasz-zimnoch:
> @lukasz-zimnoch In which places do we need `bcrypto` in `tbtc-v2.ts`? In the worst case, we can copy only the code needed.

[...]

PR #542, opened 2023-02-17
RFC 8: Cross-chain Tokenized Threshold BTC

pdyraga:
The goal of this proposal is to enable *efficient* cross-chain tBTC liquidity.

In the Ethereum ecosystem, that means bringing Bitcoin to L2s and sidechains like Arbitrum, Optimism, Polygon, Avalanche, and Binance Smart Chain. In the Cosmos ecosystem, that means Bitcoin on IBC-enabled zones (chains) like Osmosis and Penumbra. And in newer ecosystems, that means chains like Solana, Aptos, and Sui.

Efficiency requires minimizing liquidity fragmentation by avoiding the need for tBTC wrapper tokens by each chain and bridge. This situation quickly leads to an `n*m` problem, where liquidity on `n` chains is divided between `m` bridges. Each of these token variants soaks up valuable liquidity that could be better utilized elsehwere.

Because tBTC is minted "centrally" on Ethereum L1, we can do better. By deploying a canonical token on each chain, we can ensure the supply of tBTC remains sacrosanct, while enabling quick, interoperable cross-chain bridges and localizing ecosystem risk.


docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+
+=== Canonical token
+
+The canonical token we are going to deploy will be different from the L1 token
+deployed at `0x18084fbA666a33d37592fA2633fD49a74DD93a88`. The L2/sidechain token
+will be upgradeable, owned by Threshold Council, will delegate the minting
+authority to multiple parties, and have a pause functionality for mints and
+burns.
+
+Given those requirements, the address of the canonical TBTC token on L2s and
```
mhluongo, line 63: Nit, but we should stop switching between "tBTC" and "TBTC". I know this conflicts with what we did in v1, but we've agreed in the Threshold DAO Discord to move to "tBTC" everywhere a syntax formatter doesn't make us that "tBTC" > "TBTC".

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+    │      └── OptimismTBTC.sol 
+    └── polygon
+        ├── package.json
+        └── solidity
+           ├── L1PolygonGateway.sol 
+           ├── L2PolygonGateway.sol 
+           └── PolygonTBTC.sol 
+```
+
+`L2TBTC.sol` is an abstract contract doing all the heavy lifting: upgradeability,
```
mhluongo, line 279: What do you think about landing this with a consistent contract address across EVMs? I'd love to see it happen with `CREATE2`. Also, hell... grinding some 0's on the address would save gas :grin:

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+
+Each `cross-chain` project should contain L1 and L2
+contracts specific to that individual subchain. This separation will allow us to
+abstract out the complexity of deployment and avoid redeploying all L1 testnet
+contracts in case a single change in one of L2 contracts is needed.
+
+This organization of the code will also allow us to not add subchain-specific
+dependencies to the L1 project and to deploy NPM packages separately:
+
+* `@keep-network/tbtc-v2-arbitrum`,
```
mhluongo, line 235: Are the multiple NPM packages overkill?

mhluongo (approved):
Left a couple thoughts / questions. Concept ACK, but let's not merge until more people have had a look

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+* Optimism.
+
+The cross-bridge architecture and patterns we are going to choose should allow
+us to scale to other L2s, sidechains, and chains quickly.
+
+=== Canonical token
+
+The canonical token we are going to deploy will be different from the L1 token
+deployed at `0x18084fbA666a33d37592fA2633fD49a74DD93a88`. The L2/sidechain token
+will be upgradeable, owned by Threshold Council, will delegate the minting
```
mswilkison, line 59: We should get started on deploying multisigs on each of those L2s. IIRC there's already one on Optimism that can be updated/repurposed.

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+  present.
+* Delegate restricted minting authority to a short list of ecosystem bridges
+* Be paused by any one of `n` guardians, allowing avoidance of contagion in case
+  of a chain- or bridge-specific incident.
+* Be governed by the Threshold Council until we can land on a longer-term
+  cross-chain governance mechanism.
+
+== Initial implementation
+
+The goal of the initial implementation is to bring Bitcoin to the following L2s
```
mswilkison, line 45: idk if it's worth targeting initially with the other three or deferring until later but zkSync just opened up onboarding applications for their permissioned launch: https://docs.google.com/forms/d/e/1FAIpQLSfdXAoQiHLu16ykEoOb762uvC1bAhdzOG-YDk6ju0YwPWf3Fw/viewform

docs/rfc/rfc-8.adoc
mhluongo: Yes please! Bonus points if you can get the same addresses, though it's a PITA

docs/rfc/rfc-8.adoc
mhluongo: Let's do it

docs/rfc/rfc-8.adoc
pdyraga: Makes sense! Fixed in e740d15932c8b9cedc55d335538b5d197c25ede2.

docs/rfc/rfc-8.adoc
pdyraga: The dev team planned to replay the safe creation transaction on Arbitrum, Polygon, and Optimism as a part of the development work but if you @mswilkison or someone from the DAO wants to do it, please go ahead! It's captured here: https://github.com/keep-network/tbtc-v2/issues/549.

docs/rfc/rfc-8.adoc
pdyraga: I was on the fence about it. We can't have the same token address on L1 and L2 given the code is different. We could have the same address on all L2s but I am afraid this could be more confusing to users than different addresses everywhere. We don't want someone who sees 3 identical addresses on L2 to assume the address is the same on L1. What do you think about it? I am open to `CREATE2`, the potential confusion is my one and only concern.

docs/rfc/rfc-8.adoc
pdyraga: Having it all in one `solidity/` project and one `@keep-network/tbtc-v2` NPM package was my initial take but then I realized there are some complications.

First, if you `npm i @keep-network/tbtc-v2` and go to `node_modules/@keep-network/tbtc-v2/artifacts` and browse JSON artifacts, you will notice there is nothing about chain ID in JSON. Some contracts from the NPM package would be deployed on Ethereum, some others on Optimism, some others on Polygon, and some others on Arbitrum with no sign in the artifact JSON where the contract comes from. There is a potential workaround: we could reflect it in the directory structure, e.g. `artifacts/arbitrum` would be from Arbitrum, `artifacts/optimism` would be from Optimism. Though, currently, the deployer plugin puts all artifacts in a single directory with no nesting. I am not sure if this is easy to change.

Second, there are some dependencies that are L2-specific. Optimism has its modified Solidity compiler, once we start integrating native bridges there will be more. They may require a specific version of node, for example. This is already a pain for tbtc-v2 and bcrypto, see #509, #518.

Finally, I think keeping them in separate packages will make the deployment and development easier. There are already quite a few deployment steps to deploy tbtc-v2 to testnet (see [`solidity/deploy`](https://github.com/keep-network/tbtc-v2/tree/e740d15932c8b9cedc55d335538b5d197c25ede2/solidity/deploy)). If we do it all in one package, we'll have to add L2 deployment to the checklist and then go back to L1 to wire up gateways. Problems on one L2 will fail the entire deployment and NPM package publication. If we keep them separate, we do not have to repeat all the heavy L1 deployment if we need to apply a small fix to a gateway on one L2.

tl;dr; I think we keep the risk under control this way.

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+and sidechains:
+
+* Arbitrum,
+* Polygon,
+* Optimism.
```
beaushinkle, line 50: no need for punctuation here

docs/rfc/rfc-8.adoc
prestwich: I would lean towards having it at the same address everywhere. It saves you from a lot of configuration wrangling for applications. E.g. at nomad each contract address must be [recorded separately](https://github.com/nomad-xyz/config/blob/main/production.json#L474-L553) on each network

Users should (generally) not interact with the address directly unless they're relatively sophisticated, so address choice should be developer-facing and simplify devex

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+  link:https://arbiscan.io/address/0x0b2402144bb366a632d14b83f244d2e0e21bd39c[`0x0b2402144Bb366A632D14B83F244D2e0e21bD39c`].
+* Wormhole `TokenBridge` on Polygon is the contract deployed at
+  link:https://polygonscan.com/address/0x5a58505a96D1dbf8dF91cB21B54419FC36e93fdE[`0x5a58505a96D1dbf8dF91cB21B54419FC36e93fdE`].
+* Wormhole `TokenBridge` on Optimism is the contract deployed at
+  link:https://optimistic.etherscan.io/address/0x1d68124e65fafc907325e3edbf8c4d84499daa8b[`0x1d68124e65fafc907325e3edbf8c4d84499daa8b`].
+* `L2WormholeGateway` on each chain is a smart contract wrapping and unwrapping
+  Wormhole-specific tBTC representation into the canonical tBTC token on the
+  given chain. This contract needs to be implemented and deployed behind an
+  upgradeable proxy.
+* `ArbitrumTBTC`, `PolygonTBTC`, and `OptimismTBTC` are token contracts with
```
prestwich, line 341: what are the limits on the minting authority? Rate limit? Absolute amount? Other? 🤔

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+
+== Overview
+
+To achieve this goal, I propose we deploy a canonical tBTC contract on each
+chain we support. This contract should be flexible enough to:
+
+* Delegate restricted minting authority to a native bridge on the chain, if
+  present.
+* Delegate restricted minting authority to a short list of ecosystem bridges
+* Be paused by any one of `n` guardians, allowing avoidance of contagion in case
```
prestwich, line 38: can we define under what circumstances these guardians should remove minting authority?

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+localizing ecosystem risk.
+
+== Overview
+
+To achieve this goal, I propose we deploy a canonical tBTC contract on each
+chain we support. This contract should be flexible enough to:
+
+* Delegate restricted minting authority to a native bridge on the chain, if
+  present.
+* Delegate restricted minting authority to a short list of ecosystem bridges
```
prestwich, line 37: The method for delegation is to allow limited wrapping of the bridge token to the canonical token?

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+* Be paused by any one of `n` guardians, allowing avoidance of contagion in case
+  of a chain- or bridge-specific incident.
+* Be governed by the Threshold Council until we can land on a longer-term
+  cross-chain governance mechanism.
```
beaushinkle, line 41: I would love to see the explicit UX plan in the overview: that initially we're minting normally on ethereum, and then using a separate UI (wormhole) to bridge to L2s, upgrading the bridged tokens into the canonical token that we're deploying

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+This is out of the scope of the initial implementation.
+
+See link:https://github.com/ethereum-optimism/optimism-tutorial/tree/main/standard-bridge-standard-token[Optimism guide for the standard bridge]
+and link:https://community.optimism.io/docs/developers/bridge/messaging/#fees-for-l2-%E2%87%92-l1-transactions[Optimism documentation on sending data between L1 and L2]
+for more details.
+
+=== Supporting cross-ecosystem bridges
+
+The Wormhole bridge will be the first cross-ecosystem bridge to have a minting
+authority for tBTC token on Arbitrum, Polygon, and Optimism.
```
conorpp, line 146: What about other EVM chains and non-EVM chains?  Without a canonical contract on other chains, the Wormhole token representation (or another bridge) will be used and may stay that way for a long time if a lot of liquidity sinks in.  Is there a plan to try to migrate this or recommend a bridge as canonical?

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+To fully use the capabilities of the Wormhole portal and make the user
+experience seamless, we will implement a gateway contract having the authority to
+wrap Wormhole-specific tokens into tBTC tokens in a 1:1 ratio. This contract
+will have a minting authority for tBTC on L2. This way, no liquid market has to
+exist on any target chain for users to be able to cross the Wormhole bridge into
+the canonical tBTC.
+
+IMPORTANT: The requirement for a seamless user experience is that the Wormhole
+team has to integrate the step of wrapping the token using the gateway contract
+into the Wormhole bridge flow, in the same transaction as the L2 confirmation.
```
conorpp, line 199: How can this be done in the same transaction?  I believe it would need to be done in two transactions.

Option 1:
a. First, redeem "whtBTC" on L2 address using Wormhole token contract on L2.
b. Second, redeem whtBTC for tBTC on L2 using canonical token contract on L2.

Option 1:
a. First, give approval to canonical token contract to spend "whtBTC" for user on L2.
b. Second, redeem whtBTC onto the canonical token contract which can then atomically swap for tBTC.

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,410 @@
...
+Wormhole cannot magically mint all tokens on Arbitrum or other L2s. Wormhole
+mints its own token representing the bridged asset that may be swapped to the
+canonical representation on the given L2 assuming there is a market for such
+a swap. The overview of liquid markets on each target chain is available in the
+link:https://docs.wormhole.com/wormhole/overview-liquid-markets[Wormhole documentation].
+Not every chain and token pair has a market. 
+
+To fully use the capabilities of the Wormhole portal and make the user
+experience seamless, we will implement a gateway contract having the authority to
+wrap Wormhole-specific tokens into tBTC tokens in a 1:1 ratio. This contract
```
conorpp, line 192: Users will be able to send L2 tBTC to ETH using wormhole.  This will make a wormhole wrapped tBTC representation on ETH.  How will users unwrap this to BTC?

beaushinkle: Can you elaborate more here? Here's the situation I'm imagining:

1) User deposits BTC and mints TBTC on ethereum (via tbtc btc<->eth bridge)
2) User locks up the TBTC on ethereum and mints wormholeL2TBTC (via wormhole eth<->L2 bridge)
3) wormholeL2TBTC gets locked on L2 to mint L2TBTC (via L2 vending machine)

Then ideally we have:

4) User burns L2TBTC to unlock wormholeL2TBTC (via L2 vending machine)
5) User burns wormholeL2TBTC to unlock TBTC (via wormhole eth<->L2 bridge)

But if I'm understanding this correctly, they could *also*

4) User locks L2TBTC to mint wormholeL1TBTC on ethereum (via wormhole eth<->L2 bridge)

So now you have TBTC and wormholeTBTC both on ethereum, which ends up fracturing liquidity

docs/rfc/rfc-8.adoc
mhluongo: Both rate limit and absolute are useful, though if we do rate limiting it should be using something mega simple to minimize code. Just an absolute would be enough for me though

docs/rfc/rfc-8.adoc
mhluongo: I think the answer is an outside hack... I don't see another circumstance that makes sense.

mhluongo: Perhaps a combined Ethereum / Bitcoin fork, where some other chains don't want to honor the majority, maybe? It's a stretch though.

docs/rfc/rfc-8.adoc
mhluongo: Indeed. If skipping the wrapping is necessary for some ecosystem bridges though, I think that's fine... this is just the closest to existing bridge infrastructure I've seen.

docs/rfc/rfc-8.adoc
mhluongo: For non-EVM chains, I think we need a canonical bridge, period. We won't have the bandwidth to maintain those chains for the forseeable future.

 For other EVMs, though, an ideal solution would either be...
 * To stage the rollout so we can deploy canonical contracts to each one
 * To work with a canonical bridge that agrees to cede control of the token contracts when the team is ready

docs/rfc/rfc-8.adoc
mhluongo: That really depends on the bridge. We wrote this assuming there would be a way for user with address A on L1 to bridge to address B on the sidechain or L2.

If the existing bridging infra has to go from address A to address A between chains, then this wrapping idea isn't helpful, and the bridge should instead directly mint against the L2 tBTC contract.

docs/rfc/rfc-8.adoc
mhluongo: > This will make a wormhole wrapped tBTC representation on ETH. How will users unwrap this to BTC?

There should never be Wormhole-wrapped tBTC on Ethereum, right? If it lands on Ethereum it will have been minting by someone else locking tBTC on Ethereum, so it should be unwrappable.

conorpp: L2tBTC when bridged back to ethereum will produce a bridge wrapped version of L2tBTC ("wormholeL2tBTC").  In order to turn L2tBTC into proper ethereum tBTC, the user must use the L2 vending machine first before using the bridge to get L2tBTC.

I imagine users will often forget to use the vending machine when bridging back to Ethereum.  I can think of two options:

a) It would be convenient to allow swapping wormholeL2tBTC to tBTC.  However, there's still tBTC locked on Wormhole that would need to be burned or accounted for somehow.
b) It could also be up to the bridge to special case L2tBTC to always use the vending machine so hopefully the user doesn't end up with wormholeL2tBTC to begin with.  Worst case the user has to redo what they did.

docs/rfc/rfc-8.adoc
conorpp: I don't think it would be feasible for bridges (at least wormhole) to cede control of one of their assets.  It would mess with a number of security assumptions.

prestwich: > I don't think it would be feasible for bridges (at least wormhole) to cede control of one of their assets. It would mess with a number of security assumptions.

This is a day 1 feature at Nomad

https://github.com/nomad-xyz/monorepo/blob/main/packages/contracts-bridge/contracts/TokenRegistry.sol#L141-L166

prestwich: rather than ceding control, it enrolls a team-controlled token as the bridge token, and allows users to migrate to the new token

docs/rfc/rfc-8.adoc
pdyraga: I am looking at the nomad production JSON config and my understanding is that in our case you would have to provide a separate address for Ethereum L1 anyway. Wouldn't it be more confusing for dev? 🤔

docs/rfc/rfc-8.adoc
pdyraga: I was thinking about absolute limit being a governable value in the contract having an authority to mint the token. For Arbitrum, it would be `L2WormholeGateway`, and `L2ArbitrumGateway`. The gateway is upgradeable so we can always add rate limits in the future, if needed.

docs/rfc/rfc-8.adoc
pdyraga: > Maybe there's another way I'm missing?

This would require custom handling in the bridge.

I think there are three ways available:
1. Update the L2 Wormhole contract to allow for an optional wrapping into a canonical token in the L2 redemption transaction. Under the hood, the Wormhole contract would mint whtBTC and wrap it using the gateway into tBTC providing the redeemer's address. The gateway would lock wthBTC and send tBTC to the redeemer. 
2. Mint to address B on L2 providing the redeemer's address as extra info. B would be the L2 gateway and it would lock whtBTC and send tBTC to the address from the extra info provided.
3. Update Wormhole Portal to execute two transactions during the redemption: Option 1 or Option 2 @conorpp suggested.

The first way is the most user-friendly but I do not know how complicated it is on the Wormhole's side.

docs/rfc/rfc-8.adoc
pdyraga: > L2tBTC when bridged back to ethereum will produce a bridge wrapped version of L2tBTC ("wormholeL2tBTC"). In order to turn L2tBTC into proper ethereum tBTC, the user must use the L2 vending machine first before using the bridge (L2tBTC -> whtBTC on L2 -> tBTC).

Correct. The proposal assumes - although I see now it is unclear - that the Wormhole contract on L2 can be customized to allow for wrapping the token from wormholeL2tBTC into a canonical L2tBTC when redeeming on L2 and to unwrap the token from the canonical L2tBTC to wormholeL2tBTC when depositing into the L2 contract to bridge back.

I have also mentioned it in [this](https://github.com/keep-network/tbtc-v2/pull/542#discussion_r1114303689) comment.

> I imagine users will often forget to use the vending machine when bridging back to Ethereum. 

Yes, this must happen automatically - either by adding wrap/unwrap functionality to the contract - ideal for the UX. Alternatively, integrated into the Portal but it would mean more than one transaction.

>  It could also be up to the bridge to special case L2tBTC to always use the vending machine so hopefully the user doesn't end up with wormholeL2tBTC to begin with.

Yes, that's the preferred way for this proposal. From the user's perspective, the easiest way is when they never receive wormholeL2tBTC into their wallet. When they redeem on L2, wormholeL2tBTC is wrapped automatically and they receive L2tBTC. When they deposit to bridge back to L1, they deposit L2tBTC and unwrapping happens internally in the Wormhole's bridge contract.

pdyraga: @conorpp was kind to provide examples and describe the possible approaches. I summarised our options in b629bcb13b475fe26757b7ee748aa4bdcd2956e3.

docs/rfc/rfc-8.adoc
pdyraga: This is described in _3.2. Minting on L1_ section if the _Initial Implementation_. Maybe I'll move _Minting on L1_ to be the very first section of the _Initial Implementation_?

pdyraga: Done in deb66b5ae043f99f06b9a23f2f6946985df15ff5.

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,482 @@
...
+bridging back to L1, leaving no wormhole tBTC left in the contract for
+depositors who crossed the Wormhole bridge to L2 and now want to go back. Those
+depositors will have to use another bridge or redeem to BTC on L2, depending on
+available options.
+
+Example: Receiving
+```
+// Receive a wormhole token transfer from another chain
+function receiveWormhole(bytes memory encodedVm) public returns (uint64) {
+  // recieve "wormhole wrapped tBTC"
```
tomaszslabon, line 227:
```suggestion
  // receive "wormhole wrapped tBTC"
```

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,482 @@
...
+token locked in the gateway so it is potentially more friendly for native L2
+minting and other bridges.
+
+This approach is implemented in the Wormhole Foundation repository for USDC
+link:https://github.com/wormhole-foundation/wormhole-circle-integration/blob/9572867afc936bf0dde8cf93a07724a90d99ddfd/evm/src/circle_integration/CircleIntegration.sol[here].
+
+For the sake of simplicity, we will implement the gateway using the Plain token
+transfer approach. If the DAO decides to add another way of minting the token in
+the future, we can implement another gateway contract based on the Custom
+message gateway given the canonical token accepts multiple minting authorities.
```
pdyraga, line 271: We had a conversation with @conorpp and we decided this is not the optimal approach. The initial implementation would be simpler only in theory as the upgrade path to the custom-message approach would be really complicated. We decided to spend the time better on implementing the custom-message solution from the beginning and being able to support multiple minting authorities from day one.

I will rework this part of the RFC.

docs/rfc/rfc-8.adoc
sashatanase: @pdyraga I am also in favor of having the same address everywhere since most of the users will not check/remember the token contract address unless they are sophisticated. But I am pretty much sure that they will not remember the token contract address. I believe your concern can become reality only if they get to see the token contract addresses side by side and I do not know when this could happen, I am sure you can provide me with some examples, but to be honest I don't think it would be so bad. 

To avoid the DEV confusion just signal the chain name in front of the addresses in the documentation or wherever we will have all the addresses listed together. 

I wonder if having the same token address is also helpful in the wallet token import. Just a thought, nothing solid.

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,79 @@
...
+Penumbra. And in newer ecosystems, that means chains like Solana, Aptos, and
+Sui.
+
+Efficiency requires minimizing liquidity fragmentation by avoiding the need for
+tBTC wrapper tokens by each chain and bridge. This situation quickly leads to an
+stem:[n*m] problem, where liquidity on `n` chains is divided between `m`
+bridges. Each of these token variants soaks up valuable liquidity that could be
+better utilized elsehwere.
+
+Because tBTC is minted "centrally" on Ethereum L1, we can do better. By
```
sashatanase, line 25: Does it mean that we could surpass L1 and mint directly from BTC to ETH L2? instead of BTC-->L1--->L2 just have BTC --->L2

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+TODO
+
+=== Minting on L1
+
+Rather than change our L1 implementation to be aware of L2s and sidechains,
+we'll start as simply as possible. Every tBTC mint will continue to require
+transactions on Ethereum from the user — first a deposit reveal, then a lock
+to bridge to another network.
+
```
sashatanase, line 61: Does this mean that if I want to mint tBTC on say Optimism I need to first pay gas on Ethereum and also have Optimism ETH?

docs/rfc/rfc-8.adoc
sashatanase: > Yes, that's the preferred way for this proposal. From the user's perspective, the easiest way is when they never receive wormholeL2tBTC in their wallet. When they redeem on L2, wormholeL2tBTC is wrapped automatically, and they receive L2tBTC. When they deposit to bridge back to L1, they deposit L2tBTC, and unwrapping happens internally in the Wormhole's bridge contract.

This could be the best UX solution for how the gateway works on the user-facing side. The user should receive the canonical token and be unaware of the wrapping and unwrapping from wormholetBTC to L2tBTC, and this should magically happen under the hood. 

I think coercing the user to find a market to swap their non-canonical token into a canonical token is a bit of a UX nightmare, and I also believe that it may decrease the token use and its adoption if we make it too hard for the users to use their bridged tokens on the L2s.

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,482 @@
...
+deployed at `0x18084fbA666a33d37592fA2633fD49a74DD93a88`. The L2/sidechain token
+will be upgradeable, owned by Threshold Council, will delegate the minting
+authority to multiple parties, and have a pause functionality for mints and
+burns.
+
+Given those requirements, the address of the canonical tBTC token on L2s and
+sidechains will be different from the one deployed on L1.
+
+A generic implementation of the token will be provided in the `solidity` smart
+contracts directory.
```
nkuba, line 74: Do you mean this repository https://github.com/keep-network/tbtc-v2/tree/main/solidity?
Could we consider placing it in a separate package, so we don't need to install [all the dependencies](https://github.com/keep-network/tbtc-v2/blob/24b0b974bd14d576806cf57c932ade1ca1041aa4/solidity/package.json#L33-L75) of `@keep-network/tbtc-v2`?

Perhaps https://github.com/thesis/solidity-contracts would be the cleanest place to hold it? Or a dedicated subdirectory in `tbtc-v2` repository with an independent NPM package?

docs/rfc/rfc-8.adoc
nkuba: Matching contract addresses on different chains and networks seems convenient but brings a risk of confusion. Unique addresses are another level of protection for users/devs to ensure they are interacting with the right chain/network. It protects against accidentally locking funds in a contract deployed on an undesired network.

docs/rfc/rfc-8.adoc
pdyraga: I would go with `keep-network/tbtc-v2/solidity` for simplicity. We would need to import this repository and all transitive dependencies for L1 TBTC token artifact anyway.

docs/rfc/rfc-8.adoc
pdyraga: This proposal assumes TBTC is minted on L1 first, always.

There are ways to mint directly on L2 but this is something for the future as it is quite complex and requires remote mind and burn implementation for L1<>L2 bridging.

docs/rfc/rfc-8.adoc
pdyraga: Correct.

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,482 @@
...
+
+===== Plain token transfer gateway
+
+This is the easiest approach and the gateway contract can be integrated directly
+with the token bridge. This approach will benefit from Wormhole security
+measures like various accounting invariant checks and governable notional limits
+between chains. Also, the integration into the UI will be easier and cross-chain
+transfers will be automatically relayed. Last but not least, any possible
+security issues with Wormhole bridge are limited to the amount of tBTC bridged
+from L1 to L2 so far.
```
lukasz-zimnoch: They are limited to a single L2 chain or to all Wormhole-powered L2 chains tBTC can be minted on? In other words, what's the exact impact of a security issue?

docs/rfc/rfc-8.adoc
pdyraga: This section is gone now.

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,482 @@
...
+==== Wormhole
+
+The Wormhole bridging process from L1 to L2 (or sidechain) will look as follows:
+
+* There is a tBTC holder on L1. The holder goes to the Wormhole Portal and
+  selects the chain they want to bridge to.
+* The holder submits one transaction to L1. After the transaction is mined, they
+  wait about 15 minutes for the Ethereum block finality.
+* The holder submits one transaction to L2. After that transaction is mined,
+  they have their tBTC on L2.
```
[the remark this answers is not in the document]
pdyraga: See 22c09434dd8ccd40e16822f1d57011cd3acdf8dc, please.

docs/rfc/rfc-8.adoc
pdyraga: This section is gone now - see https://github.com/keep-network/tbtc-v2/pull/542/commits/fc9328355c65ddfb36d2820b4f895e085a4b6323 - and we may discuss it more during the review of the contract implementing the gateway. But let me also explain here.

With the approach as presented here, where every separate bridge has its own gateway on L1, the risk of a hack of the given gateway is limited to the amount of tokens that have been bridged from L1 to L2 through the given gateway.

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,482 @@
...
+Arbitrum, Optimism, Polygon, Avalanche, and Binance Smart Chain. In the Cosmos
+ecosystem, that means Bitcoin on IBC-enabled zones (chains) like Osmosis and
+Penumbra. And in newer ecosystems, that means chains like Solana, Aptos, and
+Sui.
+
+Efficiency requires minimizing liquidity fragmentation by avoiding the need for
+tBTC wrapper tokens by each chain and bridge. This situation quickly leads to an
+`n*m` problem, where liquidity on `n` chains is divided between `m`
+bridges. Each of these token variants soaks up valuable liquidity that could be
+better utilized elsehwere.
```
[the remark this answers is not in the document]
pdyraga: Negative. This is about the canonical token address. Every bridge - by default - mints its own representation of a token and you may end up with multiple tokens stating they are tBTC on Arbitrum or any other chain.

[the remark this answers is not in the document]
lukasz-zimnoch: Clear now. Thanks!

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,428 @@
...
```
tomaszslabon, line 23:
```suggestion
better utilized elsewhere.
```

docs/rfc/rfc-8.adoc
```diff
@@ -0,0 +1,428 @@
...
+A generic implementation of the token will be provided in the `solidity` smart
+contracts directory.
+
+=== Supporting native bridges
+
+A number of relevant chains include "native" bridges — canonical bridges for the
+chain that require interoperability. These bridges are both the most important
+and the trickiest to support.
+
+Each native bridge has its own way mechanism to manage an Ethereum-to-chain
```
tomaszslabon, line 82: Probably
```suggestion
Each native bridge has its own mechanism to manage an Ethereum-to-chain
```

lukasz-zimnoch (approved):

docs/rfc/rfc-8.adoc
beaushinkle: Do we want to include these ideas in this RFC, or flesh them out in a follow-up?

docs/rfc/rfc-8.adoc
beaushinkle: I think this is a small enough implementation detail that we don't need to block merging this RFC on resolving it

beaushinkle (approved):

r-czajkowski (approved):

mhluongo (approved):
Looking good.

PR #550, opened 2023-02-18
Cross-chain Tokenized Threshold BTC: Generic L2 ERC20 implementation

### LINKED ISSUE #543: Cross-chain Tokenized Threshold BTC: Generic L2 ERC20 implementation
See https://github.com/keep-network/tbtc-v2/pull/542

The L2/sidechain token will be upgradeable, owned by Threshold Council, will delegate the minting authority to multiple parties, and have a pause functionality for mints and burns. A generic implementation of the token will be provided in the `solidity` smart contracts directory.

pdyraga:
Closes #543 

This is the first step for the implementation of a generic L2/sidechain token contract.

The L2/sidechain token will be upgradeable, owned by Threshold Council, will delegate the minting authority to multiple parties, and have a pause functionality for mints and burns. Individual L2/sidechain token implementations will inherit from this contract setting only the token name and symbol.

So far, adding and removing minters was implemented. More to come!

### DIFF
--- a/solidity/contracts/l2/L2TBTC.sol
+++ b/solidity/contracts/l2/L2TBTC.sol
@@ -1,0 +1,81 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+// ██████████████     ▐████▌     ██████████████
+// ██████████████     ▐████▌     ██████████████
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+// ██████████████     ▐████▌     ██████████████
+// ██████████████     ▐████▌     ██████████████
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+
+pragma solidity ^0.8.17;
+
+import "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+// TODO:
+// * Be paused by any one of `n` guardians, allowing avoidance of contagion in case
+//  of a chain- or bridge-specific incident.
+// * Mint by minter.
+// * Proper documentation.
+// * Test initialization
+contract L2TBTC is ERC20Upgradeable, OwnableUpgradeable {
+    /// @notice Indicates if the given address is a Minter. Only Minters can
+    ///         mint the token.
+    mapping(address => bool) public isMinter;
+
+    /// @notice List of all Minters.
+    address[] public minters;
+
+    event MinterAdded(address indexed minter);
+    event MinterRemoved(address indexed minter);
+
+    modifier onlyMinter() {
+        require(isMinter[msg.sender], "Caller is not a minter");
+        _;
+    }
+
+    function initialize(string memory _name, string memory _symbol)
+        external
+        initializer
+    {
+        __Ownable_init();
+        __ERC20_init(_name, _symbol);
+    }
+
+    /// @notice Adds the address to the Minter list.
+    function addMinter(address minter) external onlyOwner {
+        require(!isMinter[minter], "This address is already a minter");
+        isMinter[minter] = true;
+        minters.push(minter);
+        emit MinterAdded(minter);
+    }
+
+    /// @notice Removes the address from the Minter list.
+    function removeMinter(address minter) external onlyOwner {
+        require(isMinter[minter], "This address is not a minter");
+        delete isMinter[minter];
+
+        // We do not expect too many Minters so a simple loop is safe.
+        for (uint256 i = 0; i < minters.length; i++) {
+            if (minters[i] == minter) {
+                minters[i] = minters[minters.length - 1];
+                // slither-disable-next-line costly-loop
+                minters.pop();
+                break;
+            }
+        }
+
+        emit MinterRemoved(minter);
+    }
+
+    /// @notice Allows to fetch a list of all Minters.
+    function getMinters() external view returns (address[] memory) {
+        return minters;
+    }
+}
--- a/solidity/test/l2/L2TBTC.test.ts
+++ b/solidity/test/l2/L2TBTC.test.ts
@@ -1,0 +1,261 @@
+import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers"
+import { randomBytes } from "crypto"
+import { ethers, getUnnamedAccounts, helpers, waffle } from "hardhat"
+import { expect } from "chai"
+import { ContractTransaction } from "ethers"
+
+import type { L2TBTC } from "../../typechain"
+
+const { createSnapshot, restoreSnapshot } = helpers.snapshot
+
+describe("L2TBTC", () => {
+  const fixture = async () => {
+    const { deployer, governance } = await helpers.signers.getNamedSigners()
+
+    const accounts = await getUnnamedAccounts()
+    const minter = await ethers.getSigner(accounts[1])
+    const thirdParty = await ethers.getSigner(accounts[2])
+
+    const deployment = await helpers.upgrades.deployProxy(
+      // Hacky workaround allowing to deploy proxy contract any number of times
+      // without clearing `deployments/hardhat` directory.
+      // See: https://github.com/keep-network/hardhat-helpers/issues/38
+      `L2TBTC_${randomBytes(8).toString("hex")}`,
+      {
+        contractName: "L2TBTC",
+        initializerArgs: ["ArbitrumTBTC", "ArbTBTC"],
+        factoryOpts: { signer: deployer },
+        proxyOpts: {
+          kind: "transparent",
+        },
+      }
+    )
+    token = deployment[0] as L2TBTC
+
+    await token.connect(deployer).transferOwnership(governance.address)
+
+    return {
+      governance,
+      minter,
+      thirdParty,
+      token,
+    }
+  }
+
+  let token: L2TBTC
+
+  let governance: SignerWithAddress
+  let minter: SignerWithAddress
+  let thirdParty: SignerWithAddress
+
+  before(async () => {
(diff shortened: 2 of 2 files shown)


solidity/contracts/l2/L2TBTC.sol
```diff
@@ -0,0 +1,81 @@
...
+//               ▐████▌    ▐████▌
+
+pragma solidity ^0.8.17;
+
+import "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+// TODO:
+// * Be paused by any one of `n` guardians, allowing avoidance of contagion in case
+//  of a chain- or bridge-specific incident.
```
pdyraga, line 23: @mhluongo Would you like to pause everything or just mints/burns. I assumed mint/burns given the bridges are posing most of the risk. It is not about blocking l2 DEXes, for example.

solidity/test/l2/L2TBTC.test.ts
```diff
@@ -0,0 +1,261 @@
...
+
+    const accounts = await getUnnamedAccounts()
+    const minter = await ethers.getSigner(accounts[1])
+    const thirdParty = await ethers.getSigner(accounts[2])
+
+    const deployment = await helpers.upgrades.deployProxy(
+      // Hacky workaround allowing to deploy proxy contract any number of times
+      // without clearing `deployments/hardhat` directory.
+      // See: https://github.com/keep-network/hardhat-helpers/issues/38
+      `L2TBTC_${randomBytes(8).toString("hex")}`,
```
pdyraga, line 23: Not super proud of it but I had to move forward. May be replaced with a flag if we do https://github.com/keep-network/hardhat-helpers/issues/38.

solidity/contracts/l2/L2TBTC.sol
mhluongo: Just mints and burns

solidity/contracts/l2/L2TBTC.sol
```diff
@@ -0,0 +1,81 @@
...
+        emit MinterAdded(minter);
+    }
+
+    /// @notice Removes the address from the Minter list.
+    function removeMinter(address minter) external onlyOwner {
+        require(isMinter[minter], "This address is not a minter");
+        delete isMinter[minter];
+
+        // We do not expect too many Minters so a simple loop is safe.
+        for (uint256 i = 0; i < minters.length; i++) {
```
dimpar, line 65: The alternative might be that we reuse `isMinter` map and make it e.g. `mapping(address => index)` where `index` denotes an index in the `minters` array of a given address. This way we would get rid of the loop. We can check what is the gas implication for it. Probably `isMinter` should be named a bit differently in this case. `minters[0]` should also be probably reserved for things like `isMinter[nonExistantMinterAddr]`

```
        uint256 index = isMinter[minter]
        address minter = minters[minters.length - 1]
        address minterToRemove = minters[index]
        minters[index] = minter;
        minters.pop()
        isMinter[minter] = index
        delete isMinter[minterToRemove];
```

mhluongo: The issue with the `0` index minter is pretty rough. Sticking with this construction also means an idiomatic public ABI `function isMinter(address m) view returns (bool);`

dimpar: I checked the gas implication and it seems that for a shorter set of minters, the gas usage is pretty similar.

solidity/contracts/l2/L2TBTC.sol
dimpar: What about `ERC20WithPermit, MisfundRecovery` functionality for tokens that will be inheriting from `L2TBTC.sol`? Do we want/need to add it?

solidity/contracts/l2/L2TBTC.sol
```diff
@@ -0,0 +1,81 @@
...
+    event MinterAdded(address indexed minter);
+    event MinterRemoved(address indexed minter);
+
+    modifier onlyMinter() {
+        require(isMinter[msg.sender], "Caller is not a minter");
+        _;
+    }
+
+    function initialize(string memory _name, string memory _symbol)
+        external
```
dimpar, line 44: My understanding was that this `L2TBTC` token is going to be a generic contract that actual tokens will inherit from for a given L2/sidechains like Arbitrum, Polygon etc. Shouldn't we change the visibility of `initialize` to `internal`?

solidity/contracts/l2/L2TBTC.sol
pdyraga: We want it! See c710eecaa7b2131374258801c3fa2e6c0b6fc118.

solidity/contracts/l2/L2TBTC.sol
```diff
@@ -0,0 +1,242 @@
...
+        __ERC20Permit_init(_name);
+        __Ownable_init();
+        __Pausable_init();
+    }
+
+    /// @notice Adds the address to the minters list.
+    /// @dev Requirements:
+    ///      - The caller must be the contract owner.
+    ///      - `minter` must not be a minter address already.
+    function addMinter(address minter) external onlyOwner {
```
dimpar, line 95: At the end of the day, who will be the `owner`? In RFC it says:
> The Governance will be able to add and remove minters to L2/sidechain tBTC canonical contract.

Don't we need an `onlyGovernance` modifier here instead of `onlyOwner`? The same comment for other `add*`, `remove*` functions.

solidity/contracts/l2/L2TBTC.sol
pdyraga: Are we good to resolve this conversation then?

solidity/contracts/l2/L2TBTC.sol
pdyraga: I think this approach is more flexible. We can inherit a contract and add no code. We can also not inherit a contract at all and just deploy `L2TBTC` instance. Contract deployment and initialization happen in one transaction (see `helpers.upgrades.deployProxy`) so there is no chance someone can initialize your contract with the wrong values even if they wanted to which is pretty unlikely.

solidity/test/l2/L2TBTC.test.ts
```diff
@@ -0,0 +1,1165 @@
...
+        })
+
+        it("should emit an event", async () => {
+          await expect(tx)
+            .to.emit(token, "GuardianAdded")
+            .withArgs(guardian.address)
+        })
+      })
+
+      context("when address is a guardian", () => {
```
dimpar, line 313: "when address was already a guardian"?

solidity/test/l2/L2TBTC.test.ts
```diff
@@ -0,0 +1,1165 @@
...
+
+          await token.connect(governance).addGuardian(guardian.address)
+          tx = await token.connect(governance).removeGuardian(guardian.address)
+        })
+
+        after(async () => {
+          await restoreSnapshot()
+        })
+
+        it("should take guardian role from the address", async () => {
```
dimpar, line 389: Can we check the `guardians` array too?

solidity/contracts/l2/L2TBTC.sol
dimpar: ok, I thought that we wanted the names hardcoded in the contract. I'm fine with the `initialize` approach.

solidity/test/l2/L2TBTC.test.ts
pdyraga: 205c26e348d81348278fed14debeda8c75a82cd7

solidity/test/l2/L2TBTC.test.ts
pdyraga: 205c26e348d81348278fed14debeda8c75a82cd7

solidity/test/l2/L2TBTC.test.ts
```diff
@@ -0,0 +1,1165 @@
...
+    context("when called by the owner", () => {
+      context("when address is not a minter", () => {
+        it("should revert", async () => {
+          await expect(
+            token.connect(governance).removeMinter(thirdParty.address)
+          ).to.be.revertedWith("This address is not a minter")
+        })
+      })
+
+      context("when address is a minter", () => {
```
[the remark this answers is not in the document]
pdyraga: 205c26e348d81348278fed14debeda8c75a82cd7

solidity/test/l2/L2TBTC.test.ts
```diff
@@ -0,0 +1,1165 @@
...
+        })
+
+        it("should emit an event", async () => {
+          await expect(tx)
+            .to.emit(token, "MinterAdded")
+            .withArgs(minter.address)
+        })
+      })
+
+      context("when address is a minter", () => {
```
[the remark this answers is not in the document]
pdyraga: 205c26e348d81348278fed14debeda8c75a82cd7

solidity/test/l2/L2TBTC.test.ts
```diff
@@ -0,0 +1,1165 @@
...
+    context("when called not by the owner", () => {
+      it("should revert", async () => {
+        await expect(
+          token.connect(thirdParty).addMinter(thirdParty.address)
+        ).to.be.revertedWith("Ownable: caller is not the owner")
+      })
+    })
+
+    context("when called by the owner", () => {
+      context("when address is not a minter", () => {
```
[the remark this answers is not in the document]
pdyraga: 205c26e348d81348278fed14debeda8c75a82cd7

solidity/test/l2/L2TBTC.test.ts
```diff
@@ -0,0 +1,1165 @@
...
+
+          await token.connect(governance).addMinter(minter.address)
+          tx = await token.connect(governance).removeMinter(minter.address)
+        })
+
+        after(async () => {
+          await restoreSnapshot()
+        })
+
+        it("should take minter role from the address", async () => {
```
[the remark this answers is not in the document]
pdyraga: Added the assertion in 205c26e348d81348278fed14debeda8c75a82cd7 but I am not sure about the rename - we are checking both mapping and array now.

dimpar (approved):

PR #556, opened 2023-02-23
Monitor minters health

lukasz-zimnoch:
Closes: https://github.com/keep-network/optimistic-minting/issues/71

Here we add four system events that can be produced by the minting monitor:

###  Optimistic minting not requested by designated minter

A **warning system event** indicating that an optimistic minting request was not issued by the designated minter and another minter did that job. This event is sent to Sentry hub and should get team’s attention. The default action is investigating the cause of the designated minter idleness as the designated minter may be unhealthy/malicious or there may be a bug in the minters bot code.

### Optimistic minting not finalized by designated minter

A **warning system event** indicating that an optimistic minting request was not finalized by the designated minter and another minter did that job. This event is sent to Sentry hub and should get team’s attention. The default action is investigating the cause of the designated minter idleness as the designated minter may be unhealthy/malicious or there may be a bug in the minters bot code.

### Optimistic minting not requested by any minter

A **warning system event** indicating that an optimistic minting request was not issued by any minter. This event is sent to Sentry hub and should get team’s attention. The default action is investigating the cause of the minters idleness as the underlying deposit may be invalid, minters may be unhealthy/malicious or there may be a bug in the minters bot code.

### Optimistic minting not finalized by any minter

A **warning system event** indicating that an optimistic minting request was not finalized by any minter. This event is sent to Sentry hub and should get team’s attention. The default action is investigating the cause of the minters idleness as the underlying deposit may be invalid, minters may be unhealthy/malicious or there may be a bug in the minters bot code.

### Testing

The code was tested against mainnet using a local instance. Testing of such events is not straightforward and requires some temporary hacks in the code.

To test "Optimistic minting not requested/finalized by designated minter" events, you need to:
- Find minting requested and minting finalized transactions done by non-designated minters (e.g https://etherscan.io/tx/0x4821e028fb08c9e190a5292593f2468c7867a8ff8e2984f1432402db01abf8c4 and https://etherscan.io/tx/0xe285868a4a9100bd1cf174a8a0503028ebb2f98f92c5b84fb6e010f729202b13)
- Set the monitored block window manually to enclose those transactions. This can be done by modifying the `checkpointBlock` stored in the persistence and `contracts.latestBlock` function in the code.

To test "Optimistic minting not requested/finalized by any minter", you need to:
- Find a deposit revealed transaction and a minting requested transaction
- Set the monitored block window manually in a way that the block window rewound by `mintingRequestTimeoutBlocks` encloses the deposit revealed transaction and the block window rewound by `mintingFinalizationTimeoutBlocks` encloses the minting requested transaction.
- Locate the `OptimisticMinting.getOptimisticMintingRequest` calls. Inject a `TBTCVault` mock that will return `requestedAt === 0` for the deposit revealed transaction and `finalizedAt === 0` for minting requested transaction

### DIFF
--- a/monitoring/src/minting-monitor.ts
+++ b/monitoring/src/minting-monitor.ts
@@ -107,7 +107,7 @@ const OptimisticMintingNotRequestedByDesignatedMinter = (
   chainEvent: OptimisticMintingRequestedChainEvent,
   designatedMinter: Identifier
 ): SystemEvent => ({
-  title: "Optimistic minting was not requested by designated minter",
+  title: "Optimistic minting not requested by designated minter",
   type: SystemEventType.Warning,
   data: {
     actualMinter: `0x${chainEvent.minter.identifierHex}`,
@@ -127,7 +127,7 @@ const OptimisticMintingNotFinalizedByDesignatedMinter = (
   designatedMinter: Identifier | "unknown",
   designatedMinterUnknownCause: string
 ): SystemEvent => ({
-  title: "Optimistic minting was not finalized by designated minter",
+  title: "Optimistic minting not finalized by designated minter",
   type: SystemEventType.Warning,
   data: {
     actualMinter: `0x${chainEvent.minter.identifierHex}`,
@@ -147,7 +147,7 @@ const OptimisticMintingNotFinalizedByDesignatedMinter = (
 const OptimisticMintingNotRequestedByAnyMinter = (
   chainEvent: DepositRevealedChainEvent
 ): SystemEvent => ({
-  title: "Optimistic minting was not requested by any minter",
+  title: "Optimistic minting not requested by any minter",
   type: SystemEventType.Warning,
   data: {
     btcFundingTxHash: chainEvent.fundingTxHash.toString(),
@@ -161,7 +161,7 @@ const OptimisticMintingNotRequestedByAnyMinter = (
 const OptimisticMintingNotFinalizedByAnyMinter = (
   chainEvent: OptimisticMintingRequestedChainEvent
 ): SystemEvent => ({
-  title: "Optimistic minting was not finalized by any minter",
+  title: "Optimistic minting not finalized by any minter",
   type: SystemEventType.Warning,
   data: {
     btcFundingTxHash: chainEvent.fundingTxHash.toString(),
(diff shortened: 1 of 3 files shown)


monitoring/src/minting-monitor.ts
```diff
@@ -178,4 +299,238 @@ export class MintingMonitor implements SystemEventMonitor {
...
+          mre.fundingTxHash
+        ),
+      }))
+      .filter((mre) => !mre.minter.equals(mre.designatedMinter))
+      .map((mre) =>
+        OptimisticMintingNotRequestedByDesignatedMinter(
+          mre,
+          mre.designatedMinter
+        )
+      )
```
pdyraga, line 330: So functional. Wow.

image

monitoring/src/minting-monitor.ts
```diff
@@ -178,4 +299,238 @@ export class MintingMonitor implements SystemEventMonitor {
...
+          this.tbtcVault
+        )
+      )
+    )
+
+    return chainEvents
+      .filter((ce, index) => {
+        const request = mintingRequests[index]
+        const requestExists =
+          request.status === "fulfilled" && request.value.finalizedAt !== 0
```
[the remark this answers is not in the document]
lukasz-zimnoch: IMO, this form is more readable. Otherwise, we'll need to use optional chaining (i.e. `request.value?. finalizedAt`) and check against `undefined` as well.

pdyraga (approved):

monitoring/src/minting-monitor.ts
```diff
@@ -178,4 +299,238 @@ export class MintingMonitor implements SystemEventMonitor {
...
+
+    return systemEvents
+  }
+
+  private checkDesignatedMintersRequests(cache: ChainDataCache): SystemEvent[] {
+    return cache.mintingRequestedEvents
+      .map((mre) => ({
+        ...mre,
+        designatedMinter: this.getDesignatedMinter(
+          cache.minters,
```
nkuba, line 319: If there were updates to the minters' list in the meantime we may end up with false positive errors. We should be using the minters' list from the block when the transaction we're verifying happened.

Or we could just ignore it and expect false alarms to be issued around the time the minters' list is updated.

Issue #552, opened 2023-02-21
Monitoring: token supply


pdyraga:
Enhance our monitoring system to watch the total supply of the token and send a critical alert if it goes above a certain threshold in the given period of time. What exactly is the threshold and the time period needs to be decided.

PR #560, opened 2023-02-27
Monitor TBTC token total supply

lukasz-zimnoch:
Closes: https://github.com/keep-network/tbtc-v2/issues/552

Here we add a new supply monitor that produces a new kind of system event:

### High TBTC token total supply change

A **critical system event** indicating that a high change (i.e. >=10%) of the total TBTC v2 token supply took place in the last 12 hours. This event is sent to Sentry hub and requires an immediate team’s action. The default action is checking the root cause of the supply change and making sure its source is actually a proper deposit/redemption and there are no signs of any malicious action.


pdyraga:
How this code was tested?

lukasz-zimnoch:
> How was this code tested?

Similarly, to https://github.com/keep-network/tbtc-v2/pull/556. Manual block window on points where the supply changed enough to cause an alert.

monitoring/src/supply-monitor.ts
```diff
@@ -0,0 +1,101 @@
...
+    // eslint-disable-next-line no-console
+    console.log("running supply monitor check")
+
+    // By default, the supply change is checked between the current and
+    // reference blocks.
+    const currentBlock = toBlock
+    let referenceBlock =
+      currentBlock - totalSupplyBlockSpan > 0
+        ? currentBlock - totalSupplyBlockSpan
+        : 0
```
pdyraga, line 55: I am confused about the reference block. If the time span between `fromBlock` and `toBlock` is less than 12h on average, are we checking between `fromBlock` and... `0`?

Could be worth dropping a comment about this case, ideally explaining it on a timeline.

lukasz-zimnoch: > If the time span between fromBlock and toBlock is less than 12h on average, are we checking between fromBlock and... 0?

Negative. This is just a sanity check to not nuke the monitoring if the reference block is <0. If the time span is less than 12h, the reference block is replaced by the last block the alert was fired on.

lukasz-zimnoch: Left an additional explanation: 567dd2dd

monitoring/src/supply-monitor.ts
```diff
@@ -0,0 +1,112 @@
...
+    change: `${difference.gte(0) ? "+" : "-"}${change.toString()}%`,
+    threshold: `${threshold}%`,
+    referenceBlock: referenceBlock.toString(),
+  },
+  block: currentBlock,
+})
+
+export interface SupplyMonitorPersistence {
+  lastHighTotalSupplyChangeBlock: () => Promise<number>
+  updateLastHighTotalSupplyChangeBlock: (block: number) => Promise<void>
```
pdyraga, line 32: Could be worth documenting those two functions in two sentences. 

My understanding of `lastHighTotalSupplyChangeBlock` is that it represents the end block of the check when the alert of the high total supply change` was triggered the last time. If the alert was never triggered, this is 0.

Is this correct?

monitoring/src/supply-monitor.ts
pdyraga: I think I see it now. For this particular monitor, `fromBlock` is ignored and the `referenceBlock` replaces it. 

Can we add a small comment on the top of `referenceBlock` to explain it?

monitoring/src/supply-monitor.ts
```diff
@@ -0,0 +1,112 @@
...
+      threshold = totalSupplyChangeThreshold / 2
+    }
+
+    const referenceSupply = await this.tbtcToken.totalSupply(referenceBlock)
+    const currentSupply = await this.tbtcToken.totalSupply(currentBlock)
+
+    const difference = currentSupply.sub(referenceSupply)
+    // There is a precision loss here, but it is acceptable as long as the
+    // threshold is at least 1%.
+    const change = difference.abs().mul(100).div(referenceSupply)
```
pdyraga, line 87: It is nice it works for + and - case!

monitoring/src/supply-monitor.ts
lukasz-zimnoch: Yep. Added a comment faea9554

monitoring/src/supply-monitor.ts
lukasz-zimnoch: faea9554

pdyraga (approved):

Issue #559, opened 2023-02-27
Ordinals RFC


beaushinkle:
Create a RFC describing a plan for how to bridge bitcoin ordinals

PR #561, opened 2023-02-28
RFC 9: Bitcoin Ordinals

beaushinkle:
closes https://github.com/keep-network/tbtc-v2/issues/559

This RFC outlines the beginnings of a plan to custody [Ordinals](https://docs.ordinals.com/).

Tagging @mhluongo @sashatanase @pdyraga for review


beaushinkle:
> Really excited about this! Where is the best place to have the fees discussion, you think... here in PR review, or should we treat that as future work and focus on getting this RFC merged with its current scope? I know you might want to hand this off.

I think it makes sense to focus on merging the current scope, figure out what we want to go with elsewhere, and then open an update PR to this document (or create a new document) once we've done that. What's your preference?

mhluongo:
I'm happy with that. Might even want to open up the fee discussion [on Discourse](https://forum.threshold.network/) or something after this gets merged.

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,145 @@
...
+These Inscribed Satoshis (called Inscriptions) can be transferred like normal
+BTC. The sender's wallet software needs to be careful to not merge the
+Inscriptions with "normal" UTXOs, and needs to be careful to not accidentally
+send the Inscription as part of a payment. The
+https://github.com/casey/ord[casey/ord] repo provides the primatives for the
+handling of Ordinals on BTC.
+
+A good mental model here is thinking about Inscriptions like
+https://www.apmex.com/product/117942/1932-s-washington-quarter-choice-au[special
+edition quarters]. The the linked quarter trades for ~$350, but is also legal
```
mhluongo, line 37: s/The the/The/

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,145 @@
...
+custodying Ordinals: rotating wallets, P2SH addresses, fraud proofs,
+redemptions, etc, with some slight modifications.
+
+== Proposal
+
+=== Overview
+
+We build a BTC <> Ethereum Ordinal bridge!
+
+We create 51-of-100 threshold signing groups using the same process that we use
```
mhluongo, line 51: Are they the same wallets as the current instance of the bridge, or are they separate groups?

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,145 @@
...
+for tBTCv2. To deposit to these signing groups, we use the same P2SH process
+including a refund that we use for tBTCv2. They pay a <<Fees,fee>>. We
+optimistically mint using the same Minter & Guardian process we already have. In
+the background, we sweep to disable the refund. Later, someone brings back the
```
mhluongo, line 55: Do we need a separate sweeping schedule here?

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,145 @@
...
+NFT and a <<Fees,fee>>. We burn the NFT and send them their Ordinal.
+
+=== Depositing Details
+
+Depositing works essentially the same way that it does in tBTCv2 with a few key
+differences.
+
+Once a deposit is revealed, it is **extremely** important that the wallet does
+not sweep them into the same UTXO; that would be destructive. We still need to
+sweep to disable the refundfootnote:[Whether or not the sweep can be performed
```
mhluongo, line 65: I believe it can, with a multi-output sweep tx.

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,145 @@
...
+individual satoshis.
+
+Once custodied, we would mint an ERC721 where the `metadata` field references
+the https://docs.ordinals.com/overview.html[Ordinal ID]. They offer multiple
+notations; using the Integer Notation looks simple. The reference to the Ordinal
+ID is enough for a dApp to pull the underlying content, but it needs to be
+plugged in to both Bitcoin and Ethereum.
+
+The ERC721 in the user's Ethereum Wallet points at the Ordinal ID. Then, the
+https://github.com/casey/ord[ord] program can identify the associated Ordinal on
```
mhluongo, line 76: We'll have APIs to do this easily soon

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,145 @@
...
+a <<Fees,fee>>, and an ethereum address back to contracts. The associated wallet
+sends them their Ordinal to the provided address with a 51-of-100 signature.
+
+=== Fraud Detection
+
+Rather than storing the full list of custodied Ordinals per
+walletfootnote:[Storing the full list would of Ordinals be very gas
+inefficient.], we instead:
+
+* Take the merkle root of all Ordinals under custody
```
mhluongo, line 105: When does this get updated? By whom?

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,145 @@
...
+* Propose this root on chain
+* Let anyone challenge this for a reward if it is wrong
```
mhluongo, line 107: Where does the reward come from? Is it from the slashed funds, or an independent bounty?

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,145 @@
...
+  associated with a wallet closure.
+
+(2) Not redeeming an Ordinal within the allotted time.
+
+Anyone can challenge any time (at a cost) for (1). The challenge is defeated by
+providing the legimate request within the timeout.
+
+In order to create challenge (2), the challenger must provide a proof that the
+Ordinal is being custodied and that a redemption was requested. The challenge is
+defeated by providing a SPV proof that the Ordinal was properly transferred.
```
mhluongo, line 124: Are these easy to verify? Considering we need per-satoshi tracking.

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+=== Fees
+
+Taking a percentage cut of an Ordinal does not make sense, so we need to use a
```
mhluongo, line 128: Saving my thoughts on fees for a later revision!

mhluongo (commented):
Concept ACK! Dropped some quick thoughts on how best to make the design more clear / specific

docs/rfc/rfc-9.adoc
beaushinkle: I was imagining while writing this that they were totally separate, but I think they can be in the same wallet without much risk. The v2 system is configured to only sweep things based on particular ethereum events, and only redeem from the main UTXO.

Deposited Ordinals would never trigger that same event, would never be swept into the main UTXO, and so we would never have a chance of accidentally sending them to folks. We'd save gas on wallet creation, which is nice. 

A tricky bit is wallet closures. We'd have two default conditions

1. when it has too little BTC 
2. when it has too few Ordinals

And could either close a wallet when *either* conditions are true (more wallet closures) or when *both* conditions are true (less wallet closures)

docs/rfc/rfc-9.adoc
beaushinkle: I [verified](https://bitcoin.stackexchange.com/questions/35686/how-do-you-send-to-multiple-addresses-by-paying-a-one-time-fee-which-wallet-sup) having multiple outputs is possible, so I think we should be totally fine to both sweep and redeem on a batch schedule

docs/rfc/rfc-9.adoc
beaushinkle: Yup - [double checked](https://bitcoin.stackexchange.com/questions/35686/how-do-you-send-to-multiple-addresses-by-paying-a-one-time-fee-which-wallet-sup) it

docs/rfc/rfc-9.adoc
beaushinkle: I'm thinking slashed funds - addressed in [52d038](https://github.com/keep-network/tbtc-v2/pull/561/commits/52d0388a022911408a08fb79f863e9063f08621f)

docs/rfc/rfc-9.adoc
beaushinkle: Clarified in [284318](https://github.com/keep-network/tbtc-v2/pull/561/commits/284318bc2beb0b9a25c0ad88fd4c022a1869117f)

docs/rfc/rfc-9.adoc
beaushinkle: Clarified in [284318](https://github.com/keep-network/tbtc-v2/pull/561/commits/284318bc2beb0b9a25c0ad88fd4c022a1869117f)

docs/rfc/rfc-9.adoc
beaushinkle: [184bc8](https://github.com/keep-network/tbtc-v2/pull/561/commits/184bc81094bdb3e10ff9bad7f77cbcad13b496a3)

docs/rfc/rfc-9.adoc
beaushinkle: [290037](https://github.com/keep-network/tbtc-v2/pull/561/commits/290037bcd4d467ff0b0422193e8e10d7d17bc4cd)

docs/rfc/rfc-9.adoc
beaushinkle: Oh sheesh, that's a good question; we can prove that a transaction happened in a block with SPV proofs and our relay, but i'm not sure yet if we can actually prove anything about it being the Ordinal that it says it is. Further research!

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,160 @@
...
+inefficient.], we instead:
+
+* Take the merkle root of all Ordinals under custody
+* A Proposer, determined by the ethereum hash of the mint/redeem transaction,
+  proposes this root on chainfootnote:[We reimburse the gas of the proposer here
+  like we do in v2]
+* Let anyone challenge this for a <<Optimistic Challenges,reward>> if the root is
+  wrong
+* Let anyone challenge this for a <<Optimistic Challenges,reward>> if the root does
+  not appear on chain within a timeout
```
pdyraga, line 115: How does the wallet (or Proposer) defeat the challenge? If I understand the idea correctly, they should provide proof that all the ordinals are stored in the wallet. Do we know what this proof looks like and how to validate it reflects the reality of Bitcoin?

docs/rfc/rfc-9.adoc
pdyraga: > Considering we need per-satoshi tracking.

This looks like a problem with proving the ordinal is custodied after sweeping. This could be done with SPV proof on the sweep transaction and per-ordinal tracking which feels like a one-write to a mapping on EVM. We can then construct redemption challenges.

docs/rfc/rfc-9.adoc
pdyraga: Note this is how fraud challenges work right now in tBTC. Keeping it aligned would be nice.

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,160 @@
...
+Inscriptions with "normal" UTXOs, and needs to be careful to not accidentally
+send the Inscription as part of a payment. The
+https://github.com/casey/ord[casey/ord] repo provides the primatives for the
+handling of Ordinals on BTC.
+
+A good mental model here is thinking about Inscriptions like
+https://www.apmex.com/product/117942/1932-s-washington-quarter-choice-au[special
+edition quarters]. The linked quarter trades for ~$350, but is also legal tender
+worth $0.25. An owner of such a quarter probably does not want to use it to pay
+for parking. Inscriptions work the same way.
```
sashatanase, line 39: Coin Collectibles - They inherently have a value because they are 1 satoshi but their real value is the one assigned by the market. 

I think the coin collector example is very good.

docs/rfc/rfc-9.adoc
sashatanase: > Since NFTs are unique, there is no longer a choice about which wallet an NFT is
redeemed from. It **must** be the wallet that is the custody of that Ordinal.

I may need help understanding this. Do you mean I cannot bridge back the Ordinal unless I own the wallet where the Ordinal is kept?

This means that the UI will be a smidge affected; right now, we give the liberty of the tBTC owner to send their BTC back to any address they want. 

If I understand correctly, what you are saying means that when the Ordinal ERC721 token owner wants to bridge back, the dApp will display the BTC Wallet address where the Ordinal will be sent back in a purely informative way. The input label for the BTC Wallet will be disabled and not editable.

docs/rfc/rfc-9.adoc
sashatanase: Echoing the 
> Do we know what this proof looks like and how to validate it reflects the reality of Bitcoin?

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,160 @@
...
+
+Taking a percentage cut of an Ordinal does not make sense, so we need to use a
+different fee structure.
+
+The simplest structure is to charge flat, governable fees on minting and
+redemption. For example, we could charge 0.01 BTC on mint (which we turn into
+tBTC and keep as protocol controlled value), and then 0.01 tBTC on redemption
+(which we keep as protocol controlled value).
+
+Another option is to charge dynamic redemption fees based on the amount of time
```
sashatanase, line 151: This is an intersting proposal. However, this might bring a new challenge for the people who forgot they bridged and are accruing fee debts. I wonder if one might leave the ordinal on Ethereum if the Ordinal value is much under the tBTC bridge fee value. 

I have a Magic Money Wizard on which I paid 0.0002 BTC, I bridge it on Ethereum, and then I leave it for 2 years, which means 0.02 BTC, much much more than I spent on it, which will lead me never to bridge it back, 

So how does the protocol ensure it gets the fee money?

docs/rfc/rfc-9.adoc
beaushinkle: > I may need help understanding this. Do you mean I cannot bridge back the Ordinal unless I own the wallet where the Ordinal is kept?

### BTC
Say that you bridge 10 BTC into tBTCv2 during the first week of the system. The first wallet will custody your bitcoin, and you'll get 10 tBTC. Then, a few months pass and now there are 7 alive wallets. The oldest has already been redeemed for all of its funds, so the original you deposited to is gone.

You go to redeem. The dApp will request a redemption from the oldest wallet, but from the system's perspective and in terms of fungibility, it doesn't matter. You're going to be given 10 BTC from *any* wallet, because we assume you don't care which 10 BTC you're getting - they're all the same (like barrels of oil).

### Ordinals
If you deposit an Ordinal into the system and get a NFT for that Ordinal, when you redeem that NFT you need to get exactly that same Ordinal back. That means that if there are 8 alive wallets (each the custody of many unique Ordinals), then **only one** of those Wallets has **your** Ordinal.

Ordinal-NFT holders on Ethereum don't own any of the BTC wallets that custody the Ordinals; that's the job of the node operators.

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,160 @@
...
+=== Current Functionality
+
+Ordinals, at a high level, are single satoshis with metadata stuck into their
+witness field (called Inscribing). Each satoshi has an immutable age, which
+effectively gives them serial numbers.
+
+These Inscribed Satoshis (called Inscriptions) can be transferred like normal
+BTC. The sender's wallet software needs to be careful to not merge the
+Inscriptions with "normal" UTXOs, and needs to be careful to not accidentally
+send the Inscription as part of a payment. The
```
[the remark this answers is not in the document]
beaushinkle: > Is it about the fact we can't combine 20 "standard" sat from one UTXO and 1 "ordinal" sat from another UTXO to redeem 21 "standard" sat from the wallet?

Yeah - exactly. From BTC's perspective, they're all just sats and UTXOs. But from the Ordinal perspective, they're much more valuable than 1 satoshi, so it's paramount to not pay mining fees or redeem TBTC using the Ordinal sats.

docs/rfc/rfc-9.adoc
```diff
@@ -0,0 +1,160 @@
...
+
+We build a BTC <> Ethereum Ordinal bridge!
+
+We add functionality to the current tBTCv2 client to make future signing groups
+compatible with Ordinals. To deposit to these signing groups, we use the same
+P2SH process including a refund that we use for tBTCv2. They pay a <<Fees,fee>>.
+We optimistically mint using the same Minter & Guardian process we already
+havefootnote:[The structure for Minters and Guardians is the same, but the
+validation rules would be different]. In the background, we sweep to disable the
+refund. Later, someone brings back the NFT and a <<Fees,fee>>. We burn the NFT
```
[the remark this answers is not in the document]
beaushinkle: > For example, if we have 5 inputs with ordinals, we can't sweep them to one output.

yeah exactly

> Is it just about relocking to another output to disable refunds? It feels like for n inputs we always need to create n outputs when sweeping ordinals. For tBTC, n inputs are swept to 1 output.

Completely correct. We would still need to do some sort of sweeping transaction to disable the refund. That transaction would need to have as many outputs as inputs since we can't combine them.

docs/rfc/rfc-9.adoc
beaushinkle: > How does the wallet (or Proposer) defeat the challenge? If I understand the idea correctly, they should provide proof that all the ordinals are stored in the wallet. Do we know what this proof looks like and how to validate it reflects the reality of Bitcoin?

Here's the idea I had, which is complicated in order to save us from keeping a map of all of the ordinals on chain.

(1) Each deposit and redemption is individually *proved* but not *stored* (spv proof).
(2) You take the log of deposits and redemptions and use it off-chain to calculate an ordered list. Say the log looks like:

```
deposit ord-1
deposit ord-2
deposit ord-3
redeem  ord-2
deposit ord-4
deposit ord-5
```
(3) Then you have a list of current custody that looks like `[ord-1, ord-3, ord-4, ord-5]`.

(4) Then you take the merkle root of this list:

```
hash(ord-1, ord-3) = 8eeae
hash(ord-4, ord-5) = 6ae7b

hash(8eeae, 6ae7b) = 2bb66 <- root
```

And propose that this becomes the new root. There's a challenge period where someone can ask ethereum to manually calculate the root using on-chain data. If it never gets defeated before the timeout, we assume the root is correct.

(5) In the future, someone notices that `ord-3` was moved on bitcoin via fraud. They need to prove that `ord-3` was under custody in the first place. They submit a proof that looks like:

```
hash(ord-1, ord-3) = 8eeae
hash(8eeae, 6ae7b) = 2bb66
2bb66 === root
```

Then they create a fraud challenge for the wallet that needs to be defeated the same way we do [in v2](https://github.com/keep-network/tbtc-v2/blob/6d1d695b80089da4de5a83ae01cdefad02a33faa/solidity/contracts/bridge/Fraud.sol#L201-L229)

docs/rfc/rfc-9.adoc
beaushinkle: I'm totally down to have the wallets be separate

docs/rfc/rfc-9.adoc
pdyraga: [Here is what we know](https://github.com/keep-network/tbtc-v2/blob/5d6232bb50660ae04f5c4af3fc2f53bc989ab251/solidity/contracts/bridge/BitcoinTx.sol#L24-L92) on Ethereum about Bitcoin transaction that is proven. 

I also realized I do not fully understand inscriptions. 

Each sat is unique - it is numbered. So we can say each sat is a kind of an NFT but with no metadata attached. Those single sats are... ordinals? 

Now, having uniquely numbered tokens is not enough. They need to contain some metadata. Ordinal + metadata = inscription? 

If I understand it, correctly, the metadata is stored in
```
OP_FALSE
OP_IF
  OP_PUSH "ord"
  OP_1
  OP_PUSH "<content-type>" // example: OP_PUSH "text/plain;charset=utf-8"
  OP_0
  OP_PUSH "<content>" // example: OP_PUSH "Hello, world!"
OP_ENDIF
```

this looks like a standard Bitcoin script in an input. Is it included only in the "original input" when the inscription was created and every other transfer is based on the sat number (ordinal)? 

If so - if extracting sat number from the [transaction data](https://github.com/keep-network/tbtc-v2/blob/5d6232bb50660ae04f5c4af3fc2f53bc989ab251/solidity/contracts/bridge/BitcoinTx.sol#L24-L92) is doable (I do not know if, did not validate), we can prove on Ethereum the given ordinal was deposited but we will not know the metadata. Going back 100 transactions may not be feasible given EVM limitations. We need to come up with some other algorithm for that.

I do not think we need to have it all in this RFC, but I think we should mark the spots that require future research. My understanding - after reading the RFC for the first time - was that just a few slight changes are needed but I am not really sure about it 🤔

docs/rfc/rfc-9.adoc
pdyraga: > And propose that this becomes the new root. There's a challenge period where someone can ask ethereum to manually calculate the root using on-chain data. 

If I understand the proposal correctly, Ethereum would need to calculate the root using the same data as used off-chain in (2). If this is the case, I think it is problematic because Ethereum smart contract does not have access to events (logs) emitted historically. Even if there was a way to retrieve events (logs), there could be a risk of exceeding the block gas limit given the number of operations that would have to be performed to calculate the tree . We would have to limit the number of deposits into a single wallet and it's still not perfect because there is no guarantee the block gas limit will not be lowered, I think. 

> Then they create a fraud challenge for the wallet that needs to be defeated the same way we do [in v2](https://github.com/keep-network/tbtc-v2/blob/6d1d695b80089da4de5a83ae01cdefad02a33faa/solidity/contracts/bridge/Fraud.sol#L201-L229)

In this model, why do they need to prove the `ord-3` was in the wallet? In the v2 model, the challenger says "the wallet signed a transaction that is not a valid protocol transaction". We do not care what this transaction is. Defeating the challenge requires that we store [all honestly spent UTXOs](https://github.com/keep-network/tbtc-v2/blob/main/solidity/contracts/bridge/BridgeState.sol#L306-L312).

I do not think we need to figure it out in this RFC, I'd just mark that there are some concerns and this requires further research.

docs/rfc/rfc-9.adoc
beaushinkle: > Each sat is unique - it is numbered. So we can say each sat is a kind of an NFT but with no metadata attached. Those single sats are... ordinals?

> Now, having uniquely numbered tokens is not enough. They need to contain some metadata. Ordinal + metadata = inscription?

Yup! All correct including how the metadata is stored

> Is it included only in the "original input" when the inscription was created and every other transfer is based on the sat number (ordinal)?

That's what I keep looking for; I'm having a bunch of trouble finding an ordinal explorer that links me to the underlying satoshi and where it's being kept. I might just need to send one I have around and post logs.

> If so - if extracting sat number from the [transaction data](https://github.com/keep-network/tbtc-v2/blob/5d6232bb50660ae04f5c4af3fc2f53bc989ab251/solidity/contracts/bridge/BitcoinTx.sol#L24-L92) is doable (I do not know if, did not validate), we can prove on Ethereum the given ordinal was deposited but we will not know the metadata.

**We** should never need to know the inscription metadata (or even that it has some); we only care about the Ordinal ID. For example, say you mint Ordinal #12345 and it has metadata which is a PNG of a cat. You send that to our bridge. We make a NFT with metadata "Ordinal #12345".

Then you hook up your wallet to an ethereum dapp and it says "your NFT has metadata pointing to Ordinal #12345. Ordinal #12345 points to a PNG of a cat. This is what it looks like".

The dapp would have to be aware of our NFT reference format and be able to pull the metadata from bitcoin, but that's relatively straightforward.

That said, I have no idea if we can extract the sat number from the tx data

docs/rfc/rfc-9.adoc
beaushinkle: I read through [send.rs](https://github.com/casey/ord/blob/master/src/subcommand/wallet/send.rs), [transaction_builder.rs](https://github.com/casey/ord/blob/master/src/subcommand/wallet/transaction_builder.rs) and [index.rs](https://github.com/casey/ord/blob/master/src/index.rs)

It seems like `ord` needs access to a full node because it needs to load up the whole history into its special format. When you want to send an ordinal, it tries to call `get_inscription_satpoint_by_id`, which means that its database needs to know where that particular inscription ID is located, and if we're moving it using a special P2SH script, it won't.

That isn't a dealbreaker, it just means that we need to extend or modify how the ord client works if we want to support refunds. I *think* we'd also have to do the same thing if we wanted to support batch transactions.

docs/rfc/rfc-9.adoc
beaushinkle: > If I understand the proposal correctly, Ethereum would need to calculate the root using the same data as used off-chain in (2). If this is the case, I think it is problematic because Ethereum smart contract does not have access to events (logs) emitted historically. Even if there was a way to retrieve events (logs), there could be a risk of exceeding the block gas limit given the number of operations that would have to be performed to calculate the tree . We would have to limit the number of deposits into a single wallet and it's still not perfect because there is no guarantee the block gas limit will not be lowered, I think.

well dang, I was really hoping that would work

> In the v2 model, the challenger says "the wallet signed a transaction that is not a valid protocol transaction". We do not care what this transaction is. Defeating the challenge requires that we store [all honestly spent UTXOs](https://github.com/keep-network/tbtc-v2/blob/main/solidity/contracts/bridge/BridgeState.sol#L306-L312).

Okay so in this case, someone challenges and says "this signature was for an invalid transaction". Then, the wallet shows what they were signing, and proves that the thing follows the rules.

In an Ordinal case, if the wallet commits fraud then someone can find the signature and accuse it. The wallet would be unable to prove that the thing they signed was valid.

If the wallet is acting honestly and someone falsely accuses them of fraud, then the wallet shows that the thing they were signing was valid, and in order to do that, it needs to reference a collection of properly spent ordinals.

If the wallet is acting honestly and someone accuses it of failing to redeem, we'd also need a list of currently-custodied ordinals.

beaushinkle:
I overhauled the RFC given what we know now about SPV proofs and how `ord` moves Ordinals around.

pdyraga (approved):
Looks good to me! I'll wait a few days with merging to give @mhluongo and @sashatanase a chance to take a look as well.

Issue #546, opened 2023-02-17
Cross-chain Tokenized Threshold BTC: Polygon module


pdyraga:
See https://github.com/keep-network/tbtc-v2/pull/542

Implement `cross-chain/polygon` module:
- [x] Token contract inheriting from #543 
- [x] Wormhole gateway contract inheriting from #544 
- [x] Deployment scripts for Goerli
- [x] Deployment scripts for Mainnet: L1 and L2
- [x] CI build system for contracts
- [ ] CI build system for Goerli deploy
- [x] Upgradeability for token and Wormhole gateway
- [x] Mainnet deployment scenario issue

Issue #545, opened 2023-02-17
Cross-chain Tokenized Threshold BTC: Arbitrum module


pdyraga:
See https://github.com/keep-network/tbtc-v2/pull/542

Implement `cross-chain/arbitrum` module:
- [ ] Token contract inheriting from #543 
- [ ] Wormhole gateway contract inheriting from #544 
- [ ] Deployment scripts for Goerli
- [ ] Deployment scripts for Mainnet: L1 and L2
- [ ] CI build system for contracts
- [ ] CI build system for Goerli deploy
- [ ] Upgradeability for token and Wormhole gateway
- [ ] Mainnet deployment scenario issue https://github.com/keep-network/tbtc-v2/issues/577

PR #570, opened 2023-03-07
L2WormholeGateway: send and receive tokens

### LINKED ISSUE #544: Cross-chain Tokenized Threshold BTC: Wormhole Gateway
See https://github.com/keep-network/tbtc-v2/pull/542

To fully use the capabilities of the Wormhole portal and make the user experience seamless, we will implement a gateway contract having the authority to wrap Wormhole-specific tokens into TBTC tokens in a 1:1 ratio. This contract will have a minting authority for TBTC on L2. This way, no liquid market has to exist on any target chain for users to be able to cross the Wormhole bridge into the canonical TBTC.

pdyraga:
See [RFC 8: Cross-chain Tokenized Threshold BTC](https://github.com/keep-network/tbtc-v2/blob/697d9f0971433fc8e23565894dfd29e090e5a153/docs/rfc/rfc-8.adoc)

Closes #544 

# Overview

Selected cross-ecosystem bridges are given the minting authority for tBTC token on L2 and sidechains. This contract gives a minting authority to the Wormhole Bridge.

# The implementation

The process of bridging from L1 to L2 (or sidechain) looks as follows:
1. There is a tBTC holder on L1. The holder goes to the Wormhole Portal and selects the chain they want to bridge to.
2. The holder submits one transaction to L1 locking their tBTC tokens in the bridge’s smart contract. After the transaction is mined, they wait about 15 minutes for the Ethereum block finality.
3. The holder submits one transaction to L2 that is minting tokens. After that transaction is mined, they have their tBTC on L2.

The process of bridging from L2 (or sidechain) to L1 looks as follows:
1. There is a tBTC holder on L2. That holder goes to the Wormhole Portal and selects one of the L2 chains they want to bridge from.
2. The holder submits one transaction to L2 that is burning the token. After the transaction is mined, they wait about 15 minutes for the L2 block finality.
3. The holder submits one transaction to L1 unlocking their tBTC tokens from the bridge’s smart contract. After that transaction is mined, they have their tBTC on L1.

This smart contract is integrated with step 3 of L1->L2 bridging and step 1 of L2->L1 or L2->L2 bridging. When the user redeems token on L2, this contract receives the Wormhole tBTC representation and mints the canonical tBTC in an equal amount. When the user sends their token from L1, this contract burns the canonical tBTC and sends Wormhole tBTC representation through the bridge in an equal amount.

# Deployment

This contract is supposed to be deployed behind a transparent upgradeable proxy. The deployment scripts will be developed during the work on https://github.com/keep-network/tbtc-v2/issues/545, https://github.com/keep-network/tbtc-v2/issues/546, and https://github.com/keep-network/tbtc-v2/issues/547.

The gateway needs to know Wormhole Token Bridge contract address, the canonical tBTC token address, and the Wormhole tBTC token representation address. It means that we need to move some small about of tBTC through Wormhole to all target L2s first or attest the token on L2.

# Minting limit

There is a governable, absolute minting canonical tBTC limit for the gateway. The minting limit is useful for the early days of testing the system. The gateway can not mint more canonical tBTC than the minting limit allows. The minting limit is disabled by default but we should consider enabling it during the contract deployment.

The situation when the minting limit is reached is exceptional and will hopefully never happen. I wanted to keep it simple: we send wormhole tBTC OR mint canonical tBTC when the limit is reached for the given L2 redemption. We do not want to split and send dust.


solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,192 @@
...
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import "./L2TBTC.sol";
+import "../token/TBTC.sol";
+
+/// @title IWormholeTokenBridge
+/// @notice Wormhole Token Bridge interface. Contains only selected functions
+///         used by L2WormholeGateway.
+interface IWormholeTokenBridge {
```
pdyraga, line 30: Reference for reviewers: 
https://github.com/wormhole-foundation/wormhole/blob/main/ethereum/contracts/interfaces/IWormhole.sol
https://github.com/wormhole-foundation/wormhole/blob/main/ethereum/contracts/bridge/Bridge.sol#L253

pdyraga:
@conorpp I would appreciate your review here.

conorpp (approved):

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,258 @@
...
+        uint256 balanceAfter = bridgeToken.balanceOf(address(this));
+
+        uint256 amount = balanceAfter - balanceBefore;
+        address receiver = abi.decode(payload, (address));
+        require(receiver != address(0), "0x0 receiver not allowed");
+
+        // Function is nonReentrant and we need to update the balance with
+        // a call to completeTransferWithPayload first to know the amount.
+        // slither-disable-next-line reentrancy-benign
+        mintedAmount += amount;
```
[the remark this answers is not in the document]
pdyraga: Actually, there is a question if this instruction is in the right place. We should maybe move it to be in the `else` block, next to `tbtc.mint(receiver, amount)`. I'll think more about it.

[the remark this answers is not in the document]
pdyraga: See b3da668d70444d34743283b738c634b4dd1ab606, please.

lukasz-zimnoch (commented):
Comments to `receiveWormhole` function:

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,257 @@
...
+    ///        it has to be approved for L2WormholeGateway.
+    ///      - The L2WormholeGateway must have at least `amount` of the wormhole
+    ///        tBTC.
+    /// @param amount The amount of tBTC to be sent.
+    /// @param recipientChain The Wormhole recipient chain ID.
+    /// @param recipient The address of the recipient in the Wormhole format.
+    /// @param arbiterFee The Wormhole arbiter fee.
+    /// @param nonce The Wormhole nonce used to batch messages together.
+    /// @return The Wormhole sequence number.
+    function sendWormhole(
```
lukasz-zimnoch, line 215: Just a question: Suppose the minting limit was exceeded during the `receiveWormhole` call and the user got their wormhole TBTC instead of the canonical one. What if the user wants to go back to L1? The `sendWormhole` cannot be used for that right?

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,257 @@
...
+        // we check the balance before the ITokenBridge call and the balance
+        // after ITokenBridge call. This way, we are sure this contract received
+        // Wormhole tBTC token in the given amount. This is transparent to
+        // all potential upgrades of ITokenBridge implementation and no other
+        // validations are needed.
+        uint256 balanceBefore = bridgeToken.balanceOf(address(this));
+        bytes memory payload = bridge.completeTransferWithPayload(encodedVm);
+        uint256 balanceAfter = bridgeToken.balanceOf(address(this));
+
+        uint256 amount = balanceAfter - balanceBefore;
```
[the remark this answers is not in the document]
pdyraga: I don't think this is possible, do you have any particular scenario in mind? 

`completeTransferWithPayload` is supposed to redeem tokens on L2 to the caller (the gateway contract).

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,257 @@
...
+                recipient,
+                arbiterFee,
+                nonce
+            );
+    }
+
+    /// @notice Lets the governance to update the tBTC minting limit for this
+    ///         contract.
+    /// @param _mintingLimit The new minting limit.
+    function setMintingLimit(uint256 _mintingLimit) external onlyOwner {
```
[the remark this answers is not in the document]
pdyraga: I do not understand the question - this function is callable only by the governance and is used to update the minting limit. Is the name confusing? @dimpar [proposed](https://github.com/keep-network/tbtc-v2/pull/570#discussion_r1131132628) to rename this function to `updateMintingLimit`.

solidity/contracts/l2/L2WormholeGateway.sol
pdyraga: Correct, `sendWormhole` can't be used for this but they can go back to L1 working with the Wormhole token bridge contract directly.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,257 @@
...
+    ///         tBTC representation.
+    /// @dev Requirements:
+    ///      - The receiver of Wormhole tBTC should be the L2WormholeGateway
+    ///        contract.
+    ///      - The receiver of the canonical tBTC should be abi-encoded in the
+    ///        payload.
+    ///      - The receiver of the canonical tBTC must not be the zero address.
+    /// @param encodedVm A byte array containing a Wormhole VAA signed by the
+    ///        guardians.
+    function receiveWormhole(bytes memory encodedVm) external nonReentrant {
```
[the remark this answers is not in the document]
pdyraga: The user who is moving from L1 to L2. This is the only L2 transaction they need to perform. The user retrieves `encodedVm` from one of the Wormhole guardians and calls this function.

solidity/contracts/l2/L2WormholeGateway.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Sorry, the question was: do we plan to add the governance delay? My fingers wrote something different that was in my mind 🤦

solidity/contracts/l2/L2WormholeGateway.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Ok, clear!

solidity/contracts/l2/L2WormholeGateway.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Nothing particular. Just thinking loudly.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,257 @@
...
+        // after ITokenBridge call. This way, we are sure this contract received
+        // Wormhole tBTC token in the given amount. This is transparent to
+        // all potential upgrades of ITokenBridge implementation and no other
+        // validations are needed.
+        uint256 balanceBefore = bridgeToken.balanceOf(address(this));
+        bytes memory payload = bridge.completeTransferWithPayload(encodedVm);
+        uint256 balanceAfter = bridgeToken.balanceOf(address(this));
+
+        uint256 amount = balanceAfter - balanceBefore;
+        address receiver = abi.decode(payload, (address));
```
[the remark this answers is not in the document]
pdyraga: Looks like a bug, good catch. I've got confused about two levels of the payload. In the Wormhole [`Bridge` contract implementation](https://github.com/wormhole-foundation/wormhole/blob/f311517dcf544f79bbb589aee2fac76bf21b1180/ethereum/contracts/bridge/Bridge.sol#L484), there is a call to `parseAndVerifyVM` that returns `VM memory vm` and is defined [here](https://github.com/wormhole-foundation/wormhole/blob/f311517dcf544f79bbb589aee2fac76bf21b1180/ethereum/contracts/interfaces/IWormhole.sol#L19-L33). That structure has a `payload` but it's not yet the payload the user provided but `BridgeStructs.TransferWithPayload` from where we need to extract the `payload`.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,258 @@
...
+        // ensure L2WormholeGateway received Wormhole tBTC representation.
+        // Instead of going too deep into the ITokenBridge implementation,
+        // asserting who is the receiver of the token, and which token it is,
+        // we check the balance before the ITokenBridge call and the balance
+        // after ITokenBridge call. This way, we are sure this contract received
+        // Wormhole tBTC token in the given amount. This is transparent to
+        // all potential upgrades of ITokenBridge implementation and no other
+        // validations are needed.
+        uint256 balanceBefore = bridgeToken.balanceOf(address(this));
+        bytes memory payload = bridge.completeTransferWithPayload(encodedVm);
```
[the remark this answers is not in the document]
pdyraga: I think it is a [payload 3](https://github.com/wormhole-foundation/wormhole/blob/f311517dcf544f79bbb589aee2fac76bf21b1180/ethereum/contracts/bridge/Bridge.sol#L491-L495) not described in the documentation.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,258 @@
...
+                recipient,
+                arbiterFee,
+                nonce
+            );
+    }
+
+    /// @notice Lets the governance to update the tBTC minting limit for this
+    ///         contract.
+    /// @param _mintingLimit The new minting limit.
+    function setMintingLimit(uint256 _mintingLimit) external onlyOwner {
```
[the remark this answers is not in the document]
pdyraga: Good call, done in 5b929d350e2d9c7cfee89b056a78e996dd8a900f.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,258 @@
...
+
+pragma solidity ^0.8.17;
+
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+
+import "./L2TBTC.sol";
+import "../token/TBTC.sol";
```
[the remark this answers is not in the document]
pdyraga: We do not! Removed in 8d74388adc16dc7f23b04d0c14f0a48353fda1eb.

solidity/contracts/l2/L2WormholeGateway.sol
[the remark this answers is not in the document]
dimpar: Ok, the reason why I asked is that just like @lukasz-zimnoch I wasn't sure how it works and I couldn't connect the dots just by looking at the payload. I'll close this comment in favor of the one that Łukasz added.

solidity/contracts/l2/L2WormholeGateway.sol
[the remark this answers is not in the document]
pdyraga: I would rather not complicate it and had no governance delay. I can think of no strong arguments for why we would need it here.

solidity/contracts/l2/L2WormholeGateway.sol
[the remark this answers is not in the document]
pdyraga: Fixed in https://github.com/keep-network/tbtc-v2/pull/570/commits/7f351a704c771b137f8476101e99f2a42f8e0fa3.

solidity/contracts/l2/L2WormholeGateway.sol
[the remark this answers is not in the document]
pdyraga: I think this is fine. It is not specific to Wormhole and should be the same for every other bridge we add: if the bridge starts misfunctioning, we must prevent it to mint the canonical tBTC but we shouldn't go too deep into the bridge's security and what we should allow the bridge and what not. If canonical tBTC is not minted, then the problem is limited only to the given bridge.

lukasz-zimnoch (approved):

dimpar (approved):

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,275 @@
...
+            bridge.parseTransferWithPayload(encoded).payload,
+            (address)
+        );
+        require(receiver != address(0), "0x0 receiver not allowed");
+
+        // We send wormhole tBTC OR mint canonical tBTC. We do not want to send
+        // dust. Sending wormhole tBTC is an exceptional situation and we want
+        // to keep it simple.
+        if (mintedAmount + amount > mintingLimit) {
+            bridgeToken.safeTransfer(receiver, amount);
```
conorpp, line 198: can you explain what situation would result in this?

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,275 @@
...
+        } else {
+            // Function is nonReentrant and we need to update the balance with
+            // a call to completeTransferWithPayload first to know the amount.
+            // slither-disable-next-line reentrancy-benign
+            mintedAmount += amount;
+            tbtc.mint(receiver, amount);
+        }
+
+        // Function is nonReentrant and we need to extract the payload with
+        // a call to completeTransferWithPayload first.
```
conorpp, line 208: stale comment?

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,275 @@
...
+        require(receiver != address(0), "0x0 receiver not allowed");
+
+        // We send wormhole tBTC OR mint canonical tBTC. We do not want to send
+        // dust. Sending wormhole tBTC is an exceptional situation and we want
+        // to keep it simple.
+        if (mintedAmount + amount > mintingLimit) {
+            bridgeToken.safeTransfer(receiver, amount);
+        } else {
+            // Function is nonReentrant and we need to update the balance with
+            // a call to completeTransferWithPayload first to know the amount.
```
conorpp, line 201: stale comment?

pdyraga: No, this is an explanation of why do we `// slither-disable-next-line reentrancy-benign` one line later 😉 

Slither complains:
```
Reentrancy in L2WormholeGateway.receiveWormhole(bytes) (contracts/l2/L2WormholeGateway.sol#172-210):
        External calls:
        - encoded = bridge.completeTransferWithPayload(encodedVm) (contracts/l2/L2WormholeGateway.sol#184)
        State variables written after the call(s):
        - mintedAmount += amount (contracts/l2/L2WormholeGateway.sol#202)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#reentrancy-vulnerabilities-2
```

solidity/contracts/l2/L2WormholeGateway.sol
pdyraga: The minting limit is not specific to Wormhole, I think all gateways should have the same mechanism.

Let's say we have 10k tBTC bridged to Arbitrum with Wormhole. Then we add another gateway - for example Arbitrum native bridge - capable of minting canonical tBTC on Arbitrum. That gateway should start with a minting limit until we are sure everything is stable. We do not want that second gateway to print canonical tBTC that will be later brought back to L1 with Wormhole draining Wormhole Bridge contract from L1 tBTC.

Another example is a major update. Let's say Wormhole gets one and we decide to set the minting limit to what's been minted plus... 10% (?) for the next week or so until everything is proven to be working fine.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,275 @@
...
+        uint256 balanceBefore = bridgeToken.balanceOf(address(this));
+        bytes memory encoded = bridge.completeTransferWithPayload(encodedVm);
+        uint256 balanceAfter = bridgeToken.balanceOf(address(this));
+
+        uint256 amount = balanceAfter - balanceBefore;
+        address receiver = abi.decode(
+            bridge.parseTransferWithPayload(encoded).payload,
+            (address)
+        );
+        require(receiver != address(0), "0x0 receiver not allowed");
```
a5-pickle, line 192: See comment in `sendWormhole`, but this can be removed if the `receiver` outbound is checked against `address(0)`.

solidity/contracts/l2/L2WormholeGateway.sol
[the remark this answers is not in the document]
a5-pickle: Regarding `nonReentrant`, Token Bridge actually has a protection against redeeming the same VAA again. When a Token Bridge VAA is redeemed, its message body hash is stored in a map. This map is used to check whether the hash has already been set in this map. So you can save some gas by removing this.

See https://github.com/wormhole-foundation/wormhole/blob/main/ethereum/contracts/bridge/Bridge.sol#L497.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,275 @@
...
+    /// @param amount The amount of tBTC to be sent.
+    /// @param recipientChain The Wormhole recipient chain ID.
+    /// @param recipient The address of the recipient in the Wormhole format.
+    /// @param arbiterFee The Wormhole arbiter fee.
+    /// @param nonce The Wormhole nonce used to batch messages together.
+    /// @return The Wormhole sequence number.
+    function sendWormhole(
+        uint256 amount,
+        uint16 recipientChain,
+        bytes32 recipient,
```
a5-pickle, line 236: In the body of `sendWormhole`, should `require(recipient != address(0), ...)` so you wouldn't have to check it for inbound transfers. If zero address gets passed here, there is no going back at that point.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,275 @@
...
+        uint256 arbiterFee,
+        uint32 nonce
+    ) external nonReentrant returns (uint64) {
+        require(
+            bridgeToken.balanceOf(address(this)) >= amount,
+            "Not enough liquidity in wormhole to bridge"
```
a5-pickle, line 242: This error message may be a misnomer because the method is checking the balance of `bridgeToken` on this contract (and not on Token Bridge).

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,275 @@
...
+            arbiterFee,
+            nonce
+        );
+
+        mintedAmount -= amount;
+        tbtc.burnFrom(msg.sender, amount);
+
+        bridgeToken.safeApprove(address(bridge), amount);
+        return
+            bridge.transferTokens(
```
a5-pickle, line 258: Did you mean to `transferTokensWithPayload` here? In `receiveWormhole` you are expecting a specific payload.

If so, you will not need `arbiterFee` and will have to encode the same payload as you expect to decode in `receiveWormhole`.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,275 @@
...
+        );
+        require(
+            address(_tbtc) != address(0),
+            "L2TBTC token address must not be 0x0"
+        );
+
+        bridge = _bridge;
+        bridgeToken = _bridgeToken;
+        tbtc = _tbtc;
+        mintingLimit = type(uint256).max;
```
a5-pickle, line 155: Depending on whether you want to mint everything from the get-go or if you want more control over the minting limit, you may want to set this value explicitly to zero (to be clear about the intention) and then require that the owner call `updateMintingLimit` after initialization.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,275 @@
...
+    ///      - The L2WormholeGateway must have at least `amount` of the wormhole
+    ///        tBTC.
+    /// @param amount The amount of tBTC to be sent.
+    /// @param recipientChain The Wormhole recipient chain ID.
+    /// @param recipient The address of the recipient in the Wormhole format.
+    /// @param arbiterFee The Wormhole arbiter fee.
+    /// @param nonce The Wormhole nonce used to batch messages together.
+    /// @return The Wormhole sequence number.
+    function sendWormhole(
+        uint256 amount,
```
a5-pickle, line 234: Token Bridge doesn't have a protection against `amount == 0`. So unless your contract interaction (via UI?) will protect against zero amounts, this method will succeed and cost the user some gas for transferring zero tokens.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,275 @@
...
+    ///         tBTC representation.
+    /// @dev Requirements:
+    ///      - The receiver of Wormhole tBTC should be the L2WormholeGateway
+    ///        contract.
+    ///      - The receiver of the canonical tBTC should be abi-encoded in the
+    ///        payload.
+    ///      - The receiver of the canonical tBTC must not be the zero address.
+    /// @param encodedVm A byte array containing a Wormhole VAA signed by the
+    ///        guardians.
+    function receiveWormhole(bytes memory encodedVm) external nonReentrant {
```
gator-boi, line 172: Slight gas optimization here, you could pass `encodedVM` as `calldata` instead of `memory`.

solidity/contracts/l2/L2WormholeGateway.sol
gator-boi: Based on the comment from @a5-pickle, if the intention is to use `transferTokensWithPayload` the contract should check if the recipient address is a trusted contract.

solidity/contracts/l2/L2WormholeGateway.sol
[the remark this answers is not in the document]
a5-pickle: However, the amount can be zero if the tokens bridged to this contract are not `bridgeToken`. So to protect against taking custody of irrelevant tokens, you may want to add `require(amount > 0, ...)`. And this would also save the user gas on a revert because amount zero is a no-op (since no new amount is actually being minted in this case).

solidity/contracts/l2/L2WormholeGateway.sol
a5-pickle, line 172: Regarding `nonReentrant`, Token Bridge actually has a protection against redeeming the same VAA again. When a Token Bridge VAA is redeemed, its message body hash is stored in a map. This map is used to check whether the hash has already been set in this map. So you can save some gas by removing this.

See https://github.com/wormhole-foundation/wormhole/blob/main/ethereum/contracts/bridge/Bridge.sol#L497.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,275 @@
...
+        // we check the balance before the ITokenBridge call and the balance
+        // after ITokenBridge call. This way, we are sure this contract received
+        // Wormhole tBTC token in the given amount. This is transparent to
+        // all potential upgrades of ITokenBridge implementation and no other
+        // validations are needed.
+        uint256 balanceBefore = bridgeToken.balanceOf(address(this));
+        bytes memory encoded = bridge.completeTransferWithPayload(encodedVm);
+        uint256 balanceAfter = bridgeToken.balanceOf(address(this));
+
+        uint256 amount = balanceAfter - balanceBefore;
```
a5-pickle, line 187: The amount can be zero if the tokens bridged to this contract are not `bridgeToken`. So to protect against taking custody of irrelevant tokens, you may want to add `require(amount > 0, ...)`. And this would also save the user gas on a revert because amount zero is a no-op (since no new amount is actually being minted in this case).

solidity/contracts/l2/L2WormholeGateway.sol
pdyraga: Correct me if I am wrong but I think we may need another function or additional logic inside of `sendWormhole`.

When the user bridges from L1 to L2, they set `L2WormholeGateway` contract as the recipient, and they encode the L2 receiver address in the payload. On L2, the user calls `receiveWormhole` which `completeTransferWithPayload`, and the receiver address from the payload receives the canonical tBTC.

When the user bridges back from L2 to L1, they use `sendWormhole` and `transferTokens`. They set their address on L1 as the recipient. `wormholeTBTC` goes back to the bridge, and on L1 user can unlock their canonical tBTC into the recipient address straight from the Wormhole Portal Bridge contract.

The problem I believe you are talking about is when the user bridges from L2 to L2 and we should use `transferTokensWithPayload` in this case. This functionality is currently missing.

We could have two functions: 
- `sendWormhole` for cases when the receiver is an end-address (bridge back to L1).
- `sendWormholeWithPayload` for cases when the receiver is a contract (another L2).

Does it make sense? I may be missing something from the Wormhole design.

solidity/contracts/l2/L2WormholeGateway.sol
pdyraga: Good call. Please see 3b7dbdc312fa60865573f86f8be85d4cc8eb232d.

solidity/contracts/l2/L2WormholeGateway.sol
pdyraga: See 51b1ce628e163594971c0e27b7889ac657e6946b, please.

pdyraga: Makes sense. See e6976fde27458729202134dba5911c5887939193, please.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,371 @@
...
+
+    /// @notice Lets the governance to update the tBTC gateway address on the
+    ///         chain with the given Wormhole ID.
+    /// @param chainId Wormhole ID of the chain.
+    /// @param gateway Address of tBTC gateway on the given chain.
+    function updateGatewayAddress(uint16 chainId, address gateway)
+        external
+        onlyOwner
+    {
+        bytes32 wormholeAddress = toWormholeAddress(gateway);
```
gator-boi, line 346: Is there a plan to deploy tBTC contracts on non-evm chains? If so, `updateGatewayAddress` should take a 32 byte `gateway` argument.

solidity/contracts/l2/L2WormholeGateway.sol
pdyraga: Added in 10879e035d09052a7a768543f28778e182df2912.

solidity/contracts/l2/L2WormholeGateway.sol
```diff
@@ -0,0 +1,376 @@
...
+    ///         tBTC. Useful in a situation when user received wormhole tBTC
+    ///         instead of canonical tBTC. One example of such situation is
+    ///         when the minting limit was exceeded but the user minted anyway.
+    /// @dev Requirements:
+    ///      - The sender must have at least `amount` of the Wormhole tBTC and
+    ///        it has to be approved for L2WormholeGateway.
+    /// @param amount The amount of Wormhole tBTC to deposit.
+    function depositBridgeToken(uint256 amount) external {
+        bridgeToken.transferFrom(msg.sender, address(this), amount);
+        tbtc.mint(msg.sender, amount);
```
gator-boi, line 336: Does the `mintedAmount` state variable need to be incremented here?

pdyraga: Yep, good catch. Added in da36bed40c381e94527e3265880988b7fad5c21b.

solidity/contracts/l2/L2WormholeGateway.sol
gator-boi: If you guys aren't planning to write a non-evm specific implementation (and deploy canonical tBTC on a non-evm chain) then this should work. However, if you decide to write a Solana contract (for example) the gateway address would have to be passed as bytes32 (all Solana addresses are 32 bytes).

solidity/contracts/l2/L2WormholeGateway.sol
pdyraga: Please see https://github.com/keep-network/tbtc-v2/pull/570/commits/5ecf2219d6fdfcb58d43df81e2e2822854d935fe and https://github.com/keep-network/tbtc-v2/pull/570/commits/ee8e218f9e03786f92bef6ccb50279d285812fe1.

The short version for those not up-to-date with the recent discussion: we had a missing path in `L2WormholeGateway` where tokens were moved between L2s and the target L2 had Wormhole tBTC gateway deployed capable of minting canonical tBTC. In this case, the transfer should be done with a payload allowing the user to receive tBTC on the target chain using the gateway deployed there.

The previous approach was fine when tBTC was transferred back to L1. It was not enough when tokens were transferred between L2s having canonical token minted by gateways.

The missing path has been covered in the `sendTbtc` function. Depending on whether there is a gateway on the target chain, transfer with payload or transfer with no payload is used.

solidity/contracts/l2/L2WormholeGateway.sol
pdyraga: I added the validation in `sentTbtc` (renamed the function earlier) but I think it makes sense to leave the validation here as well. If there is any bug in the code with extracting or composing the payload, the user will not be allowed to burn their tokens. This will give us a chance to fix the problem and upgrade the contract.

pdyraga:
@gator-boi @a5-pickle @conorpp @dimpar @lukasz-zimnoch This is ready for another look!

solidity/contracts/l2/L2WormholeGateway.sol
pdyraga: That makes sense, thanks for the explanation. See https://github.com/keep-network/tbtc-v2/pull/570/commits/e6b29854de931e6473076ca4f52a953aba7210bd, please.

PR #575, opened 2023-03-14
Wallet coordinator contract

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3512

Here we present the `WalletCoordinator` contract. This contract aims to facilitate the coordination of the off-chain wallet members during complex multi-chain wallet operations like deposit sweeping, redemptions, or moving funds. Such processes involve various moving parts and many steps that each individual wallet member must do. Given the distributed nature of the off-chain wallet software, full off-chain implementation is challenging and prone to errors, especially byzantine faults. This contract provides a single and trusted on-chain coordination point thus taking the riskiest part out of the off-chain software. The off-chain wallet members can focus on the core tasks and do not bother about electing a trusted coordinator or aligning internal states using complex consensus algorithms.

As mentioned above, the `WalletCoordinator` contract is designed to handle not only deposit sweeps but also future redemptions and moving funds actions. The contract is upgradeable so adding new features should be straightforward. 

### The flow 

In the first iteration, the deposit sweep process will look as follows:
1. An outside proposal submitter or designated wallet member submits a deposit sweep proposal using `submitDepositSweepProposal` function. A `DepositSweepProposalSubmitted` event holding the proposal is emitted.
2. Off-chain wallet members handle the `DepositSweepProposalSubmitted` event and confirm it.
3. Off-chain wallet members fetch necessary data and validate the proposal using a read-only call to the `validateDepositSweepProposal` function.
4. Off-chain wallet members do an additional off-chain validation.
5. Off-chain wallet members produce a proper sweep transaction and broadcast it over the Bitcoin network.
6. Once the time lock for the wallet expires, a new proposal can be submitted.

### Potential problems 

The presented approach brings some problems that must be addressed in the next iterations:
- Invalid proposals hold the wallet locked and idle for a certain period. For now, this can be mitigated with a manual `unlockWallet` call done by the `WalletCoordinator` owner but we probably need a more robust solution eventually.
- A specific variant of the above: a malicious wallet member can DoS the wallet with invalid proposals

### Remaining work
- Off-chain part of the deposit sweep feature 
- Coordinator role check


solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,182 @@
...
+pragma solidity 0.8.17;
+
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+import "./Bridge.sol";
+import "./Deposit.sol";
+import "./Wallets.sol";
+
+// TODO: Documentation and unit tests.
+// TODO: Should we make it pausable?
```
pdyraga, line 25: Nope. Let's not add any pause mechanisms.

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,182 @@
...
+    //       their own.
+    function unlockWallet(bytes20 walletPubKeyHash) external onlyOwner {
+        // Just in case, allow the owner to unlock the wallet earlier.
+        walletLock[walletPubKeyHash] = 0;
+    }
+
+    function submitDepositSweepProposal(DepositSweepProposal calldata proposal)
+        external
+        onlyProposalSubmitter
+        onlyAfterWalletLock(proposal.walletPubKeyHash)
```
pdyraga, line 132: What is the purpose of it?

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    function updateDepositSweepProposalValidity(
+        uint32 _depositSweepProposalValidity
+    ) external onlyOwner {
+        depositSweepProposalValidity = _depositSweepProposalValidity;
+        emit DepositSweepProposalValidityUpdated(_depositSweepProposalValidity);
+    }
+
+    function updateDepositSweepMaxSizeUpdated(uint16 _depositSweepMaxSize)
```
pdyraga, line 90:
```suggestion
    function updateDepositSweepMaxSize(uint16 _depositSweepMaxSize)
```

solidity/contracts/bridge/WalletCoordinator.sol
lukasz-zimnoch: We enforce 1 proposal at a time. Once we add redemptions and moving funds into the mix, that will be super important that only a single operation take place at once.

solidity/contracts/bridge/WalletCoordinator.sol
lukasz-zimnoch: Removed the TODO: a04cebe4

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,542 @@
...
+    /// @param proposal The actual deposit sweep proposal
+    /// @param walletMemberContext Optional parameter holding some data allowing
+    ///        to confirm the wallet membership of the caller. This parameter is
+    ///        relevant only when the caller is not a registered proposal
+    ///        submitter but claims to be a member of the target wallet.
+    /// @dev Requirements:
+    ///      - The caller is either a proposal submitter or a member of the
+    ///        target wallet,
+    ///      - The wallet is not time locked.
+    function submitDepositSweepProposal(
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 5cecbe18

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,542 @@
...
+    ///         validation of the incoming proposals. Thanks to it, most
+    ///         of the work can be done using a single readonly contract call.
+    ///         Worth noting, the validation done here is not exhaustive as some
+    ///         conditions may not be verifiable within the on-chain function or
+    ///         checking them may be easier on the off-chain side. For example,
+    ///         this function does not check the SPV proofs and confirmations of
+    ///         the deposit funding transactions as this would require an
+    ///         integration with the difficulty relay that greatly increases
+    ///         complexity. Instead of that, each off-chain wallet member is
+    ///         supposed to do that check on their own.
```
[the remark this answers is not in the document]
lukasz-zimnoch: For now, it's just the validation mentioned in the example. Not sure if we need something else so I wanted to avoid being too specific.

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,542 @@
...
+    /// @param proposal The validated proposal.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See e43e36e1

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,542 @@
...
+    ///      keccak256(fundingTxHash | fundingOutputIndex) for gas efficiency.
+    struct DepositKey {
+        bytes32 fundingTxHash;
+        uint32 fundingOutputIndex;
+    }
+
+    /// @notice Helper structure holding deposit extra data required during
+    ///         deposit sweep proposal validation. Basically, this structure
+    ///         is a combination of BitcoinTx.Info and relevant parts of
+    ///         Deposit.DepositRevealInfo.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See e43e36e1

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,542 @@
...
+
+    /// @notice Helper structure carrying data necessary to validate the wallet
+    ///         membership of the caller.
+    struct WalletMemberContext {
+        uint32[] walletMembersIDs;
+        uint256 walletMemberIndex;
+    }
+
+    /// @notice Helper structure representing a deposit sweep proposal.
+    ///         Holds the target wallet 20-byte public key hash and candidate
```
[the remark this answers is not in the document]
lukasz-zimnoch: See e43e36e1

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,542 @@
...
+    ///         the proposal on their own, before taking any action.
+    /// @param proposal The actual deposit sweep proposal
+    /// @param walletMemberContext Optional parameter holding some data allowing
+    ///        to confirm the wallet membership of the caller. This parameter is
+    ///        relevant only when the caller is not a registered proposal
+    ///        submitter but claims to be a member of the target wallet.
+    /// @dev Requirements:
+    ///      - The caller is either a proposal submitter or a member of the
+    ///        target wallet,
+    ///      - The wallet is not time locked.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See e43e36e1

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,542 @@
...
+        emit DepositSweepMaxSizeUpdated(_depositSweepMaxSize);
+    }
+
+    /// @notice Submits a deposit sweep proposal. Locks the target wallet
+    ///         for a specific time, equal to the proposal validity period.
+    ///         This function does not store the proposal in the state but
+    ///         just emits an event that serves as a guiding light for wallet
+    ///         off-chain members. Wallet members are supposed to validate
+    ///         the proposal on their own, before taking any action.
+    /// @param proposal The actual deposit sweep proposal
```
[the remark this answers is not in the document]
lukasz-zimnoch: See e43e36e1

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,542 @@
...
+            bytes32 ecdsaWalletID = bridge
+                .wallets(walletPubKeyHash)
+                .ecdsaWalletID;
+
+            walletMember = walletRegistry.isWalletMember(
+                ecdsaWalletID,
+                walletMemberContext.walletMembersIDs,
+                msg.sender,
+                walletMemberContext.walletMemberIndex
+            );
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yep, this is what I mentioned in the "Potential problems" section of the description. Introducing a coordinator role seems good though I think we should play with it in a follow-up PR to not block this one. Does it work for you?

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,542 @@
...
+    ///         For example, if a deposit sweep proposal was submitted at
+    ///         2 pm and depositSweepProposalValidity is 4 hours, the next
+    ///         proposal (of any type) can be submitted after 6 pm.
+    uint32 public depositSweepProposalValidity;
+
+    /// @notice The minimum time that must elapse since the deposit reveal
+    ///         before a deposit becomes eligible for a deposit sweep.
+    ///
+    ///         For example, if a deposit was revealed at 9 am and depositMinAge
+    ///         is 2 hours, the deposit is eligible for sweep after 11 am.
```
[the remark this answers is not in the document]
lukasz-zimnoch: > Does it mean the off-chain client does not have to validate confirmations of inputs

Nope, this check is not about that. The confirmations have to be checked anyway if we want to avoid producing signatures of invalid BTC transactions.

> Or is it just an additional check to ensure the wallet can't get blocked with data that will not pass the off-chain validation anyway.

This check ensures the block finality for the happy path. If we wait 2 hours before taking a proper deposit for the sweep, we are sure at least 20 ETH blocks and 6 BTC blocks have been mined in the meantime. In such a case, the BTC confirmations will be checked off-chain anyway though that will be just formality.

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,542 @@
...
+        uint32[] walletMembersIDs;
+        uint256 walletMemberIndex;
+    }
+
+    /// @notice Helper structure representing a deposit sweep proposal.
+    ///         Holds the target wallet 20-byte public key hash and candidate
+    ///         deposits that should be part of the sweep.
+    struct DepositSweepProposal {
+        bytes20 walletPubKeyHash;
+        DepositKey[] depositsKeys;
```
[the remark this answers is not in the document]
lukasz-zimnoch: I tried to avoid bothering the proposal submitter with that. Wallet members can just take the maximum allowed fee from the Bridge and use a % (50%?) of it as a starting point, then bump by another percentage if needed.

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,542 @@
...
+    /// @notice Allows to unlock the given wallet before their time lock expires.
+    ///         This function should be used in exceptional cases where
+    ///         something went wrong and there is a need to unlock the wallet
+    ///         without waiting.
+    /// @param walletPubKeyHash 20-byte public key hash of the wallet
+    /// @dev Requirements:
+    ///      - The caller must be the owner.
+    function unlockWallet(bytes20 walletPubKeyHash) external onlyOwner {
+        // Just in case, allow the owner to unlock the wallet earlier.
+        walletLock[walletPubKeyHash] = 0;
```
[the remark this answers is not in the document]
lukasz-zimnoch: Good idea, see bf609a92

solidity/contracts/bridge/WalletCoordinator.sol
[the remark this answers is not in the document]
pdyraga: I think the fee should be published to the chain. The maximum allowed fee from the Bridge is quite large by nature - we do not want to update it too often yet we need to make sure the transaction can be done in case of a fee spike in the network. It is hard to say if 50% or 25% is the right initial value.

Another problem is with bumping the fee. This is not in the scope of the initial implementation but should be done soon after. If the wallet signed a transaction but it's still in the mempool, there should be a mechanism to sign the same data with a higher fee. For the same reasons we coordinate sweep proposition on-chain, we should also coordinate the fee bump on-chain. Now, there are two options:
- Announce the fee bump (more complicated), which requires an on-chain event with a fee.
- Assume it's another sweep proposition (less complicated), also requires the fee to be known on-chain so that we do not start with the same 25% or 50% of the max allowed fee again.

I think the only logic we need is to accept the fee as a parameter, validate it is not higher than the max one, and emit it as a part of the event.

solidity/contracts/bridge/WalletCoordinator.sol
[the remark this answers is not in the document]
pdyraga: Just to clarify: this is about Ethereum block finality and Bitcoin block finality is just the bonus, correct?

If that's the case, let's add a `@dev` section explaining it, please.

solidity/contracts/bridge/WalletCoordinator.sol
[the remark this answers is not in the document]
pdyraga: Sure, let's just add a TODO to the code and I'll keep this discussion unresolved for now as a reference.

solidity/contracts/bridge/WalletCoordinator.sol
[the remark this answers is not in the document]
pdyraga: It's also about validating confirmations of deposits revealed to Ethereum. How it's done off-chain is another discussion. It might mean checking if the current`depositMinAge` is enough. I think the client should do something. There are weird scenarios like Ethereum fork with `depositMinAge` lowered to 0 and proposing a sweep of non-confirmed transactions. This is quite a complicated scenario but if we can avoid it with a simple call, I'd do it.

I'd just add something like that at the end of the `@dev Requirements` section:

```
The following off-chain validation must be performed as a bare minimum:
- Inputs used for the sweep transaction have enough Bitcoin confirmations,
- Deposits revealed to the Bridge have enough Ethereum confirmations.
```

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,571 @@
...
+                .wallets(walletPubKeyHash)
+                .ecdsaWalletID;
+
+            walletMember = walletRegistry.isWalletMember(
+                ecdsaWalletID,
+                walletMemberContext.walletMembersIDs,
+                msg.sender,
+                walletMemberContext.walletMemberIndex
+            );
+        }
```
pdyraga, line 186: How gas-expensive is the `if (!proposalSubmitter)` path compared to the `proposalSubmitter == true` path?

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,571 @@
...
+        emit DepositSweepProposalSubmitted(proposal, msg.sender);
+    }
+
+    /// @notice Wraps `submitDepositSweepProposal` call and reimburses the
+    ///         caller's transaction cost.
+    /// @dev See `submitDepositSweepProposal` function documentation.
+    function submitDepositSweepProposalWithReimbursement(
+        DepositSweepProposal calldata proposal,
+        WalletMemberContext calldata walletMemberContext
+    ) external refundable(msg.sender) {
```
pdyraga, line 360: Let's confirm in tests this is enough. Sometimes a gas offset needs to be added. [Here](https://github.com/keep-network/tbtc-v2/blob/6ac4d2d237173f85e6124bac90e70adf95bc90cc/solidity/contracts/relay/LightRelayMaintainerProxy.sol#L68-L71) is an example from #579.

solidity/test/bridge/WalletCoordinator.test.ts
```diff
@@ -0,0 +1,921 @@
...
+  })
+
+  describe("addProposalSubmitter", () => {
+    before(async () => {
+      await createSnapshot()
+    })
+
+    after(async () => {
+      await restoreSnapshot()
+    })
```
pdyraga, line 70: Are they needed if we create and restore snapshot for every test modifying the state? Not a big deal but with the number of tests we have, doing and restoring snapshots only when necessary could potentially help with `yarn test` execution time.

solidity/test/bridge/WalletCoordinator.test.ts
```diff
@@ -0,0 +1,921 @@
...
+        })
+
+        it("should do the refund", async () => {
+          // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+          expect(reimbursementPool.refund).to.have.been.calledOnce
+
+          expect(reimbursementPool.refund.getCall(0).args[0]).to.be.closeTo(
+            BigNumber.from(32000),
+            2000
+          )
```
pdyraga, line 912: This assertion is not actually checking if the caller is not net-negative, I am afraid.

How complicated would it be to implement tests with a real `ReimbursementPool`?

solidity/test/bridge/WalletCoordinator.test.ts
```diff
@@ -918,4 +944,966 @@ describe("WalletCoordinator", () => {
...
+  let depositScriptHash
+  if (witness) {
+    depositScriptHash = `220020${ethers.utils
+      .sha256(depositScript)
+      .substring(2)}`
+  } else {
+    const sha256Hash = ethers.utils.sha256(depositScript)
+    const ripemd160Hash = ethers.utils.ripemd160(sha256Hash).substring(2)
+    depositScriptHash = `17a914${ripemd160Hash}87`
+  }
```
pdyraga, line 1867: image

pdyraga:
I went through all the files, this is looking really good. I have no other comments that those already left.

solidity/contracts/bridge/WalletCoordinator.sol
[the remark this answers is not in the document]
lukasz-zimnoch: See 89a95831

solidity/contracts/bridge/WalletCoordinator.sol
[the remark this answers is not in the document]
lukasz-zimnoch: See 0848de33

solidity/contracts/bridge/WalletCoordinator.sol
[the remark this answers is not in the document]
lukasz-zimnoch: See 0848de33

solidity/contracts/bridge/WalletCoordinator.sol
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/575/commits/42b167fdc80acb876795432691a3fe9038bbeb0a

solidity/contracts/bridge/WalletCoordinator.sol
lukasz-zimnoch: The `proposalSubmitter == true` is `44401` on average. The `if !(proposalSubmitter)` path is `65338` on average.

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,617 @@
...
+    /// @notice The minimum time that must elapse since the deposit reveal
+    ///         before a deposit becomes eligible for a deposit sweep.
+    ///
+    ///         For example, if a deposit was revealed at 9 am and depositMinAge
+    ///         is 2 hours, the deposit is eligible for sweep after 11 am.
+    ///
+    /// @dev Forcing deposit minimum age ensures block finality for Ethereum.
+    ///      In the happy path case, i.e. where the deposit is revealed immediately
+    ///      after being broadcast on the Bitcoin network, the minimum age
+    ///      check also ensures block finality for Bitcoin.
```
pdyraga, line 130: That's quite a good point for [this](https://github.com/keep-network/tbtc-v2/pull/575#discussion_r1151685038) discussion.  We can't trust `depositMinAge` for Bitcoin finality validation.  👌🏻

solidity/test/bridge/WalletCoordinator.test.ts
lukasz-zimnoch: Honestly, I'd keep it as is because:
- The execution time is negligible (around 1 millisecond)
- That helps avoid future debugging in case new state modifications are added at the top level of the scenario. It happened to me to forget the snapshot instructions and lose time in finding the cause of the weird behavior of specific scenarios.

solidity/test/bridge/WalletCoordinator.test.ts
lukasz-zimnoch: I don't think using real pool changes anything in this context. After all, it's about calling `reimbursementPool.refund` with proper parameters.

solidity/test/bridge/WalletCoordinator.test.ts
pdyraga: Works for me!

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,571 @@
...
+    ///         the deposit funding transactions as this would require an
+    ///         integration with the difficulty relay that greatly increases
+    ///         complexity. Instead of that, each off-chain wallet member is
+    ///         supposed to do that check on their own.
+    /// @param proposal The sweeping proposal to validate.
+    /// @param depositsExtraInfo Deposits extra data required to perform the validation.
+    /// @dev Requirements:
+    ///      - The target wallet must be in the Live state,
+    ///      - The number of deposits included in the sweep must be in
+    ///        the range [1, `depositSweepMaxSize`],
```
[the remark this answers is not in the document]
lukasz-zimnoch: We don't bother about the main UTXO at all. We always require at least one proper deposit in the proposal.

solidity/test/bridge/WalletCoordinator.test.ts
pdyraga: Right, and the *proper parameters* are the problem. 😉 See [this](https://github.com/keep-network/tbtc-v2/pull/575#discussion_r1159593044) comment - right now we are assuming no gas offset is needed. Either way, we should confirm it in a test - gas offset is not needed or gas offset is needed and the value is `x`.

lukasz-zimnoch: Yep yep, I'll reply in the mentioned comment.

solidity/contracts/bridge/WalletCoordinator.sol
pdyraga: OK, so it's at least 30% of the cost of calling this function if called by the coordinator. Let's keep in mind when addressing this [TODO](https://github.com/keep-network/tbtc-v2/pull/575#discussion_r1151564813) that it has to be gas-optimized solution.

solidity/contracts/bridge/WalletCoordinator.sol
lukasz-zimnoch: Here is the gas consumption:
```
·---------------------------------------------------------------------|---------------------------|--------------|-----------------------------·
|                        Solc version: 0.8.17                         ·  Optimizer enabled: true  ·  Runs: 1000  ·  Block limit: 30000000 gas  │
······································································|···························|··············|······························
|  Methods                                                                                                                                     │
······················|···············································|·············|·············|··············|···············|··············
|  Contract           ·  Method                                       ·  Min        ·  Max        ·  Avg         ·  # calls      ·  eur (avg)  │
······················|···············································|·············|·············|··············|···············|··············
|  WalletCoordinator  ·  submitDepositSweepProposal                   ·      37566  ·      65338  ·       48857  ·            7  ·          -  │
······················|···············································|·············|·············|··············|···············|··············
|  WalletCoordinator  ·  submitDepositSweepProposalWithReimbursement  ·          -  ·          -  ·       59796  ·            1  ·          -  │
······················|···············································|·············|·············|··············|···············|··············

```
While the refunded amount is around `32000` as per the assertion mentioned in https://github.com/keep-network/tbtc-v2/pull/575#discussion_r1159631579. 

That is, we need an offset. However, not sure what should be the reference value for the refund?

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -0,0 +1,571 @@
...
+                depositExtraInfo.fundingTx.outputVector,
+                depositExtraInfo.fundingTx.locktime
+            )
+            .hash256View();
+
+        // Make sure the funding tx provided as part of deposit extra data
+        // actually matches the deposit referred by the given deposit key.
+        if (depositKey.fundingTxHash != depositExtraFundingTxHash) {
+            return false;
+        }
```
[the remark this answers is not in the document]
lukasz-zimnoch: See c6baec0a

solidity/contracts/bridge/WalletCoordinator.sol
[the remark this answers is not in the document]
pdyraga: Good point!

solidity/contracts/bridge/WalletCoordinator.sol
lukasz-zimnoch: Done in 205add89. Set `25000` as the initial offset value. That gives a refund slightly above `60000` which should exceed the average cost of `submitDepositSweepProposalWithReimbursement` being `59796`

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -487,8 +494,36 @@ contract WalletCoordinator is OwnableUpgradeable, Reimbursable {
...
+        uint256 depositTxFeeRemainder = sweepTxFee % depositsCount;
+        uint256 depositTxFee = (sweepTxFee - depositTxFeeRemainder) /
+            depositsCount;
+
+        (, , uint64 depositTxMaxFee, ) = bridge.depositParameters();
+
+        require(
+            depositTxFee + depositTxFeeRemainder <= depositTxMaxFee,
+            "Proposed transaction fee is too high"
+        );
```
pdyraga, line 545: I am confused about why do we add `depositTxFeeRemainder`. 

`depositTxMaxFee` is a per-deposit fee:

https://github.com/keep-network/tbtc-v2/blob/9429ad99a49ac652ead710bdc2654cecd3e97cb1/solidity/contracts/bridge/Bridge.sol#L1659-L1662

Say the proposed `sweepTxFee` is 11 and there are 5 deposits.

`depositTxFeeRemainder = 1`
`depositTxFe = 2`

Then we check whether `2 + 1 <= depositTxMaxFee` but we do not have that `1` remainder left for all deposits in the sweep batch.

Shouldn't we just take the floor (`sweepTxFee / depositsCount`) and check if it's `<= depositTxMaxFee` ?

lukasz-zimnoch: The remainder is incurred by the last deposit so `depositTxFee + depositTxFeeRemainder` is the actual maximum per-deposit fee used within the given sweep. We also do the same within `DepositSweep.submitDepositSweepProof`: https://github.com/keep-network/tbtc-v2/blob/9429ad99a49ac652ead710bdc2654cecd3e97cb1/solidity/contracts/bridge/DepositSweep.sol#L203

pdyraga: That makes sense. Can we copy the code comments from `depositSweepTxFeeDistribution` and just before we execute the `require` explain that:

```
// The transaction fee is incurred by each deposit evenly except for the last
// deposit that has the indivisible remainder additionally incurred. 
// See `DepositSweep.submitDepositSweepProof`.
// We must make sure the highest value of the deposit transaction fee does
// not exceed the maximum value limited by the governable parameter.
```

pdyraga: Can we also add comments to the specific lines? Those in `depositSweepTxFeeDistribution` helps to understand the logic behind modulo:

https://github.com/keep-network/tbtc-v2/blob/9429ad99a49ac652ead710bdc2654cecd3e97cb1/solidity/contracts/bridge/DepositSweep.sol#L564-L572

lukasz-zimnoch: https://github.com/keep-network/tbtc-v2/pull/575/commits/67035d18d3e2115a916f931edb66b853ab99fa1d

pdyraga (approved):

PR #583, opened 2023-04-03
Arbitrum: Add artifacts export script

nkuba:
We want the artifacts to be exported to `artifacts/l1` and `artifacts/l2` directories. Here we implement automatic export on package publication, with a helper task defined in https://github.com/keep-network/hardhat-helpers/pull/40.

To prepare the artifacts export run a command for a given network, e.g.:
```
yarn run export-artifacts:goerli
```
or to export and publish:
```
npm run prepublishOnly --network=goerli
```

----

Depends on: https://github.com/keep-network/hardhat-helpers/pull/40

### DIFF
--- a/cross-chain/arbitrum/package.json
+++ b/cross-chain/arbitrum/package.json
@@ -41,7 +41,7 @@
     "@openzeppelin/contracts-upgradeable": "^4.8.1"
   },
   "devDependencies": {
-    "@keep-network/hardhat-helpers": "^0.6.0-pre.18",
+    "@keep-network/hardhat-helpers": "^0.6.0-pre.20",
     "@nomiclabs/hardhat-ethers": "^2.0.6",
     "@nomiclabs/hardhat-etherscan": "^3.1.0",
     "@nomiclabs/hardhat-waffle": "2.0.3",
--- a/cross-chain/arbitrum/package.json
+++ b/cross-chain/arbitrum/package.json
@@ -30,7 +30,9 @@
     "lint:config": "prettier --check '**/*.@(json|yaml)'",
     "lint:config:fix": "prettier --write '**/*.@(json|yaml)'",
     "prepack": "tsc -p tsconfig.export.json && hardhat export-artifacts export/artifacts",
-    "prepublishOnly": "echo TODO: Copy deployment artifacts to artifacts directory",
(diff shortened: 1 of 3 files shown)


cross-chain/arbitrum/package.json
```diff
@@ -30,7 +30,9 @@
     "lint:config": "prettier --check '**/*.@(json|yaml)'",
     "lint:config:fix": "prettier --write '**/*.@(json|yaml)'",
     "prepack": "tsc -p tsconfig.export.json && hardhat export-artifacts export/artifacts",
-    "prepublishOnly": "echo TODO: Copy deployment artifacts to artifacts directory",
+    "export-artifacts:goerli": "yarn hardhat export-deployment-artifacts --network arbitrumGoerli",
```
dimpar, line 33: export-artifacts:goerli -> export-artifacts:arbitrumGoerli?

same for mainnet

dimpar:
We probably need to change either `hardhat-helpers` version in `package.json` or in the `lock.json` file.

cross-chain/arbitrum/package.json
nkuba: The reason for defining the name with `goerli` and `mainnet` is that it will combine both l1 and l2. The `export-artifacts:goerli` script should export deployment artifacts for both networks: `goerli` to `artifacts/l1` and `arbitrumGoerli` to `artifacts/l2`, to do that the script will be defined as follow:
```
"export-artifacts:goerli": "for i in goerli arbitrumGoerli; do yarn hardhat export-deployment-artifacts --network $i; done",
```

It is also a preparation for the automated deployment, where we have just `goerli` environment passed around (https://github.com/keep-network/tbtc-v2/blob/4523b83f2b4612d93d91ab55e254bb13cec632a1/.github/workflows/contracts.yml#L29-L31)

nkuba:
> We probably need to change either `hardhat-helpers` version in `package.json` or in the `lock.json` file.

It's ready.

cross-chain/arbitrum/package.json
dimpar: ah, I see now, thx for explaining.

dimpar (approved):

PR #589, opened 2023-04-11
Deployment scripts for `WalletCoordinator` contract

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3512

In #575, we introduced the `WalletCoordinator` contract. Here we add some Hardhat deployment scripts allowing us to deploy that contract.

### DIFF
--- a/solidity/deploy/34_deploy_wallet_coordinator.ts
+++ b/solidity/deploy/34_deploy_wallet_coordinator.ts
@@ -1,0 +1,43 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types"
+import { DeployFunction } from "hardhat-deploy/types"
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+  const { deployments, ethers, getNamedAccounts, helpers } = hre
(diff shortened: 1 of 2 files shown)


solidity/deploy/34_deploy_wallet_coordinator.ts
```diff
@@ -0,0 +1,43 @@
...
+    await helpers.upgrades.deployProxy("WalletCoordinator", {
+      contractName: "WalletCoordinator",
+      initializerArgs: [Bridge.address],
+      factoryOpts: {
+        signer: await ethers.getSigner(deployer),
+      },
+      proxyOpts: {
+        kind: "transparent",
+      },
+    })
```
pdyraga, line 20: Non-blocking this PR but let's verify the behavior: 

OpenZeppelin deploys only one `ProxyAdmin` contract for the given network. All transparent proxy contracts are owned by it. Since `Bridge` contract is already deployed on mainnet with its own `ProxyAdmin`, we must ensure the underlying mechanism detects the existing `ProxyAdmin` properly.

pdyraga (approved):

Issue #603, opened 2023-04-18
TS: Improve events fetching


nkuba:
We get events from a contract in one call:
https://github.com/keep-network/tbtc-v2/blob/328c7f8d3025b97e461a590e81fa6bb494daa2f8/typescript/src/ethereum.ts#L205-L222
When the blocks range for the call gets too wide some of the providers may reject such a request (e.g. QuickNode accepts max 10 000 blocks range).

We need to introduce a fallback querying in batches like we did in V1: https://github.com/keep-network/tbtc-monitoring/blob/e169357d7b8c638d4eaf73d52aa8f53ee4aebc1d/src/lib/ethereum-helper.js#L44-L73

PR #604, opened 2023-04-19
Partial events pulls

nkuba:
This changeset adds the partial events pulls mechanism that can be useful for some Ethereum providers (e.g. Alchemy, QuickNode) who limit the maximum number of blocks that can be queried or events that can be pulled at once but allows to get an arbitrary number of events in batches of a certain size. This solution is based on https://github.com/keep-network/tbtc.js/pull/141

Closes https://github.com/keep-network/tbtc-v2/issues/603


pdyraga (approved):

mhluongo (commented):
Quick question

PR #609, opened 2023-04-25
Wallet coordination: Heartbeat action

pdyraga:
~~Depends on #607~~

# Heartbeat action

With the existing, pretty computationally-expensive tECDSA signing algorithm, [we would prefer](https://github.com/keep-network/keep-core/pull/3521#discussion_r1173652203) not to execute heartbeats at the same time as other wallet actions (sweeping, redemptions, moving funds). Moreover, we would prefer not to execute heartbeats from multiple wallets at the same time.

Such coordination is complicated off-chain for the same reason as the coordination of sweeps and redemptions. We could - as we do now - execute heartbeats at fixed hours but it does not help with coordination between heartbeats and sweeps. Also, it does not solve the problem of multiple heartbeats simultaneously.

For this reason, the heartbeat has been made a first-class wallet action coordinated in the on-chain contract just like the rest of the wallet actions.

# Goodies

There are two other goodies introduced in this PR.

## Aggregate updates

I aggregated deposit sweep proposal update parameter functions into one. This approach allows to reduce the contract size by 0.3KB and makes the contract future-proof for numerous future move funds and redemption parameters. If we were updating parameters in separate functions, we'd probably stumble upon contract size problems in the future. Aggregating deposit sweep parameter update functions into one with an upgrade could be problematic given the `*Updated` events already potentially emitted. 

## Testing with Reimbursement Pool

Instead of testing ether refunds with smock mocks we now test with a real `ReimbursementPool` contract. This change allows to fine-tune the gas offset parameters. The previous ones were not very accurate it seems. The other problem was that we did not take into consideration that the first call will always cost more given that the contract fields have to be set to
non-default values.

### DIFF
--- a/solidity/contracts/bridge/WalletCoordinator.sol
+++ b/solidity/contracts/bridge/WalletCoordinator.sol
@@ -238,12 +238,13 @@ contract WalletCoordinator is OwnableUpgradeable, Reimbursable {
         (, , , reimbursementPool) = _bridge.contractReferences();
 
         heartbeatRequestValidity = 1 hours; // TODO: Check this value!
+        heartbeatRequestGasOffset = 5_000;
 
         depositSweepProposalValidity = 4 hours;
         depositMinAge = 2 hours;
         depositRefundSafetyMargin = 24 hours;
         depositSweepMaxSize = 5;
-        depositSweepProposalSubmissionGasOffset = 25000;
+        depositSweepProposalSubmissionGasOffset = 5_000;
     }
 
     /// @notice Adds the given address to the set of coordinator addresses.
@@ -309,7 +310,6 @@ contract WalletCoordinator is OwnableUpgradeable, Reimbursable {
     /// @param _depositMinAge The new value of `depositMinAge`.
     /// @param _depositRefundSafetyMargin The new value of `depositRefundSafetyMargin`.
     /// @param _depositSweepMaxSize The new value of `depositSweepMaxSize`.
-
     /// @dev Requirements:
     ///      - The caller must be the owner.
     function updateDepositSweepProposalParameters(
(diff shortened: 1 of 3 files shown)


solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -48,6 +48,8 @@ contract WalletCoordinator is OwnableUpgradeable, Reimbursable {
     enum WalletAction {
         /// @dev The wallet does not perform any action.
         Idle,
+        /// @dev The wallet is executing heartbeat.
+        Heartbeat,
```
lukasz-zimnoch, line 52: There is some client-side code that uses this enum (e.g. `parseWalletActionType` in `pkg/chain/ethereum/tbtc.go`). Until we reflect that change there, a new enum in the middle will cause a hidden bug as `Heartbeat` will be recognized as `DepositSweep`. Can we add the `Heartbeat` value at the end? This way the client code will just fail on an unrecognized value. This way, we won't forget about updating it on the client-side.

pdyraga: I forgot to explain why I decided to keep it first. This was intentional!

The reasoning is: with this order, if we need to upgrade, we will always append after what is already implemented.

Yes, we will have to update the mapping on the Go side but I am planning to integrate `WalletCoordinator`-based signing into the client right after this PR (I'll work on it today). I think the approach where `Heartbeat` and `DepositSweep` are first in the enum makes sense because they will be implemented in the client and any other functionality that we'll be adding in future releases will not require reorganization of the enum or state fields in the upgradeable `WalletCoordinator` contract. Imagine we decide to upgrade the contract and add another `WalletAction` in the future. If `Heartbeat` and `DepositSweep` are at the beginning, we will not have to worry about their values. We will also not have to care about the order of fields in the contract: we will always append after what is already implemented.

lukasz-zimnoch: Ok, makes sense!

lukasz-zimnoch (approved):

PR #610, opened 2023-04-26
Run system tests weekly

michalinacienciala:
We're decreasing frequency of running the tests (we do not develop contracts as actively as we used to so running them every 24h is too aggressive, especially that the tests use testnet funds). From now on we will run tests weekly (every Sunday at 0:00). We can still trigger the tests manually if needed.

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -2,7 +2,7 @@ name: System tests
 
 on:
   schedule:
-    - cron: "0 0 * * *"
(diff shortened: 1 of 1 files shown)


dimpar:
I think my preference would be to run them twice a week e.g. on Tuesdays and Thursdays to catch potential errors earlier, but maybe I'm being too defensive. We can leave once a week and see how it goes.

dimpar (approved):

PR #612, opened 2023-05-02
Use Human Readable Outputs in Deposit Monitoring

beaushinkle:
We currently have discord alerts that look like this:

image

This is serviceable on a laptop where one can paste the fields into a block explorer, or copy the sats and divide by 1e8, but *very* annoying on a phone. In either case, I would prefer to have a block explorer link and the BTC amount in BTC rather than sats, for easy viewing.

This PR makes that change!

Tagging @nkuba and @lukasz-zimnoch for review.

Notably, I don't see how to test this and I'm not sure how to deploy it, but I checked around in the code and I don't see anything downstream that cares about the names of these fields, so I think we'd be fine.

### DIFF
--- a/monitoring/src/deposit-monitor.ts
+++ b/monitoring/src/deposit-monitor.ts
@@ -12,19 +12,18 @@ const satsToRoundedBTC = (sats: BigNumber): string =>
 
 const hashUrls = (chainEvent: DepositRevealedChainEvent) => {
   let fundingHashUrlPrefix = ""
-  if (context.environment === Environment.Mainnet) {
-    fundingHashUrlPrefix = "https://mempool.space/tx/"
-  }
-  if (context.environment === Environment.Testnet) {
-    fundingHashUrlPrefix = "https://mempool.space/testnet/tx/"
-  }
-
   let revealHashUrlPrefix = ""
-  if (context.environment === Environment.Mainnet) {
-    revealHashUrlPrefix = "https://etherscan.io/tx/"
-  }
-  if (context.environment === Environment.Testnet) {
-    revealHashUrlPrefix = "https://goerli.etherscan.io/tx/"
+  switch (context.environment) {
+    case Environment.Mainnet: {
+      fundingHashUrlPrefix = "https://mempool.space/tx/"
+      revealHashUrlPrefix = "https://etherscan.io/tx/"
+      break
+    }
+    case Environment.Testnet: {
+      fundingHashUrlPrefix = "https://mempool.space/testnet/tx/"
+      revealHashUrlPrefix = "https://goerli.etherscan.io/tx/"
+      break
+    }
   }
 
   const fundingHash = chainEvent.fundingTxHash.toString()
--- a/monitoring/src/deposit-monitor.ts
+++ b/monitoring/src/deposit-monitor.ts
@@ -14,7 +14,7 @@ function satsToRoundedBTC(sats: BigNumber): string {
 const DepositRevealed = (
   chainEvent: DepositRevealedChainEvent
 ): SystemEvent => ({
-  title: "Deposit Revealed",
+  title: "Deposit revealed",
   type: SystemEventType.Informational,
   data: {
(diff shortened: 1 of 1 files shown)


monitoring/src/deposit-monitor.ts
```diff
@@ -7,30 +7,34 @@ import type { SystemEvent, Monitor as SystemEventMonitor } from "./system-event"
...
+
 const DepositRevealed = (
   chainEvent: DepositRevealedChainEvent
 ): SystemEvent => ({
-  title: "Deposit revealed",
+  title: "Deposit Revealed",
   type: SystemEventType.Informational,
   data: {
-    btcFundingTxHash: chainEvent.fundingTxHash.toString(),
+    btcFundingTxHash: `https://mempool.space/tx/${chainEvent.fundingTxHash.toString()}`,
```
lukasz-zimnoch, line 20: Remember that we have both testnet and mainnet so we need to use the correct links depending on the network. You can take the environment from here: https://github.com/keep-network/tbtc-v2/blob/main/monitoring/src/context.ts. Same comment to all links in this file.

nkuba, line 20: Maybe add separate properties for the URLs to keep the `btcFundingTxHash` and `ethRevealTxHash` return the hashes?

monitoring/src/deposit-monitor.ts
```diff
@@ -7,30 +7,34 @@ import type { SystemEvent, Monitor as SystemEventMonitor } from "./system-event"
...
 
+function satsToRoundedBTC(sats: BigNumber): string {
+  return (sats.div(BigNumber.from(1e6)).toNumber() / 100).toFixed(2)
+}
+
 const DepositRevealed = (
   chainEvent: DepositRevealedChainEvent
 ): SystemEvent => ({
-  title: "Deposit revealed",
+  title: "Deposit Revealed",
```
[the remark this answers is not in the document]
beaushinkle: [e06fcc](https://github.com/keep-network/tbtc-v2/pull/612/commits/e06fccd37c90eb341aade13675d8e763c88aff22)

monitoring/src/deposit-monitor.ts
beaushinkle: [e953b0](https://github.com/keep-network/tbtc-v2/pull/612/commits/e953b05feda4f54b0864ecf41e0b7d7c4aaf6e26)

beaushinkle: [80c599](https://github.com/keep-network/tbtc-v2/pull/612/commits/80c599af5e0e4ed094a39c6d52b94db6ef407950)

monitoring/src/deposit-monitor.ts
```diff
@@ -1,39 +1,79 @@
...
+    fundingHashUrlPrefix = "https://mempool.space/testnet/tx/"
+  }
+
+  let revealHashUrlPrefix = ""
+  if (context.environment === Environment.Mainnet) {
+    revealHashUrlPrefix = "https://etherscan.io/tx/"
+  }
+  if (context.environment === Environment.Testnet) {
+    revealHashUrlPrefix = "https://goerli.etherscan.io/tx/"
+  }
```
lukasz-zimnoch, line 28: I would simplify this part a bit and use a `switch` statement that would make it cleaner:
```suggestion
  let fundingHashUrlPrefix = ""
  let revealHashUrlPrefix = ""
  
  switch(context.environment) { 
   case Environment.Mainnet: { 
      fundingHashUrlPrefix = "https://mempool.space/tx/"
      revealHashUrlPrefix = "https://etherscan.io/tx/"
      break
   } 
   case Environment.Testnet: { 
      fundingHashUrlPrefix = "https://mempool.space/testnet/tx/"
      revealHashUrlPrefix = "https://goerli.etherscan.io/tx/"
      break
   }
} 
```

beaushinkle: [02d32a](https://github.com/keep-network/tbtc-v2/pull/612/commits/02d32a8f4572403b146110c23e1c579fc3ca33a5)

PR #616, opened 2023-05-10
Normalize the amount to bridge before calling Wormhole

pdyraga:
We need to eliminate the dust that cannot be bridged with Wormhole due to the decimal shift in the Wormhole Bridge contract. We cut the dust the same way as the Wormhole Bridge.

See [this part](https://github.com/wormhole-foundation/wormhole/blob/96682bdbeb7c87bfa110eade0554b3d8cbf788d2/ethereum/contracts/bridge/Bridge.sol#L276-L288) in the Wormhole Bridge contract.


a5-pickle (approved):
Nice fix

solidity/test/l2/L2WormholeGateway.test.ts
```diff
@@ -437,7 +456,7 @@ describe("L2WormholeGateway", () => {
         })
 
         context("when the target chain has a tBTC gateway", () => {
-          const amount = 998
+          const amount = to1e18(998)
```
nkuba, line 459: Do we have a test case that verifies `sendTbtc` with an amount that actually has some dust?

pdyraga: Look at the line 400, please.

nkuba: Please correct me if I'm wrong, but from my understanding, the dust dropping will change the actual amount that is transferred, and we need to ensure that it really happens.

In Line 381 we test the input amount less than the dust, so we expect a revert.
In Line 400 we test the input amount that equals dust, so the resulting amount will be the same.
We don't fully cover boundary values testing - we're missing cases where the input amount will differ from the actual transfer amount, e.g.:
- `10000000001` will result in a transfer of `10000000000`
- `19999999999` will result in a transfer of `10000000000`
But `20000000000` will transfer a matching amount of `20000000000`.

pdyraga: OK, I now see what you say. You are right, please see eeb04c2d0c35f070cfeeee617e25dac38f87b04f.

solidity/test/l2/L2WormholeGateway.test.ts
```diff
@@ -502,12 +503,170 @@ describe("L2WormholeGateway", () => {
...
+                  recipientChain,
+                  padTo32Bytes(recipient),
+                  arbiterFee,
+                  nonce
+                )
+            ).to.be.revertedWith("Amount too low to bridge")
+          })
+        })
+
+        context("when the amount just above the dust", () => {
```
nkuba, line 525:
```suggestion
        context("when the amount is just above the dust", () => {
```

nkuba (approved):
LGTM, left a tiny non-blocking comment.

solidity/test/l2/L2WormholeGateway.test.ts
pdyraga: https://github.com/keep-network/tbtc-v2/pull/616/commits/ff398b89352242751753e1bd7738c7543e309a1a

nkuba (approved):

PR #629, opened 2023-06-06
Don't run postinstall script separately from yarn install

michalinacienciala:
Background:
Previously we've been running postinstall script in a separate step from the `yarn install` command. We've been doing that as a workaround for a problem we had with the postinstall script in the `@threshold-network/solidity-contracts` package (the script in that package often randomly failed). Running `yarn upgrade` with `--ignore-scripts` flag prevented the execution of postinstall script in the main project and its depandencies. And running `yarn run postinstall` after did execute the postinstall script in the main project.

The change:
Recently we have refactored the `@threshold-network/solidity-contracts` project and got rid of the problematic script. We can go back to executing `yarn install` without the `--ignore-scripts` flag.

Ref:
https://github.com/threshold-network/solidity-contracts/pull/143
https://github.com/threshold-network/token-dashboard/pull/531

### DIFF
--- a/monitoring/Dockerfile
+++ b/monitoring/Dockerfile
@@ -24,8 +24,7 @@ COPY package.json yarn.lock ./
 COPY tsconfig.json ./
 COPY src ./src
 
-RUN yarn install --frozen-lockfile --ignore-scripts
-RUN yarn run postinstall
+RUN yarn install --frozen-lockfile
 
 RUN yarn build
 


dimpar (approved):

dimpar:
Actually, now I have second thoughts about removing it. It was not running `prepare-dependency.sh` script and the reason why we have here `yarn run postinstall` might be a different one. If so, then we should not touch anything here. @lukasz-zimnoch can you please confirm if we can remove the `--ignore-scripts` flag here? Also, we have the same flag on line 34 @michalinacienciala

lukasz-zimnoch:
> Actually, now I have second thoughts about removing it. It was not running `prepare-dependency.sh` script and the reason why we have here `yarn run postinstall` might be a different one. If so, then we should not touch anything here. @lukasz-zimnoch can you please confirm if we can remove the `--ignore-scripts` flag here? Also, we have the same flag on line 34 @michalinacienciala

The postinstall must be executed as `tbtc-v2-monitoring` depends on `tbtc-v2` which in turn depends on `bcoin` that has a problematic `bcrypto` module that contains native code and must be rebuilt after `yarn install`. This has nothing to do with `@threshold-network/solidity-contracts`

michalinacienciala:
> The postinstall must be executed as tbtc-v2-monitoring depends on tbtc-v2 which in turn depends on bcoin that has a problematic bcrypto module that contains native code and must be rebuilt after yarn install. This has nothing to do with @threshold-network/solidity-contracts. I recommend reverting this PR as it breaks the Docker build.

Well, I don't fully agree :) You're right that we need to run postinstall in order to execute `npm rebuild bcrypto`. But by default, postinstall scripts are executed during `yarn-install (--frozen-lockfile)`. It's only because of the `--ignore-scripts` flag that they are not being executed and we need to run a separate `yarn run postinstall` command.
And the reason we were running the `--ignore-scripts` was due to the misbehaving `prepare-dependency.sh` postinstall script of `threshold-network/solidity-contracts` which is an indirect dependency of the `monitoring` package (see https://github.com/keep-network/tbtc-v2/commit/5fbdc20c308dd89165ec708a9a8f22b37b1f26d3).
Running `yarn install --frozen-lockfile --ignore-scripts` and then `yarn run postinstall` makes sure only the postinstall scripts in the `monitoring` project will be executed, no postinstall scripts in subdependencies will be run.

The reason I wanted to go back to executing all postinstall scripts, even in the subdependencies is because we got rid of the misbehaving `prepare-dependencies.sh` script in `threshold-network/solidity-contracts` and I was thinking we no longer need to worry about that. However, I did not think about two matters:
1) I didn't update the `yarn.lock` and because we use `--frozen-lockfile`, still the old packages were used (including the `threshold-network/solidity-contracts` package that still contains the faulty script).
2) Only the new `development`-tagged `threshold-network/solidity-contracts` package was updated to not contain the faulty script. The latest `mainnet` and `goerli` packages still have the script (we didn't publish a new packages there). And in `monitoring` dependencies we are using those packages. So we still could hit the issue with failing installation if we'd remove the `--ignore-scripts` flag.

So generally, I agree that my change was premature and should be reverted - I just wanted to clarify for future us what exactly was the reasoning for running the install with the `--ignore-scripts` flag.

lukasz-zimnoch:
> It's only because of the --ignore-scripts flag that they are not being executed and we need to run a separate yarn run postinstall command

Nope :) the problem with `bcrypto` is more complicated. Specifically, because `bcoin` holds their `node_modules` in VCS (https://github.com/bcoin-org/bcoin/tree/master/node_modules). That means `bcrypto` dependency is already here when we run `yarn install`. For an unknown reason, `yarn install` does not build all native components in that case, hence the need for manual `postinstall` that does it again under the hood. The `--ignore-scripts` has basically nothing to do here.

Shadowfiend:
Idle observer here, but that info is captured neither in a code comment nor in the commit message that introduced the separate post install call (https://github.com/keep-network/tbtc-v2/commit/5fbdc20c308dd89165ec708a9a8f22b37b1f26d3). Probably worth fixing that given the complexity of the reason :)

michalinacienciala:
> Idle observer here, but that info is captured neither in a code comment nor in the commit message that introduced the separate post install call (https://github.com/keep-network/tbtc-v2/commit/5fbdc20c308dd89165ec708a9a8f22b37b1f26d3). Probably worth fixing that given the complexity of the reason :)

FYI, I'm adding the explanation in https://github.com/keep-network/tbtc-v2/pull/642.

PR #630, opened 2023-06-07
Find wallet for a redemption

r-czajkowski:
Add `findWalletForRedemption` function that returns the wallet details needed to request a redemption based on the redemption amount. This function looks for the oldest active wallet(wallet must be in `Live` state) that has enough BTC to handle a redemption request.


typescript/src/redemption.ts
```diff
@@ -406,3 +409,79 @@ export async function getRedemptionRequest(
...
+
+/**
+ * Finds the oldest active wallet that has enough BTC to handle a redemption request.
+ * @param amount The amount to be redeemed.
+ * @param bridge The handle to the Bridge on-chain contract.
+ * @param bitcoinClient Bitcoin client used to interact with the network.
+ * @param bitcoinNetwork Bitcoin network.
+ * @returns Promise with the wallet details needed to request a redemption.
+ */
+export async function findWalletForRedemption(
```
michalsmiarowski, line 421: We could add some tests to this function in `typescript/test/redemption.test.ts`

typescript/src/redemption.ts
```diff
@@ -406,3 +409,79 @@ export async function getRedemptionRequest(
...
+  amount: BigNumber,
+  bridge: Bridge,
+  bitcoinClient: BitcoinClient,
+  bitcoinNetwork: BitcoinNetwork
+): Promise<{
+  walletPublicKey: string
+  mainUTXO: UnspentTransactionOutput
+}> {
+  const wallets = await bridge.getNewWalletRegisteredEvents()
```
michalsmiarowski, line 430: Are those events sorted from oldest to newest by default? Can we confirm that? It looks like `getEvents` method from `ethereum-helpers` has `fromBlock` argument which indicates that it starts searching from the given block and goes "up". Does that also mean that the results are returned in the similar order (from oldest to newest)?

michalsmiarowski (changes requested):
Left some comments to look at

typescript/src/redemption.ts
r-czajkowski: Yes, I can confirm that they are sorted from oldest to newest by default. But actually, I can't find this in docs 😄.

typescript/src/redemption.ts
r-czajkowski: 97d3c65

typescript/test/utils/mock-bridge.ts
```diff
@@ -314,4 +314,21 @@ export class MockBridge implements Bridge {
...
+      utils.solidityKeccak256(
+        ["bytes32", "uint32", "uint64"],
+        [
+          utxo.transactionHash.reverse().toPrefixedString(),
+          utxo.outputIndex,
+          utxo.value,
+        ]
+      )
+    )
+  }
```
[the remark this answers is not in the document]
r-czajkowski: In `typescript` dir they are only in 2 places in Ethereum implementation and in mock-bridge. However, I think it's hard to create a reusable function for that because it will only fit for cases where we want to hash 3 types and in exactly the same order. We can create the wrapper for `utils.solidityKeccak256` in `ethereum-helpers` but not sure if it's worth it.

typescript/src/ethereum.ts
```diff
@@ -694,6 +697,57 @@ export class Bridge
       signerOrProvider: this._instance.signer || this._instance.provider,
     })
   }
+
+  async wallets(walletPublicKeyHash: string): Promise<Wallet> {
```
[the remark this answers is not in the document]
r-czajkowski: dee7020

typescript/src/ethereum.ts
```diff
@@ -694,6 +697,57 @@ export class Bridge
...
+    }
+  }
+
+  /**
+   * Builds the UTXO hash based on the UTXO components. UTXO hash is computed as
+   * `keccak256(txHash | txOutputIndex | txOutputValue)`.
+   * @param utxo UTXO components.
+   * @returns The hash of the UTXO.
+   */
+  buildUTXOHash(utxo: UnspentTransactionOutput): Hex {
```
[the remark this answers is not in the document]
r-czajkowski: dee7020

typescript/src/ethereum.ts
```diff
@@ -694,6 +697,57 @@ export class Bridge
...
+    const wallet = await backoffRetrier<Wallets.WalletStructOutput>(
+      this._totalRetryAttempts
+    )(async () => {
+      return await this._instance.wallets(walletPublicKeyHash)
+    })
+
+    return this.parseWalletDetails(wallet)
+  }
+
+  private async parseWalletDetails(
```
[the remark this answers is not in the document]
r-czajkowski: dee7020

typescript/src/redemption.ts
```diff
@@ -406,3 +409,79 @@ export async function getRedemptionRequest(
...
+ * @param bridge The handle to the Bridge on-chain contract.
+ * @param bitcoinClient Bitcoin client used to interact with the network.
+ * @param bitcoinNetwork Bitcoin network.
+ * @returns Promise with the wallet details needed to request a redemption.
+ */
+export async function findWalletForRedemption(
+  amount: BigNumber,
+  bridge: Bridge,
+  bitcoinClient: BitcoinClient,
+  bitcoinNetwork: BitcoinNetwork
```
nkuba, line 425: Instead of passing the `bitcoinNetwork` we could pull it from `bitcoinClient.getNetwork()`. But this would require an additional call to the Electrum server, so I'm not sure if it's worth it.

typescript/src/redemption.ts
```diff
@@ -406,3 +409,79 @@ export async function getRedemptionRequest(
...
+      bitcoinNetwork
+    )
+
+    const utxos = await bitcoinClient.findAllUnspentTransactionOutputs(
+      walletBitcoinAddress
+    )
+
+    // We need to find correct utxo- utxo components must point to the recent
+    // main UTXO of the given wallet, as currently known on the chain.
+    const utxo = utxos.find((utxo) =>
```
nkuba, line 462:
```suggestion
    const mainUtxo = utxos.find((utxo) =>
```

typescript/src/redemption.ts
```diff
@@ -406,3 +409,79 @@ export async function getRedemptionRequest(
...
+  for (const wallet of wallets) {
+    const prefixedWalletPublicKeyHash =
+      wallet.walletPublicKeyHash.toPrefixedString()
+    const { state, mainUtxoHash, walletPublicKey } = await bridge.wallets(
+      prefixedWalletPublicKeyHash
+    )
+
+    // Wallet must be in Live state.
+    if (state !== WalletState.Live) continue
+
```
nkuba, line 449: Let's also ensure the `mainUtxoHash` is non-zero before we proceed.

typescript/src/redemption.ts
```diff
@@ -406,3 +409,79 @@ export async function getRedemptionRequest(
...
+      walletBitcoinAddress
+    )
+
+    // We need to find correct utxo- utxo components must point to the recent
+    // main UTXO of the given wallet, as currently known on the chain.
+    const utxo = utxos.find((utxo) =>
+      mainUtxoHash.equals(bridge.buildUTXOHash(utxo))
+    )
+
+    if (!utxo) continue
```
nkuba, line 466: Would it be beneficial to add some debug logs for cases when we continue the loop execution to the next wallet?

nkuba (commented):
It looks great! I left some comments. I haven't tested it yet though.

typescript/src/redemption.ts
```diff
@@ -406,3 +409,79 @@ export async function getRedemptionRequest(
...
+ * @param bitcoinClient Bitcoin client used to interact with the network.
+ * @param bitcoinNetwork Bitcoin network.
+ * @returns Promise with the wallet details needed to request a redemption.
+ */
+export async function findWalletForRedemption(
+  amount: BigNumber,
+  bridge: Bridge,
+  bitcoinClient: BitcoinClient,
+  bitcoinNetwork: BitcoinNetwork
+): Promise<{
```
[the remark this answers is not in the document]
r-czajkowski: As I can see in `redemption.ts` we almost in all cases we pass the handles at the end eg. `getRedemptionRequest`, `submitRedemptionProof`, `requestRedemption`. Only `submitRedemptionTransaction` fn has handles in the first place. So I would reorder only `bitcoinNetwork` to be next to the `amount`. WDYT?:
```ts
function findWalletForRedemption(
  amount: BigNumber,
  bitcoinNetwork: BitcoinNetwork,
  bridge: Bridge,
  bitcoinClient: BitcoinClient,
): Promise<{
  walletPublicKey: string
  mainUTXO: UnspentTransactionOutput
}>
```

typescript/src/redemption.ts
r-czajkowski: c920964

typescript/src/redemption.ts
r-czajkowski: 96b1cfe

typescript/src/redemption.ts
r-czajkowski: fb6009a

typescript/src/redemption.ts
```diff
@@ -406,3 +409,79 @@ export async function getRedemptionRequest(
...
+      }
+    | undefined = undefined
+  let maxAmount = BigNumber.from(0)
+
+  for (const wallet of wallets) {
+    const prefixedWalletPublicKeyHash =
+      wallet.walletPublicKeyHash.toPrefixedString()
+    const { state, mainUtxoHash, walletPublicKey } = await bridge.wallets(
+      prefixedWalletPublicKeyHash
+    )
```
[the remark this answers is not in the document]
r-czajkowski: 5b42133

typescript/src/redemption.ts
```diff
@@ -406,3 +409,79 @@ export async function getRedemptionRequest(
...
+ * @returns Promise with the wallet details needed to request a redemption.
+ */
+export async function findWalletForRedemption(
+  amount: BigNumber,
+  bridge: Bridge,
+  bitcoinClient: BitcoinClient,
+  bitcoinNetwork: BitcoinNetwork
+): Promise<{
+  walletPublicKey: string
+  mainUTXO: UnspentTransactionOutput
```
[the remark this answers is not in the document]
r-czajkowski: 9a0adab

typescript/src/redemption.ts
```diff
@@ -406,3 +409,79 @@ export async function getRedemptionRequest(
 
   return redemptionRequests[0]
 }
+
+/**
+ * Finds the oldest active wallet that has enough BTC to handle a redemption request.
+ * @param amount The amount to be redeemed.
```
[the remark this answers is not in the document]
r-czajkowski: b5a8e47

typescript/src/ethereum.ts
```diff
@@ -694,6 +697,67 @@ export class Bridge
...
+  }
+
+  // eslint-disable-next-line valid-jsdoc
+  /**
+   * Builds the UTXO hash based on the UTXO components. UTXO hash is computed as
+   * `keccak256(txHash | txOutputIndex | txOutputValue)`.
+   *
+   * @see {ChainBridge#buildUTXOHash}
+   */
+  buildUTXOHash(utxo: UnspentTransactionOutput): Hex {
```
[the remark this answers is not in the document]
r-czajkowski: 8d8982a

typescript/src/ethereum.ts
```diff
@@ -639,18 +642,18 @@ export class Bridge
...
-      this._totalRetryAttempts
-    )(async () => {
-      return await this._instance.wallets(activeWalletPublicKeyHash)
-    })
+    const { ecdsaWalletID } = await this.wallets(activeWalletPublicKeyHash)
 
+    return (await this.toCompressedWalletPublicKey(ecdsaWalletID)).toString()
+  }
+
+  private async toCompressedWalletPublicKey(ecdsaWalletID: Hex): Promise<Hex> {
```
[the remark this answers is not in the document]
r-czajkowski: ea80cac

typescript/src/redemption.ts
r-czajkowski: I'm ok with both approaches. Decided to pass the `bitcoinNetwork` because we pass it as well in `calculateDepositAddress` function so we are consistent. But on the other hand the `calculateDepositAddress` does not accept the `BitcoinClient` so there is no other way to get the Bitcoin network 🤔.

typescript/src/redemption.ts
r-czajkowski: Should I address this here or opened a separate PR?

nkuba: Seems like a simple change that we can include in this PR.

typescript/src/redemption.ts
```diff
@@ -406,3 +410,103 @@ export async function getRedemptionRequest(
...
+          "0x0000000000000000000000000000000000000000000000000000000000000000"
+        )
+      )
+    ) {
+      console.debug(
+        `Wallet is not in Live state or main utxo is not set. \
+        Continue the loop execution to the next wallet...`
+      )
+      continue
+    }
```
nkuba, line 460: Could we split it into two separate checks and separate log messages?

typescript/src/redemption.ts
```diff
@@ -406,3 +410,103 @@ export async function getRedemptionRequest(
...
+    if (
+      state !== WalletState.Live ||
+      mainUtxoHash.equals(
+        Hex.from(
+          "0x0000000000000000000000000000000000000000000000000000000000000000"
+        )
+      )
+    ) {
+      console.debug(
+        `Wallet is not in Live state or main utxo is not set. \
```
nkuba, line 456: Let's add some context to the message and include the walletPublicKeyHash.

typescript/src/redemption.ts
```diff
@@ -406,3 +410,103 @@ export async function getRedemptionRequest(
...
+      walletData = {
+        walletPublicKey: walletPublicKey.toString(),
+        mainUtxo,
+      }
+
+      break
+    }
+
+    console.debug(
+      `The wallet (${wallet.walletPublicKeyHash.toPrefixedString()}) \
```
nkuba, line 500: Formatting has to be fixed, as this message was logged like this:
```
The wallet (0x328d992e5f5b71de51a1b40fcc4056b99a88a647)       cannot handle the redemption request. Continue the loop execution to the       next wallet...
```

typescript/src/redemption.ts
```diff
@@ -406,3 +410,103 @@ export async function getRedemptionRequest(
...
+      state !== WalletState.Live ||
+      mainUtxoHash.equals(
+        Hex.from(
+          "0x0000000000000000000000000000000000000000000000000000000000000000"
+        )
+      )
+    ) {
+      console.debug(
+        `Wallet is not in Live state or main utxo is not set. \
+        Continue the loop execution to the next wallet...`
```
nkuba, line 457: We need to fix formatting similar to other logs.

typescript/test/data/redemption.ts
```diff
@@ -666,3 +667,123 @@ export const redemptionProof: RedemptionProofTestData = {
...
+      ),
+      ecdsaWalletID: Hex.from(
+        "0x7a1437d67f49adfd44e03ddc85be0f6988715d7c39dfb0ca9780f1a88bcdca25"
+      ),
+      walletPublicKeyHash: Hex.from(
+        "0x328d992e5f5b71de51a1b40fcc4056b99a88a647"
+      ),
+    },
+  ],
+  wallets: {
```
nkuba, line 730: Can we add some wallets that should be skipped by the wallet selection? E.g. some non-live state, not enough utxo value, main utxo not set. 
These non-matching cases should be on top of the list to see that the mechanism skips them and finds the first wallet that matches all the criteria.

nkuba (changes requested):

typescript/src/redemption.ts
```diff
@@ -406,3 +410,103 @@ export async function getRedemptionRequest(
...
+
+    const utxos = await bitcoinClient.findAllUnspentTransactionOutputs(
+      walletBitcoinAddress
+    )
+
+    // We need to find correct utxo- utxo components must point to the recent
+    // main UTXO of the given wallet, as currently known on the chain.
+    const mainUtxo = utxos.find((utxo) =>
+      mainUtxoHash.equals(bridge.buildUtxoHash(utxo))
+    )
```
nkuba, line 496: In case a wallet is working on something (e.g. redemption) and a bitcoin transaction was already submitted by the wallet to the bitcoin chain (new utxo returned from bitcoin client), but proof hasn't been submitted yet to the Bridge (old main utxo returned from the Bridge) this script will not find such a wallet.

cc @pdyraga

typescript/test/utils/mock-bridge.ts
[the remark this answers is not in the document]
michalsmiarowski: I see, let's keep it as it is now then

typescript/src/redemption.ts
pdyraga: There is another problem with the approach currently presented in `redemption.ts`: we completely ignore the value [of pending redemptions](https://github.com/keep-network/tbtc-v2/blob/f5f300b868137b6e8075eb69ca7e11f19143a912/solidity/contracts/bridge/Redemption.sol#L141). Say the wallet has 100 BTC, redemptions are processed every 3 hours, and so far, requests for 93 BTC has been registered in the Bridge. Based on the current version of the code, that wallet would still be selected for all redemptions <= 100 BTC and the transaction to Ethereum requesting redemption higher than 7 BTC would fail.

I am not sure why do we need to `findAllUnspentTransactionOutputs` and look for main UTXOs.

Why don't we get the main UTXO from the Bridge, take the value of all pending redemptions, subtract one from another, and use this value to determine if the given wallet is the right choice or not? This would be simpler and wouldn't have the two problems mentioned.

I think we can trust the Ethereum view of the world and not care about Bitcoin view of the world in this function.

lukasz-zimnoch: @pdyraga is right but:

> Why don't we get the main UTXO from the Bridge

This is not trivial as the main UTXO is stored as hash. To get it we have two options:
- Rely on contract events from transactions that change the main UXTO (so deposit sweeps, redemptions, moving funds, moved funds sweeps)
- Use similar logic than in https://github.com/keep-network/keep-core/blob/db145b629c3b4ec2677bc4457478439c685a2eff/pkg/tbtc/wallet.go#L392C6-L392C29 which needs to talk with Bitcoin but is simpler

Nevertheless, we must take the value of pending redemptions into account.

typescript/src/redemption.ts
r-czajkowski: 0f11838

typescript/src/redemption.ts
r-czajkowski: 0f11838

typescript/src/redemption.ts
```diff
@@ -406,3 +410,103 @@ export async function getRedemptionRequest(
...
+
+    if (!mainUtxo) {
+      console.debug(
+        `Could not find matching UTXO on chains for wallet public key hash \
+        (${wallet.walletPublicKeyHash.toPrefixedString()}). Continue the loop \
+        execution to the next wallet...`
```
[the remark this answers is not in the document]
r-czajkowski: 0f11838

typescript/test/data/redemption.ts
r-czajkowski: 24ede08

typescript/src/ethereum.ts
```diff
@@ -639,18 +642,20 @@ export class Bridge
...
-    const { ecdsaWalletID } = await backoffRetrier<{ ecdsaWalletID: string }>(
-      this._totalRetryAttempts
-    )(async () => {
-      return await this._instance.wallets(activeWalletPublicKeyHash)
-    })
+    const { ecdsaWalletID } = await this.wallets(
+      Hex.from(activeWalletPublicKeyHash)
+    )
 
+    return (await this.getWalletCompressedPublicKey(ecdsaWalletID)).toString()
```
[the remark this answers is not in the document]
r-czajkowski: d6e8627

typescript/test/redemption.test.ts
```diff
@@ -1432,6 +1436,140 @@ describe("Redemption", () => {
       })
     })
   })
+
+  describe("findWalletForRedemption", () => {
+    let bridge: MockBridge
+    let bitcoinClient: MockBitcoinClient
+
+    context(
+      "when there are no wallets in the network that can hanlde redemption",
```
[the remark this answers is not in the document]
r-czajkowski: dd1af45

typescript/src/redemption.ts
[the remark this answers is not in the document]
r-czajkowski: ca2191a

typescript/src/redemption.ts
r-czajkowski: 0f11838

typescript/src/redemption.ts
r-czajkowski: 9a0f289

lukasz-zimnoch (approved):

michalsmiarowski (approved):
LGTM 🔥

PR #631, opened 2023-06-07
Update dependencies to `random-beacon`, `ecdsa` and `tbtc-v2`

michalinacienciala:
We have published new `@keep-network/random-beacon`, `@keep-network/ecdsa` and `@keep-network/tbtc-v2` packages. Those packages no longer include the `prepare-dependencies.sh` script from `@threshold-network/solidity-contracts` which was causing random failures during `yarn install`. As in some CI jobs we install the dependencies based on the lockfile (with the `--frozen-lockfile` flag), we need to update dependencies in the lockfile so that the new, improved packages would be used.

Refs:
https://github.com/threshold-network/token-dashboard/pull/533
https://github.com/keep-network/keep-core/pull/3618
https://github.com/keep-network/coverage-pools/pull/233

### DIFF
--- a/solidity/yarn.lock
+++ b/solidity/yarn.lock
@@ -108,7 +108,7 @@
   resolved "https://registry.yarnpkg.com/@celo/utils/-/utils-0.1.11.tgz#c35e3b385091fc6f0c0c355b73270f4a8559ad38"
   integrity sha512-i3oK1guBxH89AEBaVA1d5CHnANehL36gPIcSpPBWiYZrKTGGVvbwNmVoaDwaKFXih0N22vXQAf2Rul8w5VzC3w==
   dependencies:
-    "@umpirsky/country-list" "git+https://github.com/umpirsky/country-list#05fda51"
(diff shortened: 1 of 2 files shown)


dimpar (approved):

dimpar (changes requested):
Seems a lot of changes in the typescript/yarn.lock. You sure they are all needed?

michalinacienciala:
> Seems a lot of changes in the typescript/yarn.lock. You sure they are all needed?

That looked odd to me as well when I was preparing the changes... But somehow I couldn't update the `@keep-network/tbtc-v2` dependency without updating all those packages. I'll try again...

michalinacienciala:
> > Seems a lot of changes in the typescript/yarn.lock. You sure they are all needed?
> 
> That looked odd to me as well when I was preparing the changes... But somehow I couldn't update the `@keep-network/tbtc-v2` dependency without updating all those packages. I'll try again...

We'll deliver changes to `typescript` dependencies in a separate PR.

dimpar (approved):

PR #632, opened 2023-06-09
Request redemption in one transaction

r-czajkowski:
This PR refactors the `requestRedemption` function to request redemption in one transaction(instead of 2: approve + requestRedemption) using the [`approveAndCall` ](https://github.com/thesis/solidity-contracts/blob/main/contracts/token/ERC20WithPermit.sol#L207) function from tBTC token contract. Then the tBTC token contract calls the [`receiveApproval`](https://github.com/keep-network/tbtc-v2/blob/main/solidity/contracts/vault/TBTCVault.sol#L193) function from the `TBTCVault` contract which [burns tBTC tokens and requests redemption](https://github.com/keep-network/tbtc-v2/blob/main/solidity/contracts/vault/TBTCVault.sol#L338).

### DIFF
--- a/typescript/src/bitcoin.ts
+++ b/typescript/src/bitcoin.ts
@@ -1,4 +1,4 @@
-import bcoin, { TX } from "bcoin"
+import bcoin, { TX, Script } from "bcoin"
 import wif from "wif"
 import bufio from "bufio"
 import hash160 from "bcrypto/lib/hash160"
@@ -602,4 +602,13 @@ export function isPublicKeyHashLength(publicKeyHash: string): boolean {
 export function locktimeToNumber(locktimeLE: Buffer | string): number {
   const locktimeBE: Buffer = Hex.from(locktimeLE).reverse().toBuffer()
   return BigNumber.from(locktimeBE).toNumber()
+}
+
+/**
+ * Creates the output script from the BTC address.
+ * @param address BTC address.
+ * @returns The un-prefixed and not prepended with length otput script.
+ */
+export function createOutputScriptFromAddress(address: string): string {
+  return Script.fromAddress(address).toRaw().toString("hex")
 }
--- a/typescript/src/bitcoin.ts
+++ b/typescript/src/bitcoin.ts
@@ -607,7 +607,7 @@ export function locktimeToNumber(locktimeLE: Buffer | string): number {
 /**
  * Creates the output script from the BTC address.
  * @param address BTC address.
- * @returns The un-prefixed and not prepended with length otput script.
+ * @returns The un-prefixed and not prepended with length output script.
  */
 export function createOutputScriptFromAddress(address: string): string {
   return Script.fromAddress(address).toRaw().toString("hex")
--- a/typescript/src/bitcoin.ts
+++ b/typescript/src/bitcoin.ts
@@ -609,6 +609,6 @@ export function locktimeToNumber(locktimeLE: Buffer | string): number {
  * @param address BTC address.
  * @returns The un-prefixed and not prepended with length output script.
  */
-export function createOutputScriptFromAddress(address: string): string {
-  return Script.fromAddress(address).toRaw().toString("hex")
+export function createOutputScriptFromAddress(address: string): Hex {
+  return Hex.from(Script.fromAddress(address).toRaw().toString("hex"))
 }
--- a/typescript/src/chain.ts
+++ b/typescript/src/chain.ts
@@ -244,26 +244,6 @@ export interface Bridge {
    * Returns the attached WalletRegistry instance.
    */
   walletRegistry(): Promise<WalletRegistry>
-
-  /**
-   * Builds the redemption data required to request a redemption via
-   * @see TBTCToken#approveAndCall - the built data should be passed as
-   * `extraData` the @see TBTCToken#approveAndCall function.
-   * @param redeemer On-chain identifier of the redeemer.
-   * @param walletPublicKey The Bitcoin public key of the wallet. Must be in the
-   *        compressed form (33 bytes long with 02 or 03 prefix).
-   * @param mainUtxo The main UTXO of the wallet. Must match the main UTXO
-   *        held by the on-chain Bridge contract.
-   * @param redeemerOutputScript The output script that the redeemed funds will
-   *        be locked to. Must be un-prefixed and not prepended with length.
-   * @returns The
-   */
-  buildRedemptionData(
-    redeemer: Identifier,
-    walletPublicKey: string,
-    mainUtxo: UnspentTransactionOutput,
-    redeemerOutputScript: string
-  ): Hex
 }
 
 /**
@@ -402,6 +382,40 @@ export interface TBTCVault {
 }
 
 /**
+ * Represnts data required to request redemption.
+ */
+export interface RequestRedemptionData {
+  /**
+   * On-chain identifier of the redeemer.
+   */
+  redeemer: Identifier
+  /**
+   * The Bitcoin public key of the wallet. Must be in the compressed form (33
+   * bytes long with 02 or 03 prefix).
+   */
+  walletPublicKey: string
+  /**
+   * The main UTXO of the wallet. Must match the main UTXO held by the on-chain
+   * Bridge contract.
+   */
+  mainUtxo: UnspentTransactionOutput
+  /**
+   * The output script that the redeemed funds will be locked to. Must be
+   * un-prefixed and not prepended with length.
+   */
+  redeemerOutputScript: string
+  /**
+   * The amount to be redeemed with the precision of the tBTC on-chain token
+   * contract.
+   */
+  amount: BigNumber
+  /**
+   * The vault address.
+   */
+  vault: Identifier
+}
+
+/**
  * Interface for communication with the TBTC v2 token on-chain contract.
  */
 export interface TBTCToken {
@@ -418,18 +432,13 @@ export interface TBTCToken {
   totalSupply(blockNumber?: number): Promise<BigNumber>
 
   /**
-   * Calls `receiveApproval` function on spender previously approving the spender
-   * to withdraw from the caller multiple times, up to the `amount` amount. If
-   * this function is called again, it overwrites the current allowance with
-   * `amount`.
-   * @param spender Address of contract authorized to spend.
-   * @param amount The max amount they can spend.
-   * @param extraData Extra information to send to the approved contract.
+   * Requests redemption in one transacion using the `approveAndCall` function
+   * from the tBTC on-chain token contract. Then the tBTC token contract calls
+   * the `receiveApproval` function from the `TBTCVault` contract which burns
+   * tBTC tokens and requests redemption.
+   * @param requestRedemptionData Data required to request redemption @see
+   *        {@link RequestRedemptionData}.
    * @returns Transaction hash of the approve and call transaction.
    */
-  approveAndCall(
-    spender: Identifier,
-    amount: BigNumber,
-    extraData: Hex
-  ): Promise<Hex>
+  requestRedemption(requestRedemptionData: RequestRedemptionData): Promise<Hex>
 }
--- a/typescript/src/chain.ts
+++ b/typescript/src/chain.ts
@@ -244,6 +244,26 @@ export interface Bridge {
    * Returns the attached WalletRegistry instance.
    */
   walletRegistry(): Promise<WalletRegistry>
+
+  /**
+   * Builds the redemption data required to request a redemption via
+   * @see TBTCToken#approveAndCall - the built data should be passed as
+   * `extraData` the @see TBTCToken#approveAndCall function.
+   * @param redeemer On-chain identifier of the redeemer.
+   * @param walletPublicKey The Bitcoin public key of the wallet. Must be in the
+   *        compressed form (33 bytes long with 02 or 03 prefix).
+   * @param mainUtxo The main UTXO of the wallet. Must match the main UTXO
+   *        held by the on-chain Bridge contract.
+   * @param redeemerOutputScript The output script that the redeemed funds will
+   *        be locked to. Must be un-prefixed and not prepended with length.
+   * @returns The
+   */
+  buildRedemptionData(
+    redeemer: Identifier,
+    walletPublicKey: string,
+    mainUtxo: UnspentTransactionOutput,
+    redeemerOutputScript: string
+  ): Hex
 }
 
 /**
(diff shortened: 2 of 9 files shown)


typescript/src/bitcoin.ts
```diff
@@ -603,3 +603,12 @@ export function locktimeToNumber(locktimeLE: Buffer | string): number {
   const locktimeBE: Buffer = Hex.from(locktimeLE).reverse().toBuffer()
   return BigNumber.from(locktimeBE).toNumber()
 }
+
+/**
+ * Creates the output script from the BTC address.
+ * @param address BTC address.
+ * @returns The un-prefixed and not prepended with length otput script.
```
pdyraga, line 610:
```suggestion
 * @returns The un-prefixed and not prepended with length output script.
```

typescript/src/bitcoin.ts
```diff
@@ -603,3 +603,12 @@ export function locktimeToNumber(locktimeLE: Buffer | string): number {
...
+ */
+export function createOutputScriptFromAddress(address: string): string {
+  return Script.fromAddress(address).toRaw().toString("hex")
+}
```
pdyraga, line 614: Can we cover this function in `bitcoin.test.ts`? It seems `bcoin` [will cover](https://github.com/bcoin-org/bcoin/blob/b0058696cc10c8f9b17190b31fd2cd907d85d047/lib/script/script.js#L1621-L1637) all types of addresses [allowed](https://github.com/keep-network/tbtc-v2/blob/2254cf67e2b3526789f8b34c2a7d66277fb42a3c/solidity/contracts/bridge/Bridge.sol#L444-L446) by the bridge for redemption, but we should use this function as a black box and have it covered with tests to ensure this keeps working as advertised no matter the implementation.

typescript/src/chain.ts
```diff
@@ -244,6 +244,26 @@ export interface Bridge {
...
+   * @param redeemer On-chain identifier of the redeemer.
+   * @param walletPublicKey The Bitcoin public key of the wallet. Must be in the
+   *        compressed form (33 bytes long with 02 or 03 prefix).
+   * @param mainUtxo The main UTXO of the wallet. Must match the main UTXO
+   *        held by the on-chain Bridge contract.
+   * @param redeemerOutputScript The output script that the redeemed funds will
+   *        be locked to. Must be un-prefixed and not prepended with length.
+   * @returns The
+   */
+  buildRedemptionData(
```
pdyraga, line 261: I am not sure if it should be exported into a public API. The goal of `src/ethereum.ts` is to translate domain-level objects into low-level Ethereum chain primitives and perform a call. With this function we allow those low-level primitives to leak outside.

Instead of having `buildRedemptionData`, I would modify `approveAndCall` interface and replace `extraData: Hex` with domain-level objects, as accepted by `buildRedemptionData`. We could also rename `approveAndCall` to `requestRedemption` and explain in the docs it happens via `approveAndCall`. I think this is fine - although `requestRedemption` does not exist on the Solidity code of `TBTC` token, having it in `TBTCToken` Typescript API is fine given we abstract the complexity and make the API human-readable.

typescript/src/ethereum.ts
```diff
@@ -488,6 +488,39 @@ export class Bridge
...
+   *        the compressed form (33 bytes long with 02 or 03 prefix).
+   * @param mainUtxo - The main UTXO of the wallet. Must match the main UTXO
+   *        held by the on-chain contract.
+   * @param redeemerOutputScript - The output script that the redeemed funds
+   *        will be locked to. Must be un-prefixed and not prepended with
+   *        length.
+   * @returns Parsed data that can be passed to the contract to request
+   * redemption.
+   */
+  private parseRequestRedemptionTransactionData = (
```
pdyraga, line 519: I am not sure *parsing* verb is correct in this context. I would expect parsing to happen when we receive on-chain data and we want them to be translated into domain-level objects.

If we make the change from the previous comment and keep `buildRedemptionData` private, then I would suggest to rename this function to 
`buildBridgeRequestRedemptionData` and `buildRedemptionData` to `buildTbtcTokenRequestRedemptionData`.

pdyraga:
Left some early comments. I'd prefer someone more familiar with the Typescript library to perform a final code review. Let's please also make sure system tests are passing on this version of the code.

typescript/src/bitcoin.ts
```diff
@@ -603,3 +603,12 @@ export function locktimeToNumber(locktimeLE: Buffer | string): number {
   const locktimeBE: Buffer = Hex.from(locktimeLE).reverse().toBuffer()
   return BigNumber.from(locktimeBE).toNumber()
 }
+
+/**
+ * Creates the output script from the BTC address.
+ * @param address BTC address.
+ * @returns The un-prefixed and not prepended with length otput script.
+ */
+export function createOutputScriptFromAddress(address: string): string {
```
lukasz-zimnoch, line 612: Wouldn't it be better to return a `Hex`?

typescript/src/chain.ts
lukasz-zimnoch: Totally agree with @pdyraga!

typescript/src/bitcoin.ts
r-czajkowski: 3a9d842

typescript/src/bitcoin.ts
r-czajkowski: 07693eb

typescript/src/bitcoin.ts
r-czajkowski: 759cc78

typescript/src/chain.ts
r-czajkowski: 9249ee5

typescript/src/ethereum.ts
r-czajkowski: 9249ee5. I renamed the `buildRedemptionData` -> `buildRequestRedemptionData` because it lives inside the `TBTCToken` handle so IMO there is no need to add `TbtcToken`. WDYT?

typescript/test/utils/mock-bridge.ts
```diff
@@ -314,4 +326,40 @@ export class MockBridge implements Bridge {
...
+      redeemerOutputScript,
+    })
+
+    // Convert the output script to raw bytes buffer.
+    const rawRedeemerOutputScript = Buffer.from(redeemerOutputScript, "hex")
+    // Prefix the output script bytes buffer with 0x and its own length.
+    const prefixedRawRedeemerOutputScript = `0x${Buffer.concat([
+      Buffer.from([rawRedeemerOutputScript.length]),
+      rawRedeemerOutputScript,
+    ]).toString("hex")}`
```
r-czajkowski, line 349: Looks like we could create a reusable fn for this because we use it in unit tests and in Ethereum Bridge implementation. But I'm not sure which place is better- `ethereum-helpers.ts` or `bitcoin.ts? cc @lukasz-zimnoch @nkuba what do you think guys?

typescript/src/chain.ts
```diff
@@ -381,6 +381,40 @@ export interface TBTCVault {
...
+  redeemerOutputScript: string
+  /**
+   * The amount to be redeemed with the precision of the tBTC on-chain token
+   * contract.
+   */
+  amount: BigNumber
+  /**
+   * The vault address.
+   */
+  vault: Identifier
```
pdyraga, line 415: Why do we need the vault as a parameter? `requestRedemption` is exposed on `TBTCToken` and there is no other choice than working with `TBTCVault` contract for this interaction.

EDIT: see the comment in `redemption.ts`.

typescript/src/chain.ts
```diff
@@ -396,4 +430,15 @@ export interface TBTCToken {
...
+  /**
+   * Requests redemption in one transacion using the `approveAndCall` function
+   * from the tBTC on-chain token contract. Then the tBTC token contract calls
+   * the `receiveApproval` function from the `TBTCVault` contract which burns
+   * tBTC tokens and requests redemption.
+   * @param requestRedemptionData Data required to request redemption @see
+   *        {@link RequestRedemptionData}.
+   * @returns Transaction hash of the approve and call transaction.
+   */
+  requestRedemption(requestRedemptionData: RequestRedemptionData): Promise<Hex>
```
pdyraga, line 443: I would consider unwrapping parameters from the struct and passing them directly, just like we do here:
https://github.com/keep-network/tbtc-v2/blob/9249ee5ca9c71cf6754c68c5d3d891e19d77d567/typescript/src/chain.ts#L169-L174
 `RequestRedemptionData` is used only for `TBTCToken.requestRedemption` but we also have `Bridge.requestRedemption` in the same file and it is unclear why this structure is not used for both. Also, having four parameters in the function (I assume we will eliminate vault from the params) is not that bad.

typescript/src/ethereum.ts
```diff
@@ -1099,4 +1100,80 @@ export class TBTCToken
...
+  async requestRedemption(
+    requestRedemptionData: RequestRedemptionData
+  ): Promise<Hex> {
+    const { vault, amount, ...restData } = requestRedemptionData
+
+    const extraData = this.buildRequestRedemptionData(restData)
+
+    const tx = await sendWithRetry<ContractTransaction>(async () => {
+      return await this._instance.approveAndCall(
+        vault.identifierHex,
```
pdyraga, line 1117: I think we should resolve the vault address instead of requiring it to be passed in parameters.

EDIT: see the comment in `redemption.ts`.

typescript/src/redemption.ts
```diff
@@ -55,29 +56,36 @@ export interface RedemptionRequest {
...
+ *        by the on-chain Bridge contract.
  * @param redeemerOutputScript - The output script that the redeemed funds will
  *        be locked to. Must be un-prefixed and not prepended with length.
- * @param amount - The amount to be redeemed in satoshis.
- * @param bridge - Handle to the Bridge on-chain contract.
- * @returns Empty promise.
+ * @param amount - The amount to be redeemed with the precision of the tBTC
+ *        on-chain token contract.
+ * @param vault - The vault address.
+ * @param tBTCToken - Handle to the TBTCToken on-chain contract.
```
pdyraga, line 69: I don't think we should pass more than one contract as a parameter. What if someone passes `TBTCToken` contract address but `DonationVault` instead of the `TBTCVault`? This doesn't make sense, but the API should limit the surface of developer errors.

Since the redemption happens on the token contract, I think we should pass `TBTCToken` contract as a parameter and resolve the vault [from the `owner()` call](https://github.com/thesis/solidity-contracts/blob/c315b9d5f0c39253428898474e97552f5e5b9046/contracts/token/ERC20WithPermit.sol#L18).

typescript/src/redemption.ts
```diff
@@ -55,29 +56,36 @@ export interface RedemptionRequest {
 
 /**
  * Requests a redemption from the on-chain Bridge contract.
```
pdyraga, line 58: SInce we call `tBTCToken.requestRedemption` instead of `bridge.requestRedemption` now, I'd change this sentence to something like "Requests a redemption of tBTC into BTC".

typescript/src/ethereum.ts
pdyraga: Yeah, it's in a separate class, this makes sense!

typescript/src/chain.ts
pdyraga: It makes sense! 💯

typescript/src/redemption.ts
r-czajkowski: 06601b9

typescript/src/chain.ts
r-czajkowski: 06601b9

typescript/src/ethereum.ts
r-czajkowski: 06601b9

typescript/src/chain.ts
r-czajkowski: 06601b9

typescript/test/bitcoin.test.ts
```diff
@@ -464,4 +465,122 @@ describe("Bitcoin", () => {
...
+        ]) => {
+          it(`should create correct output script for ${addressType} address type`, () => {
+            const result = createOutputScriptFromAddress(address)
+
+            expect(result.toString()).to.eq(expectedOutputScript)
+            // Check if we can build the prefixed raw redeemer output script based
+            // on the result.
+            expect(buildRawPrefixedOutputScript(result.toString())).to.eq(
+              expectedRedeemerOutputScript
+            )
```
pdyraga, line 517: Isn't it guaranteed if the previous assertion passes? 

`buildRawPrefixedOutputScript` is only converting the string to a buffer and prepends the length. If the `result` is `expectedOutputScript` this will always work. I would consider removing this assertion, the most important part of this test is whether `createOutputScriptFromAddress` recognizes and handles the given address type correctly.

typescript/test/utils/mock-bridge.ts
```diff
@@ -314,4 +326,40 @@ export class MockBridge implements Bridge {
   walletRegistry(): Promise<WalletRegistry> {
     throw new Error("not implemented")
   }
+
+  buildRedemptionData(
```
pdyraga, line 330: Where is it used? 🤔

pdyraga:
Left the last comments I have. Once they are addressed, and I see `system-tests-deposit-redemption` green, I am good to merge this PR.

typescript/test/utils/mock-bridge.ts
r-czajkowski: It's a left over- no longer needed because we've refactored the `reuqestRedemption` fn in 9249ee5 and 06601b9. Removed in e814ecaeeb083f4d66a2882007de4b0b7320e4b4

typescript/src/ethereum.ts
```diff
@@ -1099,4 +1099,99 @@ export class TBTCToken
...
+   */
+  async requestRedemption(
+    walletPublicKey: string,
+    mainUtxo: UnspentTransactionOutput,
+    redeemerOutputScript: string,
+    amount: BigNumber
+  ): Promise<Hex> {
+    const redeemer = await this._instance?.signer?.getAddress()
+    if (!redeemer) {
+      throw new Error("Signer not provided.")
```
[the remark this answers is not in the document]
r-czajkowski: dbd818e

typescript/test/bitcoin.test.ts
r-czajkowski: b9633d1

typescript/test/utils/mock-bridge.ts
pdyraga: What about `buildRedemptionDataLog` and `BuildRedemptionDataLogEntry`?

r-czajkowski: > What about `buildRedemptionDataLog` and `BuildRedemptionDataLogEntry`?

🦅 👁️ thanks for catching this! See 5f0365d.

pdyraga (approved):

PR #633, opened 2023-06-12
Redemptions prototype: On-chain coordination

lukasz-zimnoch:
Refs https://github.com/keep-network/keep-core/issues/3609

Here we add the on-chain coordination for redemptions. This changeset enhances the `WalletCoordinator` contract with three methods:
- `submitRedemptionProposal` that can be used to submit redemption proposals for wallets
- `submitRedemptionProposalWithReimbursement` which is the reimbursable version of `submitRedemptionProposal`
- `validateRedemptionProposal` which can be used by off-chain wallet operators to validate proposals using static calls

Moreover, during the work on `validateRedemptionProposal` we noticed the need to harden the validation rules of `validateDepositSweepProposal` function. We added a rule which enforces that each deposit in the proposal is used exactly once.

### DIFF
--- a/solidity/contracts/bridge/WalletCoordinator.sol
+++ b/solidity/contracts/bridge/WalletCoordinator.sol
@@ -219,6 +219,39 @@ contract WalletCoordinator is OwnableUpgradeable, Reimbursable {
     ///         proposal (of any type) can be submitted after 4 pm.
     uint32 public redemptionProposalValidity;
 
+    /// @notice The minimum time that must elapse since the redemption request
+    ///         creation before a request becomes eligible for a processing.
+    ///
+    ///         For example, if a request was created at 9 am and
+    ///         redemptionRequestMinAge is 2 hours, the request is eligible for
+    ///         processing after 11 am.
+    ///
+    /// @dev Forcing request minimum age ensures block finality for Ethereum.
+    uint32 public redemptionRequestMinAge;
+
+    /// @notice Each redemption request can be technically handled until it
+    ///         reaches its timeout timestamp after which it can be reported
+    ///         as timed out. However, allowing the wallet to handle requests
+    ///         that are close to their timeout timestamp may cause a race
+    ///         between the wallet and the redeemer. In result, the wallet may
+    ///         redeem the requested funds even though the redeemer already
+    ///         received back their tBTC (locked during redemption request) upon
+    ///         reporting the request timeout. In effect, the redeemer may end
+    ///         out with both tBTC and redeemed BTC in their hands which has
+    ///         a negative impact on the tBTC <-> BTC peg. In order to mitigate
+    ///         that problem, this parameter determines a safety margin that
+    ///         puts the latest moment a request can be handled far before the
+    ///         point after which the request can be reported as timed out.
+    ///
+    ///         For example, if a request times out after 8 pm and
+    ///         redemptionRequestTimeoutSafetyMargin is 2 hours, the request is
+    ///         valid for processing only before 6 pm.
+    uint32 public redemptionRequestTimeoutSafetyMargin;
+
+    /// @notice The maximum count of redemption requests that can be processed
+    ///         within a single redemption.
+    uint16 public redemptionMaxSize;
+
     /// @notice Gas that is meant to balance the redemption proposal
     ///         submission overall cost. Can be updated by the owner based on
     ///         the current conditions.
@@ -256,6 +289,9 @@ contract WalletCoordinator is OwnableUpgradeable, Reimbursable {
 
     event RedemptionProposalParametersUpdated(
         uint32 redemptionProposalValidity,
+        uint32 redemptionRequestMinAge,
+        uint32 redemptionRequestTimeoutSafetyMargin,
+        uint16 redemptionMaxSize,
         uint32 redemptionProposalSubmissionGasOffset
     );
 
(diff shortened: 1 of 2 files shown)


solidity/test/bridge/WalletCoordinator.test.ts
```diff
@@ -1956,83 +1956,1191 @@ describe("WalletCoordinator", () => {
...
+              ],
+              redemptionTxFee: 6000,
+            })
+        })
+
+        after(async () => {
+          await restoreSnapshot()
+        })
+
+        it("should time lock the wallet", async () => {
```
pdyraga, line 2326:
```suggestion
        it("should time-lock the wallet", async () => {
```

solidity/test/bridge/WalletCoordinator.test.ts
```diff
@@ -1956,83 +1956,1191 @@ describe("WalletCoordinator", () => {
...
+                          .redeemerOutputScript,
+                      ],
+                      // Exceed the max per-request fee by one.
+                      redemptionTxFee: bridgeRedemptionTxMaxTotalFee + 1,
+                    })
+                  ).to.be.revertedWith("Proposed transaction fee is too high")
+                })
+              }
+            )
+          })
```
pdyraga, line 2647: What about the per-redemption fee test?

solidity/test/bridge/WalletCoordinator.test.ts
```diff
@@ -1956,83 +1956,1191 @@ describe("WalletCoordinator", () => {
...
+            context("when there is a non-pending request", () => {
+              let requestOne
+              let requestTwo
+
+              before(async () => {
+                await createSnapshot()
+
+                requestOne = createTestRedemptionRequest(
+                  walletPubKeyHash,
+                  5000 // necessary to pass the fee share validation
```
pdyraga, line 2661: The default value is not enough? Shouldn't we change it?

solidity/test/bridge/WalletCoordinator.test.ts
pdyraga: OK, I see it's covered in "when there is a request that incurs an unacceptable tx fee share" test later. Maybe we should add a comment to the end of this block?

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -195,6 +209,55 @@ contract WalletCoordinator is OwnableUpgradeable, Reimbursable {
...
+
+    /// @notice The minimum time that must elapse since the redemption request
+    ///         creation before a request becomes eligible for a processing.
+    ///
+    ///         For example, if a request was created at 9 am and
+    ///         redemptionRequestMinAge is 2 hours, the request is eligible for
+    ///         processing after 11 am.
+    ///
+    /// @dev Forcing request minimum age ensures block finality for Ethereum.
+    uint32 public redemptionRequestMinAge;
```
nkuba, line 231: Can we make it block number related instead of a timestamp? We usually used block numbers to confirm the emitted event.

nkuba: Oh I see, we don't track on-chain the block number for the request but a timestamp.

lukasz-zimnoch: Yeah, the Bridge itself is mostly timestamp-based.

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -259,6 +335,12 @@ contract WalletCoordinator is OwnableUpgradeable, Reimbursable {
         depositRefundSafetyMargin = 24 hours;
         depositSweepMaxSize = 5;
         depositSweepProposalSubmissionGasOffset = 20_000; // optimized for 10 inputs
+
+        redemptionProposalValidity = 1 hours;
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, may be too short. Setting it to 2 hours: e0bcc8a3

solidity/test/bridge/WalletCoordinator.test.ts
lukasz-zimnoch: The default value of `txMaxFee` set by `createTestRedemptionRequest` is `0` so it's not enough.

solidity/test/bridge/WalletCoordinator.test.ts
lukasz-zimnoch: See d668d5ca

solidity/test/bridge/WalletCoordinator.test.ts
lukasz-zimnoch: See 41476d20

solidity/test/bridge/WalletCoordinator.test.ts
```diff
@@ -1956,83 +1956,1191 @@ describe("WalletCoordinator", () => {
...
+                                          .returns(depositThree.request)
+                                      })
+
+                                      after(async () => {
+                                        bridge.deposits.reset()
+
+                                        await restoreSnapshot()
+                                      })
+
+                                      it("should succeed", async () => {
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 41476d20

solidity/contracts/bridge/WalletCoordinator.sol
```diff
@@ -687,4 +786,213 @@ contract WalletCoordinator is OwnableUpgradeable, Reimbursable {
 
         revert("Extra info funding output script does not match");
     }
+
+    /// @notice Updates parameters related to redemption proposal.
+    /// @param _redemptionProposalValidity The new value of `redemptionProposalValidity`.
+    /// @param _redemptionProposalSubmissionGasOffset The new value of `redemptionProposalSubmissionGasOffset`.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 0bb0fc10

pdyraga (approved):

PR #635, opened 2023-06-14
RFC-11: Solana tBTC minting experience

pdyraga:
There is a clear massive demand funnel present in the Solana ecosystem for tBTC. This RFC aims at providing a bespoke user experience for minting tBTC on Solana.

Even today it is possible to bridge tBTC to Solana with Wormhole. The whole user experience is far from perfect, though.
First, the user has to deposit their BTC using Threshold Dashboard on Ethereum. This requires executing one Ethereum transaction. Next, once tBTC is minted on Ethereum, the user has to go to the Wormhole Token Bridge portal and bridge their Ethereum tBTC to Solana. This requires several transactions - one transaction on Ethereum and several transactions on Solana.

Users should mint tBTC on Solana in the Threshold Dashboard with two transactions: one transaction on BTC and one transaction on Solana.

The redemption process should require just one Solana transaction from the user executed in the Threshold Dashboard.


dimpar:
Are there any changes needed on the Wormhole contracts side? My understanding is that there are no changes on the WH side needed, but I'd like to make sure.

dimpar:
What happens to the Portal TBTC functionality Ethereum<->Solana if we implement this? Should it be the alternative way of bridging or should the WH team block it in their Portal?

docs/rfc/rfc-10.adoc
```diff
@@ -0,0 +1,248 @@
...
+to use two applications and execute transactions on three networks: Bitcoin,
+Ethereum, and Solana between them. Moreover, there is a period of time between
+the time BTC was deposited and the time tBTC is minted on Ethereum that requires
+the user to wait idly before they can bridge their Ethereum tBTC to Solana.
+
+== Proposal
+
+=== Goal
+
+Users should mint tBTC on Solana in the Threshold Dashboard with two
```
dimpar, line 35: > "mint tBTC on Solana"
Correct me if I'm wrong, but this means that there will be deployed a new canonical tBTC on Solana, right? Which also inherits from L2TBTC.sol? However, if the canonical tBTC on Solana is going to be a `wormholeTBTC`, then do we still need to mint it?

dimpar:
What would be the experience bridging from L2/sidechain <-> Solana? Should this be covered with the existing WH Portal?

docs/rfc/rfc-10.adoc
pdyraga: Negative. We may deploy a canonical tBTC and we may not deploy it and consider wormhole tBTC as the canonical one. This sentence is just an abstract statement - we want users to mint their tBTC on Solana with the same experience as they have on Ethereum. The existence or non-existence of a gateway is just a technical detail. In the minimum version, the gateway is not required.

pdyraga:
> Are there any changes needed on the Wormhole contracts side? My understanding is that there are no changes on the WH side needed, but I'd like to make sure.

Correct, no changes are required.

> What happens to the Portal TBTC functionality Ethereum<->Solana if we implement this? Should it be the alternative way of bridging or should the WH team block it in their Portal?

It can stay as-is. This is not breaking the bridging flow in other dApps. 

> What would be the experience bridging from L2/sidechain <-> Solana? Should this be covered with the existing WH Portal?

This RFC is only for BTC -> Solana but we can extend this pattern to any other chain. We could use it for BTC -> L2s as well, for example.

erdogan (approved):

docs/rfc/rfc-10.adoc
```diff
@@ -0,0 +1,248 @@
...
+
+==== Solana redeemer
+
+`SolanaRedeemer` contract will act as tBTC redeemer on L1 Ethereum and should be
+placed in `tbtc-v2/solidity` project. The contract will receive tBTC from Solana
+via Wormhole Token Bridge and request redemption in the tBTC `Bridge` contract. 
+
+The contract needs to be upgradeable.
+
+==== Relayer bot
```
nkuba, line 166: Can we consider adding some guardians bots that will protect against the relayer bot going south?

docs/rfc/rfc-10.adoc
```diff
@@ -0,0 +1,248 @@
...
+
+To optimize the user experience, a relayer bot needs to be implemented. The
+relayer's responsibility will be:
+
+- Deliver Wormhole VAA to `SolanaDepositor` contract once the deposit was
+  revealed on Solana.
+- Request bridging tBTC from Ethereum to Solana once the tBTC for the deposit
+  has been optimistically minted or the deposit was swept.
```
nkuba, line 174: Can we be more unambiguous?

```suggestion
  has been optimistically minted or the deposit was swept on Ethereum.
```

docs/rfc/rfc-10.adoc
```diff
@@ -0,0 +1,248 @@
...
+- Redeem bridged tBTC on Solana to the depositor address once the tBTC minted on
+  L1 has been successfully bridged to Solana.
```
nkuba, line 176: I don't follow this point, can you check it's valid?

Shouldn't it be:
```suggestion
- Transfer bridged tBTC on Solana to the depositor address once the tBTC minted on
  L1 has been successfully bridged to Solana.
```

nkuba, line 176: Let's be consistent and use `Ethereum` or `L1` or `Ethereum L1` across the document.

One of the examples:
```suggestion
- Redeem bridged tBTC on Solana to the depositor address once the tBTC minted on
  Ethereum has been successfully bridged to Solana.
```

docs/rfc/rfc-10.adoc
```diff
@@ -0,0 +1,248 @@
...
+   and associates it with Solana depositor address.
+7. `SolanaDepositor` contract reveals the deposit to the `Bridge`
+8. tBTC is optimistically minted to `SolanaDepositor` contract address.
+9. Relayer bot notifies `SolanaDepositor` contract the deposit was
+   optimistically minted. `SolanaDepositor` inspects `TBTCVault.optimisticMintingRequests`
+   to confirm the amount and requests bridging tBTC from its own balance to the
+   Solana depositor address associated with the given deposit using Wormhole
+   Token Bridge. The same happens when the deposit was not optimistically minted
+   but swept.
+10. Relayer bot redeems the tBTC bridged to Solana to the depositor's Solana
```
nkuba, line 221: "redeems" is confusing to me. I link the redemption with a process of exiting tBTC to BTC. Is this a different kind of redemption?

```suggestion
10. Relayer bot transfers the tBTC bridged to Solana to the depositor's Solana
```

docs/rfc/rfc-10.adoc
```diff
@@ -0,0 +1,248 @@
...
+```
+/// @notice Represents tBTC deposit request data.
+struct DepositRequest {
+    // (...)
+
+    // Optional extra data used by depositors who are smart contracts. The smart
+    // contract depositor can use the extra data to represent additional
+    // information, such as the depositor's address on a foreign (non-Ethereum)
+    // chain.
+    bytes32 depositorExtraData;
```
nkuba, line 119: Will `bytes32` be enough? To be more generic don't we need to include a target chain ID and allow providing a public key (could be compressed format 33-bytes)?

docs/rfc/rfc-10.adoc
pdyraga: What particular problematic scenario do you have in mind? The relayer should not have any extra rights, it is calling the same functions any anonymous user would call. The only role of the relayer is to call those functions on behalf of the user to improve the UX.

docs/rfc/rfc-10.adoc
pdyraga: Good point, done in a4180c2bfb2bec5db75d437e16d8f59a4e190708.

docs/rfc/rfc-10.adoc
pdyraga: To be precise, we should say the deposit was swept *on Bitcoin* but the proof was submitted to Ethereum. Please take a look at a4180c2bfb2bec5db75d437e16d8f59a4e190708.

docs/rfc/rfc-10.adoc
pdyraga: I think the original version is more precise. This RFC is pretty heavily oriented around Wormhole bridge and "redeem" is what is happening in this step in Wormhole, see https://www.portalbridge.com/#/redeem.

tl;dr; is that when you bridge assets, you *deposit* them in the Wormhole Token Bridge contract on the particular source chain. Then, Wormhole guardians produce VAA and you need to take that VAA and *redeem* the wrapped token on the target chain using it.

docs/rfc/rfc-10.adoc
pdyraga: Yes, it's Wormhole redemption as described [here](https://github.com/keep-network/tbtc-v2/pull/635#discussion_r1387711972). Please take a look at b1d234e98ffcc82c3a8e73257d3d8312eff70c6a. Hopefully, it clarifies it a bit.

docs/rfc/rfc-10.adoc
pdyraga: I tried to disambiguate them in b1d234e98ffcc82c3a8e73257d3d8312eff70c6a.

pdyraga:
> Has the team considered that if we promote wormhole-bridged tBTC as the canonical asset on solidity, then the security assumptions of tBTC on solana rely on wormhole's security rather than the Threshold Network itself?
> 
> This gets worse over time, as moving away from whtBTC to a possible future tBTC gets more and more difficult. You can see this happening with USDC on arbitrum for example.
> 
> This gets EVEN worse if we incentivise this bridged tBTC.

Hey @jommi9! The whtBTC (Wormhole-wrapped token) is not what the user receives on Solana. On [Solana](https://github.com/keep-network/tbtc-v2/releases/tag/cross-chain%2Fsolana%2Fv1.0.0), Arbitrum, Optimism, Base, and Polygon we have our canonical tBTC token contracts deployed. The canonical token is minted by a gateway contract that - currently - accepts Wormhole-wrapped tokens. This mechanism is explained in [RFC 8: Cross-chain Tokenized Threshold BTC](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-8.adoc#rfc-8-cross-chain-tokenized-threshold-btc).

It is true the existing mechanism partially relies on Wormhole security assumptions and this RFC only aims at providing a better user experience. It is certainly possible that in the future we will add other bridges or implement Solana minting without bridging. The canonical token offers the flexibility of not requiring protocols to migrate tokens when this happens. In other words, the tBTC canonical token address on Solana is never going to change.

lukasz-zimnoch (approved):

PR #636, opened 2023-06-15
Update dependencies to `keep-ecdsa` and `tbtc-v2` packages

michalinacienciala:
We have published new `@keep-network/ecdsa` and `@keep-network/tbtc-v2` packages. Those packages no longer include the `prepare-dependencies.sh` script from `@threshold-network/solidity-contracts` which was causing random failures during `yarn install`. As in some CI jobs we install the dependencies based on the lockfile (with the `--frozen-lockfile` flag), we need to update dependencies in the lockfile so that the new, improved packages would be used.

Ref:
https://github.com/threshold-network/solidity-contracts/pull/143
https://github.com/keep-network/tbtc-v2/pull/631

### DIFF
--- a/typescript/yarn.lock
+++ b/typescript/yarn.lock
@@ -406,21 +406,6 @@
     "@ethersproject/properties" "^5.6.0"
     "@ethersproject/strings" "^5.6.0"
 
-"@ethersproject/abi@5.7.0", "@ethersproject/abi@^5.7.0":
-  version "5.7.0"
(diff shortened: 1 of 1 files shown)


michalinacienciala:
Most of the diff is a result of getting rid of the `@tenderly/hardhat-tenderly` dependency in the newer version of `@keep-network/tbtc-v2`.

dimpar (approved):

Issue #641, opened 2023-07-04
WalletCoordinator contract upgrade scenario


nkuba:
Team | Address
------------ | -------------
**Deployment Team** | `0x123694886DBf5Ac94DDA07135349534536D14cAf`
**Threshold Council** | `0x9f6e831c8f8939dc0c830c6e492e7cef4f9c2f5f`
**Dev Team Coordinator** | `0x0595acCca29654c43Bd67E18578b30a405265234`

# Prerequisites 
_Deployment Team_:
- [x] Implement deployment script for `WalletCoordinator` contract upgrade (https://github.com/keep-network/tbtc-v2/pull/643).

# Tag the new version
_Deployment Team_:
- [x] Create a release branch (`releases/mainnet/solidity/v1.5.0`) and pin mainnet versions in `package.json`.
- [x] Tag `solidity/v1.5.0` in `keep-network/tbtc-v2`.
- [x] Sign `solidity/v1.5.0` commit hash.
- [x] Commit deployment artifacts to `main` branch (https://github.com/keep-network/tbtc-v2/pull/644).

# Smart contracts deployment
_Deployment Team_:
- [x] Deploy the latest `WalletCoordinator` contract implementation (`0x10Fb5943E2F4F67Ee6a533DaE49B6d4cC443ffE5`).

_Threshold Council_:
- [x] Upgrade implementation address for `WalletCoordinator` proxy via `ProxyAdmin`.

# Etherscan verification
_Deployment Team_:
- [x] Verify the `WalletCoordinator` contract

# Package publication
_Deployment Team_:
- [x] Publish `v1.5.0` package of `@keep-network/tbtc-v2` to NPM.

# Configure parameters
_Threshold Council_:
- [x] Configure initial values for redemptions with `updateRedemptionProposalParameters` method and arguments:
  ```solidity
  _redemptionProposalValidity = 7200
  _redemptionRequestMinAge = 600
  _redemptionRequestTimeoutSafetyMargin = 7200
  _redemptionMaxSize = 20
  _redemptionProposalSubmissionGasOffset = 20000
  ```

# Contracts verification
_Deployment Team_:
- [x] Confirm **Dev Team Coordinator** is registered as a coordinator in the contract.

# Release notes
_Deployment Team_:
- [x] Publish `solidity/v1.5.0` release notes in `keep-network/tbtc-v2` (https://github.com/keep-network/tbtc-v2/releases/tag/solidity%2Fv1.5.0).

# Documentation
[...]

PR #643, opened 2023-07-05
Deployment script for WalletCoordinator upgrade

nkuba:
We add a script for WalletCoordinator contract upgrade.

The current WalletCoordinator contract deployed to mainnet is owned by Threshold Council, so it cannot be upgraded right away by the deployer.

The deployer will deploy a fresh WalletCoordinator contract implementation, then Threshold Council needs to execute transactions to upgrade the WalletCoordinator proxy to the new implementation and set the initial values for newly added parameters.

Refs: https://github.com/keep-network/tbtc-v2/issues/641

### DIFF
--- a/solidity/deploy/81_upgrade_wallet_coordinator_v2.ts
+++ b/solidity/deploy/81_upgrade_wallet_coordinator_v2.ts
@@ -1,0 +1,98 @@
+import { Artifact, HardhatRuntimeEnvironment } from "hardhat/types"
+import { DeployFunction, Deployment } from "hardhat-deploy/types"
+import { ContractFactory } from "ethers"
+import { ProxyAdmin, WalletCoordinator } from "../typechain"
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+  const { ethers, helpers, deployments } = hre
+
+  const { deployer } = await helpers.signers.getNamedSigners()
+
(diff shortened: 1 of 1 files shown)


solidity/deploy/81_upgrade_wallet_coordinator_v2.ts
```diff
@@ -0,0 +1,98 @@
...
+      "updateRedemptionProposalParameters",
+      [7200, 600, 7200, 20, 20000]
+    )
+
+  deployments.log(
+    `WalletCoordinator owner ${walletCoordinatorOwner} is required to update redemption proposal parameters with transaction:\n` +
+      `\t\tfrom: ${walletCoordinatorOwner}\n` +
+      `\t\tto: ${walletCoordinator.address}\n` +
+      `\t\tdata: ${updateRedemptionProposalParametersTxData}`
+  )
```
pdyraga, line 63: It is nice to be able to double-check the data in the simulation but worth noting, this part is not strictly required.

pdyraga (approved):

solidity/deploy/81_upgrade_wallet_coordinator_v2.ts
nkuba: Definitely, we will simulate with Tenderly first. But we need the new contract to be deployed for that.

PR #646, opened 2023-07-06
Add `getAddressFromScriptPubKey` function

r-czajkowski:
~Depends on: #632~

Add the new function that returns the Bitcoin address based on the script pub key placed on the output of a Bitcoin transaction.


typescript/src/bitcoin.ts
```diff
@@ -624,3 +624,20 @@ export function locktimeToNumber(locktimeLE: Buffer | string): number {
...
+
+/**
+ * Returns the Bitcoin address based on the script pub key placed on the output
+ * of a Bitcoin transaction.
+ * @param scriptPubKey Scirpt pub key placed on the output of a Bitcoin
+ *        transaction.
+ * @param network Bitcoin network.
+ * @returns The Bitcoin address.
+ */
+export function getAddressFromScriptPubKey(
```
lukasz-zimnoch, line 636: The `scriptPubKey` is a field specific to transaction outputs but basically, this function converts any script to a Bitcoin address. We can make this function more generic and explicitly the opposite of `createOutputScriptFromAddress`. That said, I propose naming it `createAddressFromOutputScript`.

typescript/src/bitcoin.ts
```diff
@@ -624,3 +624,20 @@ export function locktimeToNumber(locktimeLE: Buffer | string): number {
...
+  scriptPubKey: string,
```
lukasz-zimnoch, line 637:
```suggestion
  script: Hex,
```

typescript/src/bitcoin.ts
```diff
@@ -624,3 +624,20 @@ export function locktimeToNumber(locktimeLE: Buffer | string): number {
 export function createOutputScriptFromAddress(address: string): Hex {
   return Hex.from(Script.fromAddress(address).toRaw().toString("hex"))
 }
+
+/**
+ * Returns the Bitcoin address based on the script pub key placed on the output
+ * of a Bitcoin transaction.
+ * @param scriptPubKey Scirpt pub key placed on the output of a Bitcoin
+ *        transaction.
```
lukasz-zimnoch, line 632:
```suggestion
 * @param script The unprefixed and not prepended with length output script.
```

typescript/src/bitcoin.ts
r-czajkowski: c6f94f6

typescript/src/bitcoin.ts
r-czajkowski: 1c9b91b

typescript/src/bitcoin.ts
r-czajkowski: 1c9b91b

lukasz-zimnoch (approved):

PR #647, opened 2023-07-07
Update system tests

r-czajkowski:
Request the redemption via the `depositorBridgeHandle.requestRedemption`
directly instead of  `requestRedemption` function. After refactoring in
https://github.com/keep-network/tbtc-v2/pull/632, the `requestRedemption` requests redemption via tBTC token
contract. In the current deposit scenario, we do not pass the vault
address so the `depositor` does not actually have any tBTC tokens so we
can't request redemption via `requestRedemption` fn with changes from
https://github.com/keep-network/tbtc-v2/pull/632. We are going to add a new scenario where we pass the TBTC vault
address to the `revealDeposit` function to test the new mechanism of
requesting the redemption via tBTC token contract in follow-up work.


pdyraga:
Looks like we have a system test failure: https://github.com/keep-network/tbtc-v2/actions/runs/5484145856

lukasz-zimnoch (approved):
LGTM! Will merge once https://github.com/keep-network/tbtc-v2/actions/runs/5529864532 turns green!

PR #651, opened 2023-07-11
Improve find wallet for redemption logic

lukasz-zimnoch:
Refs: https://github.com/threshold-network/token-dashboard/issues/553

Here we improve the logic of `findWalletForRedemption` function to handle the corner case when a wallet's main UTXO known to the `Bridge` was actually spent on Bitcoin but that change was not yet reflected to the `Bridge`. So far, `findWalletForRedemption` relied on `findAllUnspentTransactionOutputs` which does not return spent UTXOs so determining the actual main UTXO was not possible in the aforementioned corner case. Now, the `findWalletForRedemption` looks for recent wallet transaction history on Bitcoin and tries to match the actual main UTXO based on that. This change will allow submitting new redemption requests without downtime, even if the target wallet is performing an action at the moment.

### DIFF
--- a/typescript/src/electrum.ts
+++ b/typescript/src/electrum.ts
@@ -2,6 +2,7 @@ import bcoin from "bcoin"
 import pTimeout from "p-timeout"
 import {
   Client as BitcoinClient,
+  createOutputScriptFromAddress,
   RawTransaction,
   Transaction,
   TransactionHash,
@@ -232,7 +233,7 @@ export class Client implements BitcoinClient {
   ): Promise<UnspentTransactionOutput[]> {
     return this.withElectrum<UnspentTransactionOutput[]>(
       async (electrum: Electrum) => {
-        const script = bcoin.Script.fromAddress(address).toRaw().toString("hex")
+        const script = createOutputScriptFromAddress(address).toString()
 
         // eslint-disable-next-line camelcase
         type UnspentOutput = { tx_pos: number; value: number; tx_hash: string }
@@ -262,7 +263,7 @@ export class Client implements BitcoinClient {
     limit?: number
   ): Promise<Transaction[]> {
     return this.withElectrum<Transaction[]>(async (electrum: Electrum) => {
-      const script = bcoin.Script.fromAddress(address).toRaw().toString("hex")
+      const script = createOutputScriptFromAddress(address).toString()
 
       // eslint-disable-next-line camelcase
       type HistoryItem = { height: number; tx_hash: string }
--- a/typescript/src/electrum.ts
+++ b/typescript/src/electrum.ts
@@ -255,6 +255,58 @@ export class Client implements BitcoinClient {
 
   // eslint-disable-next-line valid-jsdoc
   /**
+   * @see {BitcoinClient#getTransactionHistory}
+   */
+  getTransactionHistory(
+    address: string,
+    limit?: number
+  ): Promise<Transaction[]> {
+    return this.withElectrum<Transaction[]>(async (electrum: Electrum) => {
+      const script = bcoin.Script.fromAddress(address).toRaw().toString("hex")
+
+      // eslint-disable-next-line camelcase
+      type HistoryItem = { height: number; tx_hash: string }
+
+      let historyItems: HistoryItem[] = await this.withBackoffRetrier<
+        HistoryItem[]
+      >()(async () => {
+        return await electrum.blockchain_scripthash_getHistory(
+          computeScriptHash(script)
+        )
+      })
+
+      // According to https://electrumx.readthedocs.io/en/latest/protocol-methods.html#blockchain-scripthash-get-history
+      // unconfirmed items living in the mempool are appended at the end of the
+      // returned list and their height value is either -1 or 0. That means
+      // we need to take all items with height >0 to obtain a confirmed txs
+      // history.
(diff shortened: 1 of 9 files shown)


typescript/src/electrum.ts
```diff
@@ -253,6 +253,58 @@ export class Client implements BitcoinClient {
...
+    address: string,
+    limit?: number
+  ): Promise<Transaction[]> {
+    return this.withElectrum<Transaction[]>(async (electrum: Electrum) => {
+      const script = bcoin.Script.fromAddress(address).toRaw().toString("hex")
+
+      // eslint-disable-next-line camelcase
+      type HistoryItem = { height: number; tx_hash: string }
+
+      let historyItems: HistoryItem[] = await this.withBackoffRetrier<
```
pdyraga, line 270: Nitpicking: Instead of having `await`, I think we could use `.then` to chain promises together.

typescript/src/wallet.ts
```diff
@@ -209,3 +217,117 @@ type DkgResult = {
...
+      const walletTransaction = walletTransactions[i]
+
+      // Find the output that locks the funds on the wallet. Only such an output
+      // can be a wallet main UTXO.
+      const outputIndex = walletTransaction.outputs.findIndex(isWalletOutput)
+
+      // Should never happen as all transactions come from wallet history. Just
+      // in case check whether the wallet output was actually found.
+      if (outputIndex < 0) {
+        continue
```
pdyraga, line 302: Should we `console.error` before continuing?

typescript/test/electrum.test.ts
```diff
@@ -108,6 +108,28 @@ describe("Electrum", () => {
         })
       })
 
+      describe("getTransactionHistory", () => {
+        it("should return proper transaction history for the given address", async () => {
+          // https://live.blockcypher.com/btc-testnet/address/tb1qumuaw3exkxdhtut0u85latkqfz4ylgwstkdzsx
```
pdyraga, line 113: Just confirming: this is an address controlled by us and it will not execute any new transactions in the future? 😆

typescript/test/wallet.test.ts
```diff
@@ -0,0 +1,246 @@
...
+          // Set the main UTXO hash based on the oldest transaction from walletWitnessTransactionHistory.
+          actualMainUtxo: {
+            transactionHash: Hex.from(
+              "3ca4ae3f8ee3b48949192bc7a146c8d9862267816258c85e02a44678364551e1"
+            ),
+            outputIndex: 0,
+            value: BigNumber.from(100000),
+          },
+          expectedMainUtxo: undefined,
+        },
```
pdyraga, line 178: We missed the test case for an old legacy transaction.

pdyraga (commented):
The code looks good to me. I added some minor comments. 

Leaving it in @r-czajkowski's hands.

typescript/src/electrum.ts
```diff
@@ -253,6 +253,58 @@ export class Client implements BitcoinClient {
...
+  // eslint-disable-next-line valid-jsdoc
+  /**
+   * @see {BitcoinClient#getTransactionHistory}
+   */
+  getTransactionHistory(
+    address: string,
+    limit?: number
+  ): Promise<Transaction[]> {
+    return this.withElectrum<Transaction[]>(async (electrum: Electrum) => {
+      const script = bcoin.Script.fromAddress(address).toRaw().toString("hex")
```
r-czajkowski, line 265: Can't we use `createOutputScriptFromAddress` function?

typescript/src/wallet.ts
```diff
@@ -209,3 +217,117 @@ type DkgResult = {
...
+    return undefined
+  }
+
+  // The most common case is that the wallet uses a witness address for all
+  // operations. Try to determine the main UTXO for that address first as the
+  // chance for success is the highest here.
+  const mainUtxo = await determine(true)
+  if (mainUtxo) {
+    return mainUtxo
+  }
```
r-czajkowski, line 328: We can simplify `rerturn mainUtxo ?? determine(false)`

r-czajkowski (commented):
Left minor comments. Overall LGTM 🚀

typescript/src/electrum.ts
lukasz-zimnoch: I think `async/await` style leads to more readable code and is generally preferred over `then`. It is also widely used in our codebase and I think mixing both styles is not necessarily the best idea.

typescript/src/wallet.ts
lukasz-zimnoch: We can. See 0e369b2f

typescript/test/electrum.test.ts
lukasz-zimnoch: Yep, it is :)

typescript/src/electrum.ts
lukasz-zimnoch: Good catch! See 056cac5b

typescript/src/wallet.ts
lukasz-zimnoch: See 7b0b9bdb

typescript/test/wallet.test.ts
lukasz-zimnoch: See f2e5b9f9

r-czajkowski (approved):
🚀

PR #652, opened 2023-07-12
Improve errors in `findWalletForRedemption` function

r-czajkowski:
Throw error when currently, there are no live wallets in the network and when the user requested redemption for all live wallets in the network using the same Bitcoin address - in that case, a user should use another Bitcoin address.

### DIFF
--- a/typescript/src/redemption.ts
+++ b/typescript/src/redemption.ts
@@ -412,11 +412,12 @@ export async function getRedemptionRequest(
 }
 
 /**
- * Finds the oldest active wallet that has enough BTC to handle a redemption request.
+ * Finds the oldest live wallet that has enough BTC to handle a redemption
+ * request.
  * @param amount The amount to be redeemed in satoshis.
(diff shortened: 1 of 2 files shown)


typescript/src/redemption.ts
```diff
@@ -441,6 +441,7 @@ export async function findWalletForRedemption(
       }
     | undefined = undefined
   let maxAmount = BigNumber.from(0)
+  let activeWalletsCounter = 0
```
lukasz-zimnoch, line 444: There is always only one "active" wallet in the system. This is the one that accepts new deposits. What we are talking about here are "live" wallets that are operable from the Bridge's standpoint. I suggest naming this variable `liveWalletsCounter` and adjusting the nomenclature everywhere accordingly.

r-czajkowski: 19e2dff

lukasz-zimnoch (approved):

PR #677, opened 2023-08-01
Expose past redemption requested events getter

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/3664

Here we expose the `Bridge.getRedemptionRequestedEvents` method that allows fetching past redemption requested events from the `Bridge` contract.


typescript/src/ethereum.ts
```diff
@@ -762,6 +763,42 @@ export class Bridge
...
+   */
+  async getRedemptionRequestedEvents(
+    options?: GetEvents.Options,
+    ...filterArgs: Array<unknown>
+  ): Promise<RedemptionRequestedEvent[]> {
+    const events: EthersEvent[] = await this.getEvents(
+      "RedemptionRequested",
+      options,
+      ...filterArgs
+    )
```
r-czajkowski, line 779: It probably won't work if we want to filter events by the indexed `walletPubKeyHash` param. See https://github.com/ethers-io/ethers.js/pull/4244 - I ran into this issue in T daap with ethers `v5.4.7`. Here we use `v5.5.2` but I think this problem still exists in this version. As a workaround, you can take a look at [this](https://github.com/threshold-network/token-dashboard/blob/main/src/threshold-ts/tbtc/index.ts#L1041C1-L1093C1).

r-czajkowski (commented):
Overall LGTM! Left one comment to take a look before the merge.

typescript/src/ethereum.ts
lukasz-zimnoch: Added a note about that: cce9d9c6. Thanks!

r-czajkowski (approved):

PR #692, opened 2023-09-05
Delaying reveal process in the "make and reveal a deposit" test.

dimpar:
It happens that a reveal process starts when a deposit is not captured by the Bitcoin chain because it's out of sync. A delay is added to wait a bit for a chain to sync. This should fix:

```
1) System Test - Minting and unminting
       when minting is initiated by making and revealing a deposit to the TBTCVault
         "before all" hook: make and reveal deposit for "should broadcast the deposit transaction on the Bitcoin network":
     Error: Electrum action failure: [daemon error: DaemonError({'code': -5, 'message': 'No such mempool or blockchain transaction. Use gettransaction for wallet transactions.'})]
      at Client.withElectrum (node_modules/@keep-network/tbtc-v2.ts/src/electrum.ts:190:13)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:95:5)
      at async Object.revealDeposit (node_modules/@keep-network/tbtc-v2.ts/src/deposit.ts:393:5)
      at async Context.<anonymous> (test/minting-unminting.test.ts:159:9)
```

which was observed here: https://github.com/keep-network/tbtc-v2/actions/runs/6061282789/job/16446311652

### DIFF
--- a/system-tests/test/minting-unminting.test.ts
+++ b/system-tests/test/minting-unminting.test.ts
@@ -153,6 +153,12 @@ describe("System Test - Minting and unminting", () => {
         - Output index: ${depositUtxo.outputIndex}
       `)
 
+        // It happens from time to time that a deposit reveal process starts when
+        // a deposit is not captured by the Bitcoin chain yet and a deposit is
+        // revealed with a non-existing Bitcoin tx. We should wait some time so
+        // the Bitcoin chain is in sync and then start the revealing process.
+        await new Promise((r) => setTimeout(r, 3000))
+
         // Since the reveal deposit logic does not perform SPV proof, we
         // can reveal the deposit transaction immediately without waiting
         // for confirmations.


michalinacienciala:
I'm running the system tests on the code from this PR here: https://github.com/keep-network/tbtc-v2/actions/runs/6084288848. Let see in 2 hours if they're green.

michalinacienciala:
@dimpar, take a look at https://github.com/keep-network/tbtc-v2/actions/runs/6084288848/job/16505924432 - the redemptions are still going on, but we have a failure in the minting flow, different than previously, but in the same test:
```
  1) System Test - Minting and unminting
       when minting is initiated by making and revealing a deposit to the TBTCVault
         "before all" hook: make and reveal deposit for "should broadcast the deposit transaction on the Bitcoin network":
     Error: Electrum action failure: [the transaction was rejected by network rules.
```

michalinacienciala:
Btw, there's [a GH Actions outage right now](https://www.githubstatus.com/) and some things may behave abnormally. But the error we got in the system tests doesn't look to me like something caused by that.

michalinacienciala:
I've re-run the System Tests job now that the GH outage is fixed: https://github.com/keep-network/tbtc-v2/actions/runs/6084288848. Let's see how it behaves...

michalinacienciala (approved):
I've run the System tests twice on the feature branch, after the GH outage and the tests were passing.

PR #693, opened 2023-09-06
Get rid of `bcrypto` dependency

r-czajkowski:
Refs: https://github.com/keep-network/tbtc-v2/issues/695

We use `bcrypto` to compute hashes, but since this library is a bit problematic, here we replace it with hashing algorithms from the `ethers` library. `ethers` has no `hash160` algorithm but under the hood `HASH160 = RIPEMD160(SHA256(X))` so it's possible to compute `hash160` using `ethers`.

### DIFF
--- a/typescript/src/bitcoin.ts
+++ b/typescript/src/bitcoin.ts
@@ -1,9 +1,7 @@
 import bcoin, { TX, Script } from "bcoin"
 import wif from "wif"
 import bufio from "bufio"
-import hash160 from "bcrypto/lib/hash160"
-import sha256 from "bcrypto/lib/sha256-browser.js"
-import { BigNumber } from "ethers"
+import { BigNumber, utils } from "ethers"
 import { Hex } from "./hex"
 import { BitcoinNetwork, toBcoinNetwork } from "./bitcoin-network"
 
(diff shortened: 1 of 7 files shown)


Shadowfiend:
Glory hallelujah.

mhluongo:
I am shocked at the size of this PR, wow!

Would be really nice to get test vectors confirming a few hashes :pray:

lukasz-zimnoch:
Nice work! However... `bcrypto` still remains a transitive dependency. This is because `bcoin` depends on `bcrypto` as well. Are we sure this PR fully solves the problem?

tomaszslabon:
Looks good to me.

@lukasz-zimnoch 
> Nice work! However... `bcrypto` still remains a transitive dependency. This is because `bcoin` depends on `bcrypto` as well. Are we sure this PR fully solves the problem?
I think this PR is meant to be the very first step.
If we want to get rid of `bcrypto` entirely we will have to either get rid of `bcoin` or modify `bcoin` to use some other library for cryptography.

Shadowfiend:
> Nice work! However... `bcrypto` still remains a transitive dependency. This is because `bcoin` depends on `bcrypto` as well. Are we sure this PR fully solves the problem?

You're breaking my heart 😆 (This makes a lot more sense.)

typescript/src/bitcoin.ts
```diff
@@ -525,7 +523,12 @@ export function createKeyRing(
  * @returns Hash as a 20-byte un-prefixed hex string.
  */
 export function computeHash160(text: string): string {
-  return hash160.digest(Buffer.from(text, "hex")).toString("hex")
+  const sha256Hash = utils.sha256(
+    Hex.from(Buffer.from(text, "hex")).toPrefixedString()
```
lukasz-zimnoch, line 527: I'd be very careful here as we pass the `0x`-prefixed value to the `utils.sha256` hash function. I checked Ethers docs and looks like the prefix is not taken into account while calculating hashes but, just in case, can we cover this function with some basic unit tests to ensure it always works as expected? 

Same for `computeHash256` and `computeScriptHash`.

lukasz-zimnoch (approved):

Issue #698, opened 2023-09-18
 Revamp the typescript library API


lukasz-zimnoch:
Rework our current `typescript` API to be developer-friendly. 
```[tasklist]
### Tasks
- [ ] https://github.com/keep-network/tbtc-v2/pull/705
```

Issue #695, opened 2023-09-15
Replace `bcoin` with `bitcoinjs-lib`


lukasz-zimnoch:
As decided in https://github.com/keep-network/tbtc-v2/issues/694, we aim to replace `bcoin` with `bitcoinjs-lib`. This task is capturing all the required work.

```[tasklist]
### Tasks
- [ ] https://github.com/keep-network/tbtc-v2/pull/693
- [ ] https://github.com/keep-network/tbtc-v2/pull/696
- [ ] https://github.com/keep-network/tbtc-v2/pull/700
- [ ] https://github.com/keep-network/tbtc-v2/pull/702
- [ ] https://github.com/keep-network/tbtc-v2/pull/703
- [ ] https://github.com/keep-network/tbtc-v2/pull/706
- [ ] https://github.com/keep-network/tbtc-v2/pull/707
- [ ] https://github.com/keep-network/tbtc-v2/pull/708
```

PR #705, opened 2023-09-29
General revamp of the typescript library

lukasz-zimnoch:
Closes: https://github.com/keep-network/tbtc-v2/issues/698

Here we present a comprehensive refactoring of the tBTC v2 typescript library (a.k.a. typescript SDK). So far, the library was basically a bag of various utility functions, and the responsibility of using them correctly was on the library clients' side. This approach was enough for the first phases of protocol development where the library was used mainly for internal purposes. However, as the protocol becomes more mature,  it makes sense to encourage external integrators. This, in turn, requires us to improve the developer experience provided by the SDK and reduce the overhead for users. This pull request aims to achieve that goal.

The initial plan was to open several PRs one after another but, given the fact that https://github.com/keep-network/tbtc-v2/issues/695 was happening simultaneously and had precedence, it was easier to track progress and resolve conflicts within a single pull request. Because of the size, I strongly recommend reviewing commit by commit. The history of changes clearly shows all the steps that were taken. Worth noting that there are very few changes in the logic itself. Most of them are file/directory structure changes, renames, and public API improvements. Here is a detailed description of specific steps:

### Rework of the file and directory structure

So far, the file structure of the library was flat and grouped related functions into files like `bitcoin.ts` or `deposit.ts`. This quickly led to poor readability and encapsulation. The new approach uses a hierarchical approach dividing code into two distinct categories:
- Shared libraries living under `lib` directory
- Feature services living under `services` directory

#### Shared libraries

This category groups all code components that are used across different features. Specific modules are:
- `lib/bitcoin`: Contains all interfaces and utilities necessary to interact with the Bitcoin chain
- `lib/contracts`: Provides chain-agnostic interfaces allowing to interact with tBTC v2 smart contracts 
- `lib/electrum`: Contains an Electrum-based implementation of the Bitcoin client
- `lib/ethereum`: Provides implementations of tBTC v2 smart contract interfaces for Ethereum chain
- `lib/utils`: Holds some general utility functions

Each aforementioned module contains an `index.ts` file that re-exports components from specific sub-modules in order to flatten import paths used by library clients.

#### Feature services

This category holds services providing access to specific features of the tBTC v2 system:
- `services/deposits`: Provides access to the tBTC v2 deposit flow for BTC depositors through the `DepositsService` component
- `services/redemptions`: Provides access to the tBTC v2 redemption flow for TBTC redeemers through the `RedemptionsService` component
- `services/maintenance`: Provides access to authorized maintenance actions for maintainers (e.g. optimistic minting guardians) through the `MaintenanceService` component

Just as for shared libraries, each module contains an `index.ts` that re-exports all exported components from specific sub-modules.

### Provide a simple entry point with embedded chain configurations

So far, the library has not provided any unified entry point that could be used by external clients. The root `index.ts` just exported the most important components without any further guidance. Here we aim to change that by introducing the `TBTC` entry point class. The main purpose of it is to provide some static functions allowing to quickly initialize the SDK and expose specific feature services in a simple way. As for now, the `TBTC` component can be initialized using the following functions:
- `initializeMainnet` which initializes the SDK to point tBTC v2 contracts on Ethereum and use mainnet as Bitcoin network
- `initializeGoerli` which initializes the SDK to point tBTC v2 contracts on Goerli and use testnet as Bitcoin network
- `initializeCustom` which allows using custom tBTC v2 contracts (e.g. local) and use an arbitrary Bitcoin network (e.g. regtest or simnet)

Once initialized, the `TBTC` component provides easy access to specific feature services. To address unforeseen needs, it also gives low-level direct access to the underlying tBTC v2 smart contracts and the Bitcoin client:

```
// Initialize SDK
const signer = (...) 
const sdk = await TBTC.initializeMainnet(signer) 

// Access feature services
sdk.deposits.(...)
sdk.redemptions.(...)

// Access tBTC v2 smart contracts and Bitcoin client directly
sdk.tbtcContracts.bridge.(...)
sdk.bitcoinClient.(...)
```

### Other changes

Apart from the ones mentioned above, there were other important changes that had to be made. First and foremost, this PR integrates `bcoin` removal done as part of https://github.com/keep-network/tbtc-v2/issues/695. The integration was achieved by applying changes from specific PRs in separate commits here. Secondly, the unit tests file structure was adjusted to reflect the new approach with `lib` and `services` top-level directories.

### Usage examples

Changes made as part of this work greatly facilitated access to the two core use cases of the tBTC v2 system: depositing and redeeming. Here are some examples of how it looks like now:
```
// Initialize SDK
const signer = (...) 
const sdk = await TBTC.initializeMainnet(signer) 

// Deposit flow
const bitcoinRecoveryAddress = "" // take this from user input
const deposit = await sdk.deposits.initiateDeposit(bitcoinRecoveryAddress)
const depositAddress = await deposit.getBitcoinAddress() // present that to the user
await deposit.initiateMinting() // this will initiate minting if BTC were sent to the deposit address

// Redemption flow
const bitcoinRedeemerAddress = "" // take this from user input
const amount = "" // take this from user input
await sdk.redemptions.requestRedemption(bitcoinRedeemerAddress, amount)
```


typescript/src/lib/bitcoin/csuint.ts
```diff
@@ -0,0 +1,48 @@
...
+function read(varLenData: Hex): {
+  value: number
+  byteLength: number
+} {
+  // The varLenData is prefixed with the compact size uint. According to the docs
+  // (https://developer.bitcoin.org/reference/transactions.html#compactsize-unsigned-integers)
+  // a compact size uint can be 1, 3, 5 or 9 bytes. To determine the exact length,
+  // we need to look at the discriminant byte which is always the first byte of
+  // the compact size uint.
+  const discriminant = varLenData.toString().slice(0, 2)
```
tomaszslabon, line 22: Would adding a check against empty `varLenData` make sense?

typescript/test/bitcoin.test.ts
```diff
@@ -303,271 +348,351 @@ describe("Bitcoin", () => {
...
-                ? BitcoinNetwork.Mainnet
-                : BitcoinNetwork.Testnet
-            )
+  describe("BitcoinCompactSizeUint", () => {
+    const { read } = BitcoinCompactSizeUint
 
-            expect(result.toString()).to.eq(address)
+    describe("read", () => {
+      context("when the compact size uint is 1-byte", () => {
+        it("should return the the uint value and byte length", () => {
```
tomaszslabon, line 606:
```suggestion
        it("should return the uint value and byte length", () => {
```

typescript/src/lib/bitcoin/csuint.ts
```diff
@@ -0,0 +1,48 @@
...
+    case "ff":
+    case "fe":
+    case "fd": {
+      throw new Error(
+        "support for 3, 5 and 9 bytes compact size uints is not implemented yet"
+      )
+    }
+    default: {
+      // The discriminant tells the compact size uint is 1 byte. That means
+      // the discriminant represent the value itself.
```
tomaszslabon, line 34:
```suggestion
      // the discriminant represents the value itself.
```

typescript/src/lib/bitcoin/ecdsa-key.ts
```diff
@@ -0,0 +1,83 @@
...
+    prefix = "02"
+  } else {
+    prefix = "03"
+  }
+
+  return `${prefix}${publicKeyX}`
+}
+
+/**
+ * Utility functions allowing to perform Bitcoin ECDSA public keys.
```
tomaszslabon, line 54:
```suggestion
 * Utility functions allowing to perform operations on Bitcoin ECDSA public keys.
```

typescript/src/lib/bitcoin/ecdsa-key.ts
```diff
@@ -0,0 +1,83 @@
...
+  // The Y coordinate is the next 32 bytes.
+  const publicKeyY = publicKey.substring(64)
+
+  let prefix: string
+  if (BigNumber.from(`0x${publicKeyY}`).mod(2).eq(0)) {
+    prefix = "02"
+  } else {
+    prefix = "03"
+  }
+
```
tomaszslabon, line 49: Maybe ternary operator would be cleaner?
```
const prefix = BigNumber.from(`0x${publicKeyY}`).mod(2).eq(0) ? "02" : "03";
```

typescript/src/lib/bitcoin/tx.ts
```diff
@@ -0,0 +1,230 @@
...
+    version: getTxVersion(tx),
+    inputs: getTxInputVector(tx),
+    outputs: getTxOutputVector(tx),
+    locktime: getTxLocktime(tx),
+  }
+}
+
+/**
+ * Converts Bitcoin specific locktime value to a number. The number represents
+ * either a block height or an Unix timestamp depending on the value.
```
tomaszslabon, line 181:
```suggestion
 * either a block height or a Unix timestamp, depending on the value.
```

typescript/src/lib/bitcoin/address.ts
```diff
@@ -0,0 +1,121 @@
...
+function addressToPublicKeyHash(
+  address: string,
+  bitcoinNetwork: BitcoinNetwork
+): string {
+  const network = toBitcoinJsLibNetwork(bitcoinNetwork)
+
+  try {
+    // Try extracting hash from P2PKH address.
+    const hash = payments.p2pkh({ address: address, network }).hash!
+    return hash.toString("hex")
```
nkuba, line 69: We could use `Hex` util for conversion.
```suggestion
    return Hex.from(hash).toString()
```

I won't mention it again for other places we deal with hex :pray:

typescript/src/lib/bitcoin/address.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Converts an address to the respective output script.
+ * @param address BTC address.
+ * @param bitcoinNetwork Bitcoin network corresponding to the address.
+ * @returns The un-prefixed and not prepended with length output script.
+ */
+function addressToOutputScript(
+  address: string,
+  bitcoinNetwork: BitcoinNetwork
+): Hex {
```
nkuba, line 90: If this is an exported function do we want to expose `Hex` type or return `string` and convert it with `Hex.from(...).toString()`?

typescript/src/lib/bitcoin/address.ts
```diff
@@ -0,0 +1,121 @@
...
+    script.toBuffer(),
+    toBitcoinJsLibNetwork(bitcoinNetwork)
+  )
+}
+
+/**
+ * Utility functions allowing to perform Bitcoin address conversions.
+ */
+export const BitcoinAddressConverter = {
+  publicKeyToAddress,
```
nkuba, line 116: `publicKeyToAddress` function is additionally exported with `export` statement. Is there a reason other functions are not?

typescript/src/lib/bitcoin/csuint.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Reads the leading compact size uint from the provided variable length data.
+ *
+ * WARNING: CURRENTLY, THIS FUNCTION SUPPORTS ONLY 1-BYTE COMPACT SIZE UINTS
+ * AND WILL THROW ON COMPACT SIZE UINTS OF DIFFERENT BYTE LENGTH.
+ *
+ * @param varLenData Variable length data preceded by a compact size uint.
+ * @returns An object holding the value of the compact size uint along with the
+ *          compact size uint byte length.
+ */
+function read(varLenData: Hex): {
```
nkuba, line 13: Could it be `HexLike` I mentioned in another comment?

```suggestion
function read(varLenData: HexLike): {
```

typescript/src/lib/bitcoin/csuint.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Reads the leading compact size uint from the provided variable length data.
+ *
+ * WARNING: CURRENTLY, THIS FUNCTION SUPPORTS ONLY 1-BYTE COMPACT SIZE UINTS
+ * AND WILL THROW ON COMPACT SIZE UINTS OF DIFFERENT BYTE LENGTH.
+ *
+ * @param varLenData Variable length data preceded by a compact size uint.
+ * @returns An object holding the value of the compact size uint along with the
+ *          compact size uint byte length.
+ */
```
nkuba, line 12: Do we have unit tests for this utility?

typescript/src/lib/bitcoin/ecdsa-key.ts
```diff
@@ -0,0 +1,84 @@
...
+import { ECPairFactory, ECPairInterface } from "ecpair"
+import * as tinysecp from "tiny-secp256k1"
+import { BitcoinNetwork, toBitcoinJsLibNetwork } from "./network"
+
+/**
+ * Checks whether given public key is a compressed Bitcoin public key.
+ * @param publicKey - Public key that should be checked.
+ * @returns True if the key is a compressed Bitcoin public key, false otherwise.
+ */
+function isCompressedPublicKey(publicKey: string): boolean {
```
nkuba, line 12: `publicKey: HexLike`?

typescript/src/lib/bitcoin/ecdsa-key.ts
```diff
@@ -0,0 +1,84 @@
...
+  )
+}
+
+/**
+ * Compresses the given uncompressed Bitcoin public key.
+ * @param publicKey Uncompressed 64-byte public key as an unprefixed hex string.
+ * @returns Compressed 33-byte public key prefixed with 02 or 03.
+ */
+function compressPublicKey(publicKey: string | Hex): string {
+  if (typeof publicKey === "string") {
```
nkuba, line 26: We should make `Hex.from` accept also `Hex` to make things easier and don't need to check type here.

typescript/src/lib/bitcoin/ecdsa-key.ts
```diff
@@ -0,0 +1,84 @@
...
+ * @param bitcoinNetwork Bitcoin network the given key pair is relevant for.
+ * @returns Bitcoin key ring.
+ */
+function createKeyPair(
+  privateKey: string,
+  bitcoinNetwork: BitcoinNetwork
+): ECPairInterface {
+  // eslint-disable-next-line new-cap
+  return ECPairFactory(tinysecp).fromWIF(
+    privateKey,
```
nkuba, line 74: Does this function expect a private key or WIF?
ECDSA Private Key is different from WIF and `ECPairFactory` supports both.

typescript/src/lib/bitcoin/tx.ts
```diff
@@ -0,0 +1,237 @@
...
+}
+
+/**
+ * Represents a raw Bitcoin transaction decomposed into specific vectors.
+ */
+export interface BitcoinRawTxVectors {
+  /**
+   * Transaction version as an un-prefixed hex string.
+   */
+  version: string
```
nkuba, line 107: We're not consistent with using `Hex` and `string` types around the interfaces.

nkuba (commented):
Impressive work! :clap:
I'm leaving the first chunk of comments.

typescript/src/lib/bitcoin/address.ts
```diff
@@ -0,0 +1,121 @@
...
+export function publicKeyToAddress(
+  publicKey: Hex,
+  bitcoinNetwork: BitcoinNetwork,
+  witness: boolean = true
+): string {
+  const network = toBitcoinJsLibNetwork(bitcoinNetwork)
+
+  if (witness) {
+    // P2WPKH (SegWit)
+    return payments.p2wpkh({ pubkey: publicKey.toBuffer(), network }).address!
```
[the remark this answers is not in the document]
tomaszslabon: If the provided `publicKey` is correct, the `address` will always be correct.
If the provided `publicKey` is incorrect (e.g. is empty, the length is incorrect, the prefix is incorrect) this function will throw an error.

typescript/src/lib/bitcoin/address.ts
```diff
@@ -0,0 +1,121 @@
...
+ */
+function publicKeyHashToAddress(
+  publicKeyHash: string,
+  witness: boolean,
+  bitcoinNetwork: BitcoinNetwork
+): string {
+  const hash = Buffer.from(publicKeyHash, "hex")
+  const network = toBitcoinJsLibNetwork(bitcoinNetwork)
+  return witness
+    ? payments.p2wpkh({ hash, network }).address!
```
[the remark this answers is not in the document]
tomaszslabon: If the provided `publicKeyHash` is incorrect (it's not a hexstring of length `20`) the function will throw an error:
```
  1) Bitcoin
       BitcoinAddressConverter
         publicKeyHashToAddress
           when network is mainnet
             when witness option is false
               when proper public key hash is provided
                 should encode public key hash into bitcoin address properly:
     Error: Expected property "hash" of type Buffer(Length: 20), got Buffer(Length: 19)
```
So it either succeeds (i.e. the address is defined) or it throws.
We have unit tests covering incorrect length.

typescript/src/lib/bitcoin/script.ts
```diff
@@ -0,0 +1,67 @@
+import { payments } from "bitcoinjs-lib"
+
+/**
+ * Checks if the provided script comes from a P2PKH input.
+ * @param script The script to be checked.
+ * @returns True if the script is P2PKH, false otherwise.
+ */
+function isP2PKHScript(script: Buffer): boolean {
+  try {
+    payments.p2pkh({ output: script })
```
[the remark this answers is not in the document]
tomaszslabon: Checking if the function throws or not seems to be  the way this function was designed to be used.
See [an example in the bitcoinjs-lib library](https://github.com/bitcoinjs/bitcoinjs-lib/blob/6e6b889ece668828d2fd188f00be2ad58f0be117/src/psbt/psbtutils.js#L21-L37).

typescript/src/lib/bitcoin/ecdsa-key.ts
tomaszslabon: The private key should be in the WIF format and look like this:
```
cVbZ8ovhye9AoAHFsqobCf7LxbXDAECy9Kb8TZdfsDYMZGBUyCnm
```

typescript/src/lib/bitcoin/address.ts
```diff
@@ -0,0 +1,121 @@
...
+ * @param publicKeyHash Public key hash that will be encoded. Must be an
+ *        unprefixed hex string (without 0x prefix).
+ * @param witness If true, a witness public key hash will be encoded and
+ *        P2WPKH address will be returned. Returns P2PKH address otherwise
+ * @param bitcoinNetwork Network the address should be encoded for.
+ * @returns P2PKH or P2WPKH address encoded from the given public key hash
+ * @throws Throws an error if network is not supported.
+ */
+function publicKeyHashToAddress(
+  publicKeyHash: string,
```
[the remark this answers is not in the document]
tomaszslabon: I agree, we should choose one option: either `string` or `Hex` to represent hexadecimal values.
The `Hex` would be better, because of validation.
However, can we use `Hex` for exported functions?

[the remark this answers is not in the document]
lukasz-zimnoch: Yeah we should use `Hex` and there is ongoing work supposed to address that: https://github.com/keep-network/tbtc-v2/issues/713

typescript/src/lib/bitcoin/address.ts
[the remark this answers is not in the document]
lukasz-zimnoch: Let's add unit tests as part of https://github.com/keep-network/tbtc-v2/issues/714 @tomaszslabon

typescript/src/lib/bitcoin/address.ts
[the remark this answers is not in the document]
lukasz-zimnoch: Let's add hex conversion as part of https://github.com/keep-network/tbtc-v2/issues/713 and unit tests as part of https://github.com/keep-network/tbtc-v2/issues/714

typescript/src/lib/bitcoin/address.ts
```diff
@@ -0,0 +1,121 @@
...
+  witness: boolean,
+  bitcoinNetwork: BitcoinNetwork
+): string {
+  const hash = Buffer.from(publicKeyHash, "hex")
```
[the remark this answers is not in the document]
lukasz-zimnoch: Will be done as part of https://github.com/keep-network/tbtc-v2/issues/713

typescript/src/lib/bitcoin/csuint.ts
lukasz-zimnoch: I'd prefer to not use ether's types as part of our public API directly. Let's stick with our own `Hex`. If we need to replace ethers one day, this decision will be beneficial.

typescript/src/lib/bitcoin/csuint.ts
lukasz-zimnoch: Of course, see `bitcoin.test.ts`: https://github.com/keep-network/tbtc-v2/blob/4f766f31d784d840c9f996bc6d4cb15bd756f214/typescript/test/lib/bitcoin.test.ts#L764C45-L764C45

typescript/src/lib/bitcoin/ecdsa-key.ts
lukasz-zimnoch: Yeah but not as part of this PR. Something for the future.

typescript/src/lib/bitcoin/hash.ts
```diff
@@ -0,0 +1,59 @@
...
+  return Hex.from(hash160).toString()
+}
+
+/**
+ * Computes the double SHA256 for the given text.
+ * @param text - Text the double SHA256 is computed for.
+ * @returns Hash as a 32-byte un-prefixed hex string.
+ * @dev Do not confuse it with computeSha256 which computes single SHA256.
+ */
+function computeHash256(text: Hex): Hex {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Will be done as part of https://github.com/keep-network/tbtc-v2/issues/713

typescript/src/lib/bitcoin/hash.ts
```diff
@@ -0,0 +1,59 @@
...
+  return BigNumber.from(hash.reverse().toPrefixedString())
+}
+
+/**
+ * Computes the single SHA256 for the given text.
+ * @param text - Text the single SHA256 is computed for.
+ * @returns Hash as a 32-byte un-prefixed hex string.
+ * @dev Do not confuse it with computeHash256 which computes double SHA256.
+ */
+function computeSha256(text: Hex): Hex {
```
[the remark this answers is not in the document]
lukasz-zimnoch: I think it's better to keep it here. `computeHash256` and `computeHash160` are related to each other more as they both refer to respective Bitcoin opcodes. The `computeHash256` is just a generic hash function not related to Bitcoin opcodes.

typescript/src/lib/bitcoin/hash.ts
```diff
@@ -0,0 +1,59 @@
+import { BigNumber, utils } from "ethers"
+import { Hex } from "../utils"
+
+/**
+ * Computes the HASH160 for the given text.
```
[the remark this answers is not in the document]
lukasz-zimnoch: `HASH160` is Bitcoin opcode corresponding to `RIPEMD160(SHA256(x))` but yeah, I'll make it crystal clear.

typescript/src/lib/bitcoin/header.ts
```diff
@@ -0,0 +1,270 @@
+import { BigNumber } from "ethers"
+import { Hex } from "../utils"
+import { BitcoinHashUtils } from "./hash"
+
+/**
+ * BitcoinHeader represents the header of a Bitcoin block. For reference, see:
+ * https://developer.bitcoin.org/reference/block_chain.html#block-headers.
+ */
+export interface BitcoinHeader {
```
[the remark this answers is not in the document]
lukasz-zimnoch: I'd rather keep it short. Bitcoin headers always refer to block headers and there is no ambiguity here.

typescript/src/lib/bitcoin/header.ts
```diff
@@ -0,0 +1,270 @@
...
+  }
+}
+
+/**
+ * Deserializes a raw representation of a Bitcoin block headers chain.
+ * @param rawHeadersChain - Raw Bitcoin block headers chain.
+ * @returns Deserialized Bitcoin block headers.
+ */
+function deserializeHeadersChain(rawHeadersChain: string): BitcoinHeader[] {
+  if (rawHeadersChain.length % 160 !== 0) {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Will be changed as part of https://github.com/keep-network/tbtc-v2/issues/713

typescript/src/lib/bitcoin/header.ts
```diff
@@ -0,0 +1,270 @@
...
+
+    // Save the current block header hash to compare it with the next block
+    // header's previous block header hash.
+    previousBlockHeaderHash = currentBlockHeaderHash
+
+    // Check if the stored block difficulty is equal to previous or current
+    // difficulties.
+    const difficulty = targetToDifficulty(difficultyTarget)
+
+    if (previousEpochDifficulty.eq(1) && currentEpochDifficulty.eq(1)) {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Not sure if this change doesn't break anything in the upstream. I'd leave it until we have a good reason to change it.

typescript/src/lib/bitcoin/script.ts
```diff
@@ -0,0 +1,67 @@
+import { payments } from "bitcoinjs-lib"
+
+/**
+ * Checks if the provided script comes from a P2PKH input.
+ * @param script The script to be checked.
+ * @returns True if the script is P2PKH, false otherwise.
+ */
+function isP2PKHScript(script: Buffer): boolean {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Definitely. Will be handled in https://github.com/keep-network/tbtc-v2/issues/713

typescript/src/lib/bitcoin/tx.ts
lukasz-zimnoch: Yeah, and https://github.com/keep-network/tbtc-v2/issues/713 will solve this.

typescript/src/lib/bitcoin/csuint.ts
lukasz-zimnoch: See 4d4f45cf

typescript/test/bitcoin.test.ts
lukasz-zimnoch: See 2343c838

typescript/src/lib/bitcoin/csuint.ts
lukasz-zimnoch: See 2343c838

typescript/src/lib/bitcoin/ecdsa-key.ts
lukasz-zimnoch: See 2343c838

typescript/src/lib/bitcoin/ecdsa-key.ts
lukasz-zimnoch: See c359fef9

typescript/src/lib/bitcoin/tx.ts
lukasz-zimnoch: See ad51b2ed

typescript/src/lib/bitcoin/address.ts
```diff
@@ -0,0 +1,121 @@
+import { address as btcjsaddress, payments } from "bitcoinjs-lib"
+import { Hex } from "../utils"
+import { BitcoinNetwork, toBitcoinJsLibNetwork } from "./network"
+
+/**
+ * Creates the Bitcoin address from the public key. Supports SegWit (P2WPKH) and
+ * Legacy (P2PKH) formats.
+ * @param publicKey - Public key used to derive the Bitcoin address.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See ad51b2ed

typescript/src/lib/bitcoin/ecdsa-key.ts
lukasz-zimnoch: Added an additional explanation. See ad51b2ed

typescript/package.json
```diff
@@ -16,9 +16,7 @@
     "dev": "tsc --project tsconfig.build.json --watch"
   },
   "files": [
-    "dist/",
-    "src/",
-    "typechain/"
+    "dist"
```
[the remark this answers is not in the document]
lukasz-zimnoch: Ah, this is the reason we need to keep them. I was sure they were redundant. See cb806377

typescript/src/lib/bitcoin/address.ts
lukasz-zimnoch: No reason. Just forgot to remove `export`. See e063022e

typescript/src/lib/bitcoin/hash.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @param text - Text the HASH160 is computed for.
+ * @returns Hash as a 20-byte un-prefixed hex string.
+ */
+function computeHash160(text: string): string {
+  const sha256Hash = utils.sha256(
+    Hex.from(Buffer.from(text, "hex")).toPrefixedString()
```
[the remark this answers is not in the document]
lukasz-zimnoch: This will be simplified during https://github.com/keep-network/tbtc-v2/issues/713

typescript/src/lib/bitcoin/hash.ts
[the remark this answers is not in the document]
lukasz-zimnoch: See ad51b2ed

typescript/src/lib/bitcoin/header.ts
```diff
@@ -0,0 +1,270 @@
...
+  for (let index = 0; index < headers.length; index++) {
+    const currentHeader = headers[index]
+
+    // Check if the current block header stores the hash of the previously
+    // processed block header. Skip the check for the first header.
+    if (index !== 0) {
+      if (
+        !previousBlockHeaderHash.equals(currentHeader.previousBlockHeaderHash)
+      ) {
+        throw new Error("Invalid headers chain")
```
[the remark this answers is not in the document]
lukasz-zimnoch: See 62ac6082

typescript/src/lib/bitcoin/address.ts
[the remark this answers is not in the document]
tomaszslabon: @lukasz-zimnoch OK, let's not resolve this comment to keep it as a reminder.

tomaszslabon (approved):

Shadowfiend:
Just like the tiniest niptick here… Do we feel like `services` is the right term for that folder? Given services is often used to refer to longer-running server-side components (even though it has a broader meaning!), I wonder if simply `component` might be a term that will result in less confusion long term? I know it's a pretty internal detail, so obviously urgency is rather low, and my conviction is similarly low here—but wanted to mention it because I had a little dissonance when I saw it.

Shadowfiend:
Oh also, holy hell this is beautiful.

lukasz-zimnoch:
> Just like the tiniest niptick here… Do we feel like `services` is the right term for that folder? Given services is often used to refer to longer-running server-side components (even though it has a broader meaning!), I wonder if simply `component` might be a term that will result in less confusion long term? I know it's a pretty internal detail, so obviously urgency is rather low, and my conviction is similarly low here—but wanted to mention it because I had a little dissonance when I saw it.

I was wrapping my head around this for a while as well :D My first take was `modules` but this is causing a lot of confusion with JavaScript modules. Not sure about `components` - IMO, it's pretty generic and does not fully reflect that those are components that directly face the SDK client. Totally agree that `services` is not ideal though. My reasoning was: this is the code that provides specific services for SDK clients, so why not call it just `services`? If we come up with something better here, I'm totally open to change it. The library is still evolving.

Shadowfiend:
Yeah I don’t love components either, but in this case it’s generic nature kind of helps with the possibility for confusion. 

I feel like there’s a better word kicking around somewhere though, I’ll see if it pops into my head at some point.

PR #709, opened 2023-10-13
Use Hex for arguments and return values

tomaszslabon:
Refs: https://github.com/keep-network/tbtc-v2/issues/713
This PR replaces `string` with `Hex` for function arguments and return values in order to unify library interface.


tomaszslabon:
@lukasz-zimnoch 
I went through all review comments in https://github.com/keep-network/tbtc-v2/pull/705 and it seems that all the comments related to using `Hex` are addressed in this PR.
Also, this PR affects the refund script. I executed the refund script with some test data and checked that it works with the changes introduced in this PR.

typescript/src/lib/bitcoin/ecdsa-key.ts
```diff
@@ -9,7 +9,13 @@ import { BitcoinNetwork, toBitcoinJsLibNetwork } from "./network"
  * @param publicKey - Public key that should be checked.
  * @returns True if the key is a compressed Bitcoin public key, false otherwise.
  */
-function isCompressedPublicKey(publicKey: string): boolean {
+function isCompressedPublicKey(publicKey: string | Hex): boolean {
```
lukasz-zimnoch, line 12: Do we need to support `publicKey` as a `string`? Can we accept just `Hex`? Same for other functions in this file.

typescript/src/lib/bitcoin/hash.ts
```diff
@@ -4,21 +4,19 @@ import { Hex } from "../utils"
...
 
-  return Hex.from(hash160).toString()
+  return Hex.from(hash160)
 }
 
 /**
  * Computes the double SHA256 for the given text.
- * @param text Text the double SHA256 is computed for.
- * @returns Hash as a 32-byte un-prefixed hex string.
+ * @param text - Text the double SHA256 is computed for.
```
lukasz-zimnoch, line 18: Nitpick, but I think we can get rid of the dash between parameter name and description. No need to change that for the existing code but let's not introduce that in the new code.

typescript/src/lib/bitcoin/address.ts
```diff
@@ -54,35 +53,33 @@ function publicKeyHashToAddress(
...
+  // If neither of them succeeded, throw an error.
   throw new Error("Address must be P2PKH or P2WPKH valid for given network")
 }
 
 /**
  * Converts an address to the respective output script.
  * @param address BTC address.
  * @param bitcoinNetwork Bitcoin network corresponding to the address.
- * @returns The un-prefixed and not prepended with length output script.
+ * @returns The output script not prepended with length.
```
lukasz-zimnoch, line 82: We can update the docstring of `outputScriptToAddress` in the same way.

lukasz-zimnoch (commented):
Excellent work! Just a few comments:

typescript/src/lib/bitcoin/ecdsa-key.ts
tomaszslabon: We can just use `Hex`.
Removed `string` argument in 527cdd99f9102ddcef800d91696a720c6602a092 for all the functions in this file.

typescript/src/lib/bitcoin/hash.ts
tomaszslabon: Done in d698946e8af08f8496c982902c5a47ddadfec330.

typescript/src/lib/bitcoin/address.ts
tomaszslabon: Done in 794b32c8871c0ee064c2b05b49c6f783fa405da4.

typescript/src/lib/contracts/bridge.ts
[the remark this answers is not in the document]
tomaszslabon: We can do that. 
However, we should remember that Ethereum hashes use mixed-case checksum address encoding (see [EIP-55](https://eips.ethereum.org/EIPS/eip-55)).
If we use `Hex` to represent Ethereum transaction hashes, we will lose this checksum encoding.
If we want to use the resulting transaction hash and call `toString()` or `toPrefixedString()` the resulting hex-string will consist of only lowercase letters.
Perhaps we should keep the transaction hases as strings, or perhaps create a custom type for that?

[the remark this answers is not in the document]
tomaszslabon: Done in 995e166a4250cdd5abe66bf3cac8adcf68d3efa7.

lukasz-zimnoch (approved):

tomaszslabon:
Due to problems with an unverified commit and a complicated commit history we decided to close this PR in favour of https://github.com/keep-network/tbtc-v2/pull/723, which contains the same chages.

Issue #697, opened 2023-09-18
Update the typescript library to node 18


lukasz-zimnoch:
Once #695 is done, the `typescript` library should be updated to the latest LTS version of Node. As part of this work, we should also bump `bitcoinjs-lib` (introduced in https://github.com/keep-network/tbtc-v2/pull/696) to the latest version.

lukasz-zimnoch:
Addressed in https://github.com/keep-network/tbtc-v2/pull/708

PR #710, opened 2023-10-16
Fix the `npm-compile-publish` job

lukasz-zimnoch:
Refs: https://github.com/keep-network/tbtc-v2/issues/697

The typescript library has been migrated from Node 14 to Node LTS (18) recently. We need to reflect that change in all CI jobs.

### DIFF
--- a/.github/workflows/npm-typescript.yml
+++ b/.github/workflows/npm-typescript.yml
@@ -19,7 +19,7 @@ jobs:
 
       - uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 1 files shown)


michalsmiarowski (approved):
LGTM. It shouldn't disrupt any workflows on the token-dashboard side, because we specified the version of `tbtc-v2.ts` in package.json: 

https://github.com/threshold-network/token-dashboard/blob/5875b44992cbe0c24439d227c6ca89e0568b11ca/package.json#L21

PR #729, opened 2023-11-01
Change TBTC token contract artifact references

lukasz-zimnoch:
The current artifact reference for local environment taken from the `@keep-network/tbtc-v2` package is `TBTCToken.json` which actually represents the v1 token. We should point to `TBTC.json` which is the v2 token.

For consistency, we are also renaming artifacts for mainnet and goerli.


kpyszkowski (approved):
Approved 👍🏼

PR #738, opened 2023-11-07
Add README do NPM package

michalinacienciala:
In order to display the description of a package in NPM registry, the `README.md` file needs to be part of the project and needs to be placed in root. We have the file, but we were not including it in the project. This commit changes that.


michalinacienciala:
We don't need this change. The `README.md` does get added to package automatically once it's present in rot.

PR #740, opened 2023-11-07
Use satoshi amount while searching wallet for redemption

lukasz-zimnoch:
The `RedemptionsService.requestRedemption` function takes an amount in the TBTC token precision (1e18). However, the `findWalletForRedemption` method called to find a wallet for redemption expects an amount in the sathoshi precision (1e8). There is a missing conversion which we are adding in this changeset.


michalsmiarowski (approved):
LGTM 🔥 

I've tested it out with the dApp and it now works properly!

michalsmiarowski:
@lukasz-zimnoch Let's resolve the conflicts first. I can't merge it because this branch is out-of-date with the `main`.

Issue #714, opened 2023-10-16
SDK final touches: Increase unit tests coverage


lukasz-zimnoch:
Depends on: https://github.com/keep-network/tbtc-v2/issues/698

This task is about re-visiting TODOs around unit tests and implementing missing unit tests to increase coverage. Specifically, this includes:

```[tasklist]
### Tasks
- [x] Cover the `Deposit.detectFunding` function
- [x] Cover the `Deposit.initiateMinting` function (parameter-less version)
- [x] Cover the `Deposit.getBitcoinAddress` function
- [x] Cover the `DepositsService.initiateDeposit` function
- [x] Add missing cases for `ElectrumClient.getTransaction` function
- [x] Check for other places that can be covered with unit tests. Focus on client-facing functions available through the `TBTC` entrypoint (ignore all components marked as `experimental` for now)
```

PR #741, opened 2023-11-08
Increase unit test coverage: unit tests for Deposit

tomaszslabon:
#Refs: https://github.com/keep-network/tbtc-v2/issues/714.
This PR adds unit tests for the `Deposit` functions.


typescript/test/services/deposits.test.ts
```diff
@@ -670,11 +678,196 @@ describe("Deposits", () => {
   })
 
   describe("Deposit", () => {
-    // TODO: Implement unit tests for other functions.
+    describe("getBitcoinAddress", () => {
```
lukasz-zimnoch, line 681: Let's also add a case for mainnet. This can be done by switching the result returned from `MockBitcoinClient.getNetwork()`.

lukasz-zimnoch (approved):
LGTM! Left one non-blocking comment that can be addressed in the follow-up PR.

lukasz-zimnoch (approved):

PR #745, opened 2023-11-14
Sepolia testnet in the SDK

lukasz-zimnoch:
Refs: https://github.com/threshold-network/solidity-contracts/issues/150

We are slowly moving to our new Sepolia testnet. Contracts are already deployed there. It is a good time to add Sepolia testnet support to the SDK.

Artifacts were taken from the following npm packages:
- `@keep-network/tbtc-v2@sepolia`
- `@keep-network/ecdsa@sepolia`


typescript/src/services/tbtc.ts
```diff
@@ -74,6 +74,19 @@ export class TBTC {
     return TBTC.initializeEthereum(signer, "goerli", BitcoinNetwork.Testnet)
   }
 
+  /**
+   * Initializes the tBTC v2 SDK entrypoint for Ethereum Sepolia and Bitcoin testnet.
+   * The initialized instance uses default Electrum servers to interact
+   * with Bitcoin testnet
```
michalsmiarowski, line 80: Non-blocking, but the way we wrap those comments looks odd. The first line does not break at 80 characters and the second line breaks too early 🤔 I've got a margin set in my IDE at 80 characters and you can see it clearer on the image below:

image

I think the ideal scenario would be to keep the margin for comments at 80 characters, just like it is for the code:

image

The same thing is happening for `initializeEthereum` and `initializeGoerli` functions docs.

Just pointing that out. I'm not gonna block the merge because of that 😅

michalsmiarowski (approved):
LGTM! 🔥

PR #746, opened 2023-11-15
Increase unit test coverage: unit tests for DepositService.initiateDeposit

tomaszslabon:
#Refs: https://github.com/keep-network/tbtc-v2/issues/714.
This PR adds missing unit tests for `DepositService.initDeposit`.

### DIFF
--- a/typescript/test/services/deposits.test.ts
+++ b/typescript/test/services/deposits.test.ts
@@ -1032,6 +1032,9 @@ describe("Deposits", () => {
                 ).toNumber()
                 const currentTimestamp = Math.floor(new Date().getTime() / 1000)
                 expect(receiptTimestamp).to.be.greaterThan(currentTimestamp)
+                // Expect blinding factor to be set and 8-byte long.
+                expect(receipt.blindingFactor).not.to.be.undefined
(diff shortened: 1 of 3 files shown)


typescript/test/services/deposits.test.ts
```diff
@@ -947,7 +948,124 @@ describe("Deposits", () => {
...
+              beforeEach(async () => {
+                deposit = await depositService.initiateDeposit(
+                  "mjc2zGWypwpNyDi4ZxGbBNnUA84bfgiwYc"
+                )
+              })
+
+              it("should initiate deposit correctly", async () => {
+                // Inspect the deposit object by looking at its receipt.
+                const receipt = deposit.getReceipt()
+                expect(receipt.depositor).to.be.equal(depositor)
```
lukasz-zimnoch, line 1022: Can we make sure the blinding factor is set and has 8 bytes? Same for the case below.

tomaszslabon: Done in 1fcc4c92bdd0afb093e1d35a9876bd9e796473ff.

lukasz-zimnoch (approved):

PR #754, opened 2023-11-22
Add support for Sepolia testnet in monitoring

lukasz-zimnoch:
Refs: https://github.com/threshold-network/solidity-contracts/issues/150

Here we adjust the tBTC v2 monitoring tool to the new Sepolia testnet.

### DIFF
--- a/monitoring/package.json
+++ b/monitoring/package.json
@@ -14,8 +14,8 @@
   },
   "dependencies": {
     "@keep-network/tbtc-v2-mainnet": "npm:@keep-network/tbtc-v2@mainnet",
-    "@keep-network/tbtc-v2-testnet": "npm:@keep-network/tbtc-v2@goerli",
-    "@keep-network/tbtc-v2.ts": "development",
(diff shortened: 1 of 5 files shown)


monitoring/package.json
```diff
@@ -14,8 +14,8 @@
   },
   "dependencies": {
     "@keep-network/tbtc-v2-mainnet": "npm:@keep-network/tbtc-v2@mainnet",
-    "@keep-network/tbtc-v2-testnet": "npm:@keep-network/tbtc-v2@goerli",
-    "@keep-network/tbtc-v2.ts": "development",
+    "@keep-network/tbtc-v2-testnet": "npm:@keep-network/tbtc-v2@sepolia",
+    "@keep-network/tbtc-v2.ts": "1.4.0-dev.1",
```
michalinacienciala, line 18: Why do we need this particular package?

lukasz-zimnoch: We can't leave `development` as it will fetch an incompatible `2.4.0-dev.X` version. I changed to `1.4.0-dev.1` as this is the version used in `yarn.lock` and we are sure it is compatible.

michalinacienciala (approved):

PR #757, opened 2023-12-13
Added validation of moving funds proposal

tomaszslabon:
#Refs: https://github.com/keep-network/keep-core/issues/3733.

This PR adds moving funds proposal validation to the wallet proposal validator smart contract.
The validation performs several checks:
- the source wallet's state must be MovingFunds,
- the source wallet's commitment hash must be submitted,
- the target wallets from the proposal must match the submitted commitment hash,
- the proposed fee must be greater than zero,
- the transaction proposed fee must not exceed the maximum fee.

### DIFF
--- a/solidity/contracts/bridge/WalletProposalValidator.sol
+++ b/solidity/contracts/bridge/WalletProposalValidator.sol
@@ -603,100 +603,49 @@ contract WalletProposalValidator {
     ///         most of the work can be done using a single readonly contract
     ///         call.
     /// @param proposal The moving funds proposal to validate.
-    /// @param walletMainUtxo The main UTXO of the source wallet.
     /// @return True if the proposal is valid. Reverts otherwise.
-    /// @dev Requirements:
-    ///     - The source wallet must be in the MovingFunds state,
-    ///     - The source wallet must not have any pending redemptions,
-    ///     - The source wallet must not have any pending moved funds sweep
-    ///       requests,
-    ///     - The source wallet's BTC balance must be greater than zero,
-    ///     - The number of target wallets in the proposal must match the
-    ///       expected count, based on the Live wallets count, source wallet's
-    ///       BTC balance and maximum BTC transfer limit,
-    ///     - All target wallets must be in the Live state and ordered correctly
-    ///       (i.e. no duplicates and in ascending order of wallet addresses),
-    ///     - The proposed moving funds transaction fee must be greater than
-    ///       zero,
-    ///     - The proposed moving funds transaction fee must not exceed the
-    ///       maximum total fee allowed for moving funds.
-    function validateMovingFundsProposal(
-        MovingFundsProposal calldata proposal,
-        BitcoinTx.UTXO calldata walletMainUtxo
-    ) external view returns (bool) {
+    /// @dev Notice that this function is meant to be invoked after the moving
+    ///      funds commitment has already been submitted. This function skips
+    ///      some checks related to the moving funds procedure as they were
+    ///      already checked on the commitment submission.
+    ///      Requirements:
+    ///      - The source wallet must be in the MovingFunds state,
+    ///      - The target wallets commitment must be submitted,
+    ///      - The target wallets commitment hash must match the target wallets
+    ///        from the proposal,
+    ///      - The proposed moving funds transaction fee must be greater than
+    ///        zero,
+    ///      - The proposed moving funds transaction fee must not exceed the
+    ///        maximum total fee allowed for moving funds.
+    function validateMovingFundsProposal(MovingFundsProposal calldata proposal)
+        external
+        view
+        returns (bool)
+    {
         Wallets.Wallet memory sourceWallet = bridge.wallets(
             proposal.walletPubKeyHash
         );
 
-        // Make sure the source wallet is eligible for moving funds.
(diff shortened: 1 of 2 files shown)


solidity/contracts/bridge/WalletProposalValidator.sol
```diff
@@ -587,6 +597,152 @@ contract WalletProposalValidator {
...
+    ///     - The number of target wallets in the proposal must match the
+    ///       expected count, based on the Live wallets count, source wallet's
+    ///       BTC balance and maximum BTC transfer limit,
+    ///     - All target wallets must be in the Live state and ordered correctly
+    ///       (i.e. no duplicates and in ascending order of wallet addresses),
+    ///     - The proposed moving funds transaction fee must be greater than
+    ///       zero,
+    ///     - The proposed moving funds transaction fee must not exceed the
+    ///       maximum total fee allowed for moving funds.
+    function validateMovingFundsProposal(
```
lukasz-zimnoch, line 623: This function can be simplified as we can leverage the fact that on-chain commitment submission is also a kind of validation. Looks like it is enough that `validateMovingFundsProposal` makes sure that:
- The source wallet's state is `MovingFunds`
- The source wallet's commitment hash is submitted
- The target wallets from the `proposal ` match the submitted commitment hash (same wallets and same order)
- The proposed fee is greater than zero
- The proposed fee does not exceed the maximum fee

In that model, the coordination leader responsible for generating a moving funds proposal can do the following:
1. Gather all data necessary to assemble a moving funds proposal
2. Submit the commitment (or check if it is already submitted). Before doing that, the leader can make sure the commitment can be submitted using a non-mutating call. This can avoid burning gas if the wallet's state does not allow for commitment submission.
3. Wait for the commitment transaction to be mined
4. Call `validateMovingFundsProposal` to make sure the proposal is actually valid and can be safely broadcasted
5. Return the proposal

On the other hand, coordination followers receiving a moving funds proposal should:
1. Call `validateMovingFundsProposal` to make sure the proposal is actually submitted and valid
2. Wait several blocks to confirm proposal validity (32 blocks?). This is needed to avoid corner cases where the commitment is submitted but not included in the canonical chain due to a reorg and another front-running transaction that affects commitment validity.
3. Call `validateMovingFundsProposal` again to ensure the commitment has not been changed during the idleness period
4. Assemble, sign, and broadcast moving funds transaction on Bitcoin

tomaszslabon: Done in 3f7d6b3687df77a8402605f8575024dc95f2a33c.

lukasz-zimnoch: Leaving it open for future reference!

solidity/contracts/bridge/WalletProposalValidator.sol
```diff
@@ -587,6 +597,71 @@ contract WalletProposalValidator {
...
+    ///      Requirements:
+    ///      - The source wallet must be in the MovingFunds state,
+    ///      - The target wallets commitment must be submitted,
+    ///      - The target wallets commitment hash must match the target wallets
+    ///        from the proposal,
+    ///      - The proposed moving funds transaction fee must be greater than
+    ///        zero,
+    ///      - The proposed moving funds transaction fee must not exceed the
+    ///        maximum total fee allowed for moving funds.
+    function validateMovingFundsProposal(MovingFundsProposal calldata proposal)
```
lukasz-zimnoch, line 620: I think there is an additional thing that should be checked in this function. Namely:
```
require(
    walletBtcBalance >= movingFundsDustThreshold,
    "Wallet BTC balance must be equal or above the moving funds dust threshold"
);
```

This is because there is the `notifyMovingFundsBelowDust` for wallets whose balance is below the dust. If we don't check that here, the following may happen:
1. The leader submits a commitment
2. The leader creates a proposal
3. The leader validates the proposal using `validateMovingFundsProposal`
4. The leader broadcasts the proposal
5. Followers receive the proposal and validate it using `validateMovingFundsProposal`
6. Followers start the action that leads to a BTC moving funds transaction
7. A third party calls `notifyMovingFundsBelowDust` that moves the wallet to the `Closing` state
8. Submitting the SPV through `submitMovingFundsProof` is no longer possible as the wallet must be in the `MovingFunds` state
9. A third party accuses the wallet of committing fraud. The wallet cannot defeat the challenge as the SPV proof cannot be submitted so the main UTXO was spent illegally

The leader will not broadcast the proposal if we add the aforementioned `require` to `validateMovingFundsProposal`. It can still submit the commitment (because that happens before proposal validation, see https://github.com/keep-network/tbtc-v2/pull/757#discussion_r1431320949) but that's not a problem. On the other hand, if a leader broadcasts the proposal regardless, the followers will reject it due to this `require`.

Please verify my understanding. We can merge this PR to not block it and add this check in a follow-up if you find the above scenario correct.

lukasz-zimnoch (approved):

solidity/contracts/bridge/WalletProposalValidator.sol
tomaszslabon: Done in db47dbb09b123424f39a5d6c3f2c75a635c04d57.
This check indeed seems to be necessary.

PR #762, opened 2023-12-20
Added validation of moved funds sweep proposal

tomaszslabon:
#Refs: https://github.com/keep-network/keep-core/issues/3733.
#Depends on: https://github.com/keep-network/tbtc-v2/pull/757.

This PR adds moved funds sweep proposal validation to the wallet proposal validator smart contract.

### DIFF
--- a/solidity/contracts/bridge/WalletProposalValidator.sol
+++ b/solidity/contracts/bridge/WalletProposalValidator.sol
@@ -731,6 +731,25 @@ contract WalletProposalValidator {
         return walletBtcBalance;
     }
 
+    /// @notice View function encapsulating the main rules of a valid moved
+    ///         funds sweep proposal. This function is meant to facilitate the
+    ///         off-chain validation of the incoming proposals. Thanks to it,
+    ///         most of the work can be done using a single readonly contract
+    ///         call.
+    /// @param proposal The moved funds sweep proposal to validate.
+    /// @return True if the proposal is valid. Reverts otherwise.
+    /// @dev Requirements:
+    ///      - The source wallet must be in the Live or MovingFunds state,
+    ///      - The moved funds sweep request identified by the proposed
+    ///        transaction hash and output index must be in the Pending state,
+    ///      - The transaction hash and output index from the proposal must
+    ///        identify a moved funds sweep request in the Pending state,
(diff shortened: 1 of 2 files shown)


solidity/contracts/bridge/WalletProposalValidator.sol
```diff
@@ -662,6 +677,61 @@ contract WalletProposalValidator {
...
+            sweepRequest.state ==
+                MovingFunds.MovedFundsSweepRequestState.Pending,
+            "Sweep request is not in Pending state"
+        );
+
+        require(
+            sweepRequest.walletPubKeyHash == proposal.walletPubKeyHash,
+            "Sweep request does not belong to the wallet"
+        );
+
```
lukasz-zimnoch, line 771: I'm wondering whether we need a kind of safety margin that would prevent handling moved funds sweep requests that are close to reaching their timeout. I mean a similar margin as we use for redemptions (`REDEMPTION_REQUEST_TIMEOUT_SAFETY_MARGIN`).

For redemptions, the goal of the safety margin is to prevent the case where the redeemer reports a timeout, gets their TBTC back, and additionally receives BTC from a belated wallet.

For moved funds sweep, a safety margin would prevent the case where the sweep request timeout is notified but the belated wallet would produce an unprovable Bitcoin transaction right after. The wallet would be anyway terminated and slashed upon timeout notification. A terminated wallet cannot be accused of fraud so, it wouldn't incur any negative consequences for producing an unprovable Bitcoin transaction. Moreover, if the sweep timeout would be a result of a software error rather than an explicit malice, such a wallet may try to recover to unlock funds. Having all funds under one UTXO would be very helpful in such a case. 

That said, I think the safety margin is not needed for this function. Can you double-check and post your thoughts?

tomaszslabon: I also think a timeout margin is unnecessary in case of moved funds sweep validation.
When a wallet performs a moved funds sweep transaction, it sends the funds to itself.
Therefore It shouldn't be a problem - the funds will end up in the wallet address, but as a single UTXO.
As you mentioned, the wallet will be in the `Terminated` state, so it won't be accused of a fraud even though it won't be able to submit the transaction proof.

lukasz-zimnoch (approved):

PR #765, opened 2023-12-21
Added source wallet BTC balance check

tomaszslabon:
Refs: https://github.com/keep-network/keep-core/issues/3733

This PR is a follow-up to https://github.com/keep-network/tbtc-v2/pull/757.
The following checks were added:
- source wallet must be positive,
- source wallet BTC balance must be equal to or greater than the `movingFundsDustThreshold` parameter.

### DIFF
--- a/solidity/contracts/bridge/WalletProposalValidator.sol
+++ b/solidity/contracts/bridge/WalletProposalValidator.sol
@@ -614,7 +614,6 @@ contract WalletProposalValidator {
     ///      - The target wallets commitment must be submitted,
     ///      - The target wallets commitment hash must match the target wallets
     ///        from the proposal,
-    ///      - The source wallet BTC balance must be positive,
     ///      - The source wallet BTC balance must be equal to or greater than
     ///        `movingFundsDustThreshold`,
     ///      - The proposed moving funds transaction fee must be greater than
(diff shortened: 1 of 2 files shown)


solidity/contracts/bridge/WalletProposalValidator.sol
```diff
@@ -645,10 +648,37 @@ contract WalletProposalValidator {
...
+        // Make sure the source wallet balance is correct.
+        uint64 sourceWalletBtcBalance = getWalletBtcBalance(
+            sourceWallet.mainUtxoHash,
+            walletMainUtxo
+        );
 
+        require(
+            sourceWalletBtcBalance > 0,
+            "Source wallet BTC balance is zero"
+        );
```
lukasz-zimnoch, line 674: Isn't this check redundant? The `Bridge` enforces that `movingFundsDustThreshold > 0` so if the next check is fulfilled, that implies the balance is greater than zero.

tomaszslabon: True. Removed the check in 8f8de021a29e4e408e7933783480ac8f2d8d93d9.

lukasz-zimnoch (approved):

PR #766, opened 2024-01-09
Update `actions/cache` to a newer version (`v3`)

michalinacienciala:
As per
https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/, GitHub has started a deprecation process for the GitHub Actions that run on Node12. One of such actions was a `v2` of the `actions/cache` action. We're updating the action to a newer `v3` version which runs on Node16.

Ref: https://github.com/keep-network/optimistic-minting/pull/102


michalinacienciala:
Will be updated to even higher version in https://github.com/keep-network/tbtc-v2/pull/789.

Issue #750, opened 2023-11-16
Support for Base direct bridging


lukasz-zimnoch:
This task is about implementing support for Base direct bridging, based on the exploratory work done in https://github.com/keep-network/tbtc-v2/pull/704. 

The scope of this task includes:
- Implementation of the necessary L1 and L2 contracts
- Necessary contract deployments and upgrades
- Implementation of the support in the tBTC SDK
```[tasklist]
### Tasks
- [ ] https://github.com/keep-network/tbtc-v2/pull/778
- [ ] https://github.com/keep-network/tbtc-v2/pull/779
- [ ] https://github.com/keep-network/tbtc-v2/pull/792
- [ ] https://github.com/keep-network/tbtc-v2/pull/798
```

PR #778, opened 2024-01-26
The `AbstractTBTCDepositor` contract

lukasz-zimnoch:
Refs: https://github.com/keep-network/tbtc-v2/issues/750

Here we introduce the `AbstractTBTCDepositor ` contract meant to facilitate integration of protocols aiming to use tBTC as an underlying Bitcoin bridge. 

### Usage of the `AbstractTBTCDepositor` contract

An integrator willing to leverage the `AbstractTBTCDepositor` contract is supposed to:
- Create a child contract inheriting from this abstract contract
- Call the `__AbstractTBTCDepositor_initialize` initializer function
- Use the `_initializeDeposit` and `_finalizeDeposit` as part of their business logic in order to initialize and finalize deposits.

Example code:

```solidity                                                           
// Example upgradeable integrator contract.                   
contract ExampleTBTCIntegrator is AbstractTBTCDepositor, Initializable {          
    /// @custom:oz-upgrades-unsafe-allow constructor                           
    constructor() {                                                            
        // Prevents the contract from being initialized again.                 
        _disableInitializers();                                                
    }                                                                          
                                                                               
    function initialize(                                                       
        address _bridge,                                                       
        address _tbtcVault                                                     
    ) external initializer {                                                   
        __AbstractTBTCDepositor_initialize(_bridge, _tbtcVault);                  
    }                                                                          
                                                                               
    function startProcess(                                                     
        IBridgeTypes.BitcoinTxInfo calldata fundingTx,                         
        IBridgeTypes.DepositRevealInfo calldata reveal                         
    ) external {                                                               
        // Embed necessary context as extra data.                              
        bytes32 extraData = ...;                                               
                                                                               
        uint256 depositKey = _initializeDeposit(                               
            fundingTx,                                                         
            reveal,                                                            
            extraData                                                          
        );                                                                     
                                                                               
        // Use the depositKey to track the process.                            
    }                                                                          
                                                                               
    function finalizeProcess(uint256 depositKey) external {                    
        (uint256 tbtcAmount, bytes32 extraData) = _finalizeDeposit(depositKey);
                                                                               
        // Do something with the minted TBTC using context                     
        // embedded in the extraData.                                          
    }                                                                          
}                                                                                                                                           
```

### The `tbtcAmount` caveat

The `tbtcAmount` computed by `_calculateTbtcAmount` and returned by `_finalizeDeposit` may not correspond to the actual amount of TBTC minted for the deposit. Although the treasury fee cut upon minting is known precisely, this is not the case for the optimistic minting fee and the Bitcoin transaction fee. To overcome that problem, this contract just takes the current maximum values of both fees, at the moment of deposit finalization. For the great majority of the deposits, such an algorithm will return a `tbtcAmount` slightly lesser than the actual amount of TBTC minted for the deposit. This will cause some TBTC to be left in the contract and ensure there is enough liquidity to finalize the deposit. However, in some rare cases, where the actual values of those fees change between the deposit minting and finalization, the `tbtcAmount` returned by this function may be greater than the actual amount of TBTC minted for the deposit. If this happens and the reserve coming from previous deposits leftovers does not provide enough liquidity, the deposit will have to wait for finalization until the reserve is refilled by subsequent deposits or a manual top-up. The integrator is responsible for handling such cases.


solidity/contracts/bridge/DepositorProxy.sol
```diff
@@ -0,0 +1,151 @@
...
+// TODO: Move to another directory?
+abstract contract DepositorProxy {
+    using BTCUtils for bytes;
+
+    /// @notice Multiplier to convert satoshi to TBTC token units.
+    uint256 public constant SATOSHI_MULTIPLIER = 10**10;
+
+    Bridge public bridge;
+    TBTCVault public tbtcVault;
+    TBTC public tbtcToken;
```
nkuba, line 38: Do we need it? It doesn't seem to be used.

solidity/contracts/bridge/DepositorProxy.sol
```diff
@@ -0,0 +1,151 @@
...
+        bridge = Bridge(_bridge);
+        tbtcVault = TBTCVault(_tbtcVault);
+        tbtcToken = TBTC(_tbtcToken);
+    }
+
+    function initializeDeposit(
+        BitcoinTx.Info calldata fundingTx,
+        Deposit.DepositRevealInfo calldata reveal,
+        bytes32 extraData
+    ) internal {
```
nkuba, line 63: Can this function return the calculated `depositKey`?

solidity/contracts/bridge/DepositorProxy.sol
lukasz-zimnoch: Yeah, we don't. I was initially thinking about a way to manage the TBTC balance from the abstract contract level but it seems better to push that responsibility to the child contracts. See 8c068d4f

solidity/contracts/bridge/DepositorProxy.sol
lukasz-zimnoch: Of course! See fb2acde8

solidity/contracts/bridge/DepositorProxy.sol
```diff
@@ -0,0 +1,151 @@
...
+import {BTCUtils} from "@keep-network/bitcoin-spv-sol/contracts/BTCUtils.sol";
+
+import "./BitcoinTx.sol";
+import "./Bridge.sol";
+import "./Deposit.sol";
+import "./../vault/TBTCVault.sol";
+import "./../vault/TBTCOptimisticMinting.sol";
+import "./../token/TBTC.sol";
+
+// TODO: Make it safe for upgradeable contracts.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/tbtc-v2/pull/778/commits/7dc4048a0b91a9523690d23574a6fa3838d5f540

solidity/contracts/integrator/TBTCDepositorProxy.sol
```diff
@@ -0,0 +1,358 @@
...
+// ██████████████     ▐████▌     ██████████████
+// ██████████████     ▐████▌     ██████████████
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+
+pragma solidity 0.8.17;
```
nkuba, line 16: As this is an abstract contract that will be implemented in other projects, we should give more flexibility around pragma to the integrator.

lukasz-zimnoch: Good point. See b7ed96ea

solidity/contracts/integrator/TBTCDepositorProxy.sol
```diff
@@ -0,0 +1,358 @@
...
+
+import {BTCUtils} from "@keep-network/bitcoin-spv-sol/contracts/BTCUtils.sol";
+import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
```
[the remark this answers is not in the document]
lukasz-zimnoch: I think we can do it just for this abstract contract. ERC7201 seems to be worth it. Let me see.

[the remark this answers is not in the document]
lukasz-zimnoch: Unfortunately, this won't be so simple. Migration from OZ v4 to v5 is non-trivial and will incur the upgrade of the OZ Upgrade plugin. This may break things while dealing with existing tBTC contract deployments. I don't want to take this risk right now.

However, to make `TBTCDepositorProxy` usable with both OZ v4 and v5, I decided to remove the dependency on `Initializable` in favor of a custom guard in the `__TBTCDepositorProxy_initialize` initializer. Regarding the storage layout protection, I think we should keep using storage gaps in this case. That gives the best flexibility as child contracts of `TBTCDepositorProxy` can still inherit from OZ v5. This is safe because ERC-7201 namespaces used by OZ v5 are safe against collisions with the standard storage layout used by Solidity (source: [ERC-7201 abstract](https://eips.ethereum.org/EIPS/eip-7201#abstract)). I have also explored the possibility of using ERC-7201 in `TBTCDepositorProxy`. Unfortunately, this is not an option because the OZ Upgrade Plugin raises an error for Solidity versions `<0.8.20` ([source](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#namespaced-storage-layout)) and `TBTCDepositorProxy` must support all `0.8.x` versions.

solidity/test/integrator/TBTCDepositorProxy.test.ts
```diff
@@ -0,0 +1,317 @@
...
+            fixture.reveal,
+            fixture.extraData
+          )
+        })
+
+        after(async () => {
+          await restoreSnapshot()
+        })
+
+        it("should reveal the deposit to the Bridge", async () => {
```
nkuba, line 128: A test to check that the deposit key is returned by `initializeDeposit` is missing.

solidity/test/integrator/TBTCDepositorProxy.test.ts
```diff
@@ -0,0 +1,317 @@
...
+              .withArgs(
+                fixture.expectedDepositKey,
+                expectedTbtcAmount,
+                fixture.extraData
+              )
+          })
+        })
+      })
+    })
+  })
```
nkuba, line 316: Would you be interested in adding tests for `calculateDepositKey` function to cover a scenario where the fees are set to zero?

solidity/contracts/integrator/TBTCDepositorProxy.sol
```diff
@@ -0,0 +1,331 @@
...
+
+        // slither-disable-next-line reentrancy-events
+        emit DepositFinalized(
+            depositKey,
+            tbtcAmount,
+            /* solhint-disable-next-line not-rely-on-time */
+            uint32(block.timestamp)
+        );
+
+        onDepositFinalized(depositKey, tbtcAmount, deposit.extraData);
```
nkuba, line 222: 🤔 Since the `finalizeDeposit` function is internal and expected to be overridden in the implementation contract, would it make more sense to return the values instead of passing them in a callback?
With the current approach, we expect the contract to implement both `finalizeDeposit` and `onDepositFinalized`, and they are invoked together.

What if we rename the internal function to `_finalizeDeposit` and return the values:
```suggestion
        return (tbtcAmount, deposit.extraData);
```

the implementation could look like this:
```solidity
    function finalizeDeposit(uint256 depositKey) public override {
        uint256 tbtcAmount, bytes32 extraData = _finalizeDeposit(depositKey);

        (address receiver, uint16 referral) = decodeExtraData(extraData);

        request.amountToStake = tbtcAmount;
        request.receiver = receiver;
        request.referral = referral;
    }
```

lukasz-zimnoch: Actually `finalizeDeposit` is not expected to be overridden (it's not `virtual`) but rather called from a child's external function. I was also thinking about returning the parameters but the solution with `onDepositFinalized` seemed more clear and intuitive to me. I don't have a strong preference though. If you find return params more convenient, I can go that way.

solidity/test/integrator/TBTCDepositorProxy.test.ts
lukasz-zimnoch: Addressed as part of a31abf46

solidity/test/integrator/TBTCDepositorProxy.test.ts
lukasz-zimnoch: Yeah, it's good to cover those paths just in case. See 9efd0d0c

solidity/contracts/test/TestTBTCDepositorProxy.sol
```diff
@@ -0,0 +1,171 @@
...
+        _deposits[depositKey] = request;
+
+        emit DepositRevealed(depositKey);
+    }
+
+    function sweepDeposit(uint256 depositKey) external {
+        require(_deposits[depositKey].revealedAt != 0, "Deposit not revealed");
+        require(_deposits[depositKey].sweptAt == 0, "Deposit already swept");
+        /* solhint-disable-next-line not-rely-on-time */
+        _deposits[depositKey].sweptAt = uint32(block.timestamp);
```
nkuba, line 96: In the integrator tests, we will need tokens to be minted to the depositor. I'm wondering whether we could mint them here in `sweepDeposit` and `finalizeOptimisticMintingRequest` or leave it to the integrator.

If leaving it to the integrator it would be worth making `sweepDeposit` and `finalizeOptimisticMintingRequest` public virtual, so the integrator can reuse the test contracts from this file and just extend it with minting.

solidity/contracts/integrator/TBTCDepositorProxy.sol
```diff
@@ -0,0 +1,309 @@
...
+    ///        (in the context of this contract) yet.
+    ///      - The deposit must be finalized on the Bridge side. That means the
+    ///        deposit must be either swept or optimistically minted.
+    /// @dev IMPORTANT NOTE: The tbtcAmount returned by this function is an
+    ///      approximation. See documentation of the `calculateTbtcAmount`
+    ///      responsible for calculating this value for more details.
+    // slither-disable-next-line dead-code
+    function _finalizeDeposit(uint256 depositKey)
+        internal
+        returns (uint256 tbtcAmount, bytes32 extraData)
```
nkuba, line 183: Let's say the integrator wants to charge further fees on the deposit, what do you think about additionally returning the gross initial deposit amount here (multiplied by the satoshi multiplier)?

solidity/contracts/test/TestTBTCDepositorProxy.sol
```diff
@@ -0,0 +1,171 @@
...
+
+        require(
+            _deposits[depositKey].revealedAt == 0,
+            "Deposit already revealed"
+        );
+
+        IBridgeTypes.DepositRequest memory request;
+
+        request.depositor = msg.sender;
+        request.amount = uint64(10 * 1e8); // 10 BTC
```
nkuba, line 79: Can we set it to the actual funding deposit amount?

```solidity
        bytes memory fundingOutput = fundingTx
            .outputVector
            .extractOutputAtIndex(reveal.fundingOutputIndex);

        uint64 fundingOutputAmount = fundingOutput.extractValue();

        request.amount = fundingOutputAmount;
```

solidity/contracts/test/TestTBTCDepositorProxy.sol
```diff
@@ -0,0 +1,171 @@
...
+        /* solhint-disable-next-line not-rely-on-time */
+        request.revealedAt = uint32(block.timestamp);
+        request.vault = reveal.vault;
+        request.treasuryFee = uint64(1 * 1e8); // 1 BTC
```
nkuba, line 83: Can we calculate it based on the funding deposit amount:

```suggestion
        request.treasuryFee = _depositTreasuryFeeDivisor > 0
            ? fundingOutputAmount / _depositTreasuryFeeDivisor
            : 0;
```

And make the divisor configurable:
```solidity
    function setDepositTreasuryFeeDivisor(uint64 value) external {
        _depositTreasuryFeeDivisor = value;
    }
```

solidity/contracts/integrator/TBTCDepositorProxy.sol
```diff
@@ -0,0 +1,309 @@
...
+        // revealed deposits stay there forever. The transaction will revert
+        // if the deposit has already been revealed so, there is no need to do
+        // an explicit check here.
+        bridge.revealDepositWithExtraData(fundingTx, reveal, extraData);
+
+        return depositKey;
+    }
+
+    /// @notice Finalizes a deposit by calculating the amount of TBTC minted
+    ///         for the deposit and calling the `onDepositFinalized` callback
```
nkuba, line 167: The callback part is no longer valid.

solidity/contracts/integrator/TBTCDepositorProxy.sol
```diff
@@ -0,0 +1,309 @@
...
+///          }
+///
+///          function finalizeProcess(uint256 depositKey) external {
+///              (uint256 tbtcAmount, bytes32 extraData) = _finalizeDeposit(depositKey);
+///
+///              // Do something with the minted TBTC using context
+///              // embedded in the extraData.
+///          }
+///      }
+abstract contract TBTCDepositorProxy {
```
pdyraga, line 73: Why not call it just `TBTCDepositor` or `AbstractTBTCDepositor`? The proxy suffix could suggest it is some sort of an upgradeable proxy contract.

solidity/contracts/integrator/TBTCDepositorProxy.sol
```diff
@@ -0,0 +1,309 @@
...
+    ///        deposit must be either swept or optimistically minted.
+    /// @dev IMPORTANT NOTE: The tbtcAmount returned by this function is an
+    ///      approximation. See documentation of the `calculateTbtcAmount`
+    ///      responsible for calculating this value for more details.
+    // slither-disable-next-line dead-code
+    function _finalizeDeposit(uint256 depositKey)
+        internal
+        returns (uint256 tbtcAmount, bytes32 extraData)
+    {
+        require(pendingDeposits[depositKey], "Deposit not initialized");
```
pdyraga, line 185: Do we need `pendingDeposits` mapping? I think it should be enough to check whether [the `deposit.depositor` value](https://github.com/keep-network/tbtc-v2/blob/609b6fedc9444688092f310c38c0e03758bfdbfb/solidity/contracts/bridge/Deposit.sol#L102) is non-zero or set to `address(this)`. We retrieve `IBridgeTypes.DepositRequest` anyway. Also, the implementation of the `TBTCDepositorProxy` is already tightly coupled with the implementation of `Bridge` and `TBTCVault`. We could save some gas especially since those functions will be called by a relayer bot and not by the users.

EDIT: No, the current version is correct. We do not want `_finalizeDeposit` to be called twice for the same deposit.

solidity/contracts/integrator/TBTCDepositorProxy.sol
```diff
@@ -0,0 +1,309 @@
...
+    /// @param depositTreasuryFeeSat Deposit treasury fee in satoshi (1e8 precision).
+    ///        This is an accurate value of the treasury fee that was actually
+    ///        cut upon minting.
+    /// @return tbtcAmount Approximate amount of TBTC minted for the deposit.
+    /// @dev IMPORTANT NOTE: The tbtcAmount returned by this function may
+    ///      not correspond to the actual amount of TBTC minted for the deposit.
+    ///      Although the treasury fee cut upon minting is known precisely,
+    ///      this is not the case for the optimistic minting fee and the Bitcoin
+    ///      transaction fee. To overcome that problem, this function just takes
+    ///      the current maximum values of both fees, at the moment of deposit
```
pdyraga, line 233:
```suggestion
    ///      the current maximum allowed values of both fees, at the moment of deposit
```

solidity/contracts/integrator/TBTCDepositorProxy.sol
pdyraga: I agree with @nkuba, this could be a common use case and we can avoid reading from the storage one more time if we return the value here.

pdyraga (commented):
I did a high-level review. The contract looks good to me with some potential minor tweaks already noted. I was not reviewing tests.

solidity/contracts/integrator/TBTCDepositorProxy.sol
nkuba: I proposed a change in https://github.com/keep-network/tbtc-v2/pull/779.

solidity/contracts/test/TestTBTCDepositorProxy.sol
nkuba: I proposed a change in https://github.com/keep-network/tbtc-v2/pull/779.

solidity/contracts/test/TestTBTCDepositorProxy.sol
nkuba: I proposed a change in https://github.com/keep-network/tbtc-v2/pull/779.

solidity/contracts/test/TestTBTCDepositorProxy.sol
lukasz-zimnoch: Leaving this for the integrator will give them more flexibility while writing tests. The second argument is that this abstract contract is unaware of the TBTC token and only pure numbers matter here.

solidity/contracts/integrator/TBTCDepositorProxy.sol
lukasz-zimnoch: Good catch. See db599050

solidity/contracts/integrator/TBTCDepositorProxy.sol
lukasz-zimnoch: > EDIT: No, the current version is correct. We do not want _finalizeDeposit to be called twice for the same deposit.

Exactly. Moreover, we don't want to push this responsibility to the child contract.

solidity/contracts/integrator/TBTCDepositorProxy.sol
lukasz-zimnoch: See a6f4a927

solidity/contracts/integrator/TBTCDepositorProxy.sol
lukasz-zimnoch: I wanted to avoid pure "depositor" to not mix it with the concept of "depositor" already present in the `Bridge`. I don't have strong preferences here though. `AbstractTBTCDepositor` is completely fine.

lukasz-zimnoch: See 68d02e86

nkuba (approved):

PR #779, opened 2024-02-01
Tweaks to `AbstractTBTCDepositor` contract

nkuba:
In this PR I propose some modifications to the initial implementation of `AbstractTBTCDepositor` contract (https://github.com/keep-network/tbtc-v2/pull/778).

### Return initial deposit amount from _initializeDeposit function

Integrators may be interested in adding more fees to the deposit amount. For this reason, it may be cleanest to calculate it based on the initial deposit amount. By returning the value here they will avoid the need to read the funding deposit amount from the storage again.

### Mock contracts tweaks

I added modifications to the MockBridge and MockTBTCVault contracts, that were useful in tests of the integrator's implementation (https://github.com/thesis/acre/pull/91).

### DIFF
--- a/solidity/contracts/integrator/TBTCDepositorProxy.sol
+++ b/solidity/contracts/integrator/TBTCDepositorProxy.sol
@@ -167,7 +167,10 @@ abstract contract TBTCDepositorProxy {
     ///         for the deposit and calling the `onDepositFinalized` callback
     ///         function.
     /// @param depositKey Deposit key identifying the deposit.
-    /// @return tbtcAmount Approximate amount of TBTC minted for the deposit.
+    /// @return initialDepositAmount Amount of funding transaction deposit. In
+    ///         TBTC token decimals precision.
+    /// @return tbtcAmount Approximate amount of TBTC minted for the deposit. In
+    ///         TBTC token decimals precision.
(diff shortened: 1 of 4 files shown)


solidity/contracts/integrator/TBTCDepositorProxy.sol
```diff
@@ -204,6 +211,8 @@ abstract contract TBTCDepositorProxy {
         // slither-disable-next-line reentrancy-no-eth
         delete pendingDeposits[depositKey];
 
+        initialDepositAmount = deposit.amount * SATOSHI_MULTIPLIER;
```
lukasz-zimnoch, line 214: Let's update the usage example from the contract's docstring accordingly. Right now, there is an obsolete version with `_finalizeDeposit` returning two parameters.

lukasz-zimnoch (commented):
LGTM! Left one minor comment.

solidity/contracts/integrator/TBTCDepositorProxy.sol
nkuba: 4d74ad9fa082a4e2cf0a9a91d45d5b6b03a0e556

PR #783, opened 2024-02-08
Don't track pending deposits in AbstractTBTCDepositor contract

nkuba:
To optimize gas usage we removed `pendingDeposits` mapping. This operation reduces gas usage of `_initializeDeposit` function by ~20k gas.

After this change it is implementation's contract responsibility to ensure `_finalizeDeposit` function cannot be called twice for the same deposit.

`_initializeDeposit` function is not affected as the bridge blocks possibility of revealing the same deposit twice.

### DIFF
--- a/solidity/contracts/integrator/AbstractTBTCDepositor.sol
+++ b/solidity/contracts/integrator/AbstractTBTCDepositor.sol
@@ -64,6 +64,9 @@ import "./ITBTCVault.sol";
 ///          }
 ///
 ///          function finalizeProcess(uint256 depositKey) external {
+///              // Ensure the function cannot be called for the same deposit
+///              // twice.
+///
 ///              (
 ///                  uint256 initialDepositAmount,
 ///                  uint256 tbtcAmount,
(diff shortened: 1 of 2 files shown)


lukasz-zimnoch (approved):

solidity/contracts/integrator/AbstractTBTCDepositor.sol
```diff
@@ -180,6 +178,9 @@ abstract contract AbstractTBTCDepositor {
     ///        (in the context of this contract) yet.
     ///      - The deposit must be finalized on the Bridge side. That means the
     ///        deposit must be either swept or optimistically minted.
+    /// @dev This function doesn't validate if a deposit has been finalized before,
+    ///      it is a responsibility of the implementing contract to ensure this
+    ///      function won't be called twice for the same deposit.
```
pdyraga, line 183: I would CAPSLOCK it.

nkuba: https://github.com/keep-network/tbtc-v2/pull/784

Issue #781, opened 2024-02-05
Implement basic redemptions veto mechanism (TIP-072)


lukasz-zimnoch:
This task is about implementing the initial version of the redemptions veto mechanism, as specified in [TIP-072: Optimistic redemptions](https://forum.threshold.network/t/tip-072-optimistic-redemptions/793). The scope of this task includes all necessary contract-side changes that will allow guardians to issue objections. The scope DOES NOT include guardian automation.
```[tasklist]
### Tasks
- [ ] https://github.com/keep-network/tbtc-v2/pull/788
- [ ] https://github.com/keep-network/keep-core/pull/3788
```

PR #788, opened 2024-02-20
Redemptions veto mechanism

lukasz-zimnoch:
Closes: https://github.com/keep-network/tbtc-v2/issues/781

Here we present the implementation of the redemptions veto mechanism, specified in the [TIP-072: Optimistic redemptions](https://forum.threshold.network/t/tip-072-optimistic-redemptions/793). 

### High-level architecture

The current byte size of the `Bridge` contract (~22 kB) is close to the limit. To leave some space for future `Bridge` upgrades, we decided to encapsulate the logic of the redemption veto mechanism in a separate upgradeable contract called `RedemptionWatchtower`, and attach it to the existing `Bridge` instance. Such a design has a positive impact on the `Bridge` size, nicely separates the concerns, and reduces the amount of direct changes in the living `Bridge` instance.
 
The `RedemptionWatchtower` contract acts as a central point for redemption guardians, manages vetoed redemptions, and provides other functionality specified in the TIP. It also serves as a source of information about vetoed redemptions, banned redeemers, and processing delays that should be preserved for specific redemption requests.

The chart below presents how the `RedemptionWatchtower` fits into the existing architecture and provides an overview of interactions between relevant smart contracts and system actors:

```
                                +---------------+
                                |               |
                      +---------+    Wallets    +-------------+
                      |         |               |             |
   pendingRedemptions |         +---------------+             | validateRedemptionProposal
                      |                                       |
                      |                                       |
                      v                                       v
+--------------------------+                          +-----------------------------+
|                          |     pendingRedemptions   |                             |
|          Bridge          |<-------------------------+   WalletProposalValidator   |
|                          |                          |                             |
+----+---------------------+                          +-------+---------------------+
     |                ^                                       |
     |                |                                       |
     |                |                                       |
     |                |                                       | getRedemptionDelay
     |                | notifyRedemptionVeto                  |
     |                +---------+                             |
     |                          |                             |
     |                          |                             |
     | isSafeRedemption    +----+---------------------+       |
     +-------------------->|                          |<------+
                           |   RedemptionWatchtower   |
                  +------->|                          |<------+
                  |        +--------------------------+       |
                  |                     ^                     |
                  |                     |                     |
                  |                     |                     |
   raiseObjection |                     | management          | security
                  |                     | actions             | actions
                  |                     |                     |
                  |                     |                     |
          +-------+-------+     +-------+-------+     +-------+-------+
          |               |     |               |     |               |
          |   Guardians   |     |    Manager    |     |     Owner     |
          |               |     |               |     |               |
          +---------------+     +---------------+     +---------------+
```

### The `RedemptionWatchtower` contract

The `RedemptionWatchtower` contract interacts with several actors of the protocol. Each actor has a different set of capabilities:

**Owner (Threshold Council)**

The owner can call the following functions:
- `enableWatchtower` which enables the veto mechanism for the first time. This function must set the watchtower's manager and can establish an initial set of guardians. This function also captures the timestamp of the call. This information is necessary to ensure the proper lifecycle of the veto mechanism (shut down after 18 months). It is also crucial for determining stalled redemptions that were created in the pre-veto era and can be vetoed indefinitely.
- `removeGuardian` that can be used to remove specific guardians. This function is a safeguard against malicious guardians hence, it must be directly available to the owner. 

**Manager (Token holder DAO)**

As per the TIP, most of the governance should be the authority of the Token holder DAO. To make it possible, we are introducing the role of the watchtower's manager who can use the following functions:
- `addGuardian` which allows adding new redemption guardians
- `updateWatchtowerParameters` that can update all governable parameters that steer the veto mechanism. Those parameters are redemption veto delays, veto time and financial penalties, and the mechanism lifetime.
- `unban` which can be used to unban redeemers who were banned mistakenly. This is a safeguard for guardian mistakes that must be in place to protect honest redeemers.

**Guardians**

Guardians are the executors of the veto process. The only function they can call is `raiseObjection`. That function should be used to raise objections against specific redemption requests. Three subsequent objections lead to a redemption veto. Vetoed redemptions are not processed by the protocol, the requested amount is diminished by a penalty fee, and frozen for a specific time. Once the freeze period ends, the redeemer can claim those funds back. Moreover, redeemers whose redemptions were vetoed become banned and cannot ask for redemptions in the future. Last but not least, even a single objection (not leading to a veto) against a redemption (from a given wallet to a given BTC address) prevents asking the same wallet to redeem to the same BTC address in the future (this is similar to timed out redemptions).

**Others**

The `RedemptionWatchtower` exposes some functions available to the broad public. Those are:
- `isSafeRedemption` which determines whether a redemption involving a specific wallet, BTC address, Bank's balance owner, and redeemer can be considered as safe. This function leverages the objections history and banned redeemers to determine so. A redemption is considered safe if neither the balance owner nor redeemer is banned and past redemptions from the given wallet to the given BTC address were not subject to guardian objections. 
- `getRedemptionDelay` that informs tBTC wallets about processing delays that should be preserved for specific redemption requests. Delays are determined based on raised objections.
- `withdrawVetoedFunds` which can be used by redeemers to withdraw funds from vetoed redemptions once the freeze period ends. 
- `disableWatchtower` that can be used by anyone to disable the veto mechanism once the watchtower's lifetime elapses.

### Changes in the `Bridge` contract

Integration of the `RedemptionWatchtower` contract with the existing `Bridge` incurs several changes in the latter. 

First and foremost, the `Bridge`'s state gains a new address field (`redemptionWatchtower`) that points to the `RedemptionWatchtower` contract. That field can be set using the new `setRedemptionWatchtower` function available for the `Bridge`'s governance. 

Secondly, the `Bridge` exposes a new `notifyRedemptionVeto` callback function. This function can be called only by the `redemptionWatchtower` address. The main responsibility of this function is propagating the consequences of a redemption veto to the `Bridge`'s state, namely:
- Reduce the pending redemptions value of the target wallet (supposed to handle the vetoed redemption). This is the same action as performed upon redemption timeout. This must be done because otherwise, the wallet would hold the reserve for a redemption request that would never be processed
- Delete the redemption request from the pending redemptions mapping. This is important to avoid the vetoed redemption request being processed by the wallet or reported as timed out
- Detain the redemption's request amount and transfer it under the control of the redemption watchtower for further processing (i.e. burn the penalty fee and allow withdrawal of the rest after the freeze period)

Last but not least, the `Bridge` leverages the `redemptionWatchtower` address to determine the safety of upcoming redemption requests. This is achieved using the `isSafeRedemption` call. This way, redemptions from banned redeemers (and balance owners) are blocked. The same applies to redemptions from specific wallets to specific BTC addresses which were subject to guardian objections in the past.

### Changes in the `WalletProposalValidator` contract

The last piece of the puzzle is the integration of the veto mechanism with the off-chain tBTC wallets. Wallets must respect the new veto delay and not process redemptions during the period a redemption veto can still land. According to [RFC-12](https://github.com/keep-network/tbtc-v2/blob/main/docs/rfc/rfc-12.adoc#224-introduce-the-walletproposalvalidator-contract), the tBTC wallets always consult the `WalletProposalValidator` to validate the ongoing redemption proposal before issuing the redemption transaction on the Bitcoin chain. That means it is enough to modify the `WalletProposalValidator.validateRedemptionProposal` function and take the veto delay into account there. The `validateRedemptionProposal` function calls the `RedemptionWatchtower.getRedemptionDelay` function for each redemption in the proposal and considers it valid only if the returned veto delay (counted since the redemption creation timestamp) already elapsed. This logic guarantees that a particular redemption proposal can only be considered valid if all of its redemption requests have surpassed their respective veto delay periods. 

In light of the above, we should also prevent invalid redemption proposals (with redemptions violating the veto delay) from being issued by RFC-12 coordinators. An invalid proposal means a coordination window miss and the optimal strategy here is ensuring proposal validity at generation time. This requires adjustments in the [off-chain redemption proposal generator logic](https://github.com/keep-network/keep-core/blob/2a76f4de820eb13b2d49c67a15a59eac7552fa26/pkg/tbtcpg/redemptions.go#L70). This work is beyond the scope of this pull request and will be addressed in a follow-up PR.

The ultimate protection against wallets not obeying the aforementioned rules is the fact that vetoed redemptions are removed from the `pendingRedemption` set in the `Bridge`. That means the `Bridge` will reject an SPV proof of a redemption transaction that handles at least one vetoed redemption request (that happens in the `submitRedemptionProof` function). In such a case, the given wallet will be deemed as fraudulent and punished according to the protocol rules.

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -1957,6 +1957,7 @@ contract Bridge is
         external
         onlyGovernance
     {
+        // The internal function is defined in the `BridgeState` library.
         self.setRedemptionWatchtower(redemptionWatchtower);
     }
 
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -1961,7 +1961,7 @@ contract Bridge is
     }
 
     /// @return Address of the redemption watchtower.
-    function redemptionWatchtower() external view returns (address) {
+    function getRedemptionWatchtower() external view returns (address) {
         return self.redemptionWatchtower;
     }
 
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -1964,4 +1964,29 @@ contract Bridge is
     function redemptionWatchtower() external view returns (address) {
         return self.redemptionWatchtower;
     }
+
+    /// @notice Notifies that a redemption request was vetoed in the watchtower.
+    ///         This function is responsible for adjusting the Bridge's state
+    ///         accordingly.
+    ///         The results of calling this function:
+    ///         - the pending redemptions value for the wallet is decreased
+    ///           by the requested amount (minus treasury fee),
+    ///         - the request is removed from pending redemptions mapping,
+    ///         - the tokens taken from the redeemer on redemption request are
+    ///           detained and passed to the redemption watchtower
+    ///           (as Bank's balance) for further processing.
+    /// @param walletPubKeyHash 20-byte public key hash of the wallet.
+    /// @param redeemerOutputScript  The redeemer's length-prefixed output
+    ///        script (P2PKH, P2WPKH, P2SH or P2WSH).
+    /// @dev Requirements:
+    ///      - The caller must be the redemption watchtower,
+    ///      - The redemption request identified by `walletPubKeyHash` and
+    ///        `redeemerOutputScript` must exist.
+    function notifyRedemptionVeto(
+        bytes20 walletPubKeyHash,
+        bytes calldata redeemerOutputScript
+    ) external {
+        // The caller is checked in the internal function.
+        self.notifyRedemptionVeto(walletPubKeyHash, redeemerOutputScript);
+    }
 }
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -236,6 +236,8 @@ contract Bridge is
 
     event TreasuryUpdated(address treasury);
 
+    event RedemptionWatchtowerSet(address redemptionWatchtower);
+
     modifier onlySpvMaintainer() {
         require(
             self.isSpvMaintainer[msg.sender],
@@ -1943,5 +1945,23 @@ contract Bridge is
     ///         successfully evaluate an SPV proof.
     function txProofDifficultyFactor() external view returns (uint256) {
         return self.txProofDifficultyFactor;
+    }
+
+    /// @notice Sets the redemption watchtower address.
+    /// @param redemptionWatchtower Address of the redemption watchtower.
+    /// @dev Requirements:
+    ///      - The caller must be the governance,
+    ///      - Redemption watchtower address must not be already set,
+    ///      - Redemption watchtower address must not be 0x0.
+    function setRedemptionWatchtower(address redemptionWatchtower)
+        external
+        onlyGovernance
+    {
+        self.setRedemptionWatchtower(redemptionWatchtower);
+    }
+
+    /// @return Address of the redemption watchtower.
+    function redemptionWatchtower() external view returns (address) {
+        return self.redemptionWatchtower;
     }
 }
--- a/solidity/contracts/bridge/Redemption.sol
+++ b/solidity/contracts/bridge/Redemption.sol
@@ -402,6 +402,8 @@ library Redemption {
         bytes memory redeemerOutputScript,
         uint64 amount
     ) internal {
+        // TODO: Validate the request against the RedemptionWatchtower.
+
         Wallets.Wallet storage wallet = self.registeredWallets[
             walletPubKeyHash
         ];
@@ -1074,5 +1076,65 @@ library Redemption {
             key := keccak256(0, 52)
         }
         return key;
+    }
+
+    /// @notice Notifies that a redemption request was vetoed in the watchtower.
+    ///         This function is responsible for adjusting the Bridge's state
+    ///         accordingly.
+    ///         The results of calling this function:
+    ///         - the pending redemptions value for the wallet is decreased
+    ///           by the requested amount (minus treasury fee),
+    ///         - the request is removed from pending redemptions mapping,
+    ///         - the tokens taken from the redeemer on redemption request are
+    ///           detained and passed to the redemption watchtower
+    ///           (as Bank's balance) for further processing.
+    /// @param walletPubKeyHash 20-byte public key hash of the wallet.
+    /// @param redeemerOutputScript  The redeemer's length-prefixed output
+    ///        script (P2PKH, P2WPKH, P2SH or P2WSH).
+    /// @dev Requirements:
+    ///      - The caller must be the redemption watchtower,
+    ///      - The redemption request identified by `walletPubKeyHash` and
+    ///        `redeemerOutputScript` must exist.
+    function notifyRedemptionVeto(
+        BridgeState.Storage storage self,
+        bytes20 walletPubKeyHash,
+        bytes calldata redeemerOutputScript
+    ) external {
+        require(
+            msg.sender == self.redemptionWatchtower,
+            "Caller is not the redemption watchtower"
+        );
+
+        uint256 redemptionKey = getRedemptionKey(
+            walletPubKeyHash,
+            redeemerOutputScript
+        );
+        Redemption.RedemptionRequest storage redemption = self
+            .pendingRedemptions[redemptionKey];
+
+        // Should never happen, but just in case.
+        require(
+            redemption.requestedAt != 0,
+            "Redemption request does not exist"
+        );
+
+        // Update the wallet's pending redemptions value. This is the
+        // same logic as performed upon redemption request timeout.
+        // If we don't do this, the wallet will hold the reserve
+        // for a redemption request that will never be processed.
+        self.registeredWallets[walletPubKeyHash].pendingRedemptionsValue -=
+            redemption.requestedAmount -
+            redemption.treasuryFee;
+
+        // Capture the amount that should be transferred to the
+        // redemption watchtower.
+        uint64 detainedAmount = redemption.requestedAmount;
+
+        // Delete the redemption request from the pending redemptions
+        // mapping. This is important to avoid this redemption request
+        // to be processed by the wallet or reported as timed out.
+        delete self.pendingRedemptions[redemptionKey];
+
+        self.bank.transferBalance(self.redemptionWatchtower, detainedAmount);
     }
 }
--- a/solidity/contracts/bridge/Redemption.sol
+++ b/solidity/contracts/bridge/Redemption.sol
@@ -43,6 +43,16 @@ interface IRedemptionWatchtower {
         address balanceOwner,
         address redeemer
     ) external view returns (bool);
+
+    /// @notice Returns the applicable redemption delay for a redemption
+    ///         request identified by the given redemption key.
+    /// @param redemptionKey Redemption key built as
+    ///        `keccak256(keccak256(redeemerOutputScript) | walletPubKeyHash)`.
+    /// @return Redemption delay.
+    function getRedemptionDelay(uint256 redemptionKey)
+        external
+        view
+        returns (uint32);
 }
 
 /// @notice Aggregates functions common to the redemption transaction proof
--- a/solidity/contracts/bridge/Redemption.sol
+++ b/solidity/contracts/bridge/Redemption.sol
@@ -24,6 +24,27 @@ import "./Wallets.sol";
 
 import "../bank/Bank.sol";
 
+/// @notice Interface of the RedemptionWatchtower.
+interface IRedemptionWatchtower {
+    /// @notice Determines whether a redemption request is considered safe.
+    /// @param walletPubKeyHash 20-byte public key hash of the wallet that
+    ///        is meant to handle the redemption request.
+    /// @param redeemerOutputScript The redeemer's length-prefixed output
+    ///        script (P2PKH, P2WPKH, P2SH or P2WSH) that is meant to
+    ///        receive the redeemed amount.
+    /// @param balanceOwner The address of the Bank balance owner whose balance
+    ///        is getting redeemed.
+    /// @param redeemer The address that requested the redemption.
+    /// @return True if the redemption request is safe, false otherwise.
+    ///         Specific safety criteria depend on the implementation.
+    function isSafeRedemption(
+        bytes20 walletPubKeyHash,
+        bytes calldata redeemerOutputScript,
+        address balanceOwner,
+        address redeemer
+    ) external view returns (bool);
+}
+
 /// @notice Aggregates functions common to the redemption transaction proof
 ///         validation and to the moving funds transaction proof validation.
 library OutboundTx {
@@ -402,7 +423,18 @@ library Redemption {
         bytes memory redeemerOutputScript,
         uint64 amount
     ) internal {
-        // TODO: Validate the request against the RedemptionWatchtower.
+        if (self.redemptionWatchtower != address(0)) {
+            require(
+                IRedemptionWatchtower(self.redemptionWatchtower)
+                    .isSafeRedemption(
+                        walletPubKeyHash,
+                        redeemerOutputScript,
+                        balanceOwner,
+                        redeemer
+                    ),
+                "Redemption request rejected by the watchtower"
+            );
+        }
 
         Wallets.Wallet storage wallet = self.registeredWallets[
             walletPubKeyHash
--- a/solidity/contracts/bridge/Redemption.sol
+++ b/solidity/contracts/bridge/Redemption.sol
@@ -34,7 +34,15 @@ interface IRedemptionWatchtower {
     ///        receive the redeemed amount.
     /// @param balanceOwner The address of the Bank balance owner whose balance
     ///        is getting redeemed.
-    /// @param redeemer The address that requested the redemption.
+    /// @param redeemer The address that requested the redemption and will be
+    ///        able to claim Bank balance if anything goes wrong during the
+    ///        redemption. In the most basic case, when someone redeems their
+    ///        Bitcoin balance from the Bank, `balanceOwner` is the same
+    ///        as `redeemer`. However, when a Vault is redeeming part of its
+    ///        balance for some redeemer address (for example, someone who has
+    ///        earlier deposited into that Vault), `balanceOwner` is the Vault,
+    ///        and `redeemer` is the address for which the vault is redeeming
+    ///        its balance to.
     /// @return True if the redemption request is safe, false otherwise.
     ///         Specific safety criteria depend on the implementation.
     function isSafeRedemption(
@@ -413,6 +421,8 @@ library Redemption {
     ///        `amount - (amount / redemptionTreasuryFeeDivisor) - redemptionTxMaxFee`.
     ///        Fees values are taken at the moment of request creation.
     /// @dev Requirements:
+    ///      - If the redemption watchtower is set, the redemption request must
+    ///        be considered safe by the watchtower,
     ///      - Wallet behind `walletPubKeyHash` must be live,
     ///      - `mainUtxo` components must point to the recent main UTXO
     ///        of the given wallet, as currently known on the Ethereum chain,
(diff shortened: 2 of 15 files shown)


solidity/contracts/bridge/Bridge.sol
```diff
@@ -1944,4 +1946,47 @@ contract Bridge is
...
+    /// @param redemptionWatchtower Address of the redemption watchtower.
+    /// @dev Requirements:
+    ///      - The caller must be the governance,
+    ///      - Redemption watchtower address must not be already set,
+    ///      - Redemption watchtower address must not be 0x0.
+    function setRedemptionWatchtower(address redemptionWatchtower)
+        external
+        onlyGovernance
+    {
+        self.setRedemptionWatchtower(redemptionWatchtower);
```
theref, line 1960: for those of us not super familiar with the codebase, it would be handy to have a comment that highlights that this function is defined in `BridgeState.sol`

solidity/contracts/bridge/RedemptionWatchtower.sol
```diff
@@ -0,0 +1,469 @@
...
+
+        emit WatchtowerEnabled(enabledAt, _manager);
+    }
+
+    /// @notice Adds a redemption guardian
+    /// @param guardian Address of the guardian to add.
+    /// @dev Requirements:
+    ///      - The caller must be the watchtower manager,
+    ///      - The guardian must not already exist.
+    function addGuardian(address guardian) external onlyManager {
```
theref, line 207: what happens if the Watchtower is not yet enabled and a Manager has not yet been set?

solidity/contracts/bridge/RedemptionWatchtower.sol
```diff
@@ -0,0 +1,469 @@
...
+        if (isBanned[redeemer]) {
+            return false;
+        }
+
+        uint256 redemptionKey = Redemption.getRedemptionKey(
+            walletPubKeyHash,
+            redeemerOutputScript
+        );
+
+        if (vetoProposals[redemptionKey].objectionsCount > 0) {
```
theref, line 463: does this not need to be `>= 0`?

theref (commented):
What's the plan with `disableWatchtower`? if the DAO let's the timeout pass then any user can disable the check?
if the DAO wanted to re-enable later on, would a new Watchtower have to be deployed?

lukasz-zimnoch:
> What's the plan with `disableWatchtower`? if the DAO let's the timeout pass then any user can disable the check? if the DAO wanted to re-enable later on, would a new Watchtower have to be deployed?

This is the part I just added, please see: https://github.com/keep-network/tbtc-v2/pull/788/commits/8d7d51b7eb6b3a3f461684de057f43e50cb0934c. This switch is kind of a social contract and disables the watchtower permanently by design. The DAO can vote to increase `watchtowerLifetime` if needed. If Threshold Council decides to re-enable a disabled watchtower, we can always add this ability in a future contract upgrade.

solidity/contracts/bridge/Bridge.sol
lukasz-zimnoch: Good point! Please see 743e9c75

solidity/contracts/bridge/RedemptionWatchtower.sol
lukasz-zimnoch: In that case, `addGuardian` cannot be called by anyone. There are no guardians in the system and the veto mechanism is not functioning. The contract is designed in a way that `enableWatchtower` must be called first.

The deployment scenario will be as follows:
1. Deploy the new implementation of the `Bridge` contract and set the transparent proxy to point there
2. Deploy the `RedemptionWatchtower` contract
3. Threshold Council calls `enableWatchtower` with `manager` being the [Governor Bravo `TimelockController`](https://etherscan.io/address/0x87F005317692D05BAA4193AB0c961c69e175f45f) and some initial `guardians`. 
4. Threshold Council attaches the `RedemptionWatchtower` to the `Bridge` via `setRedemptionWatchtower`

solidity/contracts/bridge/RedemptionWatchtower.sol
lukasz-zimnoch: I don't think so. `vetoProposals[redemptionKey].objectionsCount == 0` is the case for non-objected requests (they don't have the corresponding entry in the `vetoProposals` mapping). In such a case, `isSafeRedemption` must return `true` (assuming the balance owner and redeemer are not banned). If we use `>=0`, `isSafeRedemption` would return `false` for non-objected redemptions which is wrong.

theref: ah sorry, I was reading this backwards! :D

solidity/contracts/bridge/RedemptionWatchtower.sol
lukasz-zimnoch: Can you describe such a scenario in points? I'm not sure if this is possible before calling `enableWatchtower`.

theref: If i'm paying careful attention, I will see the `RedemptionWatchtower` being deployed. Soon after (but not instant right?), the council will call `enableWatchtower` with a bunch of initial config, which includes setting the `manager`. My question is, during that small window with no `manager` set, is a malicious actor able to call this `addGuardian` function successfully?

lukasz-zimnoch: Ok I see, thanks for clarification! I think it's not possible because before calling `enableWatchtower` the `manager` will be `address(0)`. When someone calls `addGuardian`, we will jump to the `onlyManager` modifier that will check `msg.sender != address(0)` and revert. I added a test case to prove that: 6b0a76ab

theref: ok yeah, I can see how the modifier handles this. Thanks for adding the test though

solidity/contracts/bridge/RedemptionWatchtower.sol
```diff
@@ -0,0 +1,618 @@
...
+        uint32 levelTwoDelay,
+        uint64 waivedAmountLimit
+    );
+
+    event Banned(address indexed redeemer);
+
+    event Unbanned(address indexed redeemer);
+
+    event VetoedFundsWithdrawn(
+        uint256 indexed redemptionkey,
```
tomaszslabon, line 160:
```suggestion
        uint256 indexed redemptionKey,
```

solidity/contracts/bridge/WalletProposalValidator.sol
```diff
@@ -608,10 +612,22 @@ contract WalletProposalValidator {
...
+                if (delay > minAge) {
+                    minAge = delay;
+                }
+            }
+
             require(
                 /* solhint-disable-next-line not-rely-on-time */
-                block.timestamp >
-                    redemptionRequest.requestedAt + REDEMPTION_REQUEST_MIN_AGE,
+                block.timestamp > redemptionRequest.requestedAt + minAge,
```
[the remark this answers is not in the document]
lukasz-zimnoch: > Are we sure this will not lead to a situation where somebody calls notifyWalletRedemptionTimeout for a redemption that the wallet couldn't process because it had some objections?

Good point! This is not a problem as long as the redemption timeout is greater than the watchtower's level-two delay. In such a case, even if a redemption gets two objections, the wallet can process it before the timeout can be reported.

tomaszslabon (dismissed):

solidity/contracts/bridge/RedemptionWatchtower.sol
```diff
@@ -0,0 +1,618 @@
...
+        }
+
+        uint256 redemptionKey = Redemption.getRedemptionKey(
+            walletPubKeyHash,
+            redeemerOutputScript
+        );
+
+        if (vetoProposals[redemptionKey].objectionsCount > 0) {
+            return false;
+        }
```
vzotova, line 571: question regarding this block, what happens in case only one or two objections were raised and then delay is expired? 
I guess in that case redemption is still not safe but also new objections can't be raised?

vzotova (commented):
The first couple of questions on the way to understanding

solidity/contracts/bridge/RedemptionWatchtower.sol
lukasz-zimnoch: Such a redemption request is processed by the tBTC wallet after the delay. However, because there were objections to this redemption key (wallet + BTC address pair), the protocol blocks future redemptions from the same wallet to the given BTC address, just in case. Doing this is also easier from the redemption lifecycle management perspective and is consistent with what happens upon redemption timeout. That said:
- In case of the happy path (redemption processed without troubles) we allow reusing the given redemption key and ask the same wallet to redeem to the same BTC address
- In case of an unhappy path (objections raised, timeout hit) we block further usage of the given redemption key

I'm aware this may not be intuitive but this is a consequence of how tBTC identifies redemptions.

solidity/contracts/bridge/RedemptionWatchtower.sol
```diff
@@ -0,0 +1,618 @@
...
+        bytes calldata redeemerOutputScript
+    ) external onlyGuardian {
+        uint256 redemptionKey = Redemption.getRedemptionKey(
+            walletPubKeyHash,
+            redeemerOutputScript
+        );
+
+        VetoProposal storage veto = vetoProposals[redemptionKey];
+
+        uint8 requiredObjectionsCount = 3;
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, I missed it, good catch! Please see: 2556c282

solidity/contracts/bridge/RedemptionWatchtower.sol
vzotova: Thank you for the explanation, I've found code where redemption handled after delay, could you point how redeemer banned in that case (one or two objections)?

lukasz-zimnoch: The redeemer is banned only after a full veto (3 objections). We don't ban after 1 or 2 objections.

vzotova: That I saw, whole question is about 1 or 2 objections, got it

solidity/contracts/bridge/Bridge.sol
```diff
@@ -1944,4 +1946,48 @@ contract Bridge is
     function txProofDifficultyFactor() external view returns (uint256) {
         return self.txProofDifficultyFactor;
     }
+
+    /// @notice Sets the redemption watchtower address.
+    /// @param redemptionWatchtower Address of the redemption watchtower.
+    /// @dev Requirements:
+    ///      - The caller must be the governance,
+    ///      - Redemption watchtower address must not be already set,
```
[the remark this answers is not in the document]
lukasz-zimnoch: This is done on purpose to provide a kind of social contract regarding how the mechanism works. We do not expect to change anything in the mechanism itself. Any security upgrade would likely require a `Bridge` upgrade anyway.

solidity/contracts/bridge/Redemption.sol
```diff
@@ -1075,4 +1119,64 @@ library Redemption {
...
+        // same logic as performed upon redemption request timeout.
+        // If we don't do this, the wallet will hold the reserve
+        // for a redemption request that will never be processed.
+        self.registeredWallets[walletPubKeyHash].pendingRedemptionsValue -=
+            redemption.requestedAmount -
+            redemption.treasuryFee;
+
+        // Capture the amount that should be transferred to the
+        // redemption watchtower.
+        uint64 detainedAmount = redemption.requestedAmount;
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yep, this is correct. I will drop a comment.

solidity/deploy/41_transfer_redemption_watchtower_ownership.ts
```diff
@@ -0,0 +1,19 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types"
```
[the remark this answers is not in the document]
lukasz-zimnoch: > Where do we set the Token Holder DAO as the watchtower manager?

This will be done through the `enableWatchtower` call after deployment.

> Also, is the newly deployed watchtower proxy attached to the already existing ProxyAdmin instance?

Yes, this is the plan.

vzotova (approved):

solidity/contracts/bridge/WalletProposalValidator.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Addressed as part of c6c03a373ac60cd6195b3216da6665e569d1e386

solidity/contracts/bridge/Redemption.sol
```diff
@@ -24,6 +24,37 @@ import "./Wallets.sol";
...
+interface IRedemptionWatchtower {
+    /// @notice Determines whether a redemption request is considered safe.
+    /// @param walletPubKeyHash 20-byte public key hash of the wallet that
+    ///        is meant to handle the redemption request.
+    /// @param redeemerOutputScript The redeemer's length-prefixed output
+    ///        script (P2PKH, P2WPKH, P2SH or P2WSH) that is meant to
+    ///        receive the redeemed amount.
+    /// @param balanceOwner The address of the Bank balance owner whose balance
+    ///        is getting redeemed.
+    /// @param redeemer The address that requested the redemption.
```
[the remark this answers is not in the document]
lukasz-zimnoch: Addressed as part of c6c03a373ac60cd6195b3216da6665e569d1e386

solidity/contracts/bridge/Redemption.sol
```diff
@@ -402,6 +433,19 @@ library Redemption {
         bytes memory redeemerOutputScript,
         uint64 amount
     ) internal {
+        if (self.redemptionWatchtower != address(0)) {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Addressed as part of c6c03a373ac60cd6195b3216da6665e569d1e386

solidity/contracts/bridge/Redemption.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Addressed as part of c6c03a373ac60cd6195b3216da6665e569d1e386

solidity/contracts/bridge/WalletProposalValidator.sol
```diff
@@ -608,10 +612,22 @@ contract WalletProposalValidator {
...
+            if (redemptionWatchtower != address(0)) {
+                // Check the redemption delay enforced by the watchtower.
+                // slither-disable-next-line calls-loop
+                uint32 delay = IRedemptionWatchtower(redemptionWatchtower)
+                    .getRedemptionDelay(redemptionKey);
+                // If the delay is greater than the usual minimum age, use it.
+                // This way both the min age and the watchtower delay are preserved.
+                if (delay > minAge) {
+                    minAge = delay;
+                }
```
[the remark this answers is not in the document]
lukasz-zimnoch: Addressed as part of c6c03a373ac60cd6195b3216da6665e569d1e386

solidity/contracts/bridge/RedemptionWatchtower.sol
```diff
@@ -0,0 +1,619 @@
...
+        // - Objections against a redemption request created AFTER the
+        //   `watchtowerEnabledAt` timestamp can be raised only within
+        //   a certain time frame defined by the redemption delay.
+        // - Objections against a redemption request created BEFORE the
+        //   `watchtowerEnabledAt` timestamp can be raised without
+        //   any time restrictions.
+        if (redemption.requestedAt >= watchtowerEnabledAt) {
+            require(
+                // Use < instead of <= to avoid a theoretical edge case
+                // where the delay is 0 (veto disabled) but three objections
```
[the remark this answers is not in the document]
lukasz-zimnoch: Addressed as part of c6c03a373ac60cd6195b3216da6665e569d1e386

solidity/contracts/bridge/RedemptionWatchtower.sol
```diff
@@ -0,0 +1,619 @@
...
+        );
+        require(!objections[objectionKey], "Guardian already objected");
+
+        Redemption.RedemptionRequest memory redemption = bridge
+            .pendingRedemptions(redemptionKey);
+
+        require(
+            redemption.requestedAt != 0,
+            "Redemption request does not exist"
+        );
```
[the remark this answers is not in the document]
lukasz-zimnoch: Addressed as part of c6c03a373ac60cd6195b3216da6665e569d1e386

vzotova (approved):

tomaszslabon (approved):

PR #790, opened 2024-02-22
Remove support for Goerli testnet

michalinacienciala:
As the Goerli testnet became deprecated with end of year 2023 and we've already migrated to Sepolia testnet where we could, we no longer need the Goerli-related code. This also applies to Goerli-based L2s.

This is a second stage of the work started in https://github.com/keep-network/tbtc-v2/pull/755.

Ref: https://github.com/threshold-network/solidity-contracts/issues/163.


lionakhnazarov:
Closing — Goerli has been deprecated network-wide and testnet support moved to Sepolia;

PR #791, opened 2024-02-27
Expose deposit amounts in AbstractDepositorContract

nkuba:
In this PR we introduce two changes to the AbstractDepositorContract around the deposit amounts.

### Expose minimum deposit amount
Integrators may need a minimum deposit amount to introduce validation if the amount the user wants to deposit meets the Bridge requirement of a deposit to exceed the dust threshold.
Originally proposed in https://github.com/thesis/acre/pull/253#discussion_r1500715013.

### Return initial deposit amount from _initializeDeposit function

The amount may be useful for integrators for validations of the deposit amount. Originally proposed in https://github.com/thesis/acre/pull/253#discussion_r1500680454.
    
It turns out that this version where we read from the storage:
```solidity
        initialDepositAmount =
            bridge.deposits(depositKey).amount *
            SATOSHI_MULTIPLIER;
```
 costs `117 366` gas.
Which is less than an alternative approach:
```solidity
            initialDepositAmount =
                fundingTx
                    .outputVector
                    .extractOutputAtIndex(reveal.fundingOutputIndex)
                    .extractValue() *
                SATOSHI_MULTIPLIER;
```
which costs `117 601` gas.

We return `initialDepositAmount` from two functions: `_initializeDeposit` and `_finalizeDeposit`. I tried to introduce a getter function:
```solidity
    function _getInitialDepositAmount(
        uint256 depositKey
    ) internal view returns (uint256) {
        IBridgeTypes.DepositRequest memory deposit = bridge.deposits(
            depositKey
        );
        require(deposit.revealedAt != 0, "Deposit not initialized");

        return deposit.amount * SATOSHI_MULTIPLIER;
    }
```
and removed `initialDepositAmount` return from `_initializeDeposit` and `_finalizeDeposit` functions.
Unfortunately, the overall cost in the reference BitcoinDepositor implementation from https://github.com/thesis/acre/pull/253 wasn't improved:
BEFORE:
```
+ initializeStake --> 143 004
+ finalizeStake --> 195 178
= total --> 338 182
```
AFTER:
```
+ initializeStake --> 142 912
+ finalizeStake --> 197 960
= total --> 340 872
```

### DIFF
--- a/solidity/contracts/integrator/AbstractTBTCDepositor.sol
+++ b/solidity/contracts/integrator/AbstractTBTCDepositor.sol
@@ -123,6 +123,8 @@ abstract contract AbstractTBTCDepositor {
     ///         `keccak256(fundingTxHash | reveal.fundingOutputIndex)`. This
     ///         key can be used to refer to the deposit in the Bridge and
     ///         TBTCVault contracts.
+    /// @return initialDepositAmount Amount of funding transaction deposit. In
+    ///         TBTC token decimals precision.
     /// @dev Requirements:
     ///      - The revealed vault address must match the TBTCVault address,
     ///      - All requirements from {Bridge#revealDepositWithExtraData}
@@ -134,10 +136,10 @@ abstract contract AbstractTBTCDepositor {
         IBridgeTypes.BitcoinTxInfo calldata fundingTx,
         IBridgeTypes.DepositRevealInfo calldata reveal,
         bytes32 extraData
-    ) internal returns (uint256) {
+    ) internal returns (uint256 depositKey, uint256 initialDepositAmount) {
         require(reveal.vault == address(tbtcVault), "Vault address mismatch");
 
-        uint256 depositKey = _calculateDepositKey(
+        depositKey = _calculateDepositKey(
             _calculateBitcoinTxHash(fundingTx),
             reveal.fundingOutputIndex
         );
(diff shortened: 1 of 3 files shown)


lukasz-zimnoch (commented):
Regarding `_getInitialDepositAmount`, did you try to use the storage pointer:
```
IBridgeTypes.DepositRequest storage deposit = bridge.deposits(
 depositKey
);
```
instead of the presented:
```
IBridgeTypes.DepositRequest memory deposit = bridge.deposits(
 depositKey
);
```
?

The storage pointer requires 2 storage reads while the version with memory reads all 7 storage fields of the `DepositRequest`. This can explain the increased gas consumption. The current version works for me but if we prefer a separate function, we can give it a shot.

solidity/contracts/integrator/AbstractTBTCDepositor.sol
```diff
@@ -134,10 +136,10 @@ abstract contract AbstractTBTCDepositor {
         IBridgeTypes.BitcoinTxInfo calldata fundingTx,
         IBridgeTypes.DepositRevealInfo calldata reveal,
         bytes32 extraData
-    ) internal returns (uint256) {
+    ) internal returns (uint256 depositKey, uint256 initialDepositAmount) {
```
lukasz-zimnoch, line 139: Let's make sure we are covering this change in the example described in the contract's docstring.

nkuba: a3a7cf6b97139fc3af4e57523c7ccade68354d69

lukasz-zimnoch (approved):

PR #806, opened 2024-04-03
Lock pragma version for `L*BitcoinDepositor` contracts

lukasz-zimnoch:
Depends on: https://github.com/keep-network/tbtc-v2/pull/805

Here we fix the Solidity version to be `0.8.17` which is used across all tBTC contracts.

### DIFF
--- a/solidity/contracts/l2/L1BitcoinDepositor.sol
+++ b/solidity/contracts/l2/L1BitcoinDepositor.sol
@@ -13,7 +13,7 @@
 //               ▐████▌    ▐████▌
 //               ▐████▌    ▐████▌
 
-pragma solidity ^0.8.17;
+pragma solidity 0.8.17;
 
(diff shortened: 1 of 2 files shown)


solidity/contracts/l2/L1BitcoinDepositor.sol
```diff
@@ -13,7 +13,7 @@
 //               ▐████▌    ▐████▌
 //               ▐████▌    ▐████▌
 
-pragma solidity ^0.8.17;
+pragma solidity 0.8.17;
```
tomaszslabon, line 16: Shouldn't this change be applied to other files, like `tbtc-v2/solidity/contracts/l2/L2WormholeGateway.sol`?

lukasz-zimnoch: Let's leave as-is for now. The other contracts are already deployed anyway.

tomaszslabon (approved):

PR #814, opened 2024-06-04
Implement redemptions with a proxy for redemption integrations

nkuba:
In this PR we introduce a RedeemerProxy interface that can be used by integrators to route tBTC redemption through the tBTC Bridge.

An example of the integrator is Acre's [BitcoinRedeemer contract](https://github.com/thesis/acre/blob/main/solidity/contracts/BitcoinRedeemer.sol) that requires the redemption data to handle stBTC -> tBTC -> Bitcoin conversion in one transaction.

When redeeming tBTC the challenge is to assemble redemption data to contain valid wallet UTXO and other data in the format expected by the Bridge.

It is expected that the integrator will implement `RedeemerProxy` interface by exposing two functions:
- `redeemerAddress(): ChainIdentifier` - that will return the Ethereum address to which the tBTC balance will be assigned in case of problems with redemption,
- `requestRedemption(redemptionData: Hex): Promise<Hex>` - a callback function that will be called with assembled `redemptionData` to proceed with a custom integrator's logic.

The integrator should use the `requestRedemptionWithProxy` function to execute the redemption.

### DIFF
--- a/typescript/src/services/redemptions/redemptions-service.ts
+++ b/typescript/src/services/redemptions/redemptions-service.ts
@@ -1,5 +1,4 @@
 import {
-  ChainIdentifier,
   RedemptionRequest,
   TBTCContracts,
   WalletState,
@@ -12,8 +11,9 @@ import {
   BitcoinTxOutput,
   BitcoinUtxo,
 } from "../../lib/bitcoin"
-import { BigNumber } from "ethers"
+import { BigNumber, BigNumberish } from "ethers"
 import { Hex } from "../../lib/utils"
+import { RedeemerProxy } from "./redeemer-proxy"
 
 /**
  * Service exposing features related to tBTC v2 redemptions.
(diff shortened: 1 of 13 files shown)


typescript/src/services/redemptions/redemptions-service.ts
```diff
@@ -51,6 +52,73 @@ export class RedemptionsService {
...
+   *                               sent to. Only P2PKH, P2WPKH, P2SH, and P2WSH
+   *                               address types are supported.
+   * @param amount The amount to be redeemed with the precision of the tBTC
+   *               on-chain token contract.
+   * @param chainRedeemerAddress Chain identifier of the redeemer. This is the
+   *                             address that will be able to claim the tBTC tokens
+   *                             if anything goes wrong during the redemption process.
+   * @returns Data needed to request a redemption.
+   */
+  async buildRedemptionData(
```
lukasz-zimnoch, line 86: What do you think about making this function more high-level. Something like that:
```
async requestRedemptionWithProxy(
    bitcoinRedeemerAddress: string,
    amount: BigNumber,
    redeemerProxy: RedeemerProxy
): Promise<{
    targetChainTxHash: Hex
    walletPublicKey: Hex
}> {
    const chainRedeemerAddress = redeemerProxy.redeemerAddress()

    const { walletPublicKey, mainUtxo, redeemerOutputScript } =
      await this.determineRedemptionData(bitcoinRedeemerAddress, amount)

    const redemptionData = this.tbtcContracts.tbtcToken.buildRequestRedemptionData(
      chainRedeemerAddress,
      walletPublicKey,
      mainUtxo,
      redeemerOutputScript
    )

    return redeemerProxy.requestRedemption(redemptionData) // this calls approveAndCall underneath
}

interface RedeemerProxy {
    redeemerAddress(): string

    requestRedemption(redemptionData: Hex): Promise<{
        targetChainTxHash: Hex
        walletPublicKey: Hex
    }>
}
```

That way we make tBTC SDK API consistent with what we have on the deposit side. Moreover, we can expose some base classes implementing `RedeemerProxy` to handle some redemption complexity and do a favor for integrators.

typescript/src/services/redemptions/redemptions-service.ts
```diff
@@ -103,7 +161,7 @@ export class RedemptionsService {
    * @param amount The amount to be redeemed in satoshis.
    * @returns Promise with the wallet details needed to request a redemption.
    */
-  protected async findWalletForRedemption(
+  async findWalletForRedemption(
```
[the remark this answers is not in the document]
nkuba: This was an accident :) Reverted in 2c15ff8268651d18a56fc6aa38a7ca84886ee248

typescript/src/services/redemptions/redemptions-service.ts
nkuba: As you wish 😉  I pushed the changes in 97ea592b02021439989ff06645c91ef343fa827f

lukasz-zimnoch (approved):

PR #816, opened 2024-06-24
Skip retry for registered wallets

### LINKED ISSUE #815: Improve wallet lookup during redemptions
If there are closed/terminated wallets in the system, the redemption process takes a long time. This is what happens:

1. The `findWalletForRedemption` function must iterate over all wallets to check which one can handle the redemption: https://github.com/keep-network/tbtc-v2/blob/51094e4cdaefdf82a11b80f7e9e6899a6333a9ca/typescript/src/services/redemptions/redemptions-service.ts#L127
2. To do the check, the above function must call `wallets` to fetch data of the particular wallet: https://github.com/keep-network/tbtc-v2/blob/51094e4cdaefdf82a11b80f7e9e6899a6333a9ca/typescript/src/lib/ethereum/bridge.ts#L583
3. The `wallets` function calls `getWalletCompressedPublicKey` under the hood: https://github.com/keep-network/tbtc-v2/blob/51094e4cdaefdf82a11b80f7e9e6899a6333a9ca/typescript/src/lib/ethereum/bridge.ts#L607
4. Closed/terminated wallets are no longer in the contract state so, `getWalletCompressedPublicKey` throws while calling `getWalletPublicKey`: https://github.com/keep-network/tbtc-v2/blob/51094e4cdaefdf82a11b80f7e9e6899a6333a9ca/typescript/src/lib/ethereum/bridge.ts#L521
5. The above is not a problem per se as the error is caught. However, the `getWalletPublicKey` uses a backoff retry mechanism inside so the final throw is made after 3 internal retries which takes time and contributes to the long execution time of the entire process: https://github.com/keep-network/tbtc-v2/blob/51094e4cdaefdf82a11b80f7e9e6899a6333a9ca/typescript/src/lib/ethereum/wallet-registry.ts#L69

The solution is simple. The `backoffRetrier` function should not retry if the contract call reverts with `Wallet with the given ID has not been registered` error. This can be achieved by passing a custom `errorMatcher` argument to this specific `backoffRetrier` call: https://github.com/keep-network/tbtc-v2/blob/51094e4cdaefdf82a11b80f7e9e6899a6333a9ca/typescript/src/lib/utils/backoff.ts#L93

ptdatta:
Fixes #815 

The `backoffRetrier` runs retry for every error which makes the redemption process takes a long time. 

So for registered wallets, it should not retry so instead of `retryAll` we use `skipRetryWhenMatched["Wallet with the given ID has not been registered"]` so that it skips retry for registered wallets.


mhluongo:
Thanks for this @ptdatta! Giving it a look

PR #817, opened 2024-06-24
Added BTC with Satoshi

### LINKED ISSUE #759: Display maximum redemption amount in BTC when wallet have not enough funds
When user requests unminting but there's not enough funds in the wallets to process that request, we show error like this:
image

Could we display the amount in BTC, not satoshi? Or show satoshi but with BTC equivalent in brackets? User may want to request the max possible redemption after seeing the error, so it would be easier for them to already have the amount in the format that needs to be provided on the unmint form.

From @michalsmiarowski :
> This is an error thrown straight in tbtc-v2.ts sdk: https://github.com/keep-network/tbtc-v2/blob/cadead9ecd6005325ace4d64288c20733b058352/typescript/src/services/redemptions/redemptions-service.ts#L209-L212
>
>I'm not sure if we can change it on the front-end, because we might not receive the satoshi amount there 🤔  so we would have to probably change it in the lib if it's necessary

ptdatta:
Fixes #759 

Display maximum redemption amount in BTC with Satoshi


mhluongo (approved):
LGTM! Thanks @ptdatta :)

PR #819, opened 2024-07-05
Monitoring: Migrate to latest SDK

lukasz-zimnoch:
Depends on: #818 
Depends on: #820 

Here we migrate the monitoring package to the latest version of tBTC SDK. We are using the latest `development` version of the SDK for now as it contains some unreleased features and fixes that are useful for the monitoring tool. Migration of the monitoring to the new tBTC SDK pays off some tech debt but also reduces the number of requests sent to the Ethereum node due to some optimizations introduced in the recent versions of the SDK.


monitoring/src/index.ts
```diff
@@ -15,53 +19,87 @@ import { WalletMonitor } from "./wallet-monitor"
...
+}
+
+async function setup(): Promise<SystemEventManager> {
+  const sdk = await setupSDK()
+  return setupMonitoring(sdk)
+}
+
+setup().then((manager) => {
+  manager.trigger().then((report) => {
+    // eslint-disable-next-line default-case
```
tomaszslabon, line 91: Couldn't we add a default case, e.g.:
```
    switch (report.status) {
      case "success": {
        console.log(report)
        break
      }
      case "failure": {
        console.error(report)
        break
      }
      default: {
        console.error("Unknown report status")
      }
    }
```
?

lukasz-zimnoch: This is redundant. We set the result manually to one of the two values.

tomaszslabon (approved):

PR #821, opened 2024-07-26
Improve the `WalletRegistry.getWalletPublicKey` function

### LINKED ISSUE #815: Improve wallet lookup during redemptions

r-czajkowski:
Closes #815

Skip retrying when the contract throws an error saying the wallet with the given ID has not been registered. There is no need to retry this request because the contract responds with the exact reason - the wallet does not exist. As Łukasz mentioned in https://github.com/keep-network/tbtc-v2/issues/815 it happens because closed/terminated wallets are no longer in the contract state so, `getWalletCompressedPublicKey` throws while calling `getWalletPublicKey`. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wallet public key retrieval by avoiding unnecessary retries when a wallet is not registered.
  * Preserved existing retrieval behavior for other errors.

* **Documentation**
  * Updated source references for DKG result event retrieval methods.


typescript/src/lib/ethereum/wallet-registry.ts
```diff
@@ -66,13 +66,16 @@ export class EthereumWalletRegistry
...
-        return await this._instance.getWalletPublicKey(
-          walletID.toPrefixedString()
-        )
-      }
-    )
+    const publicKey = await backoffRetrier<string>(
+      this._totalRetryAttempts,
+      undefined,
+      undefined,
+      skipRetryWhenMatched(["Wallet with the given ID has not been registered"])
```
nkuba, line 73: I'm not sure we can safely skip this error. We want to work around the problem in the redemption process, but the `getWalletPublicKey` function can be used in other processes as well.
If someone calls `getWalletPublicKey` right after a fresh wallet creation there is a chance that chain endpoint hasn't synced yet and the retry would be desirable here.

Maybe we should add an optional argument `skipRetryWhenNotRegistered = false` and use this variant only in the redemption process?

mswilkison: Addressed in a68ab3156: the skip is now opt-in via a `skipRetryWhenNotRegistered` argument (default `false`, so the general-purpose path keeps its retry behavior), enabled only by the redemption-path caller `EthereumBridge.getWalletCompressedPublicKey`. Also added unit tests covering the retry-abort behavior, pinning the exact contract revert reason.

piotr-roslaniec (approved):

PR #843, opened 2025-03-14
Upgrade Arbitrum L1 Bitcoin Depositor contract

evandrosaturnino:
## Description

This pull request **upgrades the `L1BitcoinDepositor` implementation contract** to introduce a new boolean flag `reimburseTxMaxFee` and associated logic. When `reimburseTxMaxFee` is enabled, the contract will effectively **refund the deposit transaction’s `txMaxFee`** by adding it to the TBTC minted for the user. This way, the user is not burdened by the deposit transaction max fee.

### What's Changed

1. **New Storage Slot (`bool reimburseTxMaxFee`):**
   - Appended at the end of the storage layout to ensure no collisions or reordering issues when upgrading.
   - Defaults to `false`, meaning no reimbursement of `txMaxFee`.

2. **New Setter Function (`setReimburseTxMaxFee`):**
   - Only the contract owner can toggle this feature flag.
   - Allows the protocol to dynamically choose whether to reimburse or deduct the `txMaxFee`.

3. **Logic to Deduct `txMaxFee`:**
   - In `finalizeDeposit`, if `reimburseTxMaxFee == true`, the contract will increase the minted TBTC by the deposit transaction max fee so users do not lose out on that cost.
   - If `reimburseTxMaxFee == false`, the max fee is simply not added to the user’s final amount.

4. **All Other Functionality Unchanged:**
   - The upgrade is strictly backward‐compatible.
   - The contract’s storage layout and inheritance remain intact, with only one new boolean added at the end.

### Rationale

We already had this behavior on the Base network’s `L1BitcoinDepositor` and needed to consolidate that logic on Arbitrum and other networks as well. This ensures consistent behavior across all L2s for TBTC deposits, sparing users from deposit transaction fee overhead when the DAO chooses to cover these fees.

---

### Deployment Addresses

**Mainnet**  
- **Proxy Contract**: `0x75A6e4A7C8fAa162192FAD6C1F7A6d48992c619A`  
- **Old Implementation**: `0xff79fca71751a5a0c4487a1ace268d6cd2a64db1`  
- **New Implementation**: `0x3ffee7d7be970201a33d17e318e2421a06ad69f8`  

**Sepolia**  
- **Proxy Contract**: `0xD9B523fb879C63b00ef14e48C98f4e3398d3BA2D`  
- **Old Implementation**: `0xf90292a39014033DE50c81f651365cB6A1040Bb6`  
- **New Implementation**: `0xad0d13365b7b760e83cec07398fa975d6c1f67c9`  

These addresses confirm that the proxy contracts on both mainnet and sepolia now point to the upgraded implementations which include the `reimburseTxMaxFee` feature.


lrsaturnino (approved):
LGTM

PR #850, opened 2025-05-10
feat: fetch redemption wallets

evandrosaturnino:
## Description

This PR updates the RedemptionsService within our tBTC v2 SDK to leverage the new backend redemption endpoint rather than performing the entire wallet-search logic locally.

### Key changes

1. **`requestRedemption`:**
    - Tries to fetch candidate wallets via fetchWalletsForRedemption().
    - If successful, uses a new method determineValidRedemptionWallet(...) to pick the first wallet that can handle the requested amount (and has no pending redemption for the same output script).
    - Fallback: If the backend fetch fails (due to network or other issues), it reverts to the original on-chain scanning approach via determineRedemptionData(...) → findWalletForRedemption(...) for backward compatibility.

2. **`determineValidRedemptionWallet`:**
    - Takes the array of “serializable wallets” from the backend.
    - Converts them to the local ValidRedemptionWallet structure.
    - Checks balances, pending redemptions, etc.

3. **`fetchWalletsForRedemption`:**
    - Calls our new backend (fetch endpoint to retrieve live candidate wallets.
    - Only available on Mainnet (throws an error if called on test networks).

4. **Refactoring:**
    - Moved some shared logic to helper methods (getRedeemerOutputScript).
    - Updated logs and error handling to clarify whether the retrieval was local or from the backend.

### Rationale

Moving the heavy wallet-lookup logic into the backend significantly reduces overhead for users:

- **Performance**: The client no longer iterates over all wallets and fetches UTXO data from scratch.

- **Reliability**: A single, centralized server can be optimized (e.g., caching, concurrency controls) without requiring each user to replicate that logic.

- **Fallback**: By retaining the original findWalletForRedemption(...) path for errors, we ensure graceful degradation if the backend endpoint is offline or unreachable.

### DIFF
--- a/typescript/src/services/redemptions/redemptions-service.ts
+++ b/typescript/src/services/redemptions/redemptions-service.ts
@@ -16,6 +16,7 @@ import { BigNumber, BigNumberish } from "ethers"
 import { amountToSatoshi, ApiUrl, endpointUrl, Hex } from "../../lib/utils"
 import { RedeemerProxy } from "./redeemer-proxy"
 import {
+  RedemptionWallet,
   SerializableWallet,
   ValidRedemptionWallet,
 } from "../../lib/utils/types"
@@ -208,11 +209,7 @@ export class RedemptionsService {
     bitcoinRedeemerAddress: string,
     amount: BigNumber,
     potentialCandidateWallets: Array<SerializableWallet>
-  ): Promise<{
-    walletPublicKey: Hex
-    mainUtxo: BitcoinUtxo
-    redeemerOutputScript: Hex
-  }> {
+  ): Promise<RedemptionWallet> {
     let walletPublicKey: Hex | undefined = undefined
     let mainUtxo: BitcoinUtxo | undefined = undefined
     const redeemerOutputScript = await this.getRedeemerOutputScript(
@@ -266,7 +263,7 @@ export class RedemptionsService {
     }
 
     if (!walletPublicKey || !mainUtxo) {
-      return null;
+      throw new Error(`Could not find a wallet with enough funds.`)
     }
 
     return { walletPublicKey, mainUtxo, redeemerOutputScript }
--- a/typescript/src/services/redemptions/redemptions-service.ts
+++ b/typescript/src/services/redemptions/redemptions-service.ts
@@ -266,7 +266,7 @@ export class RedemptionsService {
     }
 
     if (!walletPublicKey || !mainUtxo) {
-      throw new Error(`Could not find a wallet with enough funds.`)
+      return null;
     }
 
     return { walletPublicKey, mainUtxo, redeemerOutputScript }
--- a/typescript/src/services/redemptions/redemptions-service.ts
+++ b/typescript/src/services/redemptions/redemptions-service.ts
@@ -227,11 +227,6 @@ export class RedemptionsService {
         mainUtxo: candidateMainUtxo,
       } = this.fromSerializableWallet(serializableWallet)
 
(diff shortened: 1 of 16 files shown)


typescript/src/services/redemptions/redemptions-service.ts
```diff
@@ -170,6 +184,94 @@ export class RedemptionsService {
...
+   */
+  protected async determineValidRedemptionWallet(
+    bitcoinRedeemerAddress: string,
+    amount: BigNumber,
+    potentialCandidateWallets: Array<SerializableWallet>
+  ): Promise<{
+    walletPublicKey: Hex
+    mainUtxo: BitcoinUtxo
+    redeemerOutputScript: Hex
+  }> {
```
piotr-roslaniec, line 215: Nit: This could be a type
```suggestion
  ): Promise<RedemptionWallet | null> {
```

typescript/src/services/redemptions/redemptions-service.ts
```diff
@@ -170,6 +184,94 @@ export class RedemptionsService {
...
+          `can handle the redemption request. ` +
+          `Stop the loop execution and proceed with the redemption...`
+      )
+
+      break
+    }
+
+    if (!walletPublicKey || !mainUtxo) {
+      throw new Error(`Could not find a wallet with enough funds.`)
+    }
```
piotr-roslaniec, line 270: Nit:

```suggestion
    if (!walletPublicKey || !mainUtxo) {
      return null;
    }
```

typescript/src/services/redemptions/redemptions-service.ts
```diff
@@ -53,19 +58,46 @@ export class RedemptionsService {
...
 
-    return {
-      targetChainTxHash: txHash,
-      walletPublicKey,
+      return {
+        targetChainTxHash: txHash,
+        walletPublicKey: walletPublicKey,
+      }
+    } catch (error) {
+      console.error(
```
piotr-roslaniec, line 82:
```suggestion
      console.warn(
```

piotr-roslaniec (approved):
LGTM, just a couple of nitpicks

typescript/src/services/redemptions/redemptions-service.ts
evandrosaturnino: I believe it’s important that this error is thrown, because `requestRedemption` triest to fetch the data, determine the valid redemption wallet from that data before it can proceed with the transaction. If any of those steps throws an error, the method should fall back to the active search (performed on the client) to locate and validate the wallet.

Therefore, we want the failure to surface, to be specified so we can track it, and move to the fallback.

https://github.com/keep-network/tbtc-v2/blob/0528fcd08b28a2436e507400903431a6271e66ab/typescript/src/services/redemptions/redemptions-service.ts#L55-L103

typescript/src/services/redemptions/redemptions-service.ts
evandrosaturnino: Addressed on [chore: update console in redemption service](https://github.com/keep-network/tbtc-v2/pull/850/commits/4c6c5ffad7ffaa66bb4148118f6c0c7029faa433)

typescript/src/services/redemptions/redemptions-service.ts
evandrosaturnino: Addressed on [refactor: add redemption wallet type](https://github.com/keep-network/tbtc-v2/pull/850/commits/0528fcd08b28a2436e507400903431a6271e66ab)

lrsaturnino (approved):
Following Piotr's Approval to release this PR

PR #861, opened 2025-06-09
feat: add support to crosschain redemption in sdk

evandrosaturnino:
### feat: Add support to cross-chain redemptions in SDK

This PR introduces support for cross-chain redemptions within the tBTC SDK. A new `RedemptionsService` has been added to manage and abstract the complexity of redemption operations, including both standard and cross-chain redemptions.

### Summary of Changes

- **New `RedemptionsService`**: A dedicated service at `typescript/src/services/redemptions/redemptions-service.ts` has been created to handle all redemption-related logic.
- **Cross-Chain Redemption Flow**: The SDK now supports initiating redemptions from L2 chains (like Arbitrum and Base) and processing them on L1 (Ethereum).
- **New SDK Methods**:
    - `requestCrossChainRedemption`: Allows users to request a redemption directly from a supported L2 chain. It interacts with the `L2BitcoinRedeemer` contract on the respective L2.
    - `relayRedemptionRequestToL1`: A method used by the redemption service to relay a validated redemption request from an L2 to the `L1BitcoinRedeemer` contract on Ethereum for final processing.
- **New Contract Abstractions**:
    - `L1BitcoinRedeemer`: A handle for the L1 contract that processes redemption requests relayed from L2s.
    - `L2BitcoinRedeemer`: Implementations for each supported L2 (e.g., `ArbitrumL2BitcoinRedeemer`) that initiate the cross-chain redemption process, typically via a messaging bridge like Wormhole.

### How it Works

1.  A user calls `requestCrossChainRedemption` from the `RedemptionsService`, specifying the amount, their Bitcoin address, and the source L2 chain.
2.  The service calls the `requestRedemption` function on the `L2BitcoinRedeemer` contract on the specified L2.
3.  The `L2BitcoinRedeemer` contract sends a message (e.g., via Wormhole) to the L1.
4.  An off-chain service monitors for these messages. Once a message is observed and verified, the service calls `relayRedemptionRequestToL1`.
5.  `relayRedemptionRequestToL1` calls the `requestRedemption` function on the `L1BitcoinRedeemer` contract on Ethereum.
6.  The `L1BitcoinRedeemer` contract then proceeds with the standard tBTC redemption logic on L1, ultimately releasing the BTC to the user.

This new functionality provides a more seamless experience for users on L2s, allowing them to redeem their tBTC without needing to manually bridge assets to Ethereum first.


typescript/src/services/redemptions/redemptions-service.ts
```diff
@@ -149,6 +164,81 @@ export class RedemptionsService {
...
+    if (!crossChainContracts || !crossChainContracts.l2BitcoinRedeemer) {
+      throw new Error(
+        `Cross-chain redeemer contracts for ${l2ChainName} not initialized`
+      )
+    }
+
+    const redeemerOutputScript = await this.getRedeemerOutputScript(
+      bitcoinRedeemerAddress
+    )
+    const nonce = Math.round(Math.random() * 10000)
```
piotr-roslaniec, line 195: Is there a reason to limit the `nonce` to `[1, 10 000]` range?

```suggestion
    // guarantees uniqueness over time
    const nonce = Date.now() * 1000 + Math.floor(Math.random() * 1000)
```

typescript/src/lib/arbitrum/l2-bitcoin-redeemer.ts
```diff
@@ -0,0 +1,98 @@
...
+  constructor(config: EthersContractConfig, chainId: Chains.Arbitrum) {
+    let deployment: EthersContractDeployment
+    let wormholeCoreDeployment: EthersContractDeployment
+    let recipientChain: number
+
+    switch (chainId) {
+      case Chains.Arbitrum.ArbitrumSepolia:
+        deployment = ArbitrumSepoliaL2BitcoinRedeemerDeployment
+        wormholeCoreDeployment = ArbitrumSepoliaWormholeCoreDeployment
+        recipientChain = 10002 // Ethereum Sepolia
```
piotr-roslaniec, line 37: Nitpick:
```suggestion
        recipientChain = ChainIds.ETHEREUM_SEPOLIA
```

piotr-roslaniec (dismissed):
LGTM, just one question

typescript/src/lib/arbitrum/l2-bitcoin-redeemer.ts
evandrosaturnino: Since `recipientChain` here expects the Wormhole specific Chain Id (Not the same value as Ethereum Chain Id), I created a specific constant holding these values in: [refactor: add wormhole chain ids constant](https://github.com/threshold-network/tbtc-v2/pull/861/commits/06df45a311ebf242a24cae346d1365c77db79bcc)

For more info, please refer to:  https://wormhole.com/docs/products/reference/chain-ids/

typescript/src/services/redemptions/redemptions-service.ts
evandrosaturnino: No specific requirement drove the 1‑10 000 limit, I picked it as a way to get a random integer, but it doesn’t guarantee uniqueness across rapid successive calls. Your timestamp‑based approach is more robust because it practically eliminates collisions. I updated the code accordingly by accepting your suggestion.

PR #867, opened 2025-06-14
feat: add base timelock deployment script and artifact

evandrosaturnino:
## Description 

Upgrade scenario: https://github.com/threshold-network/tbtc-v2/issues/866

The goal is to deploy a Timelock between the WormholeGateway, L2BitcoinDepositor, and L2TBTC ProxyAdmins and the Threshold Council multisig to enforce 24h delay between upgrades.


0xCVH:
@evandrosaturnino 

The setup of this Multisig is not ideal.

We are adding individual addresses as Executors, which 

1. creates a operational burden when someone leaves or needs to change their address.
2. as well, it just adds operational risk every time we need change roles/ownerships on the Timelock.

Timelock should be a set and forget type of activity, with any changes done on the Multisig itself, not the Timelock. ANY Timelock deployed should only have the TH Council Multisig as PROPOSER & EXECUTOR. Nothing else. 

Can we please get the Timelock setup properly done and generate the appropiate JSON for the ownership changes to this Timelock? Thanks.

PR #886, opened 2025-07-16
Depositor w/nonReentrant `finalizeDeposit()` and `_transferTbtc` becomes last interaction

### LINKED ISSUE #885: Checks-Effects-Interactions pattern on `AbstractL1BTCDepositor`
### Reentrancy Protection Gap in Abstract Method Pattern

**Impact**: Developers may inadvertently introduce reentrancy vulnerabilities in child implementations

**NOTE:** Transferring tBTC (ERC20) last can indeed reduce the risk of a large financial loss if a reentrancy vulnerability is present, because the tBTC transfer is typically the most valuable operation in the function. Transferring tBTC last is the most conservative and secure approach, especially if the reimbursement involves a low-level call to an untrusted or upgradable contract. This minimizes the risk of repeated tBTC transfers in the event of a reentrancy exploit.

```solidity
function finalizeDeposit(uint256 depositKey) external payable {
    // Abstract method called before state cleanup
    _transferTbtc(tbtcAmount, destinationChainDepositOwner);
    
    // State cleanup happens AFTER external interaction
    if (reimbursement.receiver != address(0)) {
        delete gasReimbursements[depositKey];
        reimbursementPool.refund(...)
    }
}

```

**Issue**: The pattern violates Checks-Effects-Interactions (CEI) by calling the abstract method before state cleanup.

**Recommendation**:

- Move state cleanup before `_transferTbtc()` call  
- Add reentrancy guard to `finalizeDeposit()`  
- Implement CEI (Checks-Effects-Interactions) pattern as template for developers

jose-blockchain:
Developers may inadvertently introduce reentrancy vulnerabilities in child implementations. So:

- Move state cleanup and reembursements before _transferTbtc() call
- Add reentrancy guard to finalizeDeposit()

Assuming the external calls in `_transferTbtc()` are more financially sensible so left to the end of `finalizeDeposit()`, and also some state cleanups now come ealier. The `nonReentrant` is more clear why it can help.

Closes #885

### DIFF
--- a/solidity/contracts/cross-chain/AbstractL1BTCDepositor.sol
+++ b/solidity/contracts/cross-chain/AbstractL1BTCDepositor.sol
@@ -19,6 +19,7 @@ import "@keep-network/random-beacon/contracts/Reimbursable.sol";
 import "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
 
 import "../integrator/AbstractBTCDepositor.sol";
 import "../integrator/IBridge.sol";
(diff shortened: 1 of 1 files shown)


jose-blockchain:
Was exploring this update but is not critical, and we can discuss later when I had more integrations done and I see better the context of this change.

PR #888, opened 2025-07-20
Feat/upgrade bob tbtc v2 token

evandrosaturnino:


cross-chain/bob/contracts/OptimismMintableUpgradableTBTC.sol
```diff
@@ -0,0 +1,337 @@
...
+///         The token is burnable by the token holder and supports EIP2612
+///         permits. Token holder can authorize a transfer of their token with
+///         a signature conforming EIP712 standard instead of an on-chain
+///         transaction from their address. Anyone can submit this signature on
+///         the user's behalf by calling the permit function, paying gas fees,
+///         and possibly performing other actions in the same transaction.
+///         The governance can recover ERC20 and ERC721 tokens sent mistakenly
+///         to OptimismMintableUpgradableTBTC token contract.
+/// @custom:oz-upgrades-from OptimismMintableUpgradableERC20
+contract OptimismMintableUpgradableTBTC is
```
jose-blockchain, line 36: looks excellent and I am learning a lot, just wondering if do we need to include the `MisfundRecovery` functionality, `recoverERC20()` and 721, to recover funds or it was deprecated?:

MisfundRecovery.sol

```
import "@thesis/solidity-contracts/contracts/token/MisfundRecovery.sol";

contract TBTC is ERC20WithPermit, MisfundRecovery {
```

cross-chain/bob/contracts/interfaces/IOptimismMintableERC20.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
+
+/// @title IOptimismMintableERC20
+/// @notice This interface is available on the OptimismMintableERC20 contract.
+///         We declare it as a separate interface so that it can be used in
+///         custom implementations of OptimismMintableERC20.
+interface IOptimismMintableERC20 is IERC165 {
```
jose-blockchain, line 10: catching up with the canonical version you mention. 

- does the canonical convergence is just partial here?
- do we need to approximate as much as possible the version `solidity/contracts/cross-chain/wormhole/L2TBTC.sol` ??
I understand that maybe we cannot converge 100% on this case.

cross-chain/bob/contracts/canonical/L2TBTC.sol
```diff
@@ -0,0 +1,284 @@
...
+///
+///         The token is burnable by the token holder and supports EIP2612
+///         permits. Token holder can authorize a transfer of their token with
+///         a signature conforming EIP712 standard instead of an on-chain
+///         transaction from their address. Anyone can submit this signature on
+///         the user's behalf by calling the permit function, paying gas fees,
+///         and possibly performing other actions in the same transaction.
+///         The governance can recover ERC20 and ERC721 tokens sent mistakenly
+///         to L2TBTC token contract.
+contract L2TBTC is
```
jose-blockchain, line 50: ah ok, is incorporated into `./contracts/canonical` folder

cross-chain/bob/contracts/canonical/L2TBTC.sol
```diff
@@ -0,0 +1,284 @@
...
+        token.safeTransferFrom(address(this), recipient, tokenId, data);
+    }
+
+    /// @notice Allows one of the guardians to pause mints and burns allowing
+    ///         avoidance of contagion in case of a chain- or bridge-specific
+    ///         incident.
+    /// @dev Requirements:
+    ///      - The caller must be a guardian.
+    ///      - The contract must not be already paused.
+    function pause() external onlyGuardian {
```
jose-blockchain, line 221: I still dont understand the `pause() onlyGuardian` concept, does a single bizantine guardian (1 of 19) pausing the bridge make sense?
thanks for the patience here, but there guardians are the same of Wormhole network or are specific to this bridge?

cross-chain/bob/contracts/OptimismMintableUpgradableTBTC.sol
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    /// @notice Initialize v2 of the contract with new functionality
+    /// @custom:oz-upgrades-validate-as-initializer
+    function initializeV2() public reinitializer(2) {
+        __Ownable_init();
+        __Pausable_init();
+
+        // Set legacyCapRemaining to current total supply
+        legacyCapRemaining = totalSupply();
```
jose-blockchain, line 95: can we document briefly a bit more the Legacy Cap mechanism? Is not 100% clear for me, I can imagine how it works, probably for others the same.

jose-blockchain (changes requested):
looks like a solid upgrade with focus on security and compatibility. I will clarify further how the `legacyCapRemaining` works and to what extent the canonical token is reached.

requesting minor changes with approval.

cross-chain/bob/contracts/OptimismMintableUpgradableTBTC.sol
evandrosaturnino: Document added in [chore: add bob upgrade technical doc](https://github.com/threshold-network/tbtc-v2/pull/888/commits/0416da8dcbe1c7a2612f153d13d85251914f3b83)

cross-chain/bob/contracts/interfaces/IOptimismMintableERC20.sol
evandrosaturnino: We are mostly porting the canonical methods, modifiers and variables to the existing BOB TBTC contract. 
We are not entirely removing the previous existing functionalities, but rather upgrading/adapting them due to the bridge transitioning aspect. So it should continue to accept the BOB Standard Bridge in the legacy format, while 100% compatible with the canonical L2TBTC token.

cross-chain/bob/contracts/OptimismMintableUpgradableTBTC.sol
jose-blockchain: got it

cross-chain/bob/docs/tbtc-bob-upgrade-technical-document.md
```diff
@@ -0,0 +1,181 @@
...
+This captures the exact amount of tokens that were minted through the legacy bridge at the time of upgrade.
+
+### Operational Mechanics
+
+#### Minting Behavior
+
+When the legacy bridge (BRIDGE) mints new tokens:
+
+```solidity
+if (msg.sender == BRIDGE) {
```
jose-blockchain, line 63: is this implemented? seems that is not implemented and it is okey to avoid new legacy mintings?

do we start directly on `Phase 2` or we have this `Phase 1` ?

jose-blockchain (approved):
LGTM docs added and more context given

cross-chain/bob/docs/tbtc-bob-upgrade-technical-document.md
evandrosaturnino: This has been implemented here: https://github.com/threshold-network/tbtc-v2/blob/913b5e6edc3d9dbea3938f839e4c8a48a1ef0643/cross-chain/bob/contracts/OptimismMintableUpgradableTBTC.sol#L236-L238

This upgrade is already the final stage of the token, considering the legacy cap remaining variable is susceptible to be incremented and decremented accordingly. We should not disable legacy minting because it is not possible to prevent bridging on the Standard BOB bridge on L1. To avoid users' funds becoming stuck on L1, we must continue to allow the BOB bridge to mint new tokens.

However, since the BOB front end will be disabled, we expect the number of new legacy bridge mints to decrease. Furthermore, any amount minted through the legacy bridge will be added to the remaining cap, meaning users will have to wait seven days to bridge those funds back via the standard bridge route dissuading users from using BOB bridge on the contract level.

cross-chain/bob/contracts/OptimismMintableUpgradableTBTC.sol
```diff
@@ -0,0 +1,337 @@
...
+                "Only bridge can burn while legacy cap remains"
+            );
+            require(
+                amount <= legacyCapRemaining,
+                "Amount exceeds legacy cap remaining"
+            );
+            legacyCapRemaining -= amount;
+        }
+
+        _spendAllowance(account, _msgSender(), amount);
```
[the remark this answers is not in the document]
evandrosaturnino: Addressed in https://github.com/threshold-network/tbtc-v2/pull/888/commits/325925487f2a2c1f8a47252fff90d24aefa0a08c

cross-chain/bob/contracts/OptimismMintableUpgradableTBTC.sol
```diff
@@ -0,0 +1,337 @@
...
+    /// @dev This overrides the burn function from OptimismMintableUpgradableERC20.
+    ///      Requirements:
+    ///      - The caller must be the bridge.
+    ///      - legacyCapRemaining must be greater than 0.
+    /// @param _from Address to burn tokens from.
+    /// @param _amount Amount of tokens to burn.
+    function burn(
+        address _from,
+        uint256 _amount
+    )
```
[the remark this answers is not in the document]
evandrosaturnino: The upgrade is primarily about bringing the BOB token contract in‑line with the canonical L2TBTC.sol implementation that is already live on other chains (Polygon, Optimism, Arbitrum, Base etc). That canonical version exposes both burn and burnFrom, so we mirrored the same surface here to keep our off‑chain tooling, audits, and bridge adapters identical everywhere.

`burn(address _from, uint256 _amount)` - legacy bridge only.
We leave this method as it existed in OptimismMintableUpgradableERC20, gated by the onlyBridge modifier and introduced the legacyCapRemaining counter to manage the transition from the standard bridge to the new one. 
Once the cap reaches zero the function will sit dormant, but we cannot remove or change its selector without breaking live infrastructure.

`burnFrom(address account, uint256 amount)` ― canonical pattern / new bridges.
This is the standard L2TBTC “burn‑from-with‑allowance” that alligns with our existing native crosschain minting infrastructure, that we've been using with wormhole, and now is going to be transposed to CCIP, with the integration of the `BurnFromMintTokenPool` contract (CCIP adapter). We also fold in the cap logic so that, while the legacy cap is still positive, preserving supply invariants throughout the hand‑over.

In short, two methods are necessary because they serve two mutually exclusive authority models: the immutable Optimism bridge (address‑based permission) versus all current and future bridges (allowance‑based permission). Keeping both lets us port the fully‑featured canonical token to BOB without disrupting the existing bridge during its wind‑down and without forcing special‑case code in the cross‑chain infrastructure we already run on every other network.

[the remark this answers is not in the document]
vzotova: makes sense, I just saw some duplicated code and wanted to check reasoning

vzotova (approved):

PR #918, opened 2025-11-17
ENG-472 Bridge controller pointer and MintBurnGuard guardrails

piotr-roslaniec:
### Summary
- Governance-set controller pointer on Bridge plus MintBurnGuard to gate controller-driven mint/burn/unmint with caps, pause, and rate limits.
- MintBurnGuard now uses satoshi-denominated accounting with owner-only migration helper and stricter validations to keep exposure in sync with execution targets.

### Motivation
- Provide a least-privilege mint path for integrators while enforcing global guardrails and clear, consistent units across controller flows.

### Changes
- Bridge/BridgeState: controller pointer storage and events; governance-only `setControllerBalanceIncreaser(address)` (zero clears); controller-guarded `controllerIncreaseBalance` and batch variant; expose `controllerBalanceIncreaser()`; storage gap adjusted.
- BridgeGovernance: owner-only forwarder `setControllerBalanceIncreaser(address)` to manage the Bridge controller pointer.
- MintBurnGuard: accounting/storage/events in TBTC satoshis (1e8) with internal 1e18 conversion for Bridge/Bank/Vault calls; removed controller accounting helpers and `reduceExposureAndBurn`; added owner-only `setTotalMinted` (cap-checked, rate window reset); enforced cap vs rate-limit consistency, pause/rate checks, zero-address guards, custom errors, and safe addition (unchecked removed); event naming aligned to `amount`.


solidity/contracts/bridge/Bridge.sol
```diff
@@ -1541,6 +1563,72 @@ contract Bridge is
...
+    ) external {
+        require(
+            self.authorizedBalanceIncreasers[msg.sender],
+            "Caller is not an authorized increaser"
+        );
+        self.bank.increaseBalances(recipients, amounts);
+    }
+
+    /// @notice Batch helper returning authorization status for given addresses.
+    /// @param increasers Addresses to check for controller authorization.
```
jose-blockchain: empty arrays are cool or not ? just for completeness: maybe add `s.
/// @param increasers Addresses to check. `

solidity/deploy/80_upgrade_bridge_v2.ts
```diff
@@ -1,52 +1,123 @@
...
+
+  // Use only the ProxyAdmin key for proxy operations; do not mix with
+  // governance key to avoid role confusion.
+  const proxyAdminPrivateKey = process.env.PROXY_ADMIN_PK
+
+  let signer = await ethers.getSigner(deployer)
+  let signerAddress = await signer.getAddress()
+  if (proxyAdminPrivateKey) {
+    signer = new ethers.Wallet(proxyAdminPrivateKey, ethers.provider)
+    signerAddress = await signer.getAddress()
```
jose-blockchain: maybe log the positive case too:

```typescript
deployments.log(`✓ Using ProxyAdmin signer: ${signerAddress}`)
```

solidity/test/bridge/Bridge.Governance.test.ts
```diff
@@ -125,6 +128,150 @@ describe("Bridge - Governance", () => {
...
+      it("should increase recipient balance in the bank", async () => {
+        expect(await bank.balanceOf(recipient())).to.equal(amount)
+      })
+
+      it("should emit BalanceIncreased event", async () => {
+        await expect(tx)
+          .to.emit(bank, "BalanceIncreased")
+          .withArgs(recipient(), amount)
+      })
+    })
```
jose-blockchain: possible diff len test:

```typescript
context("when recipients and amounts arrays have different lengths", () => {
  it("should revert", async () => {
    await bridgeGovernance.connect(governance)
      .setAuthorizedBalanceIncreaser(thirdParty.address, true)
    
    await expect(
      bridge.connect(thirdParty)
        .controllerIncreaseBalances([recipient()], [amount, amount])
    ).to.be.reverted
  })
})
```

solidity/test/bridge/Bridge.Governance.test.ts
```diff
@@ -125,6 +128,150 @@ describe("Bridge - Governance", () => {
...
+          recipients().map((addr, index) =>
+            expect(tx)
+              .to.emit(bank, "BalanceIncreased")
+              .withArgs(addr, amounts[index])
+          )
+        )
+      })
+    })
+  })
+
```
jose-blockchain: maybe some edges cases here: 

```typescript
describe("getAuthorizedBalanceIncreasers", () => {
  it("should handle empty array", async () => {
    const result = await bridge.getAuthorizedBalanceIncreasers([])
    expect(result).to.have.lengthOf(0)
  })
  
  it("should return correct mixed authorization states", async () => {
    // Authorize one, leave another unauthorized, test both
  })
})
```

jose-compu (changes requested):
This is solid. The core functionality is sound, the upgrade path looks safe, and those mass-revoke safeguards show you're thinking about the operational side. The suggested changes above are all **non-blocking**, they'd make things a bit better but nothing's broken without them.

The main ones I would prioritize:
1. Security comment on `controllerIncreaseBalance` about governance-level power
2. Maybe enforce contract-only controllers (depends on your design intent)
3. Fill in those test gaps for edge cases
4. Enhanced logging in the sync script for better audit trail

LGTM! (moduli comments)

solidity/contracts/bridge/Bridge.sol
```diff
@@ -1541,6 +1563,72 @@ contract Bridge is
...
     }
 
+    /// @notice Allows authorized controllers to increase Bank balances via the
+    ///         Bridge.
+    /// @dev "Controllers" here are addresses that have been authorized via
+    ///      `setAuthorizedBalanceIncreaser` and stored in the
+    ///      `authorizedBalanceIncreasers` mapping.
+    /// @param recipient Address receiving the balance increase.
+    /// @param amount Amount by which the balance is increased.
+    function controllerIncreaseBalance(address recipient, uint256 amount)
```
lrsaturnino: I noticed that `controllerIncreaseBalance` only check authorization but don't enforce any amount limits, rate limits, or caps, which means authorized controllers have unrestricted minting power. While the documentation indicates this is intentional and delegates amount validation to controller implementations, I'd like to confirm: is it a requirement that controller contracts implement their own minting limits? If a controller is compromised, the blast radius would be system-wide without Bridge-level circuit breakers. 

I'd consider adding explicit documentation warnings in `IBridgeMintingAuthorization.sol` about this responsibility, or Bridge-level safeguards like per-controller daily mint caps and enhanced monitoring events (Watchdog?).

antomor (commented):
if we plan in future to have contracts with an approach similar to the one adopted by `MintBurnGuard` where we keep track of other contract addresses (e.g.: Bank, Vault, Bridge) maybe we can consider using a Registry to centralize the maintenance and simplify their management. Not sure if it's needed for this though.

lrsaturnino (dismissed):
Took a deep dive into this and it LGTM. The separation between Bridge (who can mint) and MintBurnGuard (caps, pause, rate limits) is clean and makes the security model easy to reason about. I initially flagged a few things as concerns but after actually tracing through the code, they're all intentional - the Bank's handling of zero-address, the batch validation, test coverage - it all checks out. Nice touch on tracking mints AND burns together in totalMinted - that way the cap actually reflects real exposure, not just minting volume.

solidity/contracts/test/MockBurnBank.sol
```diff
@@ -0,0 +1,13 @@
...
+
+pragma solidity 0.8.17;
+
+/// @notice Minimal mock implementing the Bank-like burn interface expected
+///         by MintBurnGuard in tests.
+contract MockBurnBank {
+    uint256 public lastBurnAmount;
+
+    function decreaseBalance(uint256 amount) external {
+        lastBurnAmount = amount;
```
RobertoMSousa: What about adding a new event here to help with debugging the tests?

solidity/deploy/44_deploy_mint_burn_guard.ts
```diff
@@ -0,0 +1,38 @@
...
+  // Allow overriding the MintBurnGuard owner and controller via env vars.
+  // Defaults:
+  // - owner: deployer
+  // - controller: unset (0x0) and configured later via governance.
+  const owner =
+    process.env.MINT_BURN_GUARD_OWNER &&
+    process.env.MINT_BURN_GUARD_OWNER.length > 0
+      ? process.env.MINT_BURN_GUARD_OWNER
+      : deployer
+
```
RobertoMSousa: Having a validation if the "owner" is a valid address would help stop the process before moving to deployment

PR #924, opened 2026-01-21
feat: replace ethers v5 with @ethersproject modules

DefiCake:
This increases compatibility of the npm package with other projects that might be using different ethers versions. 

The monolithic `ethers` v5 package should be used in final projects; for libraries and SDKs, it is ill advised because it will cause problems during dependency resolution.

As a note, I had to also add a git resolution in the typescript `package.json` because my yarn setup would fail when fetching it. Feel free to fork this branch and remove that change if it is not suitable for your needs.

### DIFF
--- a/typescript/package.json
+++ b/typescript/package.json
@@ -27,6 +27,16 @@
   "dependencies": {
     "@bitcoinerlab/secp256k1": "^1.0.5",
     "@coral-xyz/anchor": "0.28.0",
+    "@ethersproject/abi": "^5.5.2",
+    "@ethersproject/abstract-signer": "^5.5.2",
+    "@ethersproject/bignumber": "^5.5.2",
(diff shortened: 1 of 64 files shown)


mswilkison:
Thanks @DefiCake — the idea here was right, and we've finally acted on it. Rather than rebasing this branch across six months of SDK churn (testnet4, StarkNet/SUI work), the same change has been regenerated on current `main` in PR #1033, crediting this PR. Two deltas from your version, per your own notes: the `resolutions` workaround is omitted, and tests keep using the `ethers` devDependency (only `src/`/`dist` matter for the runtime-dependency claim).

Closing in favor of #1033. A separate upcoming PR tackles the deeper issue your description pointed at (cross-version compatibility) by hiding the chain library behind an SDK-owned interface.

PR #937, opened 2026-03-20
fix(starknet): enforce strict fee equality in StarkNetBitcoinDepositor

Mavline:
## Summary

Replace the weak fee check (`>=`) with strict equality (`==`) in `StarkNetBitcoinDepositor._transferTbtc()` to prevent callers from accidentally losing excess ETH sent above the required StarkGate bridge fee.

## Problem

`StarkNetBitcoinDepositor._transferTbtc()` uses `require(msg.value >= fee)` (line 113), then forwards **all** `msg.value` to StarkGate via `starkGateBridge.deposit{value: msg.value}(...)`. StarkGate does not refund excess ETH — any amount above the estimated fee is permanently lost.

All Wormhole-based depositors in the same codebase already use strict equality:
- `L1BTCDepositorWormhole.sol:193` — `require(msg.value == cost)`
- `BTCDepositorWormhole.sol:114` — `require(msg.value == wormholeMessageFee)`

The StarkNet depositor is the only implementation using `>=`, creating an inconsistent and unsafe pattern.

## Fix

Single-line change: `require(msg.value >= fee)` → `require(msg.value == fee, "Incorrect L1->L2 message fee")`.

This ensures callers send exactly the required fee, consistent with all other depositor implementations. Callers that already send the exact fee are unaffected; callers that overpay will now get a clear revert instead of silent ETH loss.

### DIFF
--- a/solidity/contracts/cross-chain/starknet/StarkNetBitcoinDepositor.sol
+++ b/solidity/contracts/cross-chain/starknet/StarkNetBitcoinDepositor.sol
@@ -108,9 +108,11 @@ contract StarkNetBitcoinDepositor is AbstractL1BTCDepositor {
         override
     {
         // This function is called by finalizeDeposit which is payable
-        // The caller must send ETH to cover the StarkGate bridge fee
+        // The caller must send ETH to cover the StarkGate bridge fee.
+        // Use strict equality to prevent excess ETH loss, consistent
+        // with the Wormhole depositor implementations.
         uint256 fee = estimateFee();
-        require(msg.value >= fee, "Insufficient L1->L2 message fee");
+        require(msg.value == fee, "Incorrect L1->L2 message fee");
         require(address(tbtcToken) != address(0), "tBTC token not initialized");
 
         // Convert bytes32 to uint256 for StarkNet address format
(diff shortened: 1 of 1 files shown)


mswilkison:
Thanks @Mavline — the diagnosis here is correct and still applies to `main` unchanged (the file is byte-identical to this PR's base blob).

I've opened #1026, which is this fix rebased onto current `main`, with you credited via `Co-authored-by`. Two things it adds:

1. **The revert-string rename needed test updates.** `"Insufficient L1->L2 message fee"` is asserted at `StarkNetBitcoinDepositor.test.ts:278` and `starknet/CoreLogicTests.sol:235`, so this branch would have gone red on CI. Both are updated there.
2. **Regression coverage** for the excessive-fee case, verified non-vacuous: reverting the contract to `>=` makes it fail with *"Expected transaction to be reverted"*, confirming that on the old code overpaying succeeds and silently burns the surplus.

One correction to the reasoning in the description, in your favor: I initially worried `==` would be fragile because `estimateFee()` reads from StarkGate. It isn't — `estimateDepositFeeWei()` is `external pure`, returning a hardcoded `20000 * 5 gwei`, so there's no per-block drift and `==` is safe. Your fix was right as written.

Severity is low in practice (StarkGate caps `msg.value` at `MAX_FEE = 0.01 ETH` and reverts past it, and the relayer sends the exact constant today), but the improved failure mode is worth having.

Closing this in favour of #1026 once that lands — happy to keep it open until then.

mswilkison:
Closing as superseded by #1026, which is now merged. It carries this strict-fee fix along with the necessary regression coverage and the mirrored change in the separately deployed StarkNet package.

PR #944, opened 2026-03-30
feat(relay): Minimum-difficulty handling in retarget (testnet-style blocks)

lionakhnazarov:
Problem
During retarget, after the first header of the new epoch sets minedTarget, the contract required every following header in the post-retarget half of the bundle to use exactly that same target. On Bitcoin testnet4, a minimum-difficulty block (bits like 0x1d00ffff) can appear in that window with a different expanded target than the retargeted blocks, which triggered Unexpected target change after retarget.

Introduces a dedicated minimum difficulty target constant aligned with Bitcoin’s difficulty-1 target (BTCUtils.DIFF1_TARGET / compact 0x1d00ffff), to model minimum-difficulty blocks that can appear on Bitcoin testnets (e.g. testnet4) when block spacing is high, while other blocks in the same post-retarget window use the new epoch difficulty after a retarget.

Smart contract (LightRelay.sol)
New constant MIN_DIFFICULTY_TARGET — same 256-bit value as DIFF1_TARGET, documented as the on-chain representation of minimum-difficulty headers in those scenarios.
retarget logic — after the first post-retarget header establishes minedTarget, remaining headers in the post-retarget proof must either:
match minedTarget, or
match MIN_DIFFICULTY_TARGET
so the relay accepts mixed post-retarget windows (normal retargeted difficulty + occasional minimum-difficulty blocks) instead of reverting with "Unexpected target change after retarget".
PoW and chain validation still run first via validateHeader; the minimum-difficulty exception only applies after minedTarget is fixed and only where the contract explicitly allows it.

Tests (LightRelay.test.ts)
Cover post-retarget behavior when nbits is forced to the minimum-difficulty encoding vs an unrelated easy target, including:
revert on Invalid work when DIFF1 nbits is applied without a matching PoW (shows the exception does not skip work checks);
revert on Unexpected target change after retarget when a header uses a different easy target than minedTarget / MIN_DIFFICULTY_TARGET.
Helpers to parse compact bits and patch nbits on fixture headers for these scenarios.

### DIFF
--- a/solidity/contracts/relay/LightRelay.sol
+++ b/solidity/contracts/relay/LightRelay.sol
@@ -115,6 +115,13 @@ contract LightRelay is Ownable, ILightRelay {
 
     mapping(address => bool) public isAuthorized;
 
+    /// @dev Same value as `BTCUtils.DIFF1_TARGET` (compact bits `0x1d00ffff`).
+    /// Bitcoin testnet4 may emit blocks at minimum difficulty between retargets
+    /// while other blocks in the same retarget proof window use the new epoch
+    /// difficulty. Mainnet does not produce such blocks at meaningful heights.
+    uint256 private constant MIN_DIFFICULTY_TARGET =
+        0xffff0000000000000000000000000000000000000000000000000000;
+
     modifier relayActive() {
         require(ready, "Relay is not ready for use");
         _;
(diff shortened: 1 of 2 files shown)


lrsaturnino (commented):
Hey, nice work on the testnet4 min-difficulty handling — the constant is correct and the post-retarget tests are well-designed. A few things worth discussing before merge.

solidity/contracts/relay/LightRelay.sol
lrsaturnino: `validateChain()` at line 498 was not updated with the same `MIN_DIFFICULTY_TARGET` exception, so headers that pass `retarget()` would be rejected by `validateChain()`. If the logic stays in the base contract, this needs the same treatment — if moved to the testnet subclass, both overrides live together.

PR #957, opened 2026-05-04
feat(bridge,vault): add PSBT covenant migration debt and reveal guard

lrsaturnino:
## Audit positioning

This PR is the **Bridge + Vault audit baseline** for the Trail of Bits PSBT covenant assessment. Per the engagement-letter amendment of 2026-05-09, the SOW Scope URLs were rewritten to:

- `tlabs-xyz/tbtc-v2-ac#263` — AC smart contracts (consolidated baseline; supersedes `tbtc/#132` and `tbtc/#217`)
- **this PR** — Bridge migration-debt and reveal-guard work, including TBTCVault changes
- `threshold-network/keep-core#3882` — covenant signer

Coverage of this PR: Bridge migration-debt accounting and reveal guard (`Bridge.sol`, `BridgeState.sol`, `BridgeGovernance.sol`, `Deposit.sol`, `DepositSweep.sol`, `MigrationExtraData.sol`); vault migration-sweep hooks and optimistic-minting migration guard (`TBTCVault.sol`, `TBTCMigrationDebtOperations.sol`, `TBTCOptimisticMinting.sol`); storage-layout delta for proxy upgrade — `BridgeState.__gap` 48 → 47, new `migrationDebtVault` field.

## Where to start reading

1. `solidity/docs/BRIDGE_STORAGE_UPGRADE_NOTES.md` — required reading for SOW Smart-Contracts Q8 (storage upgrade considerations: compromised `GOVERNANCE_ROLE` upgrade, ERC-7201 namespace collisions, `_revertIfLegacyStorageOccupied`)
2. `solidity/contracts/bridge/MigrationExtraData.sol` — encoding of extra-data fields used during covenant migration
3. `solidity/contracts/bridge/Deposit.sol` — new migration-aware reveal guard with `MigrationRevealRejected(bytes32)` reason codes
4. `solidity/contracts/bridge/BridgeState.sol` — storage layout delta
5. `solidity/contracts/bridge/Bridge.sol` — migration-debt vault wiring, `setMigrationDebtVault` / `rotateMigrationDebtVault`
6. `solidity/contracts/vault/TBTCMigrationDebtOperations.sol` — migration-debt accounting library
7. `solidity/contracts/vault/TBTCOptimisticMinting.sol` — `MIN_OPTIMISTIC_MINTING_FEE_DIVISOR = 10`, migration-aware mint gate
8. `solidity/contracts/bridge/DepositSweep.sol` — migration-sweep callback path with bounded fan-out

## Invariants and reason codes (to verify)

**Invariants asserted by this PR:**

- For every revealer, `outstandingMigrationDebt(revealer) = sum(registered_deposits.amount) − sum(repaid_deposits.amount)` (per-revealer accounting in `TBTCMigrationDebtOperations`).
- `setVaultStatus(canonicalMigrationDebtVault, untrusted)` always reverts while it is the canonical pointer.
- Any vault still reporting outstanding migration debt cannot be untrusted — fail-open staticcall (vaults not implementing the interface are unaffected).
- `rotateMigrationDebtVault` refuses to leave the prior canonical vault holding outstanding debt.
- Optimistic-minting fee divisor lower bound: actual divisor ≥ `MIN_OPTIMISTIC_MINTING_FEE_DIVISOR (= 10)` — caps the maximum minting fee at 10%.
- `MAX_MIGRATION_SWEEP_BATCH_SIZE = 100` caps per-sweep callback fan-out.
- `migrationSweepNotifier` is informed exactly once when a revealer's migration debt reaches zero, with pending-completion bookkeeping that survives partial batch outcomes.

**`MigrationRevealRejected(bytes32)` reason codes** (raised in `Deposit.revealDeposit`):

| Reason | Trigger |
|---|---|
| `missing debt` | Deposit carries the migration tag but the per-revealer outstanding debt is zero |
| `missing tag` | Deposit comes from a registered migration revealer but lacks the migration tag |
| `mismatched revealer` | Migration tag references a revealer that does not match the per-deposit vault registration |
| `unauthorised revealer` | Migration revealer is not registered in the per-deposit vault |
| `vault call failure` | Staticcall to the vault's `ITBTCVaultMigrationDebt` interface reverts |
| `bad vault response` | Vault returns malformed data for the debt query |

**Fail-open staticcall semantics:** when a vault does not implement `ITBTCVaultMigrationDebt` at all (no selector), the `setVaultStatus` and migration-debt callback paths treat the vault as having zero migration debt — by design, so non-migration vaults are not punished. Vaults that *do* implement the interface but return malformed data are caught by `bad vault response`.

## Cross-repo coupling

This PR is the **ABI source-of-truth** that `tlabs-xyz/tbtc-v2-ac` vendors as `external/tbtc-v2/<sha>/`. The interfaces touched here (`ITBTCVaultMigrationDebt`, `ITBTCVaultMigrationSweepHook`, `ITBTCVaultMigrationSweepNotifier`) are consumed by covenant code in the sibling repo. After this PR merges, `tbtc-v2-ac` will re-pin its Bridge artifact tarball to the merged SHA.

## Deferred items (intentionally out of scope of this PR)

- **`setMigrationDebtVault` rotation-guard bypass** — flagged during PR #957 development and explicitly deferred to a follow-up. Auditors should note the current rotation flow and flag any new exploit path *beyond* this known concern; the known concern itself is tracked separately.
- **`Bridge` contract size exceeds EIP-170** — `Bridge` compiles to 26.156 KB vs the 24.576 KB cap. Resolution (delegatecall library extraction or `contractSizer` carve-out) is a separate change; see *Known issues for upstream policy* below.
- **3 test failures in `TBTCVault.OptimisticMinting.test.ts`** — divisor `type(uint32).max` does not actually truncate to zero against the fixture amount; documented in detail below.

---

## Introduction

This PR ports the on-chain machinery for the PSBT covenant migration flow into the canonical `threshold-network/tbtc-v2` source tree. The mechanism enables a one-way migration channel where a designated vault accumulates *migration debt* as deposits are revealed under a tagged extra-data byte string, and that debt is repaid when the matching sweep transaction is proven on-chain. The intent is for this repository to be the source of truth and audit baseline for the Bridge half of the covenant stack, in parallel with the covenant contracts and services that land in `tlabs-xyz/tbtc-v2-ac`.

## Provenance

The same code and tests merged into `tlabs-xyz/tbtc` on PR tlabs-xyz/tbtc#132 at merge commit tlabs-xyz/tbtc@02ee4bf1762ea2b8000b988345849645b15b672a. This PR ports the 37 files that live under `contracts/tbtc-v2/...` in the monorepo to their canonical paths under `solidity/...` here.

Bridge contracts that diverged between the monorepo's vendored copy and `threshold-network/tbtc-v2/main` (`Bridge.sol`, `Deposit.sol`) were three-way merged so that the TIP-109 hotfix (PR #948) is preserved alongside the migration debt additions; verified by `grep` on `initializeV5_RepairRebateStaking`, `RebateStakingRepaired`, `TreasuryFeeType.Deposit` and the new `MigrationDebtVaultUpdated` / `MigrationRevealRejected` symbols.

### Bridge changes

Five contracts modified (`Bridge.sol`, `BridgeGovernance.sol`, `BridgeState.sol`, `Deposit.sol`, `DepositSweep.sol`) plus one new (`MigrationExtraData.sol`).

A canonical migration debt vault pointer is maintained by governance via `setMigrationDebtVault` and an atomic `rotateMigrationDebtVault` that refuses to leave the prior canonical vault holding outstanding debt. `setVaultStatus` is tightened so the canonical migration debt vault cannot be untrusted while it is the canonical pointer, and any vault that still reports outstanding migration debt cannot be untrusted (fail-open staticcall — vaults not implementing the interface are unaffected).

A reveal-time guard is added in `Deposit.revealDeposit`: deposits whose extra data carries the migration tag must come from a registered migration revealer in the per-deposit vault and must have outstanding debt to repay; deposits from a registered migration revealer that lack the tag are rejected. Failures surface explicit reasons via `MigrationRevealRejected(bytes32)` with reason codes covering missing debt, missing tag, mismatched revealer, unauthorised revealer, vault call failure and bad vault response.

`__gap` in `BridgeState` is reduced from 48 to 47 to introduce `migrationDebtVault` while preserving storage layout continuity for proxy upgrades; documented in `solidity/docs/BRIDGE_STORAGE_UPGRADE_NOTES.md`.

### DepositSweep changes

After a sweep is proven, the canonical migration vault is notified via `ITBTCVaultMigrationSweepHook` with a fail-open batch call followed by bounded per-revealer fallback retries so a single bad entry cannot strand a completion batch.

### Vault changes

Two contracts modified (`TBTCOptimisticMinting.sol`, `TBTCVault.sol`) plus three new interfaces and one library.

`TBTCOptimisticMinting` now implements `ITBTCVaultMigrationDebt` with per-revealer migration debt tracking, a registration counter that enforces `hasOutstandingMigrationDebt` across rotations, and a `TBTCMigrationDebtOperations` library that isolates the debt state transitions (register, repay, residual clear). `MIN_OPTIMISTIC_MINTING_FEE_DIVISOR = 10` is introduced to bound the maximum optimistic minting fee rate at 10%, and `MAX_MIGRATION_SWEEP_BATCH_SIZE = 100` caps per-sweep callback fan-out.

`migrationSweepNotifier` and `migrationSweepReserve` allow a downstream contract to be informed exactly once when a revealer's migration debt reaches zero, with pending-completion bookkeeping that survives partial batch outcomes.

### Test infrastructure

Eleven harness contracts are added under `solidity/contracts/test/` so the migration paths can be exercised in isolation from the full Bridge fixture (`BridgeForVaultHarness`, `TBTCVaultHarness`, the `Deposit*` and `DepositSweep*` harnesses, plus mocks for migration debt vault, sweep notifier, sweep vault and a regular comparison vault). A NatSpec verification script at `solidity/test/verify-migration-debt-natspec.sh` asserts that the required documentation surface for the migration debt interface remains present.

## Solution

The 37-file diff decomposes as 22 added files and 15 modified files. For 13 of the 15 modified files the monorepo's pre-PR-132 vendored copy was byte-identical to `threshold-network/tbtc-v2/main`, so they were applied as direct overwrites. The remaining 2 (`Bridge.sol`, `Deposit.sol`) were three-way merged with `git merge-file` against the common base; both merges were clean (zero conflict markers) and verified by symbol grep to contain both branches' additions.

One deliberate adaptation versus the monorepo source: the rewrite of `solidity/test/fixtures/bridge.ts` introduced by PR tlabs-xyz/tbtc#132 was an accommodation for the monorepo's deploy-script chain. In `threshold-network/tbtc-v2/main` the existing fixture works against the in-tree deploy chain, so this PR keeps the bare-main fixture and only adapts what the migrated tests actually need. None of the migrated tests destructure additional fixture fields, so no signature change was required. Verified locally: with PR tlabs-xyz/tbtc#132's fixture rewrite, all modified tests fail in the `before all` hook with a JSON-RPC QUANTITY validator error from the inline `helpers.upgrades.deployProxy` call; with the bare-main fixture they execute cleanly.

## Tests

Local verification on Node 14.21.3 with `yarn 1.22.22` against `solidity/`.

`yarn lint` reports 0 errors and 363 warnings (warnings are pre-existing). `yarn format` (root) reports clean ("All matched files use Prettier code style!"). `npx hardhat compile` is clean. `solidity/test/verify-migration-debt-natspec.sh` reports 12/12 PASS.

The five new migration test files (`Bridge.CanonicalMigrationDebtVaultGuard.test.ts`, `Deposit.MigrationRevealGuard.Harness.test.ts`, `DepositSweep.MigrationCallback.test.ts`, `TBTCVault.MigrationDebt.test.ts`, `TBTCVault.OptimisticMintMigrationGuard.test.ts`) report **52/52 PASS** in 15s.

The three modified test files (`Bridge.Governance.test.ts`, `Bridge.Parameters.test.ts`, `TBTCVault.OptimisticMinting.test.ts`) report **539/542 PASS, 3 FAIL** — see *Known issues for upstream policy* below.

## Known issues for upstream policy

These two conditions exist in the source PR tlabs-xyz/tbtc#132 and are surfaced here intentionally rather than papered over. Both are policy decisions for this repository's reviewers.

### 1. `Bridge` contract size

`Bridge` compiles to **26.156 KB** with this PR vs **23.939 KB** on bare `main`. The added migration-vault rotation, reveal-guard helpers, custom errors and event accounting add ~2.2 KB. EIP-170 caps mainnet deployable code at 24.576 KB, so `hardhat-contract-sizer` (which is configured `strict: true` in `hardhat.config.ts`) fails the build.

The options are to extract a chunk of Bridge into a delegatecall library (preferred for long-term carrying capacity, but a non-trivial refactor); to accept a temporary `except: ["Bridge$"]` entry in `contractSizer` with a tracked follow-up issue; or to leave the build red until extraction lands. This PR does *not* modify the `contractSizer` configuration. The deployment blocker is real and should be reviewed alongside the rest of the change.

### 2. Three test failures in `TBTCVault.OptimisticMinting.test.ts`

PR tlabs-xyz/tbtc#132 introduced `MIN_OPTIMISTIC_MINTING_FEE_DIVISOR = 10` and rewrote the long-standing "when the optimistic minting fee is zero" context (which previously set divisor `0`) into a new "when the optimistic minting fee rounds to zero" context with `divisor = type(uint32).max = 4294967295`. The test comments assert that integer division truncates the fee to zero. The arithmetic does not actually truncate to zero for the deposit amount used in the fixture: `199_900_000_000_000 / 4_294_967_295 = 46_542`, so the treasury receives 46 542 wei and the depositor receives that amount less than the assertion expects. The three failing assertions are `should send no optimistic mint fee to treasury` (expected `0`, got `46542`), `should mint TBTC to depositor` (expected `199_900_000_000_000`, got `199_899_999_953_458`), and a second `should mint TBTC to depositor` (expected `200_000_000_000_000`, got `199_999_999_953_434`).

This was not caught upstream because the monorepo CI's `ci:core` script only runs `pnpm --filter @keep-network/tbtc-v2 run build` for the contracts package, not `run test`. Resolution requires either picking a divisor that actually truncates against the fixture amount (e.g. one larger than the deposit amount) or revisiting the `MIN_OPTIMISTIC_MINTING_FEE_DIVISOR` enforcement. This PR keeps the tests as PR tlabs-xyz/tbtc#132 wrote them so the situation is visible to upstream.

## Storage upgrade considerations

`BridgeState.Storage` adds `migrationDebtVault` and reduces the trailing `__gap` from `uint256[48]` to `uint256[47]`. Any future storage-layout diff against current proxies must reflect this and any proxy upgrade plan should include a layout review.


piotr-roslaniec (commented):
(superseded by inline comments)

solidity/contracts/bridge/Deposit.sol
```diff
@@ -207,6 +224,51 @@ library Deposit {
...
+                bytes32 canRevealLookupFailureReason
+            ) = getCanRevealMigration(reveal.vault, msg.sender);
+            if (!canRevealMethodAvailable) {
+                revert MigrationRevealRejected(canRevealLookupFailureReason);
+            }
+            if (!canReveal) {
+                revert MigrationRevealRejected(MIGRATION_DEBT_MISSING_REASON);
+            }
+        } else {
+            if (isRegisteredMigrationRevealer(self, reveal.vault, msg.sender)) {
```
piotr-roslaniec, line 284: The NatSpec says fail-closed is intentional because a failed canonical vault call makes "migration-tag enforcement indeterminate." Makes sense for the migration grant path. But this function is also called in the non-migration `else` branch, so a broken canonical vault locks out regular depositors too, not just migration ones. Was that consequence considered, or should the blocking check be fail-open?

solidity/contracts/bridge/DepositSweep.sol
```diff
@@ -259,6 +276,58 @@ library DepositSweep {
...
+
+        if (batchSuccess) {
+            return;
+        }
+
+        emit MigrationSweepCallbackFailed(vault, sweepTxHash);
+
+        // Best-effort fallback: retry each revealer in isolation so one bad
+        // entry cannot strand the whole completion batch.
+        for (uint256 i = 0; i < revealers.length; i++) {
```
piotr-roslaniec: The raw `revealers` array gets passed to the vault, which rejects batches over 100. There is no cap on sweep inputs at the Bridge level (`inputsCount` comes straight from the Bitcoin tx varint), so a sweep with 101+ deposits always fails the batch and falls through to O(N) individual vault calls inside `submitDepositSweepProof`. The callback tests only cover 1 revealer -- was this tested for gas with large migration sweeps?

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -496,28 +597,221 @@ abstract contract TBTCOptimisticMinting is Ownable {
...
+    }
+
+    function setMigrationSweepNotifier(address notifier) external onlyOwner {
+        address previousNotifier = migrationSweepNotifier;
+        migrationSweepNotifier = notifier;
+
+        emit MigrationSweepNotifierUpdated(previousNotifier, notifier);
+    }
+
+    function setMigrationSweepReserve(address revealer, address reserve)
```
piotr-roslaniec, line 808: The dedicated `MigrationSweepCompletionPendingCleared` event makes it clear the coupling is intentional. But `setMigrationSweepReserve` has no NatSpec at all, so an operator zeroing the reserve during an incident would have no hint that it also kills the pending callback. Worth adding a note there.

solidity/contracts/vault/TBTCMigrationDebtOperations.sol
```diff
@@ -0,0 +1,146 @@
...
+            revealer != address(0),
+            "Migration revealer can not be the zero address"
+        );
+        require(amount > 0, "Amount must be greater than 0");
+        require(
+            migrationDebt[revealer] == 0,
+            "Migration debt already registered"
+        );
+
+        uint256 nextDebt = migrationDebt[revealer] + amount;
```
piotr-roslaniec: The NatSpec return value already says "equals `amount` since registration starts from zero," and the `require(migrationDebt[revealer] == 0)` three lines above guarantees it. Safe to drop the addition and assign directly?

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -562,4 +856,73 @@ abstract contract TBTCOptimisticMinting is Ownable {
...
+    }
+
+    function requireCanonicalMigrationDebtVault() internal view {
+        require(
+            bridge.migrationDebtVault() == address(this),
+            "Bridge migration debt vault mismatch"
+        );
+    }
+
+    function isMigrationReveal(bytes32 extraData) internal pure returns (bool) {
```
piotr-roslaniec: Both `Deposit.sol` (line 478) and this file have an identical `isMigrationReveal` helper. `MigrationExtraData` already owns the tag constant; was putting the helper there considered?

solidity/contracts/vault/TBTCOptimisticMinting.sol
```diff
@@ -496,28 +597,221 @@ abstract contract TBTCOptimisticMinting is Ownable {
...
+        TBTCMigrationDebtOperations.setRevealer(
+            isMigrationRevealer,
+            revealer,
+            allowed
+        );
+
+        emit MigrationRevealerSet(revealer, allowed);
+    }
+
+    function setMigrationSweepNotifier(address notifier) external onlyOwner {
```
piotr-roslaniec, line 775: No `code.length > 0` check on the setter (no other setter in vault/bridge does this either, so may be accepted practice). Worth noting: an EOA notifier causes every `notifyMigrationSweep` to revert inside the vault via Solidity's EXTCODESIZE check. Those reverts get swallowed by the low-level `.call` in `notifyMigrationSweepCallback`, so `MigrationSweepCallbackRetryFailed` fires for every revealer silently. Intentional risk or worth a guard at set-time?

solidity/contracts/bridge/Deposit.sol
lrsaturnino: Intentional fail-closed — covered by `solidity/test/bridge/Deposit.MigrationRevealGuard.Harness.test.ts:445-478`. The non-migration-depositor lockout observation is correct, but a failed canonical lookup makes migration-tag enforcement indeterminate, so reverting is the safer policy. A future hardening could be a set-time interface probe on `setMigrationDebtVault` to convert this to a deploy-time check; tracking as a separate follow-up.

solidity/contracts/bridge/DepositSweep.sol
lrsaturnino: Addressed in commit d14c2e42 — chunked at `MAX_MIGRATION_SWEEP_BATCH_SIZE = 100` (`solidity/contracts/bridge/DepositSweep.sol:51,307-380`). Tests at `solidity/test/bridge/DepositSweep.MigrationCallback.test.ts:93,125` cover the 150-revealer happy path and the chunked fallback (2 batch failures + 150 single-hook successes).

solidity/contracts/vault/TBTCOptimisticMinting.sol
lrsaturnino: Addressed in commit d14c2e42 — NatSpec block added above `solidity/contracts/vault/TBTCOptimisticMinting.sol:799` documenting that `reserve == address(0)` clears `pendingMigrationSweepCompletion[revealer]` and emits `MigrationSweepCompletionPendingCleared`.

solidity/contracts/vault/TBTCMigrationDebtOperations.sol
lrsaturnino: Addressed in commit d14c2e42 — `solidity/contracts/vault/TBTCMigrationDebtOperations.sol:37` now assigns `amount` directly. The existing `require(migrationDebt[revealer] == 0)` makes the prior addition equivalent.

solidity/contracts/vault/TBTCOptimisticMinting.sol
lrsaturnino: Addressed in commit d14c2e42 — helper moved into `solidity/contracts/bridge/MigrationExtraData.sol:13-15`. Callers updated at `Deposit.sol:227`, `TBTCOptimisticMinting.sol:337,399`, and `DepositRevealGuardHarness.sol:73`.

solidity/contracts/vault/TBTCOptimisticMinting.sol
lrsaturnino: Addressed in commit d14c2e42 — `solidity/contracts/vault/TBTCOptimisticMinting.sol:766` now requires `notifier == address(0) || notifier.code.length > 0`. EOA notifier reverts at set-time; `address(0)` remains a valid disable state. Tests at `solidity/test/vault/TBTCVault.MigrationDebt.test.ts:458,464`.

solidity/contracts/bridge/Deposit.sol
```diff
@@ -207,6 +224,51 @@ library Deposit {
             "Vault is not trusted"
         );
 
+        if (MigrationExtraData.isMigrationReveal(extraData)) {
```
piotr-roslaniec, line 229: Migration reveals gate on `reveal.vault` directly, bypassing the canonical vault check that the non-migration `else` branch runs via `isRegisteredMigrationRevealer`. So any trusted vault implementing `ITBTCVaultMigrationDebt` can authorize a migration reveal, not just the canonical one pointed to by `migrationDebtVault`. Is that the intended trust model, or should the canonical vault be the sole authority here?

lrsaturnino: Good catch — pinned `reveal.vault` to `self.migrationDebtVault` in the migration branch in 02f448b2; tests cover non-canonical (interface or not) and unset canonical.

PR #958, opened 2026-05-05
fix(relay): isolate DIFF1 exception to Testnet4LightRelay, fix SEC1/C2/C3/S1/SP1/C4

piotr-roslaniec:
## Summary

Follow-up to #944. Addresses all valid security and correctness findings from the post-merge review.

- **SEC1 (P0)**: Remove `MIN_DIFFICULTY_TARGET` constant from `LightRelay.sol`. The base relay no longer accepts trivially-mined DIFF1 blocks, closing the mainnet attack vector where an adversary could forge a retarget proof using ~65k hashes instead of ~10^14.
- **C2 (P1)**: Add `_isTolerableTarget` hook to both `validateChain()` target checks so testnet4 DIFF1 blocks are accepted in SPV proofs too (not just in `retarget()`).
- **C3 (P1)**: Document in `Testnet4LightRelay` NatDoc that a DIFF1 block cannot appear as the first post-retarget header -- that position must carry the new epoch target.
- **S1 (P2)**: Reference `BTCUtils.DIFF1_TARGET` (the library constant) in `Testnet4LightRelay` instead of repeating the magic literal.
- **SP1 (P1)**: NatDoc security model statement in `LightRelay` is accurate again -- exception no longer lives in base.
- **C4 (P2)**: Add `Testnet4LightRelayStub` and `Testnet4LightRelay.test.ts` covering `_isTolerableTarget` unit tests and negative cases. Positive DIFF1 acceptance tests require real testnet4 block headers (documented as TODO).

## Design

`_isTolerableTarget(uint256) internal view virtual returns (bool)` is the single extension point. The base relay always returns `false`. `Testnet4LightRelay` overrides to return `true` for `BTCUtils.DIFF1_TARGET`. All four previous hardcoded `MIN_DIFFICULTY_TARGET` comparisons are replaced with this hook.

## Test plan

- [ ] `solhint` passes on all changed/new `.sol` files
- [ ] `hardhat compile` succeeds
- [ ] `hardhat test test/relay/LightRelay.test.ts test/relay/Testnet4LightRelay.test.ts` passes (104 tests)
- [ ] CI green

### DIFF
--- a/solidity/contracts/relay/LightRelay.sol
+++ b/solidity/contracts/relay/LightRelay.sol
@@ -115,13 +115,6 @@ contract LightRelay is Ownable, ILightRelay {
 
     mapping(address => bool) public isAuthorized;
 
-    /// @dev Same value as `BTCUtils.DIFF1_TARGET` (compact bits `0x1d00ffff`).
-    /// Bitcoin testnet4 may emit blocks at minimum difficulty between retargets
-    /// while other blocks in the same retarget proof window use the new epoch
-    /// difficulty. Mainnet does not produce such blocks at meaningful heights.
-    uint256 private constant MIN_DIFFICULTY_TARGET =
-        0xffff0000000000000000000000000000000000000000000000000000;
-
     modifier relayActive() {
         require(ready, "Relay is not ready for use");
         _;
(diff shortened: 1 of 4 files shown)


lionakhnazarov (approved):
LGTM

PR #949, opened 2026-04-07
feat(testnet4)!: Bitcoin testnet4 (BIP-94) support — bridge SPV, breaking SDK switch, deploy patches

lionakhnazarov:
## Summary

This PR delivers the full `feat/testnet4-deployment-support` delta, including protocol/security fixes, deployment and artifact updates, TypeScript SDK/runtime improvements, and CI/workflow hardening needed to keep the branch mergeable.

### 1) Bridge/Relay security and correctness (Testnet4 handling)
- Hardened SPV difficulty validation by removing unsafe DIFF1 fallback paths and introducing explicit requested-difficulty determination logic in bridge proof checks.
- Scoped Testnet4 retarget exceptions to Sepolia-specific relay behavior instead of broadening base relay acceptance rules.
- Added guardrails around DIFF1 handling so exception behavior only applies in intended relay-epoch conditions.
- Extended Solidity unit coverage around Bitcoin difficulty logic and Sepolia relay overrides.

### 2) Electrum + TypeScript SDK robustness
- Replaced Node-only crypto usage with browser-compatible hashing utilities.
- Improved transaction history fetching/error propagation behavior and clarified retry/default semantics.
- Added/updated Bitcoin network handling for Testnet4 and related config URL lists.
- Added `BitcoinClientWithNetworkOverride` and updated SDK-facing docs/API references accordingly.
- Updated scripts touching DKG approval and E2E flows to match new config/security expectations.

### 3) Deployment pipeline and on-chain artifacts
- Added and refreshed Sepolia deployment artifacts and metadata across Bridge/Bank/Relay/Governance/TBTC/WalletRegistry and related contracts.
- Updated deployment patches and scripts (including governance/ownership transfer paths) for more reliable execution and idempotency.
- Added/updated testnet initialization and relay setup deploy scripts.
- Updated lockfiles and generated artifacts where required for deterministic CI and reproducible deployment state.

### 4) Solidity deploy script lint/format compliance
- Fixed deploy script lint failures (`func-names`) by naming async deploy functions in affected scripts.
- Applied Prettier-required formatting adjustments (including Solidity `.gitignore`) to satisfy `contracts-format`.

### 5) CI/workflow hardening (multi-workflow)
- Standardized failing jobs onto Node `20.19.0` and `actions/setup-node@v4` where dependency engines required newer runtimes.
- Fixed Yarn Classic/Berry mismatches across workflows by explicitly preparing Berry where Berry lockfiles are consumed.
[...]

PR #959, opened 2026-05-05
fix: security and browser-compat follow-ups from testnet4 review

piotr-roslaniec:
## Summary

Follow-up to #949 addressing three validated security and compatibility findings from the multi-lens review.

### Security / correctness fixes

- **[P0]** \`BitcoinTx.determineRequestedDifficulty\`: removed the \`return 1\` fallback that let an attacker mine 6 trivial DIFF1 headers (~microseconds per ASIC) to pass SPV proof verification on mainnet, bypassing the relay oracle entirely. Now reverts if all headers are minimum-difficulty with no relay-epoch match — operators must call \`SepoliaLightRelay.setDifficultyFromHeaders\` with actual headers before accepting all-DIFF1 proofs.
- **[P1]** \`LightRelay.retarget\`: the base contract previously accepted \`MIN_DIFFICULTY_TARGET\` headers in the retarget proof window, weakening relay epoch integrity for mainnet. Extracted target checks into \`isValidPreRetargetTarget\` / \`isValidPostRetargetTarget\` virtual helpers (strict by default). \`SepoliaLightRelay\` now overrides both to permit DIFF1, which is the correct scope for the testnet4 BIP94 exception.
- **[P2]** \`ElectrumClient\`: replaced \`import * as crypto from "crypto"\` (Node.js-only) with \`BitcoinHashUtils.computeHash256\` which uses \`ethers.utils.sha256\` and works in browser and Node environments. Also replaced the \`switch\` in \`fromDefaultConfig\` with a lookup map (removes static-analysis false-positive) and converted \`getTransactionHistory\` to \`Promise.allSettled\` (explicit rejection surfacing).

### CI fixes (Yarn Berry migration from #949)

#949 introduced a Berry 4.x lockfile (\`__metadata: version: 8\`) in \`solidity/\` but left CI workflows using bare \`corepack enable\`, which cannot auto-detect the Berry version without a \`packageManager\` field. This PR completes the migration:

- **\`contracts.yml\`** — add \`corepack prepare yarn@4.12.0 --activate\` to all 7 jobs so Berry is explicitly activated before \`yarn install\`. Move the root-workspace install (Classic lockfile) before Berry activation so each lockfile is handled by the correct Yarn version. Use \`--frozen-lockfile\` (Classic flag) instead of \`--immutable\` (Berry-only) for that step.
- **\`contracts-docs.yml\`** — the reusable \`keep-network/ci\` workflow runs Yarn Classic internally and cannot be modified. Use \`preProcessingCommand\` to download Berry via corepack and install a thin shell shim at \`/tmp/yarn\` so subsequent steps see Berry. Fixed shim construction to use the corepack v1 storage path (\`v1/yarn/4.12.0/yarn.js\`) rather than a non-existent \`bin/\` subdirectory.
- **\`solidity/package.json\`** — removed \`packageManager: yarn@4.12.0\`; the reusable docs workflow treats that field as a hard version pin and fails when it encounters Classic.
- **\`@openzeppelin/upgrades-core\`** — pinned to 1.20.0 to avoid a transitive dep that requires Node >=18.17.0 while CI pins Node 18.15.0.
- **\`solidity/.yarnrc\`** — removed; was added as a band-aid for Classic accidentally running in \`solidity/\`, but is unreachable now that Berry handles all \`solidity/\` installs (Berry ignores \`.yarnrc\`).

## Test plan

- [x] \`cd solidity && yarn hardhat compile\` — compiles without errors
- [x] \`cd typescript && yarn tsc --noEmit\` — no type errors in changed files
- [ ] CI passes on this branch

### DIFF
--- a/.github/workflows/contracts-docs.yml
+++ b/.github/workflows/contracts-docs.yml
@@ -49,8 +49,7 @@ jobs:
       # command substitutes `///` + newline + `//` + newline with just `///` +
       # newline and does this in a loop.
       preProcessingCommand: >-
-        corepack enable &&
-        corepack prepare yarn@4.12.0 --activate &&
+        npm install --global yarn@4.12.0 &&
         sed -i ':a;N;$!ba;s_///\n//\n_///\n_g' ./contracts/bridge/BitcoinTx.sol
       publish: false
       addTOC: false
@@ -77,8 +76,7 @@ jobs:
       # command substitutes `///` + newline + `//` + newline with just `///` +
       # newline and does this in a loop.
       preProcessingCommand: >-
-        corepack enable &&
-        corepack prepare yarn@4.12.0 --activate &&
+        npm install --global yarn@4.12.0 &&
         sed -i ':a;N;$!ba;s_///\n//\n_///\n_g' ./contracts/bridge/BitcoinTx.sol
       publish: true
       addTOC: false
--- a/.github/workflows/contracts-docs.yml
+++ b/.github/workflows/contracts-docs.yml
@@ -50,7 +50,8 @@ jobs:
       # newline and does this in a loop.
       preProcessingCommand: |
         COREPACK_HOME=/tmp/berry-home corepack prepare yarn@4.12.0
-        BERRY=$(find /tmp/berry-home -path "*/bin/yarn*" -type f 2>/dev/null | head -1)
+        BERRY=/tmp/berry-home/v1/yarn/4.12.0/yarn.js
+        test -f "$BERRY" || { echo "ERROR: $BERRY not found"; exit 1; }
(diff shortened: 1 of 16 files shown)


lionakhnazarov (commented):
Small follow-up on `fix/testnet4-security-followup`: documents how `determineRequestedDifficulty` treats **asymmetric** relay difficulties and **DIFF1** headers, adds **two unit tests** for those cases, documents **Yarn Classic vs Berry** ordering in `contracts-format`, corrects **`ElectrumClient.fromUrl`** retry/timeout JSDoc, and **regenerates** `typescript/api-reference` so the docs check matches `client.ts`.

lionakhnazarov (approved):

PR #963, opened 2026-05-05
Sui: clarify minter capability revocation docs

mswilkison:
## Summary
- Clarify that `remove_minter` updates TokenState bookkeeping but does not destroy outstanding MinterCap objects
- Clarify that live minting also requires TreasuryCap access and can be stopped through the gateway/token pause path
- Align the Move comment and markdown docs with the live architecture

## Validation
- Documentation/comment-only change
- `git diff --check` passed for the safe-fix batch

## Notes
This is public-safe documentation cleanup for an already-duplicated report and does not include any private deployment mitigation.

## Summary by CodeRabbit

* **Documentation**
  * Updated guidance for adding and removing minters.
  * Clarified that removing a minter does not destroy existing capabilities.
  * Documented that minting requires the treasury capability and can be halted through pausing.

### DIFF
--- a/cross-chain/sui/deploy_l2/docs/tbtc.md
+++ b/cross-chain/sui/deploy_l2/docs/tbtc.md
@@ -34,7 +34,7 @@ The contract is initialized with:
 
 ## Administrative Methods
 
-### `add_minter`
+### `add_minter_with_cap`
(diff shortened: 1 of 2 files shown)


lrsaturnino (approved):
LGTM

PR #968, opened 2026-05-20
Make tx max fee reimbursement best effort

mswilkison:
## Summary

Updates `AbstractL1BTCDepositor.finalizeDeposit` so the deposit transaction max fee reimbursement is best-effort instead of mandatory.

When `reimburseTxMaxFee` is enabled, the depositor now:

- forwards the base finalized tBTC amount as before
- adds the `depositTxMaxFee` reimbursement only if the contract balance can cover the fully reimbursed amount
- emits `DepositTxMaxFeeReimbursementSkipped` when the reimbursement is skipped because the shared depositor balance is short

This keeps finalization live when a small reimbursement shortfall would otherwise block forwarding tBTC to the user's final destination address.

Review follow-ups also:

- mirror the same best-effort reimbursement logic into the separate StarkNet package copy
- mirror the same guard into the standalone Wormhole V2 depositor contracts
- update NatSpec so `reimburseTxMaxFee = true` no longer reads as a guaranteed reimbursement
- add Wormhole skip-path coverage for the inherited abstract depositor flow

## Root cause

The current code always adds `depositTxMaxFee` to the amount forwarded when reimbursement is enabled. If earlier finalizations consume enough of the shared depositor contract balance, a later finalization can have enough tBTC to cover the base deposit amount but not enough to also cover the reimbursement. In that case the final transfer reverts and the depositor waits until unrelated deposits top up the contract.

## Tradeoff

This favors liveness over exact per-deposit reimbursement. A depositor may miss the `depositTxMaxFee` reimbursement when the contract balance is short, but they still receive the base finalized deposit amount immediately instead of being blocked behind later deposits.

That means reimbursement is no longer strictly guaranteed by `reimburseTxMaxFee = true`; it becomes conditional on available contract balance. The new event makes skipped reimbursements visible for monitoring or possible off-chain reconciliation. If the contract cannot cover even the base amount, finalization can still revert; this PR only prevents the extra reimbursement amount from blocking otherwise-forwardable deposits.

## Validation

```bash
cd solidity
npm run test -- test/cross-chain/wormhole/L1BTCDepositorWormhole.test.ts test/cross-chain/wormhole/L1BTCDepositorWormholeV2Arbitrum.test.ts
npm run test -- test/depositor/NativeBTCDepositor.test.ts test/cross-chain/wormhole/BTCDepositorWormhole.test.ts
npm run format

cd ../cross-chain/starknet
L1_CHAIN_SEPOLIA_API_URL=http://127.0.0.1:8545 L1_CHAIN_MAINNET_API_URL=http://127.0.0.1:8545 npm run build
npx solhint contracts/cross-chain/AbstractL1BTCDepositor.sol
npx prettier --check contracts/cross-chain/AbstractL1BTCDepositor.sol
```

Results:

- `99 passing` for `L1BTCDepositorWormhole` + `L1BTCDepositorWormholeV2Arbitrum`
- `149 passing` for `NativeBTCDepositor` + `BTCDepositorWormhole`
- Solidity package `npm run format` exits successfully; existing repo warnings remain, with no errors
- StarkNet package build succeeds with dummy local RPC URLs; direct solhint/prettier checks pass for the touched StarkNet Solidity file, with existing solhint warnings only

### DIFF
--- a/cross-chain/starknet/contracts/cross-chain/AbstractL1BTCDepositor.sol
+++ b/cross-chain/starknet/contracts/cross-chain/AbstractL1BTCDepositor.sol
@@ -110,7 +110,8 @@ abstract contract AbstractL1BTCDepositor is
 
     /// @notice **Feature Flag** controlling whether the deposit transaction max fee
     ///         is **reimbursed** (added to the user’s tBTC) or **deducted**.
-    ///         - `true`  => Add `txMaxFee` to the minted tBTC amount
+    ///         - `true`  => Add `txMaxFee` on a best-effort basis when the
+    ///                      contract balance covers it
     ///         - `false` => Subtract `txMaxFee` from the minted tBTC amount
     bool public reimburseTxMaxFee;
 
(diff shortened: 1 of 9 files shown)


lrsaturnino (approved):
LGTM

piotr-roslaniec (approved):

PR #952, opened 2026-04-22
Implement EVM L2 cross-chain fee rebates

mswilkison:
## Summary
- Adds cross-chain rebate support in RebateStaking with beneficiary authorizations, implicit same-address mode, chain/flow/action checks, and cancellation.
- Wires Bridge deposit/redemption paths plus Arbitrum/Base Wormhole depositor and redeemer flows.
- Adds the EVM L2 fee waiver spec and focused rebate tests/mocks.

## Reviewer Notes
- `notifyRedemptionVeto` now cancels direct-L1 rebates as well as cross-chain rebates. This restores rebate capacity for vetoed direct redemptions and changes pre-PR main behavior by adding a `RebateCanceled` emission path on veto.
- The Bridge optimizer override uses `runs: 1` to stay below the EIP-170 size limit. Follow-up architectural refactoring can revisit runtime gas versus deployment size.

## Verification
- npm run build
- npm run test -- --grep RebateStaking
- npm run test -- test/cross-chain/wormhole/L1BTCDepositorWormholeV2Arbitrum.test.ts test/cross-chain/wormhole/L2BTCRedeemerWormhole.test.ts test/cross-chain/wormhole/L1BTCRedeemerWormhole.test.ts
- npm run test -- test/bridge/Bridge.Deposit.test.ts test/bridge/Bridge.Redemption.test.ts test/bridge/Bridge.RedemptionVaultRebate.test.ts
- npm run lint:sol

Notes: Hardhat warns about unsupported Node v25.9.0. Build/test still pass. Bridge bytecode is 23.310 KB after the Bridge optimizer override and the strict contract-sizer gate passes.

PR #970, opened 2026-05-22
Route rebate to extraData receiver for sponsored depositors

mswilkison:
## Summary

Closes the gap where a T staker who mints through a relayer-style depositor (e.g. `NativeBTCDepositor`) loses their fee waiver because the Bridge sees the depositor contract as `msg.sender` and routes the `RebateStaking` lookup against that contract's (empty) stake instead of the L1 receiver's.

- Adds a governance-managed `sponsoredDepositors` allowlist to `BridgeState`.
- In `Deposit._revealDeposit`, when the caller is allowlisted, `extraData != bytes32(0)`, the decoded address is non-zero and canonically left-padded (high 12 bytes zero), **and** that address has separately consented via `RebateStaking.setSponsorConsent`, the rebate is keyed off the address decoded from `extraData` rather than `msg.sender`. Any missing precondition falls back to charging the depositor's own (possibly empty) stake, never reverts.
- `deposit.depositor` is unchanged — refund and finalize accounting are untouched; delegation via `RebateStaking.getStaker` keeps working.
- New `Bridge.setSponsoredDepositor(address, bool)` (onlyGovernance) + `isSponsoredDepositor(address)` view; matching `BridgeGovernance.setSponsoredDepositor(...)` forwarder.
- New event `SponsoredDepositorSet(address indexed depositor, bool sponsored)`.
- New `RebateStaking.setSponsorConsent(address depositor, bool authorized)` (caller must be a staker) + `RebateStaking.isSponsorConsentGranted(address staker, address depositor)` view (delegation-aware — resolves through `getStaker` so it mirrors what `applyForRebate` actually charges); new event `SponsorConsentSet(address indexed staker, address indexed depositor, bool authorized)`.
- Storage: `BridgeState.__gap` reduced 48 → 47 (new `sponsoredDepositors` mapping). `RebateStaking.__gap` reduced 49 → 47 (two new mappings: `rebateAuthorizations` from a prior PR plus this PR's `stakerSponsorConsent`). Both layouts verified against the deployed mainnet baselines by a new parameterized `Bridge.StorageLayout.test.ts` covering both contracts.

## Why

`depositTreasuryFeeDivisor = 500` on mainnet today (20 bps), so this is the fee a user with T staked would expect their rebate to reduce on every deposit. Without this routing, gasless minting through `NativeBTCDepositor` (live on mainnet at [`0xad7c6d46f4a4bc2d3a227067d03218d6d7c9aaa5`](https://etherscan.io/address/0xad7c6d46f4a4bc2d3a227067d03218d6d7c9aaa5)) always pays the full 20 bps because the depositor contract has no stake. After this change, the user's existing `RebateStaking` budget is honored exactly as it would be on a direct L1 reveal.

## Trust model

Activation requires **two independent steps**, both load-bearing:

1. **Governance allowlist** (`Bridge.setSponsoredDepositor` / `BridgeGovernance.setSponsoredDepositor`): governance opts a depositor contract into having its reveals' `extraData` interpreted as an L1 rebate staker address at all.
2. **Per-staker consent** (`RebateStaking.setSponsorConsent`): the L1 staker whose stake would be charged must separately authorize that specific depositor. This is not optional hardening — without it, anyone could grief a staker's rolling-window rebate cap by embedding that staker's address in a relayed deposit's `extraData` with no action or consent from the victim.

Both gates must be satisfied (plus a canonical `extraData` encoding, see below) or the rebate falls back to charging the depositor's own (possibly empty) stake — the reveal never reverts on a missing precondition. Rollout note: after the governance allowlist call below, **no existing `NativeBTCDepositor` user's rebate routes to their own stake until that user separately submits their own `setSponsorConsent` transaction** — this is a step the depositor flow, the SDK, and this PR do not automate on a staker's behalf; communicate it to affected users as part of rollout.

## Allowlist scope

The allowlist is intentionally narrow. Membership means *"`extraData` on reveals from this contract is the L1 address whose stake should be charged, once that address also grants sponsor consent."* Only `NativeBTCDepositor` satisfies that today. The allowlist ships **empty**; after merge, a single governance call adds the mainnet proxy:

```
bridgeGovernance.setSponsoredDepositor(
    0xad7c6d46f4a4bc2d3a227067d03218d6d7c9aaa5, // NativeBTCDepositor
    true
)
```

Explicitly out of scope: cross-chain depositors (Sui, StarkNet, Arbitrum/Base via Wormhole). Their `extraData` is an L2/non-EVM identifier, not an L1 staker. Cross-chain rebates are handled by #952 with its proper EIP-712/EIP-1271 authorization model.

`extraData` must be canonically left-padded (high 12 bytes zero) to route — this matches `NativeBTCDepositor`'s own `CrosschainUtils.addressToBytes32` convention and is enforced defensively for any future depositor with a different encoding; a non-canonical payload falls back to the depositor's own stake rather than reverting. A TypeScript SDK wrapper for `RebateStaking.setSponsorConsent` is out of scope for this PR — the SDK does not currently wrap any `RebateStaking` function, and adding one is a separate, standalone piece of work.

## Test plan

- [x] `npx hardhat test test/bridge/Bridge.Deposit.test.ts test/bridge/Bridge.StorageLayout.test.ts test/bridge/Bridge.Governance.test.ts test/bridge/RebateStaking.test.ts` — **570 passing**, 0 failing
- [x] Tests verify:
  - `setSponsoredDepositor` access control + event + toggle on/off
  - `setSponsorConsent` reverts for non-stakers and the zero address; sets/revokes the per-pair consent; emits `SponsorConsentSet`
  - Rebate consumes from receiver (not depositor) when the depositor is allowlisted **and** the receiver has granted consent, using a dedicated fixture with canonically-padded `extraData`
  - Rebate falls back to charging the depositor's own stake (unchanged behavior) when the depositor is not allowlisted, consent is missing, or `extraData` is non-canonically padded
  - `deposit.depositor` and `deposit.extraData` are recorded unchanged
  - Storage layout of both `Bridge` and `RebateStaking` matches their deployed mainnet baselines (parameterized regression test; independently confirmed to fail on a reverted `__gap`/mapping-order mutation)
- [x] `npx hardhat compile` clean
- [x] `npx tsc --noEmit` clean
- [x] `npx prettier --check` clean on all changed files
- [x] `solhint` — no new warnings (4 pre-existing warnings on `Bridge.sol`/`Deposit.sol`, confirmed identical on `origin/dev`)
- [x] Deploy script (`87_upgrade_bridge_sponsored_depositor.ts`) redeploys the `Deposit` library fresh (previously read the stale existing deployment record, which would have shipped this PR's entire behavioral change inert on mainnet) with a fail-loud guard against an unchanged address

### DIFF
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -244,6 +244,8 @@ contract Bridge is
         address newRebateStaking
     );
 
+    event SponsoredDepositorSet(address indexed depositor, bool sponsored);
+
     /// @notice Emitted when a deposit's vault field is corrected via governance.
     /// @dev This event is used for transparency when fixing deposits that were
     ///      revealed with incorrect vault targets.
@@ -2035,6 +2037,36 @@ contract Bridge is
     /// @return Address of the rebate staking contract.
     function getRebateStaking() external view returns (address) {
         return self.rebateStaking;
+    }
+
+    /// @notice Adds or removes a depositor contract from the sponsored
+    ///         depositor allowlist. When a depositor is on the list, reveals
+    ///         it submits route the `RebateStaking` rebate to the L1 address
+    ///         decoded from `extraData` rather than to the depositor contract
+    ///         itself. Intended for direct-L1-receiver relays such as
+    ///         `NativeBTCDepositor`. Must not be enabled for cross-chain
+    ///         depositors whose `extraData` is an L2 user identifier rather
+    ///         than an L1 staker.
+    /// @param depositor Address of the depositor contract.
+    /// @param sponsored New allowlist membership.
+    /// @dev Requirements:
+    ///      - The caller must be the governance,
+    ///      - Depositor address must not be 0x0.
+    function setSponsoredDepositor(address depositor, bool sponsored)
+        external
+        onlyGovernance
+    {
+        self.setSponsoredDepositor(depositor, sponsored);
+    }
+
+    /// @notice Returns whether `depositor` is on the sponsored depositor
+    ///         allowlist.
+    function isSponsoredDepositor(address depositor)
+        external
+        view
+        returns (bool)
+    {
+        return self.sponsoredDepositors[depositor];
     }
 
     /// @notice Sets the redemption watchtower address.
--- a/solidity/contracts/bridge/Bridge.sol
+++ b/solidity/contracts/bridge/Bridge.sol
@@ -2040,10 +2040,17 @@ contract Bridge is
     }
 
     /// @notice Adds or removes a depositor contract from the sponsored
-    ///         depositor allowlist. When a depositor is on the list, reveals
-    ///         it submits route the `RebateStaking` rebate to the L1 address
-    ///         decoded from `extraData` rather than to the depositor contract
-    ///         itself. Intended for direct-L1-receiver relays such as
+    ///         depositor allowlist. When a depositor is on the list and a
+    ///         reveal it submits carries non-empty `extraData`, the
+    ///         `RebateStaking` rebate for that reveal routes to the L1
+    ///         address decoded from `extraData` instead of to the depositor
+    ///         contract itself -- but only if that decoded address has
+    ///         separately authorized the depositor via
+    ///         `RebateStaking.setSponsorConsent`. Activation therefore
+    ///         requires two independent steps: this governance allowlist
+    ///         call, and the L1 staker's own consent transaction. Without
+    ///         both, the rebate falls back to charging the depositor's own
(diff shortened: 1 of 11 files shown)


lrsaturnino (dismissed):
LGTM

piotr-roslaniec (approved):
LGTM. Verified the rebate routing is safe (extraData is committed to in the P2(W)SH script hash, so a sponsored relayer cannot forge a different receiver), storage layout is upgrade-safe (gap 48 -> 47), and the require() guards added in 3ba6515e make the sponsored-depositor branch fail loud on missing/zero-address extraData.

piotr-roslaniec (approved):

PR #971, opened 2026-05-26
feat(frost): FROST threshold-Schnorr (Taproot) wallets — registry + DKG lifecycle, Taproot deposits, P2TR fraud coverage, SDK & watchtower

mswilkison:
## What

This PR is the contracts + TypeScript SDK half of the tBTC v2 FROST migration: it teaches the Bridge to create, operate, and retire **FROST threshold-Schnorr (Taproot) wallets** alongside — and on a governance-controlled path, instead of — tECDSA wallets. A FROST wallet's identity is its BIP-340 x-only group key; deposits, sweeps, redemptions, and moving funds spend Taproot **key-path** outputs that are threshold-signed off-chain by the FROST/ROAST node stack (companion PRs below). Custody-equivalent security is preserved on-chain: every wallet lifecycle transition is validated by a dedicated registry with a full DKG submit/challenge/approve cycle, and any key-path signature the wallet produces outside the protocol is provably fraudulent via full BIP-341 sighash reconstruction.

The change spans four surfaces: **Solidity** (a new `frost-registry` contract suite, Bridge integration with canonical wallet IDs and Taproot deposits/redemptions, complete P2TR signature-fraud coverage delivered as EIP-170-friendly sidecar routers, and ECDSA retirement scaffolding), the **TypeScript SDK** (Taproot deposit + P2TR redemption + wallet-ID + Bitcoin testnet4 support, FROST wallet discovery, and a P2TR fraud watchtower runner), an operator-grade **watchtower service** (`services/watchtower/`), and **deploy/CI/formal gates** for all of the above.

This repo is the canonical source of truth for this surface; earlier extraction/mirror process notes in the docs are historical context only.

### FROST registry (`solidity/contracts/frost-registry/`)
- `FrostWalletRegistry` — sortition-pool-backed wallet lifecycle: DKG start, distributed-DKG result submission, challenge period, approval (registers the wallet with the Bridge via `IFrostWalletOwner.__frostWalletCreatedCallback`), inactivity claims, wallet closing, and governable parameters.
- `FrostDkgValidator` — on-chain validation of distributed FROST DKG results (result digest binding/parity with the node protocol, group key liftability, misbehaved-member bounds).
- `FrostAllowlist` — DAO-controlled operator weight allowlist implementing `IFrostAuthorizationSource`; replaces token staking for FROST operator authorization following the TIP-092/TIP-100 model used by the production ECDSA registry.
[...]

PR #982, opened 2026-06-27
fix(bridge): drop the vestigial ECDSA IDLE gate from requestNewWallet

mswilkison:
## Summary

Follow-up to #971 (security review finding). Removes a vestigial ECDSA gate that can permanently brick FROST wallet creation.

`Wallets.requestNewWallet` still required the **retired** ECDSA wallet registry to read `EcdsaDkg.State.IDLE` before starting FROST wallet creation:

```solidity
require(
    self.ecdsaWalletRegistry.getWalletCreationState() == EcdsaDkg.State.IDLE,
    "Wallet creation already in progress"
);
```

With ECDSA wallet creation removed (its `__ecdsaWalletCreatedCallback` is gone), an ECDSA DKG stuck in `CHALLENGE` — e.g. a valid result submitted just before the cutover, whose `approveDkgResult` now reverts on the removed callback, and which **no timeout path can clear** — leaves the ECDSA registry permanently out of `IDLE`. Every subsequent `requestNewWallet` then reverts, so **no FROST wallet can ever be created** (recoverable only by another Bridge upgrade). The pre-upgrade "drain all ECDSA DKGs" runbook is the only mitigation, with no on-chain enforcement.

## Fix

Remove the gate. The "wallet creation must not be in progress" guarantee is **preserved by the FROST registry itself**: `requestNewWallet()` → `FrostWalletRegistry.requestNewWallet()` → `dkg.lockState()`, which reverts (`"Current state is not IDLE"`, `FrostDkg.sol:251`) unless the FROST DKG is `IDLE`. The Bridge-level ECDSA gate is therefore redundant for FROST. The now-unused `EcdsaDkg` import is dropped (`ecdsaWalletRegistry` itself stays — still used by `seize`/`closeWallet`/the fraud-callback auth).

**Behavior note:** a concurrent-creation attempt now reverts with the FROST registry's `"Current state is not IDLE"` instead of `"Wallet creation already in progress"` — tests asserting the old string need updating.

## Verification

`hardhat compile` clean.

_Found during security review of #971._

### DIFF
--- a/solidity/contracts/bridge/Wallets.sol
+++ b/solidity/contracts/bridge/Wallets.sol
@@ -16,7 +16,6 @@
 pragma solidity 0.8.17;
 
 import {BTCUtils} from "@keep-network/bitcoin-spv-sol/contracts/BTCUtils.sol";
-import {EcdsaDkg} from "@keep-network/ecdsa/contracts/libraries/EcdsaDkg.sol";
 import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
 
 import "./BitcoinTx.sol";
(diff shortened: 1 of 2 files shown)


mswilkison:
CI fix: the failing `contracts-build-and-test` runs were 3 `Bridge - Wallets` tests that asserted the **removed** ECDSA gate (`when wallet creation is already in progress` → revert `"Wallet creation already in progress"`). That gate is deliberately gone in this PR — a stuck ECDSA DKG must not block FROST wallet creation — so those tests asserted behavior that no longer exists. Removed them with an inline note explaining the "one creation at a time" guarantee now lives in the FROST registry's `dkg.lockState()`. `Bridge.Wallets.test.ts` now passes 94/94 locally (`USE_EXTERNAL_DEPLOY=true TEST_USE_STUBS_TBTC=true hardhat test`).

PR #987, opened 2026-06-27
fix(watchtower): correct Bridge proof tx-hash byte order + enable file-backed submission mode

mswilkison:
## Summary

Two watchtower correctness fixes found by a Codex re-review of #971, **plus the CI coverage that verifies them** (the `services/watchtower/` service shipped in #971 with no CI job at all).

### 1. Bridge proof Bitcoin tx hash compared in the wrong byte order (correctness)

`extractCompletedProofBitcoinTxHash` returned the Bridge proof event's Bitcoin tx hash in Bitcoin **internal** byte order, but Esplora observations and challenge records key on **explorer/display** order (`BitcoinTxHash.from(txid)`), and the matcher compares byte-for-byte via `Hex.equals`. So a confirmed honest-spend proof **never matched** the observed transaction → the watchtower could keep, or submit, a fraud challenge against an **already-proven honest spend**. Fixed by reversing to display order (`BitcoinTxHash.from(...).reverse().toPrefixedString()`), mirroring the SDK convention used for every other Bridge-emitted tx hash.

### 2. File-backed runtime can't start in the submission mode its env parser validates (correctness/usability)

With `SUBMIT_CHALLENGES=true` the loader validated submission mode, but returned config with **no** `submissionPolicy`/`spendTypeClassifier`, so the service constructor rejected it. Since the classifier is a code predicate that can't come from env (direction confirmed with Codex), the fix splits env-vs-code: a new `…ALLOWED_SPEND_TYPES` env var → `submissionPolicy` (fail-closed types rejected), a `{ spendTypeClassifier, submissionPolicy }` injection seam on `createFileBackedP2TRSignatureFraudWatchtowerRuntime` that fails closed without a classifier, config-time enforcement of the other env-owned preconditions, and a README rewording.

### 3. CI coverage for `services/watchtower/` (new)

The service had **no** workflow (`typescript.yml`/`monitoring.yml` only watch `typescript/**` and `monitoring/**`). This PR adds `services-watchtower.yml` — a pnpm-based build+test job on node ≥22. Making it green required:
- `.npmrc` `node-linker=hoisted` so the yarn-era SDK's phantom `@ethersproject/*` deps resolve under pnpm;
- building the SDK with **yarn** (its `typechain` step resolves contract artifacts from a per-package `./node_modules/...` layout pnpm doesn't provide), then resolving the built `typescript/dist` from the pnpm-linked watchtower;
- fixing a **pre-existing** test bug: `loadFirstSignatureFraudVector` resolved the vector path from `process.cwd()` with `../../../docs/...`, which escapes the repo when the suite runs from `services/watchtower`; now anchored to the test file via `import.meta.url`.

The new job now **typechecks both fixes and runs all 66 tests green** — closing the verification gap (these changes could not be compiled/tested before, locally or in CI).

> Scope note: this PR bundles the two fixes with the CI workflow that covers them. Happy to split the CI/infra commits into a separate PR if you'd prefer to review them independently.

_Found during the Codex re-review of #971._


mswilkison:
Added a third correctness fix from the latest Codex re-review: **don't commit the Bridge lifecycle cursor when transaction indexing failed**. `commitBridgeLifecycleScanIfSafe` previously only blocked the commit on bridge-lifecycle failures, so a mempool/confirmed transaction-source or processing failure would still advance the cursor past `MovingFundsCompleted`/`RedemptionsCompleted` logs that landed in `bridgeLifecycle.ignored` (the spend wasn't indexed) — permanently dropping them, so the watchtower could keep/submit a fraud challenge for an already-proven honest spend. Now blocks on any source or transaction-batch failure; regression test added. Watchtower CI green.

(Codex's companion finding — add `BitcoinNetwork.Testnet4` — does not apply on this branch: the tbtc-v2 TypeScript SDK's `BitcoinNetwork` enum has only `testnet`/`mainnet` (no `Testnet4`); that variant exists in keep-core's Go `pkg/bitcoin`, not this SDK. Supporting testnet4 here would first require adding `Testnet4` to the SDK core.)

PR #988, opened 2026-06-27
fix(watchtower): don't mark post-broadcast submission failures as rejected

mswilkison:
## Summary

Fixes a Codex-flagged P1 correctness bug in `P2TRSignatureFraudWatchtower.submitChallenge` (`typescript/src/services/maintenance/p2tr-signature-fraud.ts`).

The on-chain broadcast (`submitter.submitSignatureFraudChallenge`) was evaluated **inside** the object literal passed to `recordP2TRWatchtowerChallengeEvent` for the `submission-accepted` event, all wrapped in one `try/catch`:

```ts
try {
  return recordP2TRWatchtowerChallengeEvent(this.store, {
    type: "submission-accepted",
    challengeTxHash: await submitter.submitSignatureFraudChallenge(observation), // broadcasts on-chain
    ...
  })
} catch (error) {
  return recordP2TRWatchtowerChallengeEvent(this.store, { type: "submission-rejected", ... })
}
```

If the broadcast **succeeds** but a later step in the `try` fails — persisting the accepted record, or any post-broadcast work — the `catch` records the challenge as **`submission-rejected`**. That drops the on-chain challenge tx hash and leaves local state retryable, so the watchtower can **resubmit a duplicate** or **raise a misleading alert** for a challenge that already exists on-chain.

## Fix

Capture the broadcast result first:
- A **pre-broadcast** failure → `submission-rejected` (the tx was genuinely not submitted; safe to retry).
- After a successful broadcast → record `submission-accepted` with the tx hash; if **that** persistence fails, the error **propagates** rather than being misattributed as a rejection.

## Test

Adds a regression test using a store that rejects the post-broadcast `submitted`-record save: asserts the persistence error propagates, the tx was broadcast exactly once, and **no** `rejected` record is written. (Against the old code the test fails — it would return a `rejected` record instead of throwing.)

Covered by the existing `typescript.yml` CI (build + test).

_Found during the Codex re-review of #971._

### DIFF
--- a/typescript/api-reference/classes/P2TRSignatureFraudWatchtowerRunner.md
+++ b/typescript/api-reference/classes/P2TRSignatureFraudWatchtowerRunner.md
@@ -60,7 +60,7 @@
 
 #### Defined in
 
-[src/services/maintenance/p2tr-signature-fraud.ts:3285](https://github.com/threshold-network/tbtc-v2/blob/main/typescript/src/services/maintenance/p2tr-signature-fraud.ts#L3285)
+[src/services/maintenance/p2tr-signature-fraud.ts:3297](https://github.com/threshold-network/tbtc-v2/blob/main/typescript/src/services/maintenance/p2tr-signature-fraud.ts#L3297)
 
 ## Properties
 
@@ -70,7 +70,7 @@
 
 #### Defined in
 
-[src/services/maintenance/p2tr-signature-fraud.ts:3287](https://github.com/threshold-network/tbtc-v2/blob/main/typescript/src/services/maintenance/p2tr-signature-fraud.ts#L3287)
+[src/services/maintenance/p2tr-signature-fraud.ts:3299](https://github.com/threshold-network/tbtc-v2/blob/main/typescript/src/services/maintenance/p2tr-signature-fraud.ts#L3299)
 
 ___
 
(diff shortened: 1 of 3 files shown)


mswilkison:
Closing as **superseded by #995** (merged into `extraction/frost-mirror-2026-05-26`).

Both PRs fix the same `submitChallenge` bug — the broadcast was `await`ed inside the `submission-accepted` object literal, so a post-broadcast failure was mislabeled `submission-rejected`. This PR (#988) was the first fix: it stops the mislabeling and propagates the error instead.

However, propagating the error leaves the record in the **`submitting`** state, which is replayable — so the next cycle can still re-broadcast a duplicate. That residual is exactly what the Codex re-review flagged ("the post-broadcast persistence failure path still leaves the challenge in a replayable state, so duplicate on-chain submissions remain possible"), and it's what #995 closes by introducing a non-replayable **`broadcast-pending`** status (recorded via an `onBroadcast` callback), plus confirmed-revert handling and bounded persist retries.

#995 is a functional superset of #988 and is already merged. This PR is now conflicting against the mirror, and its regression test asserts the error *propagates* — which contradicts #995's behavior (it records `broadcast-pending` instead). Nothing here is missing from #995, so closing rather than reworking.

PR #991, opened 2026-06-28
fix(deploy): route lifecycle-router wiring through governance

mswilkison:
## Summary

Fixes a Codex **P2**: `49_deploy_bridge_lifecycle_router.ts` wired the router from the **deployer** even on networks where Bridge governance is `BridgeGovernance` owned by the governance **Safe**. `BridgeGovernance.setLifecycleRouter` is `onlyOwner`, so the deployer's call reverts with `Ownable` and **aborts the production deployment** — with no calldata emitted for manual governance execution. (The `FrostWalletRegistry.lifecycleOwner` update had the same class of issue: it hard-threw when registry governance wasn't the deployer.)

## Fix

Mirror the established `51_deploy_frost_allowlist` pattern via a `getConfiguredSigner` helper, for **both** governance-gated steps:
- Resolve the authorized caller — Bridge governance when it's the deployer, else `BridgeGovernance.owner()` for the router, and `FrostWalletRegistry.governance()` for the registry.
- Send the tx **only** when that account is a configured signer.
- Otherwise emit `target` + ABI-encoded `calldata` and, on **mainnet**, log `skipping for manual governance execution` and continue (so the deploy completes, router deployed + verified, wiring deferred to a governance Safe tx); on other networks, throw (the executor is expected to be configured in test).

## Scope / testing

- The **dev/test path** (Bridge governance is the deployer) is unchanged and is the one exercised by the deployment fixtures — so `contracts-build-and-test` continues to cover it.
- The **production skip path** (Safe-owned governance) is not fixture-exercised; it's a straightforward calldata-emit-and-skip that mirrors `51_deploy_frost_allowlist`, but worth a staging-deploy check before mainnet.
- `prettier` clean; `eslint` 0 errors (the `no-console` warnings match the original deploy-script style).

_Found during the Codex re-review of PR #971._

### DIFF
--- a/solidity/deploy/48_deploy_frost_wallet_registry.ts
+++ b/solidity/deploy/48_deploy_frost_wallet_registry.ts
@@ -1,6 +1,26 @@
 import type { HardhatRuntimeEnvironment } from "hardhat/types"
 import type { DeployFunction } from "hardhat-deploy/types"
 
+// Returns a signer for `address` only when that account is configured for the
+// current network. Used to decide whether a governance-gated call can be sent by
+// this deployment, or must instead be emitted as calldata for manual governance
+// execution (the governance owner -- e.g. a Safe -- is not a deployer-controlled
+// signer in production). Mirrors 49_deploy_bridge_lifecycle_router.
+async function getConfiguredSigner(
+  hre: HardhatRuntimeEnvironment,
+  address: string
+) {
+  const configuredAccounts = (await hre.ethers.provider.listAccounts()).map(
+    (account) => account.toLowerCase()
+  )
+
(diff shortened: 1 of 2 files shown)


mswilkison:
Folded in a sibling fix (commit 8e84c37): `48_deploy_frost_wallet_registry.ts` had the same governance-wiring bug this PR fixes for `49` — it sent the `BridgeGovernance.onlyOwner` `setFrostWalletRegistry` call from `deployer`, which reverts on post-handoff networks (owner = governance multisig), aborting the deploy with the registry unwired. Applied the same pattern: resolve the governance owner / emit calldata for manual governance execution. (Codex P1.)

PR #998, opened 2026-06-30
Remove Sei support

mswilkison:
## Summary

- Removes Sei cross-chain support, including the dedicated workflow, cross-chain package, SDK module/export, docs, artifacts, and tests.
- Removes Sei chain mappings and L1 Bitcoin depositor artifact/encoder wiring.
- Removes Sei from the NTT deployment configuration and renames generic NTT test fixtures/mocks so they no longer imply Sei support.

## Testing

- `npm run build` from `solidity`
- `yarn build` from `typescript` using Yarn 4
- `yarn test --grep "NTT Utilities"` from `typescript` under Node 18: 20 passing
- Focused Solidity NTT suite: 159 passing, 3 setup failures before assertions in legacy smock-based tests with `provider.init is not a function`; a clean Yarn 4 reinstall was blocked locally by DNS resolution for `registry.yarnpkg.com`

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed Sei destination-chain support across the TypeScript API (exports, chain mappings, artifacts) and related NTT utilities/examples.
  * Removed Sei L2 token and depositor integration artifacts and scripts, including associated deployment/verification workflow automation.
* **Documentation**
  * Updated Wormhole/NTT receiver and encoding examples to use the Ethereum Sepolia “destination” chain (and new chain-id examples).
  * Removed Sei-specific API reference pages and integration guide content.
* **Chores**
  * Retired Sei deployment tooling and formatting/linting workflow configuration.

PR #999, opened 2026-06-30
Configure NTT migration chains

mswilkison:
## Summary

- Add fixed-destination L1 NTT depositors for EVM destinations while preserving proxy-upgrade storage slots and legacy in-flight recipient decoding.
- Add Sui and Solana NTT mint-authority migration primitives with paused/drained guards.
- Keep Sui/Solana blocked in the L1 config script until spoke-side NTT deployment, token-authority handoff, and legacy lockbox migration are complete.
- Rewrite the executor fee model for `L1BTCDepositorNttWithExecutor` (field renames/retirements, `setDefaultParameters` arity 5→3, `>=` → `==` platform-fee pin, cached-quote enforcement instead of a live re-quote at finalize time), several ABI-breaking changes (`getNttConfiguration`, `quoteFinalizeDeposit`/`quoteFinalizedDeposit`, `setExecutorParameters`, `initialize` arities), and switch `retrieveTokens` from `.transfer` to `.call`. A published SDK export (`TransceiverStructs.sol`) is removed as dead code. This is a materially larger scope than the original bullet list above; flagging it explicitly per review feedback (see below).

## Notes

This is stacked on #998 because the NTT target set intentionally excludes Sei.

The Base mainnet configuration now uses Wormhole chain ID `30` instead of the EVM chain ID.

Executor transfers default the destination gas refund address to the destination recipient so refunds remain deliverable on non-EVM chains. EVM deployments can set `destinationRefundAddress` if they want a relayer or treasury refund account instead.

## Post-review fixes

A multi-agent review (8 lenses + 2 independent adversarial validation passes) found 28 confirmed issues (4 P1, 11 P2, 13 P3) across Solidity, Sui Move, Solana/Anchor, and the TS SDK. All have been addressed:

**Solidity (`L1BTCDepositorNttWithExecutor`, `L1BTCDepositorNtt`, `AbstractFixedDestinationNttDepositor`)**
- Fixed a permissionless fee-ceiling regression: `setExecutorParameters`' platform-fee check was loosened from an equality pin to a range, letting any address stage `dbps=10000` (10%) against someone else's Initialized deposit. Restored the equality pin.
- Restored a cached required-payment field on `ExecutorParameterSet`, removed at finalize time by this PR, which reintroduced a quote-drift revert between staging and finalization plus a redundant on-chain re-quote call.
- Deduplicated `initializeV2DestinationChain` (previously copy-pasted in both children) into the shared abstract base, and added an owner-only `setDestinationChainId` corrective setter for the storage-slot-reuse upgrade edge case.
- Deleted the orphaned `TransceiverStructs.sol` (zero remaining references after this PR).
- Documented (rather than implemented) two design tradeoffs surfaced in review: keeping a separate `fixedDestinationDeposits` mapping instead of folding the flag into the shared `AbstractL1BTCDepositor.DepositState` enum (the enum approach would touch the state-transition checks inherited by every other live depositor proxy on this L1 — Base, Arbitrum, StarkNet — for a ~20k-gas one-time saving on two not-yet-deployed contracts), and the legacy-format decode branch (confirmed via deployment-history cross-check that no `L1BTCDepositorNtt`/`L1BTCDepositorNttWithExecutor` proxy has ever been deployed, so it currently ships as forward-compatibility infrastructure rather than an active migration path).
- Added test coverage for the previously-untested nonce-based executor-parameter refresh/expiry/multi-user-isolation workflow, `retrieveTokens`, the NTT-manager-address setters, `setExecutorParameters`' input-validation revert branches, and the remaining default-parameter setters' boundary/revert paths.

**Sui (`wormhole_gateway.move`)**
- `retire_gateway_for_ntt` now also requires the TBTC token itself to be paused (`TBTC::TokenState.paused`, not just `GatewayState.paused` — `TBTC::mint` gates on the former, so pausing only the gateway left the retiring admin free to mint canonical tBTC directly during the handoff window).
- Added a `TreasuryCap<TBTC>.total_supply() == 0` check alongside the existing `minted_amount == 0` check, catching canonical tBTC minted before `initialize_gateway` while the deployer held the raw `TreasuryCap`.
- `retire_gateway_for_ntt` now requires a `TBTC::GuardianCap` co-signer in addition to the gateway `AdminCap`, for on-chain parity with the Solana tBTC program's authority+guardian dual-signer requirement on its equivalent instruction.
- Updated the 3 existing `retire_gateway_for_ntt` tests and `README.adoc`'s migration checklist for the new signature and pre-conditions.

**Solana (`transfer_mint_authority.rs`)**
- `transfer_mint_authority` now requires at least two registered guardians. A lone guardian only proves the co-signer holds a pubkey distinct from the authority, which a compromised or malicious authority can satisfy trivially by registering one puppet key it also controls and immediately co-signing with it.
- Updated the new `01__tbtc.ts` test suite and `README.adoc` for the new precondition.

**TypeScript SDK / deploy script**
- Fixed a case-sensitivity bug in `configure-l1-btc-depositor-ntt.ts`'s `selectDestination`: the blocked-destination-set check lowercased its input but the config lookup didn't, so `NTT_DESTINATION=Base` skipped the block path and fell through to a misleading "unsupported" error.
- Documented that the Sui/Solana block in that script is a release-gating checklist, not technical enforcement (it's an off-chain script that sends no transactions; actual readiness is enforced independently by each spoke chain's own on-chain gates).
- Added doc comments for `DestinationChainName`'s Optimism exclusion and `WORMHOLE_NTT_CHAIN_IDS`, and a text-based CI consistency check between the TS SDK's and the Solidity deploy script's independently-maintained Wormhole chain-ID tables.

**Not fixed / follow-up needed**
- Attempted to add `cross-chain-solana.yml` / `cross-chain-sui.yml` CI workflows (both packages currently have zero CI coverage). Dropped both after hands-on verification: Anchor 0.28.0 is incompatible with every currently-installable Solana CLI (`build-bpf` was renamed `build-sbf`; the resulting toolchain then hits an unrelated `#![feature(stdsimd)]` nightly-feature error in a transitive dependency), and this Sui package's Move dependency graph (wormhole-foundation/wormhole) has drifted across old-style/new-style Move.toml formats such that no tagged release nor `main` reliably resolves against this package's manifest layout in a fresh CI checkout. Both are genuine, pre-existing toolchain-compatibility gaps unrelated to this PR's code changes. Recommend a follow-up ticket to pin compatible Rust/Solana/Anchor and Sui/Wormhole versions before adding CI for these two packages.
- Bumped `cross-chain/solana/Cargo.lock`'s `wasm-bindgen` family to 0.2.92 (semver-compatible, within existing `Cargo.toml` ranges) to unblock local `cargo check` — it was failing before any project code compiled due to an incompatibility with the currently-installed Rust.

## CI

Pushed and monitored CI to green. Two failures surfaced, both fixed:

- `contracts-build-and-test`: `L1BTCRedeemerWormhole` was missing its `__ReentrancyGuard_init()` call despite inheriting `ReentrancyGuardUpgradeable`, tripping the OpenZeppelin upgrade-safety validator. Pre-existing, unrelated to this PR (a different contract this PR never touches), but blocking green CI — fixed alongside since it was a one-line, well-understood gap.
- `typescript-docs`: this PR's `chain.ts`/`wormhole.ts`/`ntt.ts` edits shifted line numbers and doc comments; regenerated `typescript/api-reference/**` via `yarn build && yarn docs` to match.

All CI checks now pass (0 failing).

## Validation

- `yarn build`, `yarn lint`, and the full `solidity` test suite (`npx hardhat test`) — 1085 passing, 26 pre-existing failures unrelated to this PR (verified identical on the unmodified branch: all 26 are `"before all"` hook failures in unrelated suites — Timelock/Bridge/VendingMachine/TBTCVault/etc. — from a missing `ReimbursementPool` deployment fixture in this environment).
- `npx hardhat test test/cross-chain/wormhole/*.test.ts test/scripts/wormhole-chain-ids-consistency.test.ts` — 54 passing (0 failing).
- `npx eslint` / `npx prettier --check` clean on all touched TS/Solidity files.
- `npx solhint` clean on all touched contracts.
- `cargo check -p tbtc` passes (after the `wasm-bindgen` bump above).
- Sui: `sui move build --path . --test` passes cleanly against a CI-only dependency manifest exercising the same source files (see follow-up note above on why that manifest isn't used for the actual `cross-chain-sui.yml` CI workflow).

Local toolchain blockers (unchanged from before, see follow-up note above for what was additionally investigated):

- `anchor test` / `sui move test` still not runnable end-to-end in this environment; the above are the closest achievable local verification.


piotr-roslaniec:
Pushed a merge commit (`cb278a04`) from `main` to resolve the merge conflict and bring this branch up to date (branch protection is `strict`).

Rationale for how it was resolved: since the merge-base, `main` only advanced by one commit — `c10e824` (Remove Sei support, #998). Every hunk of that commit is Sei-only (Sei enum/field/mappings, Sei config blocks, Sei doc-comment examples; the encoding logic is untouched). This branch had already dropped Sei as part of its fixed-destination NTT redesign, so all conflicts were resolved in favor of this branch's redesign. The resulting tree is **byte-identical to the previous branch tip except for removing the residual `cross-chain/sei/` artifacts** — i.e. the merge changes none of this PR's actual content, it only clears the conflict and syncs with main. CI is green on the merge commit.

Note this does not touch the open design question of the destination-recipient encoding (full 32-byte vs chain-ID-packed) — that remains as designed in this PR.

piotr-roslaniec (approved):

PR #1000, opened 2026-06-30
Deprecate Polygon bridge support

mswilkison:
## Summary

- Deprecate Polygon as an exit-only Wormhole bridge path.
- Clear Polygon peer mappings on Base, Arbitrum, Optimism, and Polygon gateway deployment scripts.
- Add a Polygon gateway deprecation deploy script that sets `mintingLimit` to `0`.
- Remove Polygon from new Solana gateway initialization and add a Solana helper to clear the existing Polygon peer.
- Document the operational exit-only flow and governance caveats.

## Notes

The package is intentionally kept in place. Existing Polygon liquidity should not be stranded: users and protocol-owned liquidity can exit through `PolygonWormholeGateway.sendTbtc`, route back to Ethereum, then use supported paths from there.

For existing live deployments where gateway ownership has already moved to governance, the equivalent `updateMintingLimit(0)` and `updateGatewayAddress(<chainId>, DISABLED_GATEWAY())` calls need to be executed by governance. The deploy scripts apply these settings only in the normal fresh-deployment ownership flow.

## Validation

- `git diff --check`
- `prettier --write` on touched TypeScript files
- `prettier --check` on touched TypeScript files

Package-level lint/test dependencies were not installed in this isolated worktree.


piotr-roslaniec:
## Fix: gateway-less wrapped tBTC sends were bricked

Pushed `6a0c3a59` addressing a P0 regression found during review.

**Problem:** `send_tbtc_wrapped` was changed to require a mandatory, initialized `GatewayInfo` PDA. The wrapped-transfer path serves chains with **no** registered tBTC gateway (e.g. Ethereum), which have no such PDA — so those sends revert with `AccountNotInitialized`, breaking the Solana→Ethereum exit route. (The happy-path test had been modified to pre-create the PDA, which masked this.)

**Fix:** keep `gateway_info` **mandatory and seeds-pinned** — so a disabled route cannot be bypassed by omitting or substituting the account — but accept it as an `UncheckedAccount` and branch on its on-chain state in the handler:
- program-owned (initialized `GatewayInfo`) → reject if the address is the disabled sentinel;
- uninitialized (empty/system-owned) → no registered gateway → allow.

This mirrors the EVM `gateways[chain]` storage read, which the caller likewise cannot bypass. Note: an earlier `Option<Account>` approach was rejected because on Solana an optional account is *caller-controlled*, letting a caller pass `null` for a disabled chain and skip the check — the seeds-pinned mandatory account closes that hole.

Also un-masked the gateway-less wrapped-send test and added a test asserting the seeds-pin blocks a disabled-route bypass.

⚠️ **Verification caveat:** the Rust program is **compile-checked only** (`cargo check`). This environment has no Solana test-validator / `anchor` CLI, and there is no Solana CI in this repo, so the mocha tests were **not executed**. Please run `anchor test` before taking this out of draft.

piotr-roslaniec (approved):

PR #1003, opened 2026-07-01
Trail of Bits tBTC v2 AC Extension — security remediation

lrsaturnino:
Applies the remediation for the Trail of Bits "tBTC v2 AC Extension" security assessment. Every in-scope finding whose target code lives in this repository is fixed, tagged with its public report ID (`TOB-TBTCACEXT-<n>`) so every commit maps to exactly one finding — findings that needed more than one remediation round carry multiple commits, each still traceable to its finding via the tag.

## Base / stacking

This branch was created off the frozen review commit `adefd664`, which is the head of `feat/psbt-covenant-bridge-port` (the reviewed PR, #957). The PR therefore targets that branch so the diff is exactly the remediation (23 commits) rather than PR #957's own changes. Retarget to `main` once #957 lands.

## Findings remediated (16)

- High: 1 (bridge-side covenant spend defense path for fraud challenges), 17 (deploy + activate a migration-debt `TBTCVault` in the hotfix, plus a full legacy-vault retirement path — see below), 18 (deploy new `BridgeGovernance` and transfer governance before seeding fraud escrow), 23 (Sui: reject minting with a `MinterCap` whose minter was removed)
- Medium: 19 (guard vault upgrade finalization against outstanding optimistic-minting debt, plus an enforceable legacy-vault migration coordinator — see below), 21 (Bank-balance rescue for cross-chain redemption refunds), 24 (deterministic moving-funds target selection), 28 (reconcile Account Control exposure on legacy redemptions), 30 (block untrust/rotation while a vault has optimistic-minting debt, plus exact-legacy fail-closed retirement — see below), 37 (keep wallets open while fraud challenges can still mature), 59 (recover vetoed cross-chain redemption funds from the watchtower), 64 (stop persisting precomputed fraud-challenge escrow seed calldata)
- Low: 60 (process moved-funds sweep timeouts for `Closing`/`Closed` target wallets)
- Informational: 81 (drive migration sweep completions only from migration-tagged deposits), 90 (cover fail-open migration sweep callback retry completion), 91 (drop migration completion flags left without a configured reserve)

Finding 37 is a pre-existing issue outside the reviewed diff; it is included because the scoped TIP-109 hotfix re-enables the fraud-challenge path it affects.

Finding 1 is targeted at `keep-core/pkg/tbtc/covenant_signer.go` by the report's Target field, but the report's recommended remediation is a bridge-level defense in this repository — recognizing a covenant active UTXO spend as honest when it is proven through the account-control covenant proof flow. This is implemented as a new `CovenantSpendAuthorization` registry plus `Bridge.defeatFraudChallengeWithCovenantSpend`. The account-control covenant authority attests, before the covenant signer signs, that a specific source outpoint is a covenant active UTXO spendable by a specific wallet, for a specific value, into a specific set of outputs (the BIP-143 `hashOutputs`); the defeat path recognizes the spend only against a matching attestation. The defense is keyed by the source outpoint's provenance rather than any output-side heuristic, so a diversion of an authorized covenant UTXO to other outputs cannot defeat a challenge. As defense in depth against a mis-issued or malicious attestation, the defeat path also rejects any outpoint the Bridge already tracks (a revealed deposit, a spent main UTXO, a moved-funds sweep request, or the challenged wallet's current main UTXO); a genuine external covenant UTXO is never one of these, but a wallet's own backing UTXO is, which blocks the most direct laundering of a backing UTXO through this path. Authorizations are write-once and permanent (no overwrite, no revoke), so once a covenant spend is authorized the wallet stays defensible for its whole challengeable lifetime and the authority cannot pull the defense out from under a legitimate signature. The registry owner is a trusted fraud-defense root (documented on the contract and the governance setter); the registry address is set through governance, and while unset the defeat path reverts, so the keep-core covenant signer keeps its fail-closed stopgap (keep-core PR #4126, `bridgeCovenantFraudDefenseConfirmed` defaulting to `false`) until the registry is wired and the authority is posting attestations. Completing the end-to-end feature additionally requires the covenant signer to gate on the on-chain authorization (in keep-core); removing the residual owner trust would replace the owner with an on-chain account-control proof verifier.

## Additional hardening (post-review internal audit)

Several rounds of internal review surfaced further issues, each folded into its originating finding's commit(s) — still tagged, still traceable to the finding, but no longer strictly one commit per finding where a finding needed more than one round:

- `TOB-24`: `seedWalletRegistrationOrder` could double-count a wallet that registered in the upgrade→seed window, permanently bricking moving-funds commitments for `count >= 2` sources. Fixed with a dedup check against the on-chain snapshot before appending.
- `TOB-17`/`TOB-19`/`TOB-30` — round 1: the deployed legacy `TBTCVault` predates `hasOutstandingOptimisticMintingDebt()`, so the new debt guards fail-open for the actual migration target. The hotfix runbook computes and hard-blocks on outstanding legacy OM debt before `finalizeUpgrade`, and the new vault's ownership transfer to governance is an explicit, ordered runbook step.
- `TOB-28` — round 1: `_reconcileAccountControlRedemption` calls the AC notifier unconditionally with no fail-safe — confirmed intentional (isolating the failure would reopen the accounting gap TOB-28 exists to close) and documented as a fail-closed trust-model dependency; the "required + zero notifier" misconfiguration path was also closed.
- `TOB-17`/`TOB-19`/`TOB-30` — round 2 (enforceable legacy-vault retirement): the round-1 runbook precondition was advisory only and unsatisfiable as an "all OM debt zero" gate (thousands of harmless residual dust debtors never clear). Replaced with an enforceable design: a new immutable `LegacyTBTCVaultMigrationCoordinator` becomes the legacy vault's owner and permanently pauses optimistic minting — only it, via `BridgeGovernance`, can finalize the legacy vault's upgrade (closes TOB-19: governance can no longer finalize directly). A new `BridgeState.legacyVaultOptimisticMintingDebtCoordinator` mapping plus an address-and-codehash-pinned attestation makes the known legacy vault fail-closed on its undecodable debt selector unless a valid attestation and locked coordinator exist, without changing the existing fail-open behavior for any other pre-interface vault (closes TOB-30). The untrust/rotation guard logic moved into a new `VaultManagement` library shared by `Bridge.sol` and its test harness, keeping `Bridge.sol` under the EIP-170 size limit. `BridgeGovernance.finalizeLegacyVaultMigration` performs the cutover as five sequential, non-try/catch calls (attest → untrust legacy → finalize coordinator → trust successor → set canonical vault) that roll back atomically on any failure. The TIP-109 hotfix deploy script gained an unconditional Phase-I preflight (chain ID, exact legacy address, exact legacy runtime codehash) before any deploy call in either mode, a reorg-safe pinned scan for the dangerous "finalized-but-unswept" deposit set, and a two-phase preparation/cutover split. A new `Bridge.LegacyVaultRetirement.fork.test.ts` suite (gated on `FORKING_URL`) exercises the cutover against a real forked mainnet Bridge proxy/ProxyAdmin/Timelock.
- `TOB-28` — round 3 (cross-chain redemption regression, caught and fixed before merge): round 1's fail-closed restriction on plain `unmint` silently broke the production cross-chain redemption path — `L1BTCRedeemerWormhole.requestRedemption` → `AbstractBTCRedeemer._requestRedemption` called `tbtcVault.unmint(amount)` directly, the only vault entry point the integrator used, which now reverted unconditionally once an Account Control deployment activated reconciliation. The existing regression suite never caught it because the redeemer's tests exercise a mock vault with no reconciliation semantics at all. Fixed by routing `_requestRedemption` through the vault's existing atomic, reconciled `unmintAndRedeem(amount, redemptionData)` entry point instead of the removed manual unmint-then-Bridge-call sequence; `ITBTCVault` gained the corresponding declaration. New tests (`AbstractBTCRedeemer.AccountControl.test.ts`) exercise the real `TBTCVault`/`Bridge`/`Bank` stack — not the mock — proving the redemption path now succeeds with reconciliation active and still fails closed if the notifier is missing or reverts.

## Verification

`yarn build` passes and the full Solidity suite is green (3238 passing, 0 failing, 40 pending), reproduced fresh including once from a clean rebuild. Integration suite: 27 passing, 0 failing. `yarn format` (ESLint + solhint + Prettier): 0 errors. Slither: 57 results, matching the pre-existing baseline (0 new). The Sui Move change is covered by the isolated `token_test` harness (`sui move test -p token_test`); the full `deploy_l2` package does not compile independently of this PR (unresolved Wormhole dependency), unchanged from before this remediation. Each finding's commit(s) carry their own regression test(s).


solidity/contracts/bridge/Wallets.sol
```diff
@@ -225,9 +225,66 @@ library Wallets {
...
+        // preserved regardless of the upgrade-to-seed timing.
+        bytes20[] memory postUpgradeWallets = self.walletRegistrationOrder;
+        delete self.walletRegistrationOrder;
+
+        for (uint256 i = 0; i < preUpgradeWallets.length; i++) {
+            self.walletRegistrationOrder.push(preUpgradeWallets[i]);
+        }
+        for (uint256 j = 0; j < postUpgradeWallets.length; j++) {
+            self.walletRegistrationOrder.push(postUpgradeWallets[j]);
+        }
```
miquelcabot, line 327: This reconstruction appends the on-chain `postUpgradeWallets snapshot` after the caller-supplied `preUpgradeWallets`, with no dedup, but the runbook in `86_deploy_tip109_hotfix.ts` tells the operator to recompute the full event-ordered list (`computeWalletRegistrationOrder` scans `NewWalletRegistered` from Bridge-deploy -> head), which already includes any post-upgrade wallet.

So if a wallet completes DKG in the upgrade->seed window (registration is permissionless) and the operator follows the runbook literally, that wallet ends up twice in `walletRegistrationOrder`. `selectMovingFundsTargetWallets` then returns a target set with a repeated entry, which a strictly-ascending commitment can never match -> `submitMovingFundsCommitment` reverts permanently for the affected counts.

The defensive non-empty branch was added exactly to survive an in-window registration, but as written it produces the corruption it means to prevent. Suggest deduping against the existing on-chain order before appending, e.g.:

```
for (uint256 i = 0; i < preUpgradeWallets.length; i++) {
    self.walletRegistrationOrder.push(preUpgradeWallets[i]);
}
for (uint256 j = 0; j < postUpgradeWallets.length; j++) {
    bytes20 w = postUpgradeWallets[j];
    bool seen = false;
    for (uint256 k = 0; k < preUpgradeWallets.length; k++) {
        if (preUpgradeWallets[k] == w) { seen = true; break; }
    }
    if (!seen) self.walletRegistrationOrder.push(w);
}
```

(alternatively, keep the seed dedup-free and change the runbook to pass only wallets not already on-chain). Either way, also fix the runbook/NatSpec text that claims the seed "requires an empty order" / reverts with `"Wallet registration order already populated"`. The actual guard is `walletRegistrationOrderSeeded` / `"already seeded"` and it does accept a non-empty order.

lrsaturnino: tks @miquelcabot. This is addressed

miquelcabot (commented):
Reviewed the full diff against the ToB report. The dedup fix on `seedWalletRegistrationOrder` landed exactly as discussed, and I found no correctness defects in the new work. The BIP-143 offsets in `defeatFraudChallengeWithCovenantSpend` check out (min length 157, value at `-52`, `hashOutputs` at `-40`, outpoint at `68`), the write-once registry is tight, and the fail-closed rework of the TOB-30 staticcall correctly closes a hole the earlier fail-open version left open for the immutable mainnet legacy vault.

**Not approving on code review alone**, because TOB-1 adds a new fraud-defense trust root that I don't think a reviewer LGTM should carry:

`CovenantSpendAuthorization`'s owner can make any wallet signature over an attested `(outpoint, wallet, value, outputsHash)` tuple immune to fraud slashing. The registry does not verify covenant provenance on-chain. The owner vouches for it. The commit message is admirably honest about this ("a malicious owner colluding with a compromised threshold wallet could pardon a theft"), and the known-tracked-UTXO rejection is a good mitigation, but the residual trust is real and new.

The report's recommendation was to recognize the spend "when a covenant spend is **proven** through the account-control covenant proof flow". This PR ships an owner-attested reflection of that flow instead, because the proof verifier lives outside this repo. That's a defensible interim design, but it's a governance decision about who gets to pardon fraud, not a code-quality one.

**Ask:** explicit sign-off from the Threshold Council on (a) who owns the registry at deployment, (b) whether the owner-attested model is accepted as the shipping design or only as a stopgap until the on-chain proof verifier exists, and (c) the operational rule already noted in the commit. That governance must never repoint or zero `covenantSpendAuthorization` while any wallet that relied on the current registry is still challengeable. With that sign-off I'm happy to approve.

**Process note (non-blocking):** this branch now carries 15 findings plus three new subsystems (`VaultManagement`, `CovenantSpendAuthorization`, `LegacyTBTCVaultMigrationCoordinator`) at +11.9k lines. The TOB-28 break-then-repair pair (the plain-unmint guard silently broke `AbstractBTCRedeemer`, caught a commit later only because the suite ran against `MockTBTCVault`) is the symptom, it's a lot to review as one unit. TOB-1 in particular would be easier to reason about as its own PR. Not asking for a split at this stage, just flagging it for the next one.

solidity/contracts/vault/TBTCVault.sol
```diff
@@ -351,8 +482,62 @@ contract TBTCVault is IVault, Ownable, TBTCOptimisticMinting {
...
+        // via `Bank.transferBalance`/`transferBalanceFrom`, which could then
+        // remint it as TBTC through `mint`/`receiveBalanceApproval`. There is
+        // no way to tag that Bank balance with the AC decrement it should owe,
+        // so once reconciliation is required, a *reversible* per-unmint AC
+        // decrement here could be detached from the balance it was meant to
+        // track and never restored — permanently understating AC exposure.
+        // Block plain unmint outright instead; `_unmintAndRedeem` remains the
+        // irreversible, reconciled exit for AC-origin supply.
+        require(
+            !accountControlReconciliationRequired,
```
miquelcabot, line 534: Agreed this is the right call. The transferable-Bank-balance defeat of the per-unmint decrement is real, and there genuinely is no vault-side hook on `Bank.transferBalance` to preserve provenance. Blocking outright beats an unsound pending-decrement pool.

Flagging the blast radius rather than the logic: this turns `TBTCVault.unmint()`, a **public entry point**, into an unconditional revert once `accountControlReconciliationRequired` is set. In-repo that's covered (`AbstractBTCRedeemer` was the only production caller and 9c1f8049 migrated it to `unmintAndRedeem`), but external integrators calling `unmint` and then doing something else with the resulting Bank balance will break the moment AC is activated, with no on-chain signal beforehand.

The default-`false` flag means this is a scheduled break, not an immediate one, which is the right shape. Two things I'd want before AC activation:

1. A note in the release/upgrade docs that activating AC reconciliation disables plain `unmint`, with `unmintAndRedeem` named as the supported replacement.
2. A check of known integrators against the `unmint` selector before flipping the flag — worth doing precisely because the in-repo break here was found by a follow-up commit, not by the suite.

No code change requested.

lrsaturnino: Acknowledged — tracking the release-note and integrator selector-check as pre-activation items before AC reconciliation is enabled; no code change needed here.

PR #1009, opened 2026-07-07
test(system-tests): FROST testnet4 lifecycle runbook + e2e template

mswilkison:
## What

Preserves the FROST/Schnorr **testnet4** funds-movement e2e as a committable
artifact. Covers the full lifecycle driven by live keep-core FROST nodes:

```
Taproot deposit → sweep → MINT → redemption → FROST→FROST moving funds
```

with real testnet4 (BIP-94) proof-of-work under every SPV proof.

**Three files, all under `system-tests/`:**
- **`FROST_TESTNET4_E2E.md`** — the verified runbook. Architecture (local anvil
  for Ethereum, real testnet4 for Bitcoin — no Sepolia/faucet/RPC), prerequisites,
  the exact SDK/contract call sequence for each stage, and the on-chain/Bitcoin
  evidence from a reference run.
- **`test/utils/frost.ts`** — reusable helpers: `waitForNewTransactionAtAddress`
  (a FROST wallet is threshold-signed, so the test *waits for the nodes* to
  broadcast rather than signing with a WIF like the ECDSA tests) and
  `submitMovingFundsProof` (the SDK's `maintenance.spv` has `submitDepositSweepProof`
  + `submitRedemptionProof` but **not** moving-funds — mirrored here).
- **`test/frost-testnet4-lifecycle.test.ts`** — gated behind `RUN_FROST_TESTNET4_E2E`,
  `describe.skip` by default. A full-stack e2e template, not part of unit CI.

## Why it's structured this way

Unlike the ECDSA system tests, a FROST wallet has no single key the test can
sign with — the sweep/redemption/move are threshold-Schnorr-signed by the running
keep-core group. So the test only drives the depositor + SPV maintainer and waits
for the nodes on Bitcoin. And SPV proofs need real PoW, so the Bitcoin side must
be a real network (testnet4) while Ethereum stays local.

## Status / caveats (please read)

This is intentionally shipped as a **draft**:
- The runbook is **verified end-to-end** (evidence tx hashes included) and is the
  high-confidence part.
- The mocha test is a **template**. Un-skipping it requires two things I could not
  do in isolation:
  1. Wiring `system-tests` to the FROST SDK — it currently pins
     `@keep-network/tbtc-v2.ts@^2.3.0`, but the Taproot-deposit/testnet4 APIs live
     in this branch's `typescript/` (`4.0.0-dev`), which isn't in the pnpm
     workspace. The `TODO(frost-e2e)` markers flag the SDK surface to confirm.
  2. A live run (FROST nodes + testnet4 deployment + funded wallet).
- Not exercised in CI; `tsc`/`hardhat test` were not run here.

## Suggested follow-ups
- Promote `submitMovingFundsProof` into the SDK `maintenance.spv` service.
- Add `system-tests` to the pnpm workspace / point its SDK dep at the FROST
  package so this (and future FROST SDK coverage) compiles + runs.


mswilkison:
Thanks — all three are valid; fixed in 91f6dda61.

- **[P2] compile break** — right: `tsconfig` includes `test/` and `.mocharc` uses `ts-node/register/files`, so every `.ts` under `test/` compiles at load, before `describe.skip` can help. The test file now imports nothing (no `Testnet4`/4.x symbols), and the reusable helpers moved out of the compiled tree into the runbook appendix (`test/utils/frost.ts` removed). `yarn test` compiles against the pinned `^2.3.0`.
- **[P2] ECDSA WIF** — right: it went through `setupSystemTestsContext()`, which mandates `WALLET_BITCOIN_WIF`. No setup now; a FROST context keyed on the compressed group key is noted as required when this is implemented.
- **[P2] false green** — right: the body was a no-op after commented TODOs. It is now a single `it(...)` with no callback → reported **pending**, never a passing lifecycle.

Kept as a pending template + verified runbook rather than pretending it runs; wiring the runnable test still needs system-tests linked to the FROST 4.x SDK + a live node stack (both called out in the description).

mswilkison:
Both fixed in the latest push.

- **[P2] unknown lint rule** — removed the `mocha/no-pending-tests` disable; system-tests has no `eslint-plugin-mocha`, so the disable itself broke `eslint .`. The pending `it()` needs no suppression there.
- **[P2] requestRedemption arg** — right, `Bridge.requestRedemption` takes `bytes20 walletPubKeyHash`. The runbook snippet now derives it (`BitcoinHashUtils.computeHash160(walletPublicKey)`) before the call, and notes the contrast that `maintenance.spv.submitRedemptionProof` accepts the compressed key (it hashes internally).

mswilkison:
Fixed in the latest push — both stemmed from the raw-ethers redemption snippet needing manual serialization, so I switched it to the SDK wrapper you suggested.

- **[P2] computeHash160 arg / [P2] raw Bridge arg formatting** — the snippet now calls `depositorSdk.tbtcContracts.bridge.requestRedemption(walletPublicKey, walletMainUtxo, redeemerOutputScript, redemptionAmount)`. That wrapper derives the `bytes20` walletPubKeyHash from the compressed key, puts the main-UTXO tx hash in Bitcoin-internal byte order, and length-prefixes the redeemer script — so the manual `computeHash160(...)` (which wanted a `Hex`, not a string) is gone and the raw struct/byte-order formatting is no longer hand-rolled.

PR #1025, opened 2026-07-14
fix(starknet): correct SDK relayer conflict handling

lrsaturnino:
## Problem

When the relayer responded with HTTP 409 to an already-in-flight deposit, the SDK returned the relayer's deposit ID as a Hex, consumed as a reveal tx hash. The conflict lookup also trusted whatever deposit ID and status the relayer returned without validating it was genuine. Separately, a dead branch left local-relayer development routing permanently disabled.

## Solution

Every 409 now routes through a single handler that always throws a typed `StarkNetRelayerDepositConflictError` (renamed from `RelayerDepositConflictError`/`RelayerDepositStatus` to match this module's StarkNet-prefixed export convention); there is no longer a 409 path that returns a partial object as a successful receipt. The handler validates the deposit ID's shape against a canonical-integer/uint256-range pattern, independently re-derives the deposit ID from the funding transaction and warns on (and downgrades) a mismatch against the relayer's reported ID, and only marks a status verified when the relayer's response echoes the same ID and reports one of the three recognized on-chain states. Deposit-ID derivation itself was also corrected as part of this fix: the previous code mislabeled a double-keccak hash as "double SHA-256" and packed the output index as `uint256`; it now uses a real Bitcoin SHA-256d digest and packs the index as `uint32`, matching this SDK's on-chain deposit-key formula (`EthereumBridge.buildDepositKey`) - this changes the exact decimal value logged for every deposit ID. StarkNet chain-ID recognition is now an explicit allow-list (rejecting prototype-pollution-style keys such as `"toString"`/`"__proto__"` and unrecognized/mistyped chain IDs) instead of a silent `map[key] || "StarknetTestnet"` fallback. Local-relayer development routing was **re-enabled** by removing the dead `&& false` guard that had permanently disabled it (this activates the local-dev routing path; it does not remove any code path). Generated API docs were updated for the new error type.

## Breaking Changes

- Constructing a `StarkNetBitcoinDepositor` (directly, or via `loadStarkNetCrossChainInterfaces`) with an unrecognized chain ID now throws synchronously at construction time instead of silently defaulting to the testnet relayer route. Only chain IDs in the SDK's known StarkNet chain-ID map (mainnet, Sepolia, legacy Goerli) default correctly; a caller passing any other chain ID (including non-hex identifiers like the string `"SN_MAIN"`, which never worked as intended) must now supply both `relayerUrl` and `relayerStatusUrl` explicitly.
- New optional `relayerStatusUrl` field on `StarkNetBitcoinDepositorConfig`, also newly accepted as an optional parameter on `loadStarkNetCrossChainInterfaces` (falls back to `STARKNET_RELAYER_STATUS_URL`).
- `RelayerDepositStatus` and `RelayerDepositConflictError` are renamed to `StarkNetRelayerDepositStatus` and `StarkNetRelayerDepositConflictError` to match this module's naming convention (both are new in this PR, so this is a rename of unreleased API, not a break of prior behavior).
- initializeDeposit now rejects with StarkNetRelayerDepositConflictError on every HTTP 409, including the QUEUED (no-on-chain-record) case. Previously, the QUEUED case resolved with a Hex deposit ID. Integrators should branch on `error instanceof StarkNetRelayerDepositConflictError` and inspect `depositId`, `status`, and `statusVerified`.
- The success-path response check is now strictly typed: only the literal boolean `true` is accepted; truthy non-boolean values such as 'true', 1, {}, or [] now throw instead of resolving to a receipt. The deployed relayer emits a JSON boolean.

## Tests

New unit tests cover every terminal relayer status arriving via 409, malformed and mismatched deposit IDs, restored local-relayer routing, the `relayerStatusUrl`-only defaulting combination, and the default status endpoint's local-vs-production origin detection. `yarn build`, `yarn test`, and `yarn format` pass from `typescript/` (755+ tests); docs were regenerated and confirmed to match the committed output.

## Summary by CodeRabbit

* **New Features**
  * Improved StarkNet deposit conflict handling with typed errors, status verification, canonical deposit IDs, and configurable relayer status endpoints.
  * Added automatic relayer URL routing for supported StarkNet networks.
  * Extended cross-chain contract loading to accept an optional relayer status URL.

* **Documentation**
  * Added API references for relayer deposit statuses and conflict errors.
  * Corrected source links throughout the API reference.

* **Tests**
  * Expanded coverage for conflicts, retries, validation, routing, and deposit ID derivation.

* **Chores**
  * Updated ignore rules for Yarn, Python, Rust, and build artifacts.


piotr-roslaniec (commented):
This review is framed as questions for the author. The core change, where a 409 no longer fabricates a success `Hex` and instead always throws a typed `RelayerDepositConflictError`, reads as correct and well-motivated. A few questions below, most inline.

**Propagation test coverage**
The stated goal is that a 409 surfaces a typed throw to the SDK consumer (`CrossChainDepositor.revealDeposit`) instead of a fabricated `Hex`. The 409 tests in `typescript/test/lib/starknet/starknet-depositor-implementation.test.ts` assert one level lower, at `StarkNetBitcoinDepositor.initializeDeposit`. Propagation currently holds only because `revealDeposit` (`typescript/src/services/deposits/cross-chain.ts:89-131`) has no try/catch, but nothing pins that. Would a test that wraps the depositor in `CrossChainDepositor` and asserts `RelayerDepositConflictError` propagates out of `revealDeposit` be worth adding, so a future change inside `revealDeposit` cannot silently swallow the error?

**Dead 409 branch in isRetryableError**
`typescript/src/lib/starknet/starknet-depositor.ts:676` still special-cases `error.response?.status === 409`, but a 409 is now short-circuited into `handleDepositConflict` at `:478` before this formatter runs. Is that branch still reachable, or is it dead now and safe to remove? (It falls outside the diff hunks, so noting it here rather than inline.)

typescript/src/lib/starknet/starknet-depositor.ts
```diff
@@ -98,32 +231,19 @@ export class StarkNetBitcoinDepositor implements BitcoinDepositor {
...
-      } else {
-        // Production URLs with chain-specific endpoint
-        enhancedConfig.relayerUrl = `https://tbtc-crosschain-relayer-swmku.ondigitalocean.app/api/${chainName}/reveal`
-      }
+      enhancedConfig.relayerUrl = getDefaultStarkNetRelayerUrl(config.chainId)
+    }
+    // Deposit-status endpoint, sibling of the reveal endpoint above. Used
+    // to verify the real outcome of a deposit the relayer reports as a
+    // conflict (HTTP 409) instead of trusting the conflict response alone.
+    if (!enhancedConfig.relayerStatusUrl) {
```
piotr-roslaniec: When a consumer overrides `relayerUrl` but not `relayerStatusUrl`, `relayerStatusUrl` still defaults to the Threshold production/localhost endpoint. A 409 from the custom relayer would then be verified against an unrelated service. Is that intended, or should `relayerStatusUrl` default only when `relayerUrl` was also defaulted, so a custom `relayerUrl` implies a custom status endpoint?

typescript/src/lib/starknet/starknet-depositor.ts
```diff
@@ -390,6 +507,156 @@ export class StarkNetBitcoinDepositor implements BitcoinDepositor {
...
+
+    if (depositId) {
+      try {
+        const statusResponse = await this.queryRelayerDepositStatus(depositId)
+        if (
+          statusResponse &&
+          // The status response must confirm the same canonical deposit ID
+          // that was queried - a mismatched ID means the response cannot be
+          // trusted to describe this conflict.
+          statusResponse.depositId === depositId &&
```
piotr-roslaniec: Verification here confirms the relayer echoes back the same `depositId` it just reported in the 409. It does not cross-check against the locally-derived deposit ID computed in `initializeDeposit`. Is the relayer intended to be the source of truth for the ID, or would warning on a divergence between the local and relayer IDs add a useful signal that the recovery handle is the right one?

typescript/src/lib/starknet/starknet-depositor.ts
```diff
@@ -390,6 +507,156 @@ export class StarkNetBitcoinDepositor implements BitcoinDepositor {
...
+    )}`
+
+    const response = await axios.get<RelayerDepositStatusResponse>(statusUrl, {
+      timeout: 15000,
+      headers: {
+        "Content-Type": "application/json",
+      },
+    })
+
+    if (!response.data?.success) {
```
piotr-roslaniec: `!response.data?.success` gates an untrusted JSON body, so `success: "false"` (a string) or any truthy non-boolean would pass as success. Would `response.data?.success !== true` be safer here, to require a strict boolean `true`?

typescript/src/lib/starknet/starknet-depositor.ts
```diff
@@ -66,6 +117,88 @@ export interface StarkNetBitcoinDepositorConfig {
...
+}
+
+/**
+ * Resolves the default relayer reveal URL for a StarkNet chain when no
+ * explicit relayerUrl override is configured.
+ * @param chainId The StarkNet chain ID.
+ * @returns The relayer reveal URL for the given chain.
+ */
+function getDefaultStarkNetRelayerUrl(chainId: string): string {
+  const chainName = STARKNET_RELAYER_CHAIN_NAMES[chainId] || "StarknetTestnet"
```
piotr-roslaniec: An unrecognized `chainId` falls back to `"StarknetTestnet"` here, and again at `:175` for the status URL. For an unexpected chainId this silently routes to testnet rather than failing. Is the fallback intended, or should an unrecognized chainId throw?

typescript/src/lib/starknet/starknet-depositor.ts
```diff
@@ -390,6 +507,156 @@ export class StarkNetBitcoinDepositor implements BitcoinDepositor {
...
+    if (!this.#config.relayerStatusUrl) {
+      return undefined
+    }
+
+    const statusUrl = `${this.#config.relayerStatusUrl}/${encodeURIComponent(
+      depositId
+    )}`
+
+    const response = await axios.get<RelayerDepositStatusResponse>(statusUrl, {
+      timeout: 15000,
```
piotr-roslaniec, line 852: On a 409 this status GET can add up to 15s, on top of the existing retry backoff, before the conflict is thrown. Since the outcome is always a throw regardless of the verified status, is the extra round-trip worth the latency, or is the verified status (used to tailor the three error messages and expose `statusVerified` to consumers) the intended payoff? Raising as a design question, not a defect.

PR #1035, opened 2026-07-17
feat(vault): optimistic minting exposure cap and per-minter rate limits

mswilkison:
## Summary

Adds native, governance-tunable limits to the optimistic minting path in `TBTCOptimisticMinting`, bounding the worst-case TBTC issuance through optimistic minting. Optimistic minting is currently paused on mainnet; these controls are part of the hardening required before it can be re-enabled. They apply to future deployments of `TBTCVault` — the currently deployed vault is not upgradeable in place.

Today, a single compromised Minter key can request optimistic mints for the bridge's entire TVL, with Guardian cancellation as the only backstop during the `optimisticMintingDelay` window. With these limits, the worst-case unauthorized issuance is bounded by the caps instead of being unbounded.

## Design

The primary control is a cap on **outstanding optimistic minting exposure** rather than a time-windowed rate: at any moment, the TBTC that could turn out to be unbacked is exactly the optimistic minting debt not yet repaid by swept deposits, plus the value of in-flight (requested, not yet finalized) mints. The contract now tracks both quantities globally (`optimisticMintingDebtTotal`, mirroring the per-depositor `optimisticMintingDebt` ledger, and `optimisticMintingPendingTotal`) and rejects requests that would push their sum above the cap.

This shape has three properties a pure rate limit lacks:

- **The bound holds regardless of detection time.** Debt of deposits that never get swept — the signature of a fraudulent mint — is never repaid, so it permanently consumes the cap. A rate limit refills and keeps feeding a slow attacker; the debt cap is an automatic circuit breaker that trips exactly when unverified issuance accumulates, and stays tripped until governance deliberately resolves the situation (e.g. raises the cap through the delayed update).
- **Capacity recycles at the speed of settlement.** Legitimate mints return their capacity as soon as the backing deposits are swept and the debt is repaid. Conversely, if sweeping stalls, optimistic minting throttles itself rather than growing exposure against a settlement path that is not confirming.
- **In-flight requests count.** The check includes the value of pending requests, so the cap cannot be overshot by submitting many requests in one block and finalizing them together. Cancellation by a Guardian releases the pending value (nothing was minted), and finalization converts it into outstanding debt.

Four limits, each disabled by setting it to zero:

| Parameter | Default | Meaning |
|---|---|---|
| `optimisticMintingDebtCap` | 10 BTC | Max outstanding + in-flight optimistic minting exposure at any moment |
| `optimisticMintingCapPerMinter` | 10 BTC | Per-Minter token-bucket capacity: 10 BTC, refilling continuously at 10 BTC per 24h (a full bucket can be spent immediately and again after 24h, so up to ~2x the cap can be requested within an arbitrary rolling 24h window) |
| `optimisticMintingMaxDepositSize` | 5 BTC | Max size of a single optimistically minted deposit; larger deposits follow the standard sweep flow |
| `optimisticMintingRequestLimitPerMinter` | 100 | Per-Minter token-bucket capacity: 100 requests, refilling continuously at 100 requests per 24h; bounds burst request volume (and, over time, the long-run rate) Guardians may need to validate and cancel, independent of value |

The per-Minter limits stay time-based deliberately. Debt repayment reaches the vault as per-depositor amounts with no per-Minter attribution, so a per-Minter debt cap is not cleanly implementable — and a per-Minter rate limit is independently valuable because it bounds burst and long-run value flow per Minter, keeping the Guardians' verification workload bounded over time (a continuously-refilling bucket bounds sustained rate, not a strict amount per single `optimisticMintingDelay` window — a full bucket can be spent, then spent again after a further 24h). Per-Minter caps are meant to overlap: their sum may exceed the debt cap, which remains the binding total.

Other semantics worth noting:

- **Value is measured at request time** from `Bridge.deposits().amount` (satoshi), credited gross to `optimisticMintingPendingTotal`. On finalization it converts to `optimisticMintingDebt`/`optimisticMintingDebtTotal` net of the treasury fee — pending exposure is therefore a conservative (slightly higher) estimate of the debt a request will eventually create.
- **Cancellation releases the in-flight value but does not restore the per-Minter bucket**, so request/cancel cycles from a misbehaving Minter stay bounded by that Minter's own limits. Deposits swept before finalization can no longer be finalized (finalization requires `sweptAt == 0`); their in-flight value is released via the new permissionless `releaseOptimisticMintForSweptDeposit`, safe to call by anyone precisely because a swept, unfinalized request can never be finalized or re-requested. Guardians should still proactively cancel requests that are unlikely to ever sweep (e.g. approaching refund locktime).
- **Deposits above the size limit** mint through the standard sweep flow instead. Historically, deposits above the default 5 BTC limit are roughly the p95 of individual deposit sizes; the limit also keeps any single deposit at half the default debt cap, so one large deposit cannot monopolize the optimistic minting pipeline while it settles.
- **Extension seam.** `_isOptimisticMintingThrottleExempt(address)` is an `internal virtual` hook (always `false` here) allowing derived vault implementations to exempt classes of requesters whose issuance is bounded by separate, dedicated exposure limits. Exempt requests bypass the checks but still count toward the measured totals; overriding contracts must account for that overlap.

Defaults were derived from historical mainnet optimistic minting activity (typical outstanding exposure given observed daily volume and reveal-to-sweep settlement times, and the per-deposit size distribution) and are set deliberately conservative for the initial re-enablement: hitting a cap costs only latency (deposits fall back to the sweep flow) while raising one takes the 24-hour governance delay, so the intended posture is to start tight and ratchet up based on observed cap-utilization telemetry.

## Governance

The four limits are updated together via the existing two-step idiom (`beginOptimisticMintingCapsUpdate` / `finalizeOptimisticMintingCapsUpdate`) with the same 24-hour `GOVERNANCE_DELAY` used by the fee and delay updates, mirroring the Bridge's grouped-parameter governance.

## Monitoring

- `OptimisticMintingAllowanceConsumed` is emitted on every request with the amount, the Minter's remaining value bucket, the Minter's remaining request-count bucket, and the remaining debt-cap headroom (sentinel `type(uint64).max`/`type(uint32).max` when a limit is disabled) — cap-utilization alerts double as an anomaly detector for a misbehaving Minter.
- `getOptimisticMintingAllowance(minter)` returns live remaining allowances (including accrued bucket refill) for Minter clients and dashboards; `optimisticMintingDebtTotal` and `optimisticMintingPendingTotal` are public for exposure monitoring.

## Testing

New suite `TBTCVault.OptimisticMintingCaps.test.ts` (35 tests, self-contained against a mocked Bridge): defaults; all four limits; cross-minter bucket independence; the debt cap binding across Minters; cancellation releasing headroom while per-Minter allowance stays consumed; the full settle-and-recycle cycle (in-flight → outstanding debt on finalization → capacity recycled on repayment via `receiveBalanceIncrease`); the cap lowered below current exposure clamping to zero headroom; continuous bucket refill and clamping; count-limit exhaustion and recovery; disabled-limit sentinels; finalization not double-consuming; governance update flow (authorization, delay, event, state reset); the enable-transition semantics (a limit enabled after being disabled starts every Minter from a full bucket, exercised with an exact-boundary request); fractional refill accrual for small request limits; and the exemption seam via a test stub, including exempt in-flight exposure still being measured.

The pre-existing vault suites are unchanged; default limits sit far above the fixture deposit sizes, so their behavior is unaffected.

## Compatibility

- New state variables change the storage layout relative to the deployed vault, but that is not the operative migration constraint: `TBTCVault` is a plain, non-proxy deployment (no upgradeable storage layout to preserve), so the layout change itself is safe. Delivering these controls to the live system instead requires the existing `initiateUpgrade`/`finalizeUpgrade` vault-migration path -- deploying a new `TBTCVault`, re-registering Minters/Guardians, and re-applying the optimistic minting fee/delay parameters on it. That migration is a separate, out-of-scope operational step from this PR; these controls target the next vault deployment.
- `requestOptimisticMint` gains new revert paths (size cap, per-Minter limits, debt cap). Minter clients should consult `getOptimisticMintingAllowance` before submitting.
- `optimisticMintingDebtTotal` maintenance adds one storage write to `finalizeOptimisticMint` and to debt-repaying `receiveBalanceIncrease` calls; behavior of finalization, Minter/Guardian management, and the pause switch is otherwise unchanged.
- New permissionless external function `releaseOptimisticMintForSweptDeposit(fundingTxHash, fundingOutputIndex)`, callable by anyone once a deposit with an unfinalized optimistic-mint request has been swept, releasing its in-flight pending exposure. Existing `cancelOptimisticMint` (Guardian-only) is unchanged.

https://claude.ai/code/session_01AnrPAUEuN6uKDVeASMxb4M


piotr-roslaniec:
## CI note: `contracts-build-and-test` failure is unrelated to this PR

The `contracts-build-and-test` job fails deterministically (3/3 fresh reruns) on:

```
L1BTCRedeemerWormhole (using real contract)
  "before all" hook
Error: Contract `contracts/cross-chain/wormhole/L1BTCRedeemerWormhole.sol:L1BTCRedeemerWormhole` is not upgrade safe
contracts/cross-chain/wormhole/L1BTCRedeemerWormhole.sol:143: Missing initializer calls for one or more parent contracts: `ReentrancyGuardUpgradeable`
```

**This PR never touches `L1BTCRedeemerWormhole.sol` or any wormhole file.** Evidence gathered before concluding this is out of scope:

- Isolated run of `test/cross-chain/wormhole/L1BTCRedeemerWormhole.test.ts` passes cleanly (58/58), on both Node 22 and Node 24 (CI's pinned version).
- A byte-for-byte fresh clone of this branch, with a clean `yarn install` (non-frozen, matching CI's `contracts-build-and-test` step exactly) and Node 24, does not reproduce the failure locally in the subset of the suite this sandbox can execute.
- `git diff` between the last-green commit (`bc7c921e`) and the next commit (`04fde3df`) touches only 3 files with pure Prettier reformatting (whitespace/line-break changes) — zero functional changes, zero wormhole involvement — yet the CI outcome flipped from pass to this failure.
- `@openzeppelin/upgrades-core` is pinned to an exact version (`1.20.0`) in the Berry lockfile; no dependency drift between the passing and failing runs.
- No other currently open PR (including branches that do touch wormhole/cross-chain code) shows this failure.

This sandbox could not reach CI's full-suite baseline pass count (~3123) either with or without live mainnet forking enabled, so a fully faithful local repro of CI's execution environment/order wasn't achievable to pin down root cause further (likely a test-ordering or shared-network-state interaction specific to the full CI suite, not something introduced by this PR's diff).

Flagging for a maintainer to dig into `L1BTCRedeemerWormhole.sol`'s upgrade-safety separately; recommend not blocking this PR's merge on it once the change itself is reviewed, or retrying CI.

piotr-roslaniec (approved):

piotr-roslaniec (approved):

piotr-roslaniec (approved):

PR #1043, opened 2026-07-25
feat(watchtower)!: bind the signed request into the boundary digest

mswilkison:
Blocker A, increment 1 of `BLOCKER-A-SCOPE.md`. Stacked on #1042.

## What it does

The irreversible-boundary binding — re-authorized after the outbox CAS and immediately before a signer or broadcaster runs — named the durable *slot* but not what the signer is asked to sign. Four groups were unbound, each naming something an authorizer or reconciler had no way to check:

| group | fields | why |
|---|---|---|
| lane | `laneID`, `signerIdentity` | the lane chooses which signer is invoked and which fee envelope applies, yet the authorized digest was lane-agnostic |
| intent | `intentID` (recomputed), `routerAddress`, `intentValueWei` | `intentID` commits to the router, calldata and value, but nothing at the boundary recomputed it. Router and value are carried in the clear too — an authorizer cannot invert a keccak |
| fee | `challengeValueWei`, `maxGasLimit`, `maxFeePerGas`, `maxPriorityFeePerGas`, `maxTotalFeeWei` | only the activation manifest hash was bound; the lane envelope actually handed to the signer was never named |
| replacement | `replacedTransactionHash` | the prepared-tx hash was bound for broadcast only, so nothing committed to which variant a replacement supersedes |

`intentID` is **recomputed** from the record's intent and checked against its durable identity — reading the stored ID back would prove nothing about the router and calldata. The envelope is **passed in by the caller** rather than re-derived, so the digest names what the signer receives rather than a value that ought to equal it, and the builder rejects an envelope whose lane is not the reservation's lane.

Each stage now has exactly one legal transaction-hash shape: prepare names none, replacement names the variant it supersedes, broadcast names the bytes it is about to send.

## Failure handling

Building the binding moved inside the try that already handles a rejected authorization. A binding that cannot be built names no invoked signer, so it must clear the durable pre-I/O marker. Leaving it outside meant a record whose stored intent identity no longer matched its contents threw out of `prepare()` with the marker set — and lease recovery cannot clear a marker it must assume names a real signer call, so `assertRecoveryBarrier` then blocked signing on **every lane** until an out-of-band resolution supplied attested evidence about a call that never happened.

## Breaking change

The request-binding digest domain moves v2 → v3 and the challenge / attestation / request / signed-attestation schemas move v3 → v4, because the hashed structure changed.

Nothing persists these digests and no frozen vector pins them. A reconciler signs the request-binding digest from a separate trust domain and must ship the identical v3 struct — but both provider interfaces are still test-only and automatic challenge submission is gated off, so there is no live counterparty today. That window closes once a reconciler is deployed.

## Also

- Deletes `signerBoundaryIdentity` — zero call sites on any ref, and its component order differed from both live realizations of the same identity (the SQL primary key and the in-memory double), so adopting it would have produced inconsistent keys.
- Two compiler guards replace what were hand-copied shapes: `NormalizedBoundaryBinding` becomes an alias of the exported type, and a type-level assertion pins the boundary and reconciler bindings to the same key set. Either firing turns "a field silently dropped from the digest" into a build error.
- uint256 fields are length-bounded before `BigInt` conversion, since that path parses remote attestation input before the envelope signature is verified.

## Verification

Watchtower suite with PostgreSQL: **387/387 → 390/390**, no title lost (compared as sets).

Every new guarantee is mutation-tested — stripping one field from the hashed struct, sourcing the envelope from the manifest instead of the signer's, reading the intent ID back instead of recomputing it, and building the binding outside the try each fail exactly one test.


mswilkison:

PR #1044, opened 2026-07-25
feat(watchtower)!: identify a signer invocation deterministically

mswilkison:
Blocker A, increment 2 of `BLOCKER-A-SCOPE.md`. Stacked on #1043.

## What it does

A signer boundary had no identity. What served as one was a wall-clock tuple — `(record_id, boundary_started_at_unix_ms, preparation_attempts, nonce_reservation_id)` — that was the primary key of the orphaned-boundary resolution table. It embedded `this.now()`, it was never sent to the signer, and it was not a digest, so a provider could not journal or dedupe per invocation.

`computeP2TRSignatureFraudSignerInvocationID` is a domain-separated digest **derived from the request-binding digest** rather than by walking the binding a second time. A second walk would need a third canonicalizer — precisely the silent-drop failure the key-set assertion from #1043 guards against — so deriving it means the identity inherits the binding's field coverage for free. The domain must differ, or the value would be byte-identical to a digest that already travels to an external attester and is about to become a durable primary key.

It is committed **in the same swap that sets the pre-I/O marker**. That is possible because the binding reads no marker and is built from the record `nextRecord` already produced, which carries the post-swap version. Building before the swap also inverts the failure for the better: a binding that cannot be built now strands nothing, because no durable marker exists yet.

## What the identity does *not* replace

The boundary start, the preparation attempt and the nonce reservation are still compared against the durable row, in both the TypeScript assert and the PostgreSQL guard.

An earlier revision of this branch dropped them on the theory that matching `preparation_attempts` was racy. That was wrong, and an adversarial review disproved it by reproduction: the attempt cannot drift while the marker is set (`prepare` requires status `queued`, `prepareReplacement` clears the marker in the same swap, `recoverExpiredPreparation` retains it only when quarantined), the start is immutable in flight and NULL-paired with the ID, and the reservation cannot be NULLed under an active marker. All three were race-free and free to keep — and dropping them let a resolution carrying the correct ID but fabricated start / attempt / reservation be **accepted**, writing permanently wrong forensics into two append-only tables.

PostgreSQL compares the identity against the durable column rather than re-deriving it: the binding preimage spans three tables and a TypeScript layout. That is the same standard `nonce_reservation_id` already met.

## Breaking change

Migration 003 gains `active_signer_invocation_id` and `signer_invocation_id`, re-keys the boundary resolution onto `PRIMARY KEY (record_id, signer_invocation_id)`, and moves the resolution evidence digest to `tbtc-p2tr-signer-boundary-independent-resolution-v2`. 003 is not on `main` and the table is append-only, so this is an in-place edit with no data-migration path.

A CHECK pairs the **active** marker and its identity as NULL-together: the activation barrier counts outstanding invocations by that marker's null transitions alone, and would otherwise count an invocation with no identity. The **historical** pair is deliberately not paired — a failure record may have to assert that some signer call began from a fallback clock when no identity survived.

`record_id` stays the leading PK column because the PK is the table's only index and two queries filter on it alone.

Both halves of one retirement move together: the provenance-incident resolution is written in the same transaction, so it carries the same identity and its digest moves to v2. Leaving it on the wall clock would give one boundary two identities.

## Verification

Watchtower suite with PostgreSQL: **393/393 → 397/397**, no existing title lost.

TS/SQL digest parity verified byte-for-byte on two vectors against a live PostgreSQL 16 under the v2 layout.

Three mutants each killed by exactly the intended test: writing the identity after the swap instead of within it, keying the in-memory double on the record alone, and dropping the three restored equalities.

New coverage fills gaps the review located rather than the suite: the identity is committed with the marker and promoted on completion; the replacement boundary gets its **own** distinct identity (deleting that line previously left 183/183 green); and two distinct boundaries on one record stay apart — nothing tested that before, so re-keying the in-memory map to `recordID` alone would have passed the whole suite while diverging from the primary key.

## Known, out of scope

The broadcast boundary gets no invocation ID: it owns no signer marker and has no resolution table, so an ID would have nowhere to live.

Separately noted while verifying the DDL: migration **004 does not apply to PostgreSQL at all** — `ERROR: syntax error at or near "authorization"`, a reserved word used as a table alias. Pre-existing; nothing executes 004 because the migration test only regex-matches file text.

### DIFF
--- a/services/watchtower/migrations/003_p2tr_signature_fraud_challenge_outbox.sql
+++ b/services/watchtower/migrations/003_p2tr_signature_fraud_challenge_outbox.sql
@@ -315,9 +315,33 @@ CREATE TABLE p2tr_signature_fraud_challenge_outbox (
         signer_invocation_started_at_unix_ms IS NULL
         OR signer_invocation_started_at_unix_ms BETWEEN 0 AND 9007199254740991
     ),
+    -- Deterministic identity of the boundary the marker below names, committed
+    -- in the same swap. The activation barrier counts outstanding invocations
+    -- by the marker's NULL transitions alone, so the two must be NULL together
+    -- or the counter stops meaning "a signer call may be outstanding".
+    signer_invocation_id bytea CHECK (
+        signer_invocation_id IS NULL
+        OR octet_length(signer_invocation_id) = 32
+    ),
     active_signer_invocation_started_at_unix_ms bigint CHECK (
         active_signer_invocation_started_at_unix_ms IS NULL
         OR active_signer_invocation_started_at_unix_ms BETWEEN 0 AND 9007199254740991
+    ),
+    active_signer_invocation_id bytea CHECK (
+        active_signer_invocation_id IS NULL
+        OR octet_length(active_signer_invocation_id) = 32
+    ),
+    CHECK (
+        (active_signer_invocation_started_at_unix_ms IS NULL)
+            = (active_signer_invocation_id IS NULL)
+    ),
+    -- The historical pair is deliberately NOT required to agree. The active
+    -- pair drives the barrier counter and must be exact; the historical marker
+    -- is proof that some signer call began, and a failure record may have to
+    -- assert that from a fallback clock when no identity survived.
+    CHECK (
+        signer_invocation_id IS NULL
+        OR signer_invocation_started_at_unix_ms IS NOT NULL
     ),
     latest_variant_sequence smallint CHECK (
         latest_variant_sequence IS NULL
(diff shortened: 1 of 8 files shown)


mswilkison:
Operational note for reviewers, surfaced while scoping the next increment.

This PR edits migration `003` **in place** rather than adding a new migration file. That is safe only while no database has applied `003`, because the migration runner pins a checksum per applied migration and refuses to start on drift:

`services/watchtower/src/P2TRWatchtowerMigrations.ts:292-320` — `assertAppliedMigrationPrefix` throws ``Database watchtower migration ${version} does not match ${expected.filename}`` when the file's checksum differs from the row recorded at application time.

`003` is not on `main`, the subsystem is gated off in production, and the resolution table is append-only (no `UPDATE`/`DELETE`), so there is no deployed database to migrate and no data-migration path anyway. But the moment `003` reaches any environment that runs the migrator, it is frozen and this change would have to become `005_*.sql` instead.

Worth a reviewer's explicit sign-off that no environment has applied `003` yet.

mswilkison:

PR #1045, opened 2026-07-25
feat(watchtower)!: make the signer name the request it served

mswilkison:
Blocker A, increment 3a of `BLOCKER-A-SCOPE.md`. Stacked on #1044.

## What it does

The request digest was process-local evidence plumbing: computed in the boundary authorizer to match a reconciler attestation, then discarded. Never persisted, never transmitted. The SDK signer interface took `(intent, reservation, feePolicy)` and the prepared transaction echoed nothing, so a response carried no statement about which request produced it.

Both prepare methods now take a `P2TRSignatureFraudSignerInvocationRequest` of `{ invocationID, requestDigest }`, and the prepared transaction carries the echo back. The type is deliberately **opaque** — the caller's boundary binding is a watchtower record shape and the dependency runs watchtower → SDK, so the SDK carries the two identities and nothing else.

## What this actually buys — read this before reviewing

**The echo is an assertion, not evidence.** The digest is a plaintext parameter and a signer can copy it beside bytes it signed for something else. Nothing about the signature is proven by it.

Every argument handed to the signer — intent, reservation, fee policy, previous variant — is *already* reconstructed from the signed bytes and enforced field by field (`validateP2TRSignatureFraudPreparedChallengeTransaction`, `...Reservation`, `validatePreparedTransactionFeePolicy`, the replacement validator). What the echo adds is exactly the coordinates that never appear in a transaction: record version, generation, attempt, stage, provenance fingerprint, reservation ID.

So the honest value is: it catches a response that does not belong to this request — a crossed transport, a cached reply from an earlier attempt, another tenant's queued job, a signer that silently ignores parameters it does not understand — and it makes the request identity a value the **store** and the out-of-band boundary resolver can check, rather than process-local state. That last part is what pays for the increment.

## Classification of a mismatch

A wrong echo is neither `ambiguous-signer-invocation` (the call did not throw) nor `malformed-signed-envelope` (nothing about the bytes is wrong). By that point the transaction is fully authenticated: it parses, recovers to the reserved sender, and consumes the reserved nonce.

So the echo is compared **in the outbox after validation**, not inside the SDK validator — which is what lets the authenticated artifact be captured as an unexpected signed artifact instead of being dropped by the malformed path. The boundary stays marked invoked, because the signer *was* called and may hold further bytes never seen. New reason code `wrong-signer-invocation-request`, added to the durable CHECK enum.

## No new column, no new migration

`activeSignerInvocationID` is already durable, already committed in the same swap as the pre-I/O marker, and is itself `sha256(domain ‖ requestDigest)` — so an echo is verifiable against state that already exists. `computeP2TRSignatureFraudSignerInvocationRequest` returns both identities from one canonicalization, which keeps the outbox off a second canonicalization path; the digest is otherwise reachable only inside the authorizer.

## Why the SDK validator's third parameter is optional

The durable variant ledger re-validates historical variants long after their request is gone. Supplied → the echo must match; omitted → it is preserved untouched and never invented. Preservation is load-bearing: that validator returns a freshly built literal, so a field it does not name is silently dropped.

## Verification

- Watchtower with PostgreSQL: **397/397 → 400/400**, no title lost.
- SDK suite: **983 passing, 1 failing — identical to baseline.** The failure (`Previous transaction raw bytes do not match the requested input txid`) is pre-existing and untouched; the SDK is not modified by #1043 or #1044, so it predates this stack's watchtower work.
- Two mutants each caught: never comparing the echo, and letting the SDK validator drop it from its return literal.
- SDK `dist` rebuild is required for the watchtower to see the new types (`tsc --project tsconfig.build.json`); the watchtower resolves `@keep-network/tbtc-v2.ts` to the built output, not `src`.

## Follow-ups this surfaced, not fixed here

**The EIP-1559 access list is never validated anywhere.** A signer can attach an arbitrary access list today. The fee-policy check bounds the money but not intrinsic gas consumption inside the signer's own `gasLimit`, so a stuffed list can push the call to out-of-gas — a revert that consumes the nonce and drives the record to `terminal-nonce-consumed`. Worth fixing on its own merits.

**Binding the digest into the signed bytes is feasible and would make this evidence rather than assertion.** An EIP-2930 access-list entry `{address, storageKeys:[digest]}` serializes, is covered by the signature (it changes the transaction hash), and round-trips through this repo's ethers 5.5 — verified. Cost is ~4300 intrinsic gas and the digest becomes public on-chain. Calldata is *not* an option: it is committed by `intentID`, which the digest is derived from, so it is circular. Validating the access list is the prerequisite for both, which is why they are worth doing together.

### DIFF
--- a/services/watchtower/migrations/003_p2tr_signature_fraud_challenge_outbox.sql
+++ b/services/watchtower/migrations/003_p2tr_signature_fraud_challenge_outbox.sql
@@ -2111,6 +2111,7 @@ CREATE TABLE p2tr_signature_fraud_challenge_signer_quarantine (
         'wrong-sender',
         'wrong-nonce',
         'malformed-signed-envelope',
+        'wrong-signer-invocation-request',
         'invalid-replacement-envelope',
         'reservation-binding-mismatch',
         'reservation-provider-failure'
--- a/services/watchtower/src/P2TRSignatureFraudChallengeOutbox.ts
+++ b/services/watchtower/src/P2TRSignatureFraudChallengeOutbox.ts
@@ -30,7 +30,7 @@ import {
 
 // A value import, unlike the authorization module's type-only import of this
 // file, so the dependency runs one way at run time and there is no cycle.
-import { computeP2TRSignatureFraudSignerInvocationID } from "./P2TRSignatureFraudIrreversibleBoundaryAuthorization.js"
+import { computeP2TRSignatureFraudSignerInvocationRequest } from "./P2TRSignatureFraudIrreversibleBoundaryAuthorization.js"
 import type { P2TRSignatureFraudWatchtowerStoreProfileProvider } from "./types.js"
 
 export const P2TR_SIGNATURE_FRAUD_OUTBOX_MAX_PAGE_SIZE = 1_000
@@ -1193,6 +1193,7 @@ export type P2TRSignatureFraudSignerQuarantine = {
     | "wrong-sender"
     | "wrong-nonce"
     | "malformed-signed-envelope"
+    | "wrong-signer-invocation-request"
     | "invalid-replacement-envelope"
     | "reservation-binding-mismatch"
     | "reservation-provider-failure"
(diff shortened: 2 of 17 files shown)


mswilkison:

PR #1046, opened 2026-07-25
fix(watchtower)!: require a provider tombstone to claim never-invoked

mswilkison:
Stacked on #1045. Fixes a soundness defect in the orphaned-boundary resolver introduced in #1041.

## The defect

`resolveOrphanedSignerBoundary` accepts `never-invoked` on two attestations from distinct trust *and* independence domains. That is the right shape for a claim about something observable — and the wrong shape for this one.

**Nobody can observe the absence of a request.** Delivery delay between the outbox and the key is unbounded: a request can sit in a proxy or a queue and arrive *after* the outbox crashed, after it recovered, and after both attesters inspected the signer and found nothing. Reading provider state cannot settle it at any instant, because the answer can change afterwards.

Only a **write** at the provider settles it — a one-shot transition that both reports "no bytes escaped" and refuses any later arrival of the same invocation.

`never-invoked` matters because it is the outcome that releases the reserved nonce. Getting it wrong while signed bytes exist is exactly the two-transactions-one-nonce failure the whole boundary machinery exists to prevent.

## The fix

`never-invoked` now requires a provider tombstone naming the invocation, and is refused without one.

The two attestations keep their job, but **the proposition they attest changes** — from "the signer was never invoked", which they are not positioned to know, to "the provider produced this receipt", which they can check.

This is expressible only because #1044 gave the boundary a deterministic identity. Before that there was no stable name for a provider to tombstone or for an attester to check a receipt against. That is why the fix lands here rather than in #1041 where the defect is.

## Enforced at three layers, deliberately

1. The **digest function** refuses to produce a digest for the inconsistent shape at all, so a `never-invoked` resolution without a tombstone cannot be constructed.
2. The **resolver** refuses it.
3. The **database** refuses it independently — by trigger and by CHECK — for a caller that bypasses TypeScript entirely.

The tombstone is bound into the resolution evidence digest, so both attestations commit to it and a resolution cannot be re-pointed at another receipt.

## Verification

- Watchtower with PostgreSQL: **400/400 → 402/402**, no title lost.
- Digest domain v2 → v3; TS/SQL mirror verified **byte-for-byte against a live PostgreSQL 16 on both branches** (tombstoned and not).
- Removing *both* database defences leaves the raw-SQL guard test failing, so the layers are independently covered rather than shadowing each other.

## What this does not do

It does not specify the provider contract — that is increment 3b. It makes the *shape* of the requirement explicit and unforgeable-by-omission, so the outcome cannot be claimed without the evidence that would justify it.

The design analysis behind this also produced two things worth their own work:

- **Journal-before-egress, not before key use.** The scope doc's phrasing ("provider journals its result before key use") is incoherent — the result does not exist before key use. The decisive predicate is not "was the key used" but "can bytes for this invocation ever cross the egress boundary". A crash between computing σ and committing is *safe* under the right ordering.
- **Consuming the contested nonce on-chain** dissolves most of this ambiguity rather than resolving it. Ethereum already guarantees at most one transaction confirms per nonce, so a self-transfer at the contested nonce, priced above the authorized envelope's committed cap, terminates every ambiguous case in bounded, chain-observable time with zero provider cooperation.

### DIFF
--- a/services/watchtower/migrations/003_p2tr_signature_fraud_challenge_outbox.sql
+++ b/services/watchtower/migrations/003_p2tr_signature_fraud_challenge_outbox.sql
@@ -1737,6 +1737,31 @@ CREATE TABLE p2tr_signature_fraud_challenge_signer_boundary_resolution (
     signer_invocation_id bytea NOT NULL CHECK (
         octet_length(signer_invocation_id) = 32
     ),
+    -- Proof that the provider itself closed this invocation. Required exactly
+    -- for 'never-invoked': two attesters can report what they observed, but
+    -- nobody can observe the absence of a request still buffered in transit --
+    -- only a fencing write at the provider settles that.
+    provider_tombstone_receipt bytea CHECK (
+        provider_tombstone_receipt IS NULL
+        OR octet_length(provider_tombstone_receipt) BETWEEN 1 AND 2048
+    ),
+    provider_tombstone_at_unix_ms bigint CHECK (
+        provider_tombstone_at_unix_ms IS NULL
+        OR provider_tombstone_at_unix_ms BETWEEN 0 AND 9007199254740991
+    ),
+    CHECK (
+        (outcome = 'never-invoked')
+            = (provider_tombstone_receipt IS NOT NULL)
+    ),
+    CHECK (
+        (provider_tombstone_receipt IS NULL)
+            = (provider_tombstone_at_unix_ms IS NULL)
+    ),
+    CHECK (
+        provider_tombstone_at_unix_ms IS NULL
+        OR provider_tombstone_at_unix_ms
+            BETWEEN boundary_started_at_unix_ms AND resolved_at_unix_ms
+    ),
     boundary_started_at_unix_ms bigint NOT NULL CHECK (
         boundary_started_at_unix_ms BETWEEN 0 AND 9007199254740991
     ),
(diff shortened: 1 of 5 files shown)


mswilkison:
Two consequences of this change that reviewers should weigh, from the design analysis behind it.

## `never-invoked` is now unreachable in production, and that is a DoS lever

No provider implements the tombstone, so the outcome cannot currently be produced. That was the right trade — the previous state let an attester assert an unobservable absence, and a wrong `never-invoked` releases the nonce while signed bytes exist, which is exactly the two-artifacts-one-nonce failure the boundary machinery exists to exclude. A silent safety hole is worse than a loud availability one.

But the cost should be named plainly: **one orphaned invocation now wedges the entire challenge pipeline** until an operator writes `terminal-unsafe`, which itself keeps activation blocked. The freeze is three-part and global — an orphan keeps `hasExpiredPreparationLeases` true so `assertRecoveryBarrier` throws on every lane; it pins `active_signer_invocation_count >= 1` which the release trigger requires to be `0`; and any pending release sets `unresolved_release_count > 0` which the signer-increment gate requires to be `0`. For a fraud watchtower, availability is the mission, so this state is acceptable only briefly.

Anyone triaging an orphan should know the resolution they would naturally reach for is currently refused by the trigger, by design.

## A clock-skew hazard for whoever implements the provider

The resolver requires `tombstonedAtUnixMs` to fall within `[boundaryStartedAtUnixMs, resolvedAtUnixMs]`. A provider that stamps its own clock can therefore produce a **correctly signed receipt that this guard rejects**, purely from skew. The provider contract should clamp the tombstone timestamp to a caller-supplied floor rather than trusting its local clock.

Worth writing into the provider spec before anyone builds against this, since the failure looks like a forged receipt rather than a clock problem.

mswilkison:

PR #1047, opened 2026-07-25
feat(watchtower)!: settle an orphaned boundary on finalized nonce consumption

mswilkison:
Blocker A, increment 3b-i. Stacked on #1046.

## Why

#1046 gated `never-invoked` on a provider tombstone, correctly — nobody can observe the absence of a request still in transit. But no provider implements tombstones, so that outcome is unreachable, and one orphan wedges challenge signing on **every lane** until an operator writes `terminal-unsafe`.

There is a fourth exit that needs nobody's cooperation. Once the reserved nonce is consumed on-chain at finality, any signed bytes for it are permanently inert — they can never confirm. The chain has already settled the question. That is a **decidable base case**, which `never-invoked` is not.

## What it claims, precisely

A statement about the **bytes**, not the invocation. The signer may have signed and may still hold an envelope; this says only that whatever it holds can no longer land. Three consequences, each a requirement rather than a refinement:

**The record moves to `quarantined`, not back to `preparing`.** Clearing the marker and stopping would leave a live reservation in the state lease recovery treats as pre-signer — it would void the nonce guard and hand the nonce back to the allocator, dropping the `(chain_id, sender, nonce)` partial-unique index that stops a second reservation at it. The whole claim is that the nonce is **spent, not free**.

`generation-required` would be more informative but requires a linked disposition row, and producing one here means reconciling a record whose signer may still be live. `quarantined` is itself reconcilable, so the ordinary loop carries it onward through the normal path.

**The reservation is retained and `captureEscapedSignedArtifact` is re-anchored on it.** The live marker was its only authorization; clear it without this and a late-returning envelope is dropped with no alert and no escaped-envelope row. The bytes are harmless once the nonce is spent — the silence is not.

**It inherits none of `never-invoked`'s privileges.** No zero-escape precondition — it is permitted *precisely because* bytes may have escaped — and no incident retirement. Anyone who can transact from the sender can force this outcome with a fee bump, so if it also asserted cleanliness, a boundary where bytes escaped could be laundered clean using only the fee market. A record carrying signed state still raises its activation-blocking alert; a returned-signer marker still records an `ambiguous-signer-invocation` quarantine.

## Evidence

Binds `chainID`, which **nothing in the existing nonce-consumption evidence names** — without it two attestations over "sender S nonce N is consumed" would replay against any record on any chain sharing that pair.

Reuses `validateCanonicalNonceConsumingTransaction` unchanged, and deliberately does *not* import the terminal path's "the consuming transaction is not one of ours" check: the prepared-hash set is empty for an orphan, so it would be vacuous, and the recovery is identical either way.

## Breaking change

Digest domain → `tbtc-p2tr-signer-boundary-independent-resolution-v4`; outcome enum gains `nonce-consumed`. Migration 003 gains seven consumption columns with a biconditional CHECK against the outcome, an all-or-none CHECK across the seven, a strictly-past-the-nonce CHECK, and a read-at-the-finality-boundary CHECK. The guard additionally verifies the consumed nonce and chain against the durable outbox row.

## Verification

- Watchtower with PostgreSQL: **402/402 → 405/405**, no title lost.
- TS/SQL digest parity verified byte-for-byte against a live PostgreSQL 16 on both branches.
- Two mutants caught: leaving the record in `preparing`, and dropping the chain binding. The second one initially **survived** — the scenario that kills it had silently failed to insert, which is exactly what mutation testing is for.

## Known limitations

- Finality is a configured depth (`observedHead - finalizedThrough >= finalityConfirmationBlocks`), not a finalized tag, with no floor and no per-chain minimum. If this outcome is what unwedges a global barrier, that number deserves one.
- A deep reorg below the claimed finality is not retractable: both tables are append-only. The honest worst case is a liveness failure, not a double-submit.
- The lane is not released here; the sender stays pinned to this record until the ordinary reconcile loop produces a disposition. Safe, slightly less live than routing through the disposition machinery directly.

Next in the sequence is the burn itself (3b-ii), which produces this evidence automatically instead of waiting for someone to consume the nonce.

### DIFF
--- a/services/watchtower/migrations/003_p2tr_signature_fraud_challenge_outbox.sql
+++ b/services/watchtower/migrations/003_p2tr_signature_fraud_challenge_outbox.sql
@@ -1762,6 +1762,67 @@ CREATE TABLE p2tr_signature_fraud_challenge_signer_boundary_resolution (
         OR provider_tombstone_at_unix_ms
             BETWEEN boundary_started_at_unix_ms AND resolved_at_unix_ms
     ),
+    -- Evidence that the chain, not the provider, settled this boundary: the
+    -- reserved nonce is consumed at finality, so any bytes for it are inert.
+    -- chain_id is carried because nothing else in nonce-consumption evidence
+    -- names a chain, and without it an attestation over (sender, nonce) would
+    -- replay against any record on any chain sharing that pair.
+    nonce_consumption_chain_id numeric(78, 0) CHECK (
+        nonce_consumption_chain_id IS NULL OR nonce_consumption_chain_id > 0
+    ),
+    nonce_consumption_nonce bigint CHECK (
+        nonce_consumption_nonce IS NULL OR nonce_consumption_nonce >= 0
+    ),
+    nonce_consumption_account_nonce bigint CHECK (
+        nonce_consumption_account_nonce IS NULL
+        OR nonce_consumption_account_nonce >= 0
+    ),
+    nonce_consumption_read_at_block bigint CHECK (
+        nonce_consumption_read_at_block IS NULL
+        OR nonce_consumption_read_at_block >= 0
+    ),
+    nonce_consumption_transaction_hash bytea CHECK (
+        nonce_consumption_transaction_hash IS NULL
+        OR octet_length(nonce_consumption_transaction_hash) = 32
+    ),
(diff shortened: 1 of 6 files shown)


mswilkison:

PR #1049, opened 2026-07-25
feat(watchtower)!: key the nonce-allocator barrier by nonce lane

mswilkison:
Stacked on #1048.

## What was wrong

The nonce-allocator safety barrier serialized signer I/O against release I/O through a **single row**. An orphaned signer boundary on one sending account therefore froze challenge signing *and* nonce release for every account — a three-part cycle:

1. The boundary CAS increments `active_signer_invocation_count` to 1.
2. Every release-invocation claim requires that counter `= 0`, so no release can be attempted anywhere.
3. Every release stays non-terminal, so `unresolved_release_count > 0`, and the signer-increment gate requires *that* `= 0` — freezing a completely clean lane.

What the two claims actually exclude is signing against releasing **on the same Ethereum account**: a release hands that account's reserved nonce back to its allocator. Nothing in the schema states, or needs, an invariant relating account A's in-flight release to account B.

## What changed

**The claim is keyed by `(chain_id, sender)`** — the nonce lane.

**The columns that are not about an account moved out** to `p2tr_signature_fraud_nonce_allocator_global_barrier`. `contract_mismatch_blocked` condemns the allocator's API implementation, not one lane, and `incident_epoch` counts the events that set it. Both stay global and still block every lane. The already-global unresolved-alert gate is untouched.

**Two bindings the singleton made vacuous are now explicit:**

- The FK to the release *attempt* carries no lane, so on its own it would let one lane's row hold another lane's claim. A second FK pins the claimed request to the row's own key (a new `UNIQUE (chain_id, sender, release_request_id)` on the request table makes that declarative — no denormalizing the attempt table).
- Lane rows are **seeded eagerly** from the signer lane configuration, never lazily on first use, because every gate reads a missing row as fail-closed and the activation rollup needs the row set to be ground truth. Rows outlive the manifest that introduced them, so a lane rotated out keeps its barrier state.

## The part that needed care: activation safety

Activation is a whole-service event. Narrowing the barrier's key must **not** narrow activation safety — the handshake must still refuse if *any* lane has outstanding I/O.

Per-lane equality alone does not achieve that. A lane holding a live signer invocation whose barrier row is **absent** contributes no failing group and no counters — it reads exactly like a clean store, and activation passes. The orphan is invisible to everything else in the handshake, so the result would be a signed "activation-ready" attestation over a live irreversible boundary.

The audit therefore pairs **grouped per-lane equality** with **whole-store SUM equalities**. Together they force any lane without a row to account for nothing, which is exactly the completeness the single-row read used to get for free from the singleton. The retained `rows.length !== 1` guard no longer proves completeness and now says so.

Note the asymmetry that makes this easy to get wrong: the release counter already had an independent whole-store JS twin (`unresolvedReleaseBarrierCount !== pendingNonceReleaseCount`), so a reviewer testing the missing-row case against a *release* sees it correctly blocked. The signer counter had no such twin.

## Tests

No test in the suite drove the barrier audit to a positive mismatch — every assertion was a negative ("mismatch is 0"), which a rewrite that makes mismatch constant-0 satisfies perfectly. Four new tests close that:

| test | kills |
|---|---|
| lane's row deleted under a live boundary | SUM equalities |
| barrier table emptied with I/O outstanding | SUM equalities |
| counters drifted between lanes, total still agrees | grouped equality |
| global flag set with no supporting evidence | invariants D/E over the split-out singleton |

Verified by mutation: removing the two SUM equality terms fails exactly the first two; neutralizing `lane_mismatch_count` fails exactly the third.

**412/412 green** (baseline 408 + 4).

## Two things worth knowing

- The `-1` decrement keys off `OLD`. That is provably the lane that was incremented: the reservation cannot be voided while the marker is live (`003:4774`), so the lane-change guard refuses to clear `selected_sender` in the same update. `NEW` would work today for that reason — `OLD` just doesn't depend on the argument. A mutation swapping them survives, correctly.
- **Known limitation:** within one manifest, `(chain_id, sender)` is 1:1 with a lane. Across manifests a lane that keeps its identity but changes sender gets two barrier rows while the triggers still treat it as one lane. That is a mutual-exclusion weakening, not an activation weakening.

## Not in this PR

This is the durable half. The JS half of the freeze is unchanged and still store-wide: `hasExpiredPreparationLeases` / `hasPendingNonceReleases` take no lane, and `assertRecoveryBarrier` at the `prepare()` entry point runs before lane selection. Lane-scoping those needs store-interface changes (sender-filtered paging) and is the next increment. Until then the service is strictly no *less* safe than before — the JS layer remains stricter than the SQL one.


mswilkison:

PR #1050, opened 2026-07-25
fix(watchtower): make migration 004 applicable to PostgreSQL

mswilkison:
Stacked on #1049.

Migration 004 has never applied to a database. It aliases a table `authorization` in the candidate-enqueue resolution trigger:

```sql
FROM p2tr_candidate_enqueue_authorizations authorization
WHERE authorization.manifest_hash = NEW.manifest_hash
```

`AUTHORIZATION` is a reserved key word, so PostgreSQL rejects the first qualified reference and with it the whole file:

```
ERROR:  syntax error at or near "."
```

The effect is that everything migration 004 declares is absent from every database: the exact-authority key, the outbox-intent foreign key, `p2tr_candidate_enqueue_transaction_guard`, `p2tr_candidate_enqueue_transaction_resolution`, `p2tr_candidate_enqueue_retry_exhaustion_alert`, and their append-only triggers.

## Why no test caught it

Nothing applies it. Every assertion over migration 004 reads the file as **text** — regexes over table names and error strings, plus a lexer-level quote validator — and the adapter suites build their schemas from 001 through 003 only. Under those tests a file the server will not parse is indistinguishable from a correct one.

## The fix

The alias becomes `granted` (one statement, lines 123–128). With that, all four migrations apply cleanly — it was the only defect in the file.

The durable half is a new test that applies **every** migration in directory order to a real server, holding each body to the same `validateP2TRWatchtowerMigrationBody` contract the runner uses so the two cannot drift. It is gated on `P2TR_WATCHTOWER_TEST_POSTGRES_URL` like the other Postgres suites.

Verified by reverting the alias: the new test fails with the original `syntax error at or near "."`, naming the file.

**413/413** (412 + this test).

### DIFF
--- a/services/watchtower/migrations/004_p2tr_candidate_enqueue_retry_alerts.sql
+++ b/services/watchtower/migrations/004_p2tr_candidate_enqueue_retry_alerts.sql
@@ -120,12 +120,12 @@ BEGIN
         RAISE EXCEPTION 'candidate enqueue guard is already exhausted';
     END IF;
 
-    SELECT authorization.consumed_at, authorization.outbox_intent_id
+    SELECT granted.consumed_at, granted.outbox_intent_id
       INTO authorization_consumed_at, authorization_outbox_intent_id
-      FROM p2tr_candidate_enqueue_authorizations authorization
-     WHERE authorization.manifest_hash = NEW.manifest_hash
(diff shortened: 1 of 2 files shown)


mswilkison:

PR #1051, opened 2026-07-25
feat(watchtower)!: refuse signing per nonce lane, not store-wide

mswilkison:
Stacked on #1050. Completes the wedge fix begun in #1049.

## What was still broken

#1049 keyed the durable barrier by `(chain_id, sender)`. The JavaScript half stayed store-wide, so one orphaned signer boundary still froze challenge signing on **every** sending account — the practical wedge was untouched.

## The split that makes it work

`assertRecoveryBarrier` was doing two jobs:

1. **Drive recovery** — `establishRecoveryBarrier` pages the whole store, draining pending releases and expired leases until convergence.
2. **Refuse to sign** on anything left over.

Only (2) is harmful across lanes. Recovering account A's backlog while preparing on account B is work that had to happen anyway and touches nothing of B's safety. So the sweep stays store-wide — `listPage` and `listPendingNonceReleases` are **not** touched — and only the refusal moves to the lane.

## Why the sweep must stay global

A `preparing` record whose lane selection is still `NULL` — the window between the claim swap and the lane swap — belongs to **no lane**. No per-lane predicate, and no rollup over every lane, can see it; only the store-wide sweep reclaims it.

This is the same shape as the activation hole fixed in #1049, where a lane with no barrier row read exactly like a clean store. There the fix was whole-store SUM equalities; here it is keeping the sweep global. A test pins it directly.

## The throw that was actually firing

The error a wedged store raised was `establishRecoveryBarrier`'s `"unresolved durable recovery work"` — **not** either message in `assertRecoveryBarrier`. The flag starts `false`, so establish runs first and throws before the assert's live re-read is ever reached. Narrowing only the assert would have unwedged nothing. That throw is gone; residue now only decides whether the store-wide fast path may latch.

No new per-lane cache was needed: `recoveryBarrierEstablished` already means "the sweeps ran and the store is globally clean", which is a sound fast path — globally clean implies no lane is wedged.

## Call sites

- **`prepare()` entry** — asserts nothing; no lane is selected there. It only drives recovery.
- **No per-candidate wedge check**, deliberately. A lane wedged by an orphan is already held by that orphan under the unique `(chain, sender)` lane slot, so its swap fails and the loop moves on; a lane wedged by an unacknowledged release or a quarantine is refused by `lockAndAssertSignerLaneAvailable` in the same swap. I wrote such a check, found a mutation removing it survived every test, confirmed both existing mechanisms cover it, and dropped it rather than ship two queries per candidate of untested defence.
- **Both irreversible boundaries** assert the selected lane. This is the refusal that matters.
- **`prepareReplacement()`** asserts once its reservation and preparer are resolved, not at entry where the lane is not yet known.
- **`burnContestedNonce`** stays exempt — its lane is by definition the wedged one, so asserting there would deadlock the escape hatch.

The unresolved `reservation-release-failed` disjunct in the durable lane gate stays **global** on purpose: it mirrors the SQL gate at `003:4337` and the `contract_mismatch_blocked` singleton kept global in #1049. An allocator whose contract is in doubt condemns every lane, and the alert table carries no lane key.

## Tests

Nothing pinned any of this. No test asserted a barrier message in either direction, the adapter's two predicates were executed by **no** test, and every fixture was single-lane — so a regression here would have shipped silently. Five tests come with it:

| test | mutation it kills |
|---|---|
| orphan on lane A no longer freezes lane B | restoring the store-wide throw |
| orphan still freezes its own lane | restoring the store-wide throw |
| lane-less `preparing` record still swept | neutralising the lane filter |
| both predicates lane-scoped in SQL, across chains and address casing | neutralising the adapter's filter |

Verified by mutation in both directions. A two-lane dispatcher fixture is new — `validateSignerLanes` requires distinct lane/identity/sender and the reservation is EIP-712 signed by the sender, so lane B needs its own wallet and its own signed blob.

**418/418** (413 + 5).

## Notes for review

- `lane_released_at_unix_ms IS NULL` in the lane-scoped lease predicate is not an optimisation: a record that surrendered its lane is not occupying it, so freezing the lane on its behalf would be wrong. It also happens to match the partial unique index.
- Both store implementations route through one `normalizeP2TRSignatureFraudSigningLane`. A lane is a lookup key, and a casing disagreement between the double and the adapter would have failed in the unsafe direction — double reporting clear while PostgreSQL reported blocked.
- The lane is **required** on the dispatcher assert, not optional. An absent lane silently meaning "no check" is the same shape of hole as a missing barrier row reading as a clean lane.

### DIFF
--- a/services/watchtower/src/P2TRSignatureFraudChallengeOutbox.ts
+++ b/services/watchtower/src/P2TRSignatureFraudChallengeOutbox.ts
@@ -1515,6 +1515,35 @@ export type P2TRSignatureFraudNonceReleasePageRequest = {
   cursor?: string
 }
 
+/**
+ * One nonce lane: the sending account on one chain. This is the unit the
+ * durable nonce-allocator barrier is keyed by, because a nonce reservation and
+ * its release both belong to exactly one account.
+ */
+export type P2TRSignatureFraudSigningLane = {
+  chainID: number
+  sender: string
+}
+
+/**
+ * The single normalizer both store implementations must route through. A lane
+ * is a lookup key, so a checksummed address and its lower-case spelling have to
+ * collapse to one value; if the two implementations disagreed here, the
+ * in-memory double would report a lane clear while PostgreSQL reported it
+ * blocked, and the divergence would favour signing.
+ */
+export function normalizeP2TRSignatureFraudSigningLane(
+  lane: P2TRSignatureFraudSigningLane
+): P2TRSignatureFraudSigningLane {
+  return {
+    chainID: normalizePositiveSafeIntegerLike(
+      lane.chainID,
+      "Signing lane chain ID"
+    ),
+    sender: normalizeAddress(lane.sender, "Signing lane sender"),
+  }
+}
+
 export type P2TRSignatureFraudNonceReleasePage = {
   requests: P2TRSignatureFraudNonceReleaseRequest[]
   nextCursor?: string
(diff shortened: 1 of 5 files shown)


mswilkison:

PR #1055, opened 2026-07-26
build(solidity): pin the toolchain to the versions smock actually works with

mswilkison:
The solidity manifest currently permits dependency combinations that break
the **entire** test suite. Only the lockfile is holding it together.

## Measured

One test file per data point, everything else constant:

| hardhat | smock | result |
|---|---|---|
| 2.12.5 | 2.3.4 | pass — what the lockfile pins today |
| 2.19.5 | 2.3.4 | **pass** |
| 2.20.1 | 2.3.4 | FAIL |
| 2.29.0 | 2.3.4 | FAIL — `Cannot read properties of undefined (reading 'bind')` |
| 2.12.5 | 2.4.1 | FAIL — `provider.init is not a function` |
| 2.29.0 | 2.4.1 | FAIL — `provider.init is not a function` |

`@defi-wonderland/smock` reaches into Hardhat's provider internals, so it breaks
on Hardhat >= 2.20 — and its own 2.4.x line breaks against *every* Hardhat we
can run, including the one we are on.

## The problem

The manifest declared `hardhat: ^2.10.0` and `smock: ^2.3.4`. **Both ranges
resolve to broken versions.** A lockfile refresh, a dependabot bump, or a fresh
resolution on a new machine produces a suite where all 34 fixture-using files
fail in `before all` — with an error that points at neither package.

## This PR

- pins `@defi-wonderland/smock` to `2.3.4` exactly
- constrains `hardhat` to `~2.19.5` — the measured ceiling, which also picks up
  seven minor releases of node/EVM fixes over the 2.12.5 the lockfile had

Full suite unchanged: **2949 passing, 0 failing**.

## Note for anyone going further

Hardhat >= 2.20 (and therefore Hardhat 3, `hardhat-viem`, or
`hardhat-chai-matchers`) requires replacing smock first. No published smock
version works above this ceiling.

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -69,7 +69,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: "20.19.0"
+          node-version: "24.11.1"
 
       - name: Enable Corepack (Yarn from lockfile)
         run: |
@@ -89,6 +89,9 @@ jobs:
 
       - name: Build contracts
         run: yarn build
+
+      - name: Typecheck
+        run: yarn typecheck
 
       # We don't want to execute the tests for dapp-friendly contracts (either
       # when deploying from `dapp-development` or during PR checks on that branch).
@@ -114,7 +117,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: "20.19.0"
+          node-version: "24.11.1"
 
       - name: Enable Corepack (Yarn from lockfile)
         run: |
@@ -150,7 +153,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: "20.19.0"
+          node-version: "24.11.1"
           registry-url: "https://registry.npmjs.org"
 
       - name: Enable Corepack (Yarn from lockfile)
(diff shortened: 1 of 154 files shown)


mswilkison:
Context found after opening this PR, worth recording here since it bears on the pin.

`@defi-wonderland/smock` is **deprecated by its maintainers**. The last commit to
the default branch (2025-05-13) added a `DEPRECATED – DO NOT USE` banner:
unmaintained, "may contain outdated, insecure, or vulnerable code", no further
support or security patches. The npm package carries no `deprecated` flag, so
installs are silent about it.

That does not change what this PR should do — pinning to the last version that
works is still correct, and pinning is *more* important for an abandoned
dependency, not less. But it reframes it as a **stopgap rather than a
resolution**: there will be no smock that supports Hardhat >= 2.20, so the
ceiling this PR documents is permanent until smock is replaced.

Replacement path and cost are in #1056 (`solidity/docs/hardhat-3-migration.md`).
Short version: 29 test files, and the repo already hand-rolls Solidity stubs
(`contracts/test/*Stub.sol`) that are toolchain-independent, so the conversion
can start whenever and does not depend on any Hardhat 3 decision.

Worth noting the exposure is test-only — dev/CI supply chain, nothing on-chain.

mswilkison:
Closing as superseded by changes already merged into dev. After updating this branch with the latest dev, the PR has no remaining content diff: the Hardhat upgrade and smock-removal work are already incorporated upstream.

PR #1056, opened 2026-07-26
chore(solidity)!: Hardhat 3 upgrade spike — blocked, scoped

mswilkison:
**Draft. Not landable.** Opened so the migration gets scoped from measurements
rather than estimates. `solidity/docs/hardhat-3-migration.md` is the deliverable.

The manifest bump to `hardhat@^3.11.1` is included to make the attempt
reproducible. The lockfile is deliberately **not** updated, so this branch does
not carry a broken resolution.

## Stops before a contract compiles

```
ERROR: You are using Node.js 18.20.8 which is not supported by Hardhat.
Please upgrade to Node.js 22.13.0 or later.
```

```
Hardhat only supports ESM projects.
Please make sure you have `"type": "module"` in your package.json.
```

We build and test on Node 18, and the `solidity` package is CommonJS
throughout — config, 60 deploy scripts, 70 test files, `tasks/`.

## Every loaded plugin is Hardhat 2 only

`@nomiclabs/hardhat-waffle`, `@nomiclabs/hardhat-etherscan`,
`@typechain/hardhat`, `hardhat-gas-reporter`, `hardhat-contract-sizer`,
`hardhat-dependency-compiler`, `solidity-docgen`, `@tenderly/hardhat-tenderly`,
plus **two `@keep-network` plugins** that gate on our own release cycle. Full
table with peer ranges in the doc.

## Two have no forward path at all

- **`@defi-wonderland/smock`** — 29 test files. No Hardhat 3 version, no viem
  equivalent. It reaches into Hardhat provider internals, which is also why
  Hardhat >= 2.20 is already unreachable today (see #1055).
- **`hardhat-deploy`** — the Hardhat 3 line is `2.x`, a rewrite on `rocketh`
  with a different API. That takes all **60 deploy scripts** with it.

## Surface

| surface | files |
| --- | --- |
| deploy scripts | 60 |
| tests using ethers v5 API | 75 |
| tests using typechain types | 72 |
| tests using waffle matchers | 53 |
| tests using smock | 29 |

## Recommendation

Treat as a project with prerequisites in order: Node 22 in CI → **settle the
mocking story (no off-the-shelf answer)** → ESM conversion → deploy-layer port →
ethers/typechain/waffle replacement → update the two `@keep-network` plugins.

Nothing here blocks current work. The suite is green on the pinned toolchain
(#1055), and the fixture-loader defect that prompted looking at waffle has
already been fixed independently in #1054 with no dependency change.

### DIFF
--- a/solidity/docs/hardhat-3-migration.md
+++ b/solidity/docs/hardhat-3-migration.md
@@ -59,11 +59,34 @@ Replacements exist for some, but they are replacements, not upgrades:
 
 ## The two that have no answer at all
 
-**`@defi-wonderland/smock`** — used by 29 test files. There is no Hardhat 3
-version, and no viem equivalent. It works by reaching into Hardhat's provider
-internals, which is also why it breaks on Hardhat >= 2.20 (see
+**`@defi-wonderland/smock`** — used by 29 test files. It works by reaching into
+Hardhat's provider internals, which is why it breaks on Hardhat >= 2.20 (see
 `build(solidity): pin the toolchain to the versions smock actually works with`).
-Every fake in the suite would have to be rebuilt on something else.
+
+It is also **deprecated by its maintainers**. The last commit to the default
+branch, 2025-05-13, added this to the README:
+
+> ⚠️ **DEPRECATED – DO NOT USE**
+>
+> This repository is no longer maintained and is **deprecated**. It may contain
+> **outdated, insecure, or vulnerable code** and should **not** be used in
+> production or as a dependency in any project. The repository is retained
+> solely for historical reference. No support, updates, or security patches will
+> be provided.
+
+The npm package itself carries no `deprecated` flag, so installs are silent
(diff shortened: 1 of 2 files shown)


mswilkison:
Superseded by #1065. Closing.

This carried a 127-line Hardhat 3 assessment. #1065 re-measures the same ground
after smock was removed and the toolchain moved, and the document is now 438
lines with the numbers checked rather than inferred — including three of this
version's own figures that turned out to be wrong:

- ESM was described as the blocker across 130 files. Measured across all 151
  TypeScript files it is `"type": "module"`, a tsconfig setting, ten `__dirname`
  usages and two runtime `require()` calls.
- `require(` was recorded as 0 files; it is 2.
- Plugin support, EOL dates and the deploy-layer port were all re-derived.

All four checks on this branch are also failing. Nothing to carry over.

PR #1057, opened 2026-07-26
chore(ci): move off end-of-life Node to 22 LTS

mswilkison:
> **Corrected from Node 24 → 22.** The branch name still says `24`; the content
> and title are 22. See "Why 22, not 24" below — the first revision was verified
> against a warm local solc cache and did not survive CI's cold cache.

Every Node version this repo builds and tests on is past end of life.

| Node | EOL | workflows |
| --- | --- | --- |
| 14 | **2023-04-30** | cross-chain-optimism, cross-chain-polygon |
| 18 | **2025-04-30** | monitoring, npm-contracts, npm-typescript, system-tests, typescript |
| 20 | **2026-04-30** | contracts, cross-chain-arbitrum, cross-chain-bob, deployment-artifacts, format, yearn |

Including `20.19.0`, the newest pin in the tree, which went EOL three months ago.

## Why 22, not 24

24 has 21 months of support against 22's 9, and was the first choice. It does not
work here. Measured, cold solc cache, one variable at a time:

| hardhat | node | result |
| --- | --- | --- |
| 2.12.5 | **22** | compiles, **2949 passing** |
| 2.12.5 | **24** | `HH502` — `InvalidArgumentError: maxRedirections is not supported` |
| 2.29.0 | 24 | compiles |
| 2.29.0 + smock 2.3.4 | 18 | **34 failing** — `reading 'bind'` |

Hardhat's `undici` call is incompatible with Node 24 until ~2.20, and
`@defi-wonderland/smock` breaks on hardhat >= 2.20. **The two windows do not
overlap.** Node 22 is the ceiling until smock is replaced — and smock is archived
upstream (`wonderland-archive/smock`, `archived=true`), so that is a project of
its own rather than a bump.

The 9-month window is therefore a *consequence* of the smock pin, not a reason to
prefer 22 on its merits.

## On the cold cache

The first revision of this PR claimed Node 24 worked. It did — locally, against a
warm 295 MB solc cache. CI starts cold, cannot download the compiler, and dies.
Every number above was re-measured with the cache cleared.

## Verified

Node v22.23.1, cold cache, no dependency changes:

```
solidity     2949 passing, 31 pending, 0 failing
typescript    738 passing,  63 pending, 0 failing
```

## Scope

- 13 workflows → Node 22. Pin style preserved: exact pins become `22.23.1`,
  ranges become `22.x`.
- `engines.node` raised in `solidity` (`>=18.0.0`) and `typescript` (`>=16`).
[...]

PR #1060, opened 2026-07-26
test: prototype replacing smock fakes with Solidity stubs (1 of 29 files)

mswilkison:
**Draft — pattern-setter.** One type, one file, so the per-type cost of removing
smock is measured rather than estimated before committing to the other 14.

## Why remove smock at all

`@defi-wonderland/smock` is **archived upstream** — the repo is now
`wonderland-archive/smock`, `archived=true`, with a README banner: *"no longer
maintained… may contain outdated, insecure, or vulnerable code… No support,
updates, or security patches."* The npm package carries no `deprecated` flag, so
installs say nothing.

It is also load-bearing on the toolchain. One dependency, three constraints:

| | |
| --- | --- |
| smock breaks on hardhat >= 2.20 | `Cannot read properties of undefined (reading 'bind')` |
| hardhat < ~2.20 cannot run Node 24 | `InvalidArgumentError: maxRedirections is not supported` |
| ⇒ | CI is capped at **Node 22** (9 months of support) instead of 24, and Hardhat 3 is unreachable |

## Why a hand-written stub rather than a library

Every third-party option is dead or incompatible:

| candidate | last publish | verdict |
| --- | --- | --- |
| `@ethereum-waffle/mock-contract` | 2023-10 | part of abandoned waffle |
| `@gnosis.pm/mock-contract` | 2022-04 | 4 years stale |
| `@clrfund/waffle-mock-contract` | 2024-05 | needs **ethers v6** — which smock is what blocks |

The category is dead because the ecosystem moved to Foundry, where mocking is
native. A plain Solidity stub has **no JavaScript dependency**, so it survives
hardhat 2 → 3, ethers → viem, and a Foundry move.

## The migration is the tractable half of smock

Across 29 files: **78 `smock.fake` calls over 15 distinct types**, and **zero**
uses of `smock.mock`, `setVariable` or `getVariable` — the storage-manipulation
APIs with no equivalent. Everything is interface fakes with programmable
returns, which is exactly what a stub contract is. Four types cover 41 of the 78
sites.

## The translation

`ReimbursementPoolStub` records `refund` calls instead of transferring, so
smock's spy matchers become state reads:

| smock | stub |
| --- | --- |
| `refund.getCall(0).args` | `refundCall(0)` |
| `expect(refund).to.have.been.calledTwice` | `refundCallCount() == 2` |
| `refund.reset()` | `resetRefundCalls()` |
| `maxGasPrice.returns(x)` | `setMaxGasPrice(x)` |

That also removes the need for `smock.matchers` in files that only fake this type.

## Result

`L1BTCDepositorNtt.ntt.test.ts`: **14 passing, 0 failing**. Three edited lines
plus one new contract. The other 8 `ReimbursementPool` call sites follow the same
shape and are left for follow-ups so this stays reviewable.

Draft because the sequencing is yours: 29 files is a wide test-only change and
the 13-PR watchtower chain is currently green against main.

### DIFF
--- a/solidity/contracts/test/ReimbursementPoolStub.sol
+++ b/solidity/contracts/test/ReimbursementPoolStub.sol
@@ -1,0 +1,104 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+pragma solidity 0.8.17;
+
+/// @notice Test-only stand-in for `@keep-network/random-beacon`'s
+///         ReimbursementPool.
+/// @dev Replaces `smock.fake<ReimbursementPool>("ReimbursementPool")`.
+///
+///      smock is archived upstream (`wonderland-archive/smock`, no further
+///      support or security patches) and it is the reason this package cannot
+///      move past hardhat 2.19.x, which in turn caps the toolchain at Node 22.
+///      A plain Solidity stub carries none of that: it has no JavaScript
+///      dependency, so it survives hardhat 2 -> 3, ethers -> viem, and a move
+///      to Foundry.
+///
+///      Behaviour is deliberately the subset the tests use, not the whole
+///      contract:
+///
+///      - `refund` records every call instead of transferring anything. The
+///        recorded calls replace smock's `refund.getCall(n).args` and the
+///        `calledOnce` / `calledTwice` matchers, via `refundCallCount()` and
+///        `refundCall(i)`.
+///      - `maxGasPrice`, `staticGas`, `owner` and `isAuthorized` return values
+///        the test sets, replacing smock's `.returns(...)`.
+///      - `resetRefundCalls()` replaces `refund.reset()`.
+///
+///      Assertions become state reads rather than spy matchers, which is why
+///      this needs no chai plugin.
+contract ReimbursementPoolStub {
+    struct RefundCall {
+        uint256 gasSpent;
+        address receiver;
+    }
+
+    RefundCall[] internal refundCalls;
+
+    uint256 public maxGasPrice = 20000000000;
+    uint256 public staticGas = 40800;
+    address public owner;
+    bool internal authorizedDefault = true;
+
+    mapping(address => bool) internal authorizedOverride;
+    mapping(address => bool) internal authorizedOverrideSet;
+
+    /// @notice Records the call. The production contract would transfer ETH;
+    ///         every test that uses this stub asserts on the call, not on a
+    ///         balance change.
+    function refund(uint256 gasSpent, address receiver) external {
+        refundCalls.push(RefundCall(gasSpent, receiver));
+    }
+
+    function isAuthorized(address account) external view returns (bool) {
+        if (authorizedOverrideSet[account]) {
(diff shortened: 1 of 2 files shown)


mswilkison:
Scope note, since the previous title read like this removed smock.

It does not. This is a single-file prototype of the stub pattern: 1 of **29** test
files that import smock, 1 of **77** `smock.fake<>` call sites, 1 of **15** distinct
types. The rest is real work that has not started.

Measured today, and it corrects two things I had wrong earlier:

**1. The pinned Hardhat is 2.12.5, not 2.19.x.** `hardhat: "^2.10.0"` resolves to
2.12.5 in `yarn.lock`.

**2. Node 24 needs Hardhat >= ~2.20, and that is what smock blocks.** Cold compile
of this repo on Node 24:

| hardhat | result |
| --- | --- |
| 2.12.5 (pinned) | `HH502` |
| 2.19.5 | `HH502` |
| 2.29.0 | compiles |

The underlying error is dependency skew in the solc download path, not anything
to do with smock:

```
InvalidArgumentError: maxRedirections is not supported, use the redirect interceptor
    at new Request (node:internal/deps/undici/undici:2581:17)      <- Node 24 built-in undici
    at Agent.request (solidity/node_modules/undici/lib/api/api-request.js:193)  <- npm undici 5.10.0
```

npm `undici@5.10.0` and Node 24 bundled undici share a global-dispatcher symbol, so
Hardhat 2.12.x/2.19.x route their download through Node internal dispatcher, which
dropped `maxRedirections`. Bumping only `undici` to 5.29.0 does **not** fix it;
Hardhat 2.29.0 does.

**3. smock really does break there.** Verified on Node 20, so Node is not a variable:

```
smock=2.3.4  hardhat=2.29.0
TypeError: Cannot read properties of undefined (reading "bind")
  at Sandbox.create (node_modules/@defi-wonderland/smock/src/sandbox.ts:73:56)
```

So the ordering is: finish this migration -> bump Hardhat to 2.29.x -> Node 24.
Node 22 (#1057) does not depend on any of it and should land on its own.

mswilkison:
Superseded by #1062 and #1063. Closing.

This prototyped one approach — a hand-written Solidity stub per mocked
contract, `ReimbursementPoolStub.sol`, done for 1 of 29 files as proof of
concept. What shipped instead is a generic `contracts/test/MockContract.sol`
plus `test/helpers/mock.ts`, which works against any ABI and so covered all 29
files without a stub each.

The strongest argument here was toolchain independence: a Solidity stub carries
no JavaScript dependency, so it survives hardhat 2 → 3, ethers → viem and a
move to Foundry. That was a good point and it has been addressed rather than
ignored — `MockContract.sol` is Solidity too, and #1066 includes an interop
test showing the TypeScript helper drives it unchanged from a viem test, in
both directions. Foundry can use the contract directly.

Nothing here needs salvaging. The generic replacement is merged into the stack
at #1062 → #1063 and the suite is green on it.

PR #1062, opened 2026-07-26
test(solidity): toolchain-independent replacement for smock (foundation)

mswilkison:
Foundation only — the mock plus its own 19-test suite. **No existing test is migrated here.**
Follow-ups migrate the 29 files; this is the piece they all depend on.

## Why smock has to go

`@defi-wonderland/smock` is archived upstream ("may contain outdated, insecure, or
vulnerable code"). It is also what pins the toolchain. Measured:

| hardhat | Node 24 cold compile | smock 2.3.4 |
| --- | --- | --- |
| 2.12.5 (what is pinned) | `HH502` | works |
| 2.19.5 | `HH502` | works |
| 2.29.0 | compiles | `TypeError` in `Sandbox.create` |

So Node 24 needs hardhat >= ~2.20, and hardhat >= 2.20 needs smock gone. See #1060
for the full measurement.

## Why not Foundry / Hardhat 3 / an off-the-shelf package

Both Foundry and [Hardhat 3](https://hardhat.org/docs/reference/cheatcodes/environment/mock-call)
offer `vm.mockCall`, and it is better than anything we would write. But cheatcodes
only exist inside **Solidity** tests. These 29 files are TypeScript on typechain and
ethers v5, and 17 suite tests drive `deployments.fixture()`, which has no Solidity
equivalent. Reaching `vm.mockCall` means porting the tests first.

`@term-finance/ethers-mock-contract` is the one maintained JS option and does not fit:
v0.1.9 last published 14 months ago, no declared peer deps, and no call recording or
call-count assertions — which this suite uses **69** times (`getCall` 33, `to.have.been.*` 36).

## What this is

Foundry’s *model*, not its implementation. `MockContract.sol` is an ordinary deployed
contract holding a calldata→returndata table plus call recording, driven from TypeScript
over the public provider API. It touches no Hardhat internals — precisely what broke
smock, and what will break any JS mock at the next toolchain move.

The survey said this was viable: **78** `smock.fake`, **0** stateful `smock.mock`,
**0** `.returns()` taking a callback. Every use is static.

## Three findings that each forced a redesign

**1. STATICCALL.** Recording is a storage write, so a stubbed `view` function reached by
STATICCALL would revert — every such stub in the suite. Recording now goes through a
self-call whose failure is swallowed, making it best-effort. Sound here because *every*
call assertion in the suite targets a state-changing function (`refund`, `revealDepositWithExtraData`,
`seize`, `transferTokensWithPayload`, `sendVaasToEvm`, `requestNewWallet`); a view reached
by STATICCALL could not have been counted anyway.

**2. `reset` cannot infer.** Since STATICCALL calls are not recorded, reset cannot work out
what to clear from the calls it saw. It tracks what was *configured* instead.

**3. Empty returndata is not zero.** Solidity compares `returndatasize` against the size its
ABI expects and reverts on a short answer, so an unstubbed function cannot answer with `0x`
— it reverts the caller. The helper computes the correctly encoded zero for every return
type and installs it as a base layer `reset` leaves alone, reproducing smock’s behaviour.

## The one behavioural difference

smock configured fakes by mutating in-process JavaScript, so its API was **synchronous**.
Configuration here is a transaction, so setup and inspection must be `await`ed. That is what
the per-file migration changes at call sites — mechanical, but it is not a no-op, and I want
it stated up front rather than discovered in review.

## Incidental

The fallback reads `msg.data` and returns through assembly rather than using a typed
fallback, because **prettier-plugin-solidity silently rewrites**
`fallback(bytes calldata callData) external payable returns (bytes memory)` into
`fallback() external payable` — dropping the parameter and the return type. A formatter
quietly changing semantics is worth knowing about independently of this PR.

## Verification

```
19 passing
```

covering STATICCALL views, struct and multi-value returns, `whenCalledWith` precedence,
reverts (blanket and per-argument), call recording and argument decoding, all four `reset`
behaviours, `wallet` impersonation, and pinned-address deployment.

## Summary by CodeRabbit

* **New Features**
  * Added a programmable Solidity mock framework for stubbing returns/reverts (including argument-specific overrides), supporting static-call friendly `view` behavior, call tracking, and reset/clear controls, with safety checks for selector collisions and pinned deployments.
  * Added typed mock creation for contract interfaces and async call assertion helpers.
* **Tests**
  * Added comprehensive coverage for static calls (including struct/multi-value returns), overrides, reverts, call recording (including no-return flows), reset behavior, and pinned-address safety and assertion handling.

### DIFF
--- a/solidity/contracts/test/IMockTarget.sol
+++ b/solidity/contracts/test/IMockTarget.sol
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+pragma solidity ^0.8.0;
+
+/// @notice Test-only interface exercising the shapes `MockContract` has to
+///         answer: a view function reached by STATICCALL, a state-changing
+///         function whose calls are asserted on, a function returning nothing,
+///         a struct return and a multi-value return.
+interface IMockTarget {
+    struct Info {
+        address owner;
+        uint64 createdAt;
+        bool active;
+    }
+
+    function doThing(address who, uint256 amount) external returns (bool);
+
+    function noReturn(uint256 value) external;
+
+    function readValue(uint256 key) external view returns (uint256);
+
+    function readInfo(uint256 key) external view returns (Info memory);
+
+    function readPair(uint256 key) external view returns (uint32, uint32);
+}
--- a/solidity/contracts/test/MockContract.sol
+++ b/solidity/contracts/test/MockContract.sol
@@ -49,42 +49,64 @@ contract MockContract {
         bytes data;
     }
 
-    /// @dev keccak256(full calldata) => response.
-    mapping(bytes32 => Response) private responseByCalldata;
-    /// @dev selector => response.
-    mapping(bytes4 => Response) private responseBySelector;
-
-    /// @dev selector => response of last resort, installed once when the mock
-    ///      is created and never cleared by `reset`.
+    /// @dev All mock state lives behind one hashed base slot.
     ///
-    ///      Solidity checks returndatasize against the size its ABI expects and
-    ///      reverts on a short answer, so an unconfigured function cannot
-    ///      simply return nothing: it has to return a correctly encoded zero
-    ///      value. Only the helper knows the mocked ABI, so it computes those
-    ///      encodings and installs them here. That reproduces smock, where an
-    ///      unstubbed function yields the zero value of its return type.
-    mapping(bytes4 => Response) private baseResponseBySelector;
+    ///      Mocks are routinely installed over an address that already holds a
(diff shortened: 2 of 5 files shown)


piotr-roslaniec (approved):
Self-approve as repo admin. MockContract.sol foundation for the smock removal. 5 files, 1330 +/- 0. The 19 self-tests pass but code review surfaced 3 P0 latent bugs (selector masking, setBaseReturns-before-relocation, encodeReturn single-array). These are inherited by 1063, 1064, and 4203. Whether to fix before merge or merge-as-is given the migration paths don't currently trigger them is a separate decision to be made next.

PR #1065, opened 2026-07-26
docs(solidity): re-measured Hardhat 3 assessment

mswilkison:
**Draft, docs only.** Stacked on #1064. Re-measures the Hardhat 3 blockers now that
smock is gone and the toolchain is on hardhat 2.29 / Node 24. Supersedes the
assessment in #1056.

## No longer blockers

- **smock** — was archived with no Hardhat 3 version ever coming. Fully replaced.
- **Node floor** — Hardhat 3 wants `>= 22.13.0`; CI is on 24.
- **ESM** — this was the largest number in the original assessment and it was wrong.
  Across all 151 TypeScript files: `module.exports` 0, `__filename` 0,
  `import x = require()` 0, `require(` **2**, `__dirname` **10**. The package is
  CommonJS only by compilation target. The conversion is `"type": "module"`, a
  tsconfig setting, ten `__dirname` replacements and two `require()` calls. An
  afternoon, not a quarter — and it is the *cheapest* item, not the gate.

## Still decisive

**Plugin support, measured — 3 of 11:**

| has a Hardhat 3 line | does not |
| --- | --- |
| `@nomicfoundation/hardhat-ethers` 4 | `@nomiclabs/hardhat-waffle`, `@typechain/hardhat` |
| `@openzeppelin/hardhat-upgrades` 4 | `hardhat-gas-reporter`, `hardhat-contract-sizer` |
| `hardhat-deploy` 2 | `hardhat-dependency-compiler`, `solidity-docgen` |
| | `@keep-network/hardhat-helpers` *(ours)* |

**The deploy layer is a port, not a bump.** `hardhat-deploy@2` supports Hardhat 3,
but on a `rocketh` rewrite with a different API, against 60 scripts.

## New in this revision: the waffle replacement is decided

Item 1 of the work order forks between viem and ethers v6. It was being argued
from a premise that does not hold — that the repo's viem work gives the contract
suite a head start. #1036 is 177 files and #1033 is 60; **all 237 are in
`typescript/` and zero in `solidity/`**. Consistency is still a fair argument for
viem; reuse is not one.

So it was priced. #1066 migrates one real test file to viem. viem *works* — the
plugin coexists with waffle, typechain, hardhat-deploy and ethers v5, and the
migrated file keeps all 24 tests. What it costs is the assertion layer (no viem
chai matchers exist on Hardhat 2 — `hardhat-viem-assertions` is stable but peers
`hardhat@^3.8.0`) and the types (TypeScript 5 + `skipLibCheck` + `strict` +
`paths.artifacts` in `include`; `strict` alone is 378 errors in existing files).

`hardhat-chai-matchers@2.1.2` needs none of that and keeps 931 + 571 + 522
existing assertion sites working. **Decision: ethers v6 for `solidity/`, viem for
`typescript/`** — revisit if the TypeScript 5 / `strict` work lands on its own
merits, since that chain is most of viem's cost.

## Also new: this document said the suite was green, and it is not

CI has been red since #1063 — 2466/31/**188** there, 2690/31/**102** on #1064 and
#1065, against a green #1062. A local run reproduces the 102 exactly.

Four defects account for all of them: a bare `deployments.fixture()` reverting
below the mocks `fixtures/bridge.ts` installs, combined with waffle's
`loadFixture` discarding `evm_revert`'s result (43); `allowBlocksWithSameTimestamp`
removing the per-block second `increaseTime` relied on to satisfy a strict guard
(51); `expectCalledOnceWith` comparing representation rather than value (6); and
the `Mock` handle having no `.connect` (2). The document now records the
mechanisms and the sequencing. **The fixes belong on #1063, not here.**

## Smaller corrections to this document's own numbers

`require(` was recorded as 0 and is 2; `test/` holds 89 TypeScript files, not 70;
and `bridgeGovernanceCompilerConfig` drops the optimizer to 200 from a default of
1000 — the `runs: 1` override belongs to `L1BTCDepositorNttWithExecutor`.

## Worth deciding separately

What Hardhat 3 offers that Hardhat 2 cannot is **multichain simulation** (this repo
tests Base and Arbitrum depositors against a mainnet-behaving network) and **build
profiles** (the config hand-rolls this today). **Not** EDR — that was backported to
2.21 and is already in use. Solidity tests with cheatcodes are reachable *now* via
Foundry, which is #1061.

### DIFF
--- a/solidity/docs/hardhat-3-migration.md
+++ b/solidity/docs/hardhat-3-migration.md
@@ -340,104 +340,99 @@ Nothing here blocks current work. The archived dependency is gone, and the
 fixture-loader defect that started this line of investigation was fixed
 independently with no dependency change at all.
 
-## The suite is not green, and this document said it was
+## The suite was not green, and this document said it was
 
 An earlier revision closed by calling the suite green on hardhat 2.29 and
-Node 24. Measured on `chore/hardhat-3-assessment`, on the TypeScript it pins,
-twice with identical results:
+Node 24. It was not, and CI had been saying so for as long as the stack existed:
 
-```
-2690 passing   31 pending   102 failing
-```
-
-Not the 23 previously recorded, and not the same failures — none of the 102 is
-the `deployed ReimbursementPool contract not found` error that accounted for
-all 23. CI says the same thing and has been saying it for as long as the stack
-has existed:
-
-| PR      | Node | CI                                      |
+| PR      | Node | CI before                               |
 | ------- | ---- | --------------------------------------- |
 | `#1062` | 20   | green                                   |
 | `#1063` | 20   | 2466 passing / 31 pending / 188 failing |
 | `#1064` | 24   | 2690 / 31 / **102**                     |
 | `#1065` | 24   | 2690 / 31 / **102**                     |
 
-So the regression enters at `#1063`, the smock removal itself, and `#1064`'s
-toolchain bump repaired 86 of the 188 rather than causing any. A local run
-reproduces CI's 102 exactly, failure for failure and in the same order.
+Not the 23 previously recorded, and not the same failures — none of the 102 was
+the `deployed ReimbursementPool contract not found` error that accounted for
+all 23. The regression entered at `#1063`, the smock removal itself; `#1064`'s
+toolchain bump repaired 86 of the 188 rather than causing any.
 
-They are four defects, not one, and they account for all 102 between them.
+It is fixed, in three commits on `#1063`. **2971 passing, 31 pending, 0
+failing**, three consecutive runs. The four defects, and what each was:
 
-**A `deployments.fixture()` rollback un-installs the fixture's mocks — 43.**
-`test/fixtures/bridge.ts` runs `deployments.fixture()` and only then installs
-MockContract bytecode over the real `WalletRegistry` and `LightRelay`
-addresses with `hardhat_setCode`. A bare `deployments.fixture()` later in the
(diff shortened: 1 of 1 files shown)


mswilkison:
## For the record: this PR's content landed via #1113

This PR reads as closed and unmerged with an empty diff. That record is misleading. What actually happened:

- **2026-07-27** — last real push, head `f1a7abe7816650c4b6fa373133c7b079a984bf5d`: the hardhat 2.29 + Node 24 CI change and the re-measured Hardhat 3 assessment in `solidity/docs/hardhat-3-migration.md`, on top of the #1063 smock-removal chain.
- **2026-08-26 15:29Z** — every branch in the #1063 → #1070 stack was force-pushed onto the base line at once, emptying this PR. On 2026-08-27 the branch was pointed at the empty squash `2401e6bb5` and the PR closed.
- **2026-08-27** — #1113 restored the whole linear chain #1063 → #1070 to `dev` as `7bf95ecdd`, which contains everything this PR held. Verified: every file where `7bf95ecdd` differs from this PR's tip is explained by an unrelated `dev` commit that landed in between.

So the content is on `dev` and has since been edited further (#1115, #1054, and the 3 September commits). Only two things were lost: this PR's record, and the per-commit authorship, which `dev` carries as a single commit with a co-author trailer.

The branch was restored to `f1a7abe78` on 2026-09-07 only because #1066 and #1067 use it as their base. Once #1067 is rebased onto `dev` and #1066 is resolved, the branch will be tagged to keep the original commits reachable, then deleted. Nothing to reopen here.

